PR #40, opened 2018-04-10
Add .travis.yml, with Appengine creds for Travis

lukebjerring:
Travis creds are from Appengine.

- Downloaded a copy of the travis-ci service account key
- Copied to `client-secret.json`
- Ran `travis encrypt-file client-secret.json`
- Remove the original (unencrypted) client-secret.json

PR #41, opened 2018-04-10
Add Travis staging deploy job

lukebjerring:
Building on #40 

Adds an Appengine staging deployment step to travis, which re-uses the `git-deploy.sh` standard version naming to name the Appengine `--version` flag after the `[user]-[branch]` combo.

The deploy also comments on the PR that triggered the build, providing a link to the deployed staging environment (using the GITHUB_TOKEN, which was added to Travis using `travis encrypt GITHUB_TOKEN=secret_sauce`.


Makefile
```diff
@@ -58,3 +58,8 @@ go_deps: $(find .  -type f | grep '\.go$' | grep -v '\.pb.go$')
 dev_data:
 	cd $(WPTD_GO_PATH)/util; go get -t ./...
 	go run util/populate_dev_data.go
+
+webapp_deploy_staging:
+	gcloud config set project wptdashboard
+	gcloud auth activate-service-account --key-file $(WPTD_PATH)/client-secret.json
+	cd $(WPTD_PATH)/git; ./git-deploy -q
```
Hexcles: What is `git-deploy`? I don't see the file under `git/`.

lukebjerring: It's moved to `util/deploy.sh` since I originally authored this.

Hexcles:
In fact, all Travis jobs are failing. Something is wrong with the Makefile/Docker setup. Do you copy the Makefile into Docker? The Travis jobs try to make *inside* Docker and fail because Makefile can't be found.

The other jobs are hard failures (red) because they fail inside the `script` section (they have non-empty `MAKE_TEST_TARGET`), while `webapp_deploy_staging` only has `MAKE_DEPS_TARGET` (which is in the `install` section).

lukebjerring:
Added ` || exit $?` to the script to ensure the failed task fails the build.

util/deploy.sh
```diff
@@ -23,30 +27,67 @@ info "Installing dependencies..."
...
+if [[ "${QUIET}" != "true" ]]
+then
+    confirm "Execute?"
+    if [ "${?}" != "0" ]; then exit "${?}"; fi
+fi
 
 info "Executing..."
-${COMMAND}
-exit 0
+${COMMAND} || exit $?
```
Hexcles: If you find yourself repeating this pattern, you can try to add `set -e` at the top of your Bash script, which will exit the script immediately if any command fails. Then, for the commands that you think may (and can) fail, add `|| true` at the end of those lines instead.

(In fact, I'd recommend doing so, but I'm not sure if it'll break anything existing; you can try it out.)

lukebjerring:
OK, build is now (correctly) red on a failed deploy :+1:

lukebjerring:
Running at https://travis-staging-deployment-dot-wptdashboard.appspot.com

lukebjerring:
@Hexcles - turns out one of the many plaguing problem was just `\r` chars. As you can see from https://github.com/web-platform-tests/wpt.fyi/pull/41#issuecomment-382545589 the the task now succeeds - and the token is commenting as me, which is a blocker for sure.

BUT, it's ready for review!

.travis.yml
```diff
@@ -52,41 +56,53 @@ before_install: |
...
+    if [ "${MAKE_DEPS_TARGET}" != "" ]; then
+      docker exec -t -u 0:0 "${DOCKER_INSTANCE}" make "${MAKE_DEPS_TARGET}" || exit $?
+    fi
 
 script:
-  - |
+  - | # Deploy PR to staging environment
+    if [ "${MAKE_DEPLOY_TARGET}" != "" ]; then
+      echo "Copying output to ${TEMP_FILE:=$(mktemp)}"
+      docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_DEPLOY_TARGET}" BRANCH_NAME=${TRAVIS_PULL_REQUEST_BRANCH} 2>&1 | tee ${TEMP_FILE}
```
Hexcles: Can we add a comment here for why we need to redirect stderr to stdout, given the huge amount of time spent on figuring this out?

.travis.yml
```diff
@@ -52,41 +56,53 @@ before_install: |
...
+      if [ "${EXIT_CODE:=${PIPESTATUS[0]}}" != "0" ]; then exit ${EXIT_CODE}; fi
+      DEPLOYED_URL="$(grep -Po 'Deployed to \K.*$' ${TEMP_FILE})"
```
Hexcles: What about

```bash
grep -Po 'Deployed to \K\S$' ${TEMP_FILE}
```

`\S` for non whitespace, so that we don't need to `tr` in `deploy-comment.sh`. This would be more intuitive and future-proof IMHO. For example, if later we want to use `${DEPLOYED_URL}` in some other command/script, we don't need to trim it again.

util/deploy-comment.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# Helper script for posting a GitHub comment pointing to the deployed environment,
+# from Travis CI. Also see deploy.sh
+
+# Ensure no carriage returns or newlines
+STAGING_URL="$(echo $1 | tr -d '\r\n')"
```
Hexcles: See comment above.

.travis.yml
lukebjerring: Done

.travis.yml
lukebjerring: grep prints a trailing newline. Moved the `tr` into .travis.yml though.

util/deploy-comment.sh
lukebjerring: Removed.

lukebjerring:
Submitting with myself as the commenter until https://github.com/web-platform-tests/wpt-pr-bot/issues/17 is resolved.

Issue #39, opened 2018-04-10
Change UI incentives to focus on interoperability, not test pass rate


lukebjerring:
_From @jeffcarp on August 16, 2017 20:56_

According to http://wpt.fyi/about, the stated purpose of the WPT Dashboard is:

> to promote viewing the web platform as one entity and to make identifying and fixing interoperability issues as easy as possible.

However, the way the UI works today explicitly rewards passing tests over failing tests by displaying green for 100% passing results and shades of red for anything else.[1] If a browser came along and magically made all their tests 100% green, that wouldn't entirely satisfy the goal of platform predictability.

Ideally, as I understand the goals, the "opinion" of the dashboard UI should be:

- Tests on all platforms passing: `GOOD`
- Tests on all platforms failing: `OK`
- Tests on two platforms passing, other two failing: `BAD`

My concrete suggestions are:
1. Move away from using the colors green and red with test results. To maintain the ability to quickly glance and see passing vs. failing tests, we could map test passing percentage to a shade of blue on a linear scale
2. Calculate the standard deviation of test results per directory and per test file (normalized for number of total subtests) and use red & green colors to reward directories that have a low deviation.[2] We can also highlight rows more prominently that have a high deviation and therefore need more interop focus.

I have a demo of this up here: http://sigma-dot-wptdashboard.appspot.com/

![screenshot from 2017-08-16 13 46 29](https://user-images.githubusercontent.com/916028/29384559-65cc18b6-8289-11e7-8fc0-ed3d7651253b.png)

[1] The code that determines the color based on pass rate lives at [components/wpt-results.html#L320](https://github.com/GoogleChrome/wptdashboard/blob/master/components/wpt-results.html#L320)
[2] The green=good red=bad connotation applies only in Western cultures, however I can't think of a better alternative

_Copied from original issue: w3c/wptdashboard#83_

lukebjerring:
_From @drufball on August 17, 2017 17:57_

Amazing - LOVE this idea. Even just briefly scanning through the demo the sigma's were highliting good areas of focus.

lukebjerring:
_From @gsnedders on August 18, 2017 15:5_

One option is to show something derived from per-directory (presumably pre-computed) interop data, and don't do anything based on pure pass/fail data.

lukebjerring:
_From @jgraham on August 18, 2017 17:9_

[...]

Issue #13, opened 2018-04-10
wpt.fyi/* routes are future hostile


lukebjerring:
_From @rwaldron on January 18, 2018 20:40_

If https://github.com/w3c/web-platform-tests were to ever grow a new top level directory for some hypothetical spec named "api", "interop", "results", or "test-run", then routing in webapp/main.go becomes ambiguous. 

_Copied from original issue: w3c/wptdashboard#374_

lukebjerring:
_From @boazsender on January 19, 2018 17:38_

+1, test results routes should be name spaced.

For the `/interop` example in #372, it's worth mentioning that the [compat spec](https://compat.spec.whatwg.org) could have just as easily been called "interop". We have tests for that spec, and the `/compat` route is in the global name space (see https://wpt.fyi/compat).

lukebjerring:
_From @rwaldron on January 19, 2018 17:48_

Honestly, the only truly safe solution is the one that breaks all the existing inbound links, ie. wpt.fyi/test/path.html => wpt.fyi/some-safe-path/test/path.html, but that might be the price to pay for rushing this out.

lukebjerring:
We can redirect the current wildcard matches to /some-safe-path/, mitigating breakages to current links to clashes only.

lukebjerring:
_From @rwaldron on January 19, 2018 19:23_

@lukebjerring yep, that's what was I thinking (and what brought me back here just now to follow up on). I don't want to rush into this yet.

lukebjerring:
This issue is related to the Q1 OKR of switching the default view to the interop view:
https://docs.google.com/document/d/1ojlLDiRruI4e95Z7JQ10JQPRMiQHd_40rBkYCL6IJqw/edit#heading=h.v93ez0y7fvmy

Hexcles:
I came across this when writing my design doc for the results receiver.

Currently, we have a few API endpoints in `wptf.fyi/api`. Shall we perhaps use `wpt.fyi/_api/` instead, especially considering most of these APIs are internal to the project?

lukebjerring:
Not an important distinction, IMHO. After https://github.com/web-platform-tests/wpt.fyi/pull/47, there won't be a clash, and there currently isn't a clash, so should be no existing (incorrect) links.

`/api/` links will be in a fair few pieces of code already.

lukebjerring:
Resolved by https://github.com/web-platform-tests/wpt.fyi/pull/47

PR #47, opened 2018-04-13
Add results/ prefix to landing page

lukebjerring:
Working towards https://github.com/web-platform-tests/wpt.fyi/issues/13 (current path scheme is future-hostile) and https://github.com/web-platform-tests/wpt.fyi/issues/39 (interop focus)

Redirects the `/*` path wildcards to `/results/*`, which will be _mooooostly_ backward-compatible for existing links, except for the clashes which are possible (and hence the change moving forward).

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -151,10 +151,6 @@ found in the LICENSE file.
   <script>
     /* global TestRunsBase */
     class WPTResults extends TestRunsBase {
-      static get UrlToResultsPath() {
-        return location => location.pathname.replace(/\/results(.+)(\/)?$/, '$1');
-      }
-
       static get is() {
         return 'wpt-results';
       }
@@ -177,10 +173,7 @@ found in the LICENSE file.
             type: String,
             computed: 'computeTestScheme(path)'
           },
-          path: {
-            type: String,
-            value: WPTResults.UrlToResultsPath(location)
-          },
+          path: String,
           pathIsATestFile: {
             type: Boolean,
             computed: 'computePathIsATestFile(path)'
@@ -194,6 +187,11 @@ found in the LICENSE file.
             value: []
           }
         };
+      }
+
+      ready() {
+        super.ready();
+        this.path = this.urlToResultsPath(window.location);
       }
 
       computeIsLatest(sha) {
@@ -211,11 +209,16 @@ found in the LICENSE file.
         return /(\.(html|htm|py|svg|xhtml|xht|xml)$)/.test(path);
       }
 
+      urlToResultsPath(location) {
+        return location.pathname.replace(/\/results\/(.+)?$/, '/$1');
+      }
+
       async connectedCallback() {
         await super.connectedCallback();
 
+        const that = this;
         window.onpopstate = () => {
-          this.path = WPTResults.UrlToResultsPath(window.location);
+          this.path = that.urlToResultsPath(window.location);
           this.refreshDisplayedNodes();
         };
 
@@ -318,7 +321,7 @@ found in the LICENSE file.
 
       navigate(event) {
         event.preventDefault();
-        let path = WPTResults.UrlToResultsPath(event.target);
+        let path = this.urlToResultsPath(event.target);
         if (path === this.path) {
           return;
         }
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -126,7 +126,7 @@ found in the LICENSE file.
           <template is="dom-repeat" items="{{displayedNodes}}" as="node">
             <tr>
               <td>
-                <path-part path="{{ node.path }}" is-dir="{{ node.isDir }}" navigate="{{ bindNavigate() }}"></path-part>
+                <path-part prefix="/results" path="{{ node.path }}" is-dir="{{ node.isDir }}" navigate="{{ bindNavigate() }}"></path-part>
               </td>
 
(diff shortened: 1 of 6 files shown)


lukebjerring:
Running at https://lukebjerring-results-prefix-dot-wptdashboard.appspot.com/

foolip:
https://lukebjerring-results-prefix-dot-wptdashboard.appspot.com/ has two of the big red warnings, an extra at the bottom. Related to this change?

lukebjerring:
No, the two warnings is not related to this PR (as you can tell by the touched files).
Introduced here: https://github.com/web-platform-tests/results-collection/commit/d3f336c9c2ccc56d8d728be4a78e58235c503f52

webapp/components/wpt-results.html
```diff
@@ -126,7 +126,7 @@
           <template is="dom-repeat" items="{{displayedNodes}}" as="node">
             <tr>
               <td>
-                <path-part path="{{ node.path }}" is-dir="{{ node.isDir }}" navigate="{{ bindNavigate() }}"></path-part>
+                <path-part prefix="/results" path="{{ node.path }}" is-dir="{{ node.isDir }}" navigate="{{ bindNavigate() }}"></path-part>
```
foolip: Documented as "Domain path-prefix, e.g. '/interop/'", so this already existed, then :)

webapp/components/wpt-results.html
```diff
@@ -151,6 +151,10 @@
   <script>
     /* global TestRunsBase */
     class WPTResults extends TestRunsBase {
+      static get UrlToResultsPath() {
+        return location => location.pathname.replace(/\/results(.+)(\/)?$/, '$1');
```
foolip: This returns a functions, so why it worked puzzled me under I saw the `get` keyword: getter returns a functions which is invoked.

Does it need to be a getter, or can it just be a static function? That'd be more obvious.

Some possibly unintentional things about the regex:
* the `(\/)` group won't be used AFAICT. If the input ends with a slash, that slash will stell end up in the first group because of greedy matching. And the second group isn't referenced as $2, so need not be a group.
* Not sure if it's an invariant that `location` begins with "/results/" if this code is run, but this will map "/resultsabc/def" to "abc/def". If that's a bug, looks like `/\/results(\/.+)?$/` followed by slash normalization might be the way? (I do fairly frequently end up with double slashes in wpt.fyi URLs when pasting file paths into them, and it doesn't work, so normalizing that *somewhere* would be great.)

webapp/components/wpt-results.html
```diff
@@ -151,6 +151,10 @@
   <script>
     /* global TestRunsBase */
     class WPTResults extends TestRunsBase {
+      static get UrlToResultsPath() {
```
foolip: lowerCamelCase as in rest of file, or is there a convention here I'm missing?

webapp/test_results_handler.go
```diff
@@ -21,7 +22,12 @@ import (
 //
 // The browsers initially displayed to the user are defined in browsers.json.
 // The JSON property "initially_loaded" is what controls this.
-func testHandler(w http.ResponseWriter, r *http.Request) {
+func testResultsHandler(w http.ResponseWriter, r *http.Request) {
+	if strings.Index(r.URL.Path, "/results/") != 0 {
+		http.Redirect(w, r, fmt.Sprintf("/results%s", r.URL.Path), 308)
```
foolip: Never seen 308 Permanent Redirect before, but if it's cached like 301 Moved Permanently and unlike 302 Found that could be a problem. (https://whatwg.org/faq was once a 301 and in many people's caches, broke things a bit when we wanted to host the FAQ there.) We might want to put something on / later, right?

webapp/main.go
```diff
@@ -14,7 +14,8 @@ var templates = template.Must(template.ParseGlob("templates/*.html"))
 func init() {
 	// Test run results, viewed by browser (default view)
 	// For run results diff view, 'before' and 'after' params can be given.
-	http.HandleFunc("/", testHandler)
+	http.HandleFunc("/", testResultsHandler)
```
foolip: Would it make more sense to just have two handlers, if `strings.Index(r.URL.Path, "/results/") != 0` can only match the first case with an early return?

foolip:
@mdittmer, take a look if there's some I might have missed regarding the /interop/ paths? I only confirmed that https://lukebjerring-results-prefix-dot-wptdashboard.appspot.com/interop/ doesn't redirect to /results/interop/.

webapp/components/wpt-results.html
lukebjerring: Yes, planned in advance.

webapp/components/wpt-results.html
lukebjerring: Added some tests to cover the behaviour of the function.

As a result, the function can't be static (WCT can't do ECMA6, so far as I can tell).

webapp/components/wpt-results.html

webapp/test_results_handler.go
lukebjerring: > similar to 301 (Moved Permanently) but does not allows the request method to be changed from POST to GET .

Changed to 307.

webapp/main.go
lukebjerring: I don't think it matters much. More boilerplate in a function + docs than in an if-statement.

foolip: OK :)

lukebjerring:
It was actually me who added those paths, so I'll go ahead and approve non-clashyness.

PR #48, opened 2018-04-13
Simplify path collapse code, fix WPT navigate hook

lukebjerring:
Code was unnecessarily complex; In particular, the following was essentially just parts[0].path:

        let dirParts = [parts[0]];	
        let dirPath = dirParts[dirParts.length - 1].path;

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -80,7 +80,7 @@ found in the LICENSE file.
         class="query"
         placeholder="Search test files, like cors/allow-headers.htm">
       <div class="path">
-        <a href="/" on-click="navigate">WPT</a>
+        <a href="/" onclick="{{ bindNavigate() }}">WPT</a>
         <template is="dom-repeat" items="{{ splitPathIntoLinkedParts(path) }}" as="part">
           <span class="path-separator">/</span>
           <a href="{{ part.path }}" on-click="navigate">{{ part.name }}</a>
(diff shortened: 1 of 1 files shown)


webapp/components/wpt-results.html
```diff
@@ -243,7 +243,8 @@
         /* Recomputes the list of displayed directories and test files. */
         const displayedNodeMap = new Map();
         const currentPathParts = this.splitPathIntoLinkedParts(this.path);
-
+        const partCountIfTestFile = this.path == '/' ? 1 : currentPathParts.length + 1;
```
foolip: === was in removed code, == in added. I seem to remember an eslint setup, is it running and does it enforce any style? Should it?

webapp/components/wpt-results.html
```diff
@@ -243,7 +243,8 @@
...
+        const nextSubdirIndex = this.path == '/' ? 0 : currentPathParts.length;
```
foolip: This is always the above - 1. Incidental, or can one be defined in terms of the other?

webapp/components/wpt-results.html
```diff
@@ -80,7 +80,7 @@
         class="query"
         placeholder="Search test files, like cors/allow-headers.htm">
       <div class="path">
-        <a href="/" on-click="navigate">WPT</a>
+        <a href="/" onclick="{{ bindNavigate() }}">WPT</a>
```
foolip: This bit I'll admit to not understanding. Reviewing on phone, so I'll assume it works :)

Issue #5, opened 2018-04-10
Use as short version numbers as is relevant for each browser


lukebjerring:
_From @foolip on February 25, 2018 7:54_

The front page now says chrome 63.0 and firefox 58.0, should be just 63 and 58. While there are updates to the stable channel, at least in chrome the change is not in the second position, but in the 4th. Example: https://chromereleases.googleblog.com/2018/02/stable-channel-update-for-desktop_22.html

(We should probably store the exact version somewhere, but not display it.)

_Copied from original issue: w3c/wptdashboard#494_

lukebjerring:
_From @foolip on February 25, 2018 9:41_

For Safari I think we want to show the minor version, and for Edge not, and that's already not shown.

PR #50, opened 2018-04-13
Truncate browser version

jugglinmike:
This is intended to resolve gh-5 (previously https://github.com/web-platform-tests/results-collection/issues/494).

We recently extended the results collection project [to include multiple versions of the same browser](https://github.com/web-platform-tests/results-collection/commit/c8952d37985b44ebba85585f5436051cec425f4a), and part of that effort involved publishing the complete version identifier reported by the binary. @foolip was originally motivated by the string "chrome 63.0", and the change in results collection exacerbates the problem--the UI now reports "chrome 65.0.3325.181". Since this is due to my modification of the reported data, it seemed appropriate for me to follow up with a fix for the UI.

Before:

screenshot from 2018-04-13 17-20-07

After:

screenshot from 2018-04-13 19-02-29


webapp/test/test-run.html
```diff
@@ -46,6 +47,65 @@
             assert.isFalse(trf.isDiff('literally anything'));
           });
         });
+        suite('shortVersion', () => {
+          test('valid, major only', () => {
+            trf.set('testRun.browser_name', 'chrome');
+
+            trf.set('testRun.browser_version', '33');
+            assert.equal(trf.get('testRun.browser_version'), '33');
```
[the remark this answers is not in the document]
jugglinmike: Agree completely. This is a vestige of my learning how to use Polymer

jugglinmike:
The target itself has definitely flown the coop, but [in `master` today, a couple other recipes still reference it as a dependency](https://github.com/web-platform-tests/wpt.fyi/blob/e0f26f25de958a3dce2ab4ff85c3146636c73bc3/Makefile#L36-L42). The first commit on this branch just removes those references.

lukebjerring:
Can you fix up the WCT tests?

webapp/components/test-run.html
```diff
@@ -49,7 +49,7 @@
 
     <div>
       <div><img src="/static/{{testRun.browser_name}}_64x64.png" /></div>
-      <div>{{displayName(testRun.browser_name)}} {{testRun.browser_version}}</div>
+      <div>{{displayName(testRun.browser_name)}} {{shortVersion()}}</div>
```
lukebjerring: The function `shortVersion` should probably depend on `testRun.browser_version` as a param so that if the `test-run` element is re-used (currently not the case) then the function's output value won't get stale.

webapp/components/test-run.html
```diff
@@ -59,6 +59,11 @@
   </template>
 
   <script>
+    const versionPatterns = {
```
lukebjerring: From https://eslint.org/docs/developer-guide/code-conventions

> Variables that act as constants (values that won’t be changed) should be formatted using all uppercase letters with words separated by a single underscore.

webapp/components/test-run.html
```diff
@@ -104,6 +116,9 @@
         }
 
         return match[1];
+
```
jugglinmike: Looks like we missed a `}` while merging `master` back into this feature branch. I'll rebase and resolve this

webapp/components/test-run.html
jugglinmike: It's surprising that the ESLint project would recommend that without offering a linting rule to enforce it.

webapp/components/test-run.html
jugglinmike: Shows what I know about Polymer. I've updated it, also explicitly depending on `browser_name` so that rendering of the minor version likewise respects changes.

jugglinmike:
Thanks for the review, Luke! I've rebased this branch and made the original version available here:

https://github.com/web-platform-tests/wpt.fyi/compare/master...bocoup:truncate-browser-version-orig

jugglinmike:
I don't believe so, but I also don't see a merge conflict. Can you say more about what's going wrong?

jugglinmike:
That's intentional. It's technically unrelated to the feature being implemented, but since it was necessary for me to validate my changes, I included it in this pull request, though it is a distinct commit:

https://github.com/web-platform-tests/wpt.fyi/pull/50/commits/b5404661d613dd8d7aaa8101d325f89e19f708de

If you would prefer, I can open a dedicated pull request with that fix, and then we could merge that and remove the commit from this pull request. Let me know!

lukebjerring:
Split it into a separate PR, just makes things cleaner in commit history (as we'll squash this one).
Also, fix the failing WCT.

P.S. if you are now experienced at setting up the requirements for WCT locally... docs would be lovely.

jugglinmike:
Because I rebased this branch on `master`, the pull request would be merged with git's "fast forward" strategy. From the perspective of the commit history, there would be no difference between that procedure and merging the two commits from two separate branches.

That said, it looks like you've resolved the issue in `master` just now, so that commit is no longer relevant. I've rebased this branch, removing the commit and squashing the merge commits that have been added since.

lukebjerring:
Ping - there's still a few failing WCT tests.

jugglinmike:
Looks like I was a bit careless with that last change. I've updated the tests to accommodate it. Thanks for the reminder!

PR #67, opened 2018-04-18
Use correct scheme in w3c-test.org link

zcorpan:
Tests should use http scheme by default and https scheme if the path has ".https.".

(This is rebased on current master. Please "squash and merge".)

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -218,7 +218,7 @@ found in the LICENSE file.
         },
         scheme: {
           type: String,
-          value: window.location.pathname.indexOf('.https.') !== -1 ? 'https' : 'http'
+          computed: 'computeTestScheme(path)'
         },
         path: {
           type: String,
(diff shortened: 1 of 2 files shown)


webapp/components/interop.html
```diff
@@ -324,6 +328,10 @@ <h2>Interoperability</h2>
       stateFunc(state, '', `/interop${path}?q=${encodedQuery}`);
     }
 
+    computeTestScheme(path) {
+      return path.indexOf('.https.') !== -1 ? 'https' : 'http';
```
foolip: A link to the logic in wpt run which this needs to be in sync with would be nice. I'd be surprised if there are any mismatches even if the rule isn't exactly the same, so not asking to centralize the definition or anything :)

zcorpan: Added.

lukebjerring:
Deployed to https://web-platform-tests-zcorpan-https-dot-here-lift-this.appspot.com

zcorpan:
Seems to do what it should there :)

https://web-platform-tests-zcorpan-https-dot-wptdashboard.appspot.com/WebCryptoAPI/idlharness.https.html links to https
https://web-platform-tests-zcorpan-https-dot-wptdashboard.appspot.com/2dcontext/building-paths/canvas_complexshapes_arcto_001.htm links to http

PR #72, opened 2018-04-19
Serve complete run by default

lukebjerring:
Replacement for https://github.com/web-platform-tests/results-collection/pull/322/files
For the [Q1 KR](https://docs.google.com/document/d/1ojlLDiRruI4e95Z7JQ10JQPRMiQHd_40rBkYCL6IJqw/edit#heading=h.2nhv4te6hv4x)

Don't merge this until the runs are consistently complete.

## Description
Changes the run fetch to grab a complete run by default.

## Review Information
Running at https://web-platform-tests-complete-runs-dot-wptdashboard.appspot.com


lukebjerring:
Running at https://complete-runs-dot-wptdashboard.appspot.com

jugglinmike:
@mdittmer Thanks for asking! By requiring SHA alignment, we will not be displaying the most recent results for Chrome and Firefox 75% of the time. That's because we collect results from those browsers every 6 hours, but we only collect from Edge and Safari once every 2 days (I wrote "4 days" above, but that was a typo--sorry about that). I assumed this was undesirable, but it might just reflect a different perspective on how the site should be used.

lukebjerring:
@foolip - can you weigh in here on the lesser of 2 evils:
 - 1. We compare more-recent runs with different SHAs
 - 2. We compare less-recent runs with same SHA

foolip:
I would say that there's some tipping point for "freshness" where it starts to make sense to use the same SHAs even if there are more recent runs for some browsers. I think my tipping point is around 24 hours, i.e., if freshness (latency) will never go (much) beyond 24 hours by using aligned SHAs, then we should.

Related question: if we use aligned runs by default, what does the UI look like to see the most recent, if any? Looking at https://wpt.fyi/test-runs kinda does the trick, but it doesn't allow comparing two runs of different shas.

foolip:
And although it's obvious, it's a liability that this kind of view is limited by the slowest and least reliable browser. Those are the most important to improve, of course, but if there's a weeks-long problem we would ideally have an automatic warning shown, or something.

lukebjerring:
As discussed offline today, we're now seeing daily runs from Safari & Edge, which is frequent enough to merge this in!

foolip:
A wonderful milestone, thank you @jugglinmike and @lukebjerring!!

jugglinmike:
Hooray!

foolip:
https://wpt.fyi/results/ is still showing results from different runs. Is that because it hasn't been deployed since this change, or isn't it working as intended?

lukebjerring:
Hasn't been deployed. See https://staging.wpt.fyi for master state.

PR #77, opened 2018-04-23
Add HSTS header to every request

### LINKED ISSUE #58: Redirect http://wpt.fyi/ to https://wpt.fyi/ and use HSTS
Moved from https://github.com/web-platform-tests/results-collection/issues/128.

This would depend on builds.wpt.fyi using HTTPS, or moving it to another domain.

jeffcarp:
This adds an HSTS header to every request that expires after 1 year.
It also applies the rule to all subdomains (we should check if this is
feasible) and opts-in to the HSTS preload list.

Resolves #58

### DIFF
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -54,7 +54,7 @@ func init() {
 
 func wrapHSTS(h http.HandlerFunc) http.HandlerFunc {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		value := "max-age=31536000; includeSubDomains; preload"
+		value := "max-age=31536000; preload"
 		w.Header().Add("Strict-Transport-Security", value)
 		h(w, r)
 	})
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -33,18 +33,6 @@ func init() {
 
 		// Admin-only manual results upload.
 		"/admin/results/upload": adminUploadHandler,
-
-		// API endpoint for diff of two test run summary JSON blobs.
-		"/api/diff": apiDiffHandler,
-
-		// API endpoint for listing all test runs for a given SHA.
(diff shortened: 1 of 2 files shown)


jeffcarp:
IIRC this can't land until the subdomain situation is worked out (either TLS is added to http://builds.wpt.fyi or `includeSubDomains` is removed from this PR)

foolip:
I guess let's just not use `includeSubDomains` for now.

jeffcarp:
Rebased & removed `includeSubDomains`

Hexcles:
Given that proper TLS is required by #147 . We should enable HSTS.

What's the blocker of this PR? Does it just need to be resurrected & rebased?

jeffcarp:
No blockers I know of. I can rebase by EOW.

jeffcarp:
Rebased yesterday but looks like it's not through the woods yet. Will take another look as soon as I can.

jeffcarp:
Rebased again but Travis is still failing. I won't have time to dig into the error messages for another 1.5-2 weeks.

Hexcles:
Hmm, looks like after @mdittmer 's recent Docker cleanup, Travis doesn't know how to fold gcloud SDK installation logs, which makes it very hard to find out what exactly goes wrong. Perhaps we need to add the special markups to teach Travis to fold properly.

More directly related to this change: we now have another package that also installs a bunch of routes in the `init` function (`api/routes.go`). Do we need to repeat the same setup there (in which case we should probably just export `WrapHSTS`)? Or are you aware of any cleaner ways?

lukebjerring:
I'll clone this change and take it off your hands, Jeff. Thanks for authoring!

jeffcarp:
Thanks so much for taking this Luke!

Issue #60, opened 2018-04-17
Expose results for experimental browsers in a way that is easily linked


foolip:
This is the frontend part of https://github.com/web-platform-tests/results-collection/issues/146.

@lukebjerring has suggested we could use a label system to be able to mix-and-match. Anything that will allow us to have one link for Chrome+Edge+Firefox+Safari stable, and another for their most recent pre-stable builds, would do the job.

foolip:
Issues for getting experimental results:
* [Chrome + Firefox](https://github.com/w3c/web-platform-tests/pull/9226) on Taskcluster
* [Chrome Dev](https://github.com/web-platform-tests/results-collection/issues/388)
* [Edge Insider Preview](https://github.com/web-platform-tests/results-collection/issues/109)
* [Firefox Nightly](https://github.com/web-platform-tests/results-collection/issues/521)
* [Safari Technology Preview](https://github.com/web-platform-tests/results-collection/issues/112)

jugglinmike:
Results for experimental builds of Chrome and Firefox are now available in the same format/location as the results for stable builds.

foolip:
@jugglinmike, are the unsplit wptreport JSON files also preserved somewhere?

jugglinmike:
Nope. We can start doing this if you like, but storing the complete WPT report is part of @Hexcles' [proposal for the results receiver](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork). Unless there is some specific value in the unsplit files themselves (and depending on how much time Robert will need), it might be wise to wait for the new results receiver.

foolip:
Nah, if you don't happen to have it then let's treat it like all other results and transition by re-merging them.

lukebjerring:
I'm going to do this in a few steps.
First up, I'll add `?label=unstable` but spoof it under the hood to change the browser names to `[name]-experimental` during the query, and we'll set that sit in staging until `results-receiver` progresses.

foolip:
@orphis asked me about experimental runs on wpt.fyi today and I pointed him at https://wpt.fyi/results/?label=experimental.

@lukebjerring, is all that remains of this issue to make experimental runs more discoverable?

lukebjerring:
Closing this as a dupe of #379, which will be fixed via https://github.com/web-platform-tests/wpt.fyi/issues/535

foolip:
[...]

PR #82, opened 2018-04-23
Fake-support the label=experimental param

lukebjerring:
Working towards https://github.com/web-platform-tests/wpt.fyi/issues/60

Spoofs the `?label=unstable` functionality support by querying for `[browser]-experimental` during query execution. This is to ensure that when we develop real labels, the UI doesn't need to change behaviour (just the datastore query).

Notes:
- `?labels=A,B,C` and `?label=A&label=B` forms of the param are both supported.
- Param is supported by both the `/api/runs` and the `/results` endpoints (and bubbles through).

CC @jugglinmike who may be happy to be able to view the experimental runs in the UI.

### DIFF
--- a/webapp/api_handlers.go
+++ b/webapp/api_handlers.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 	"time"
 
 	models "github.com/web-platform-tests/wpt.fyi/shared"
@@ -48,6 +49,9 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	labels := ParseLabelsParam(r)
+	unstable := labels != nil && labels.Contains("unstable")
+
 	var testRuns []models.TestRun
 	var limit int
 	if limit, err = ParseMaxCountParam(r); err != nil {
@@ -61,6 +65,9 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 
 	for _, browserName := range browserNames {
 		var testRunResults []models.TestRun
+		if unstable {
+			browserName = browserName + "-experimental"
+		}
 		query := baseQuery.Filter("BrowserName =", browserName)
 		if runSHA != "" && runSHA != "latest" {
 			query = query.Filter("Revision =", runSHA)
@@ -68,6 +75,11 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		if _, err := query.GetAll(ctx, &testRunResults); err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
+		}
+		if unstable {
+			for i := range testRunResults {
+				testRunResults[i].BrowserName = strings.Replace(testRunResults[i].BrowserName, "-experimental", "", 1)
+			}
 		}
 		testRuns = append(testRuns, testRunResults...)
 	}
--- a/webapp/api_handlers.go
+++ b/webapp/api_handlers.go
@@ -20,6 +20,8 @@ import (
 	"google.golang.org/appengine/datastore"
 )
 
+const experimentalLabel = `experimental`
+
 // apiTestRunsHandler is responsible for emitting test-run JSON for all the runs at a given SHA.
 //
 // URL Params:
@@ -50,7 +52,7 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	labels := ParseLabelsParam(r)
-	unstable := labels != nil && labels.Contains("unstable")
+	experimentalBrowsers := labels != nil && labels.Contains(experimentalLabel)
 
 	var testRuns []models.TestRun
 	var limit int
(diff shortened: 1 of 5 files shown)


lukebjerring:
Oh, turns out (for obvious reasons) forks can't access the Travis secrets, so this didn't autodeploy.

Manually deployed to https://lukebjerring-unstable-dot-wptdashboard.appspot.com/results/?label=unstable

jugglinmike:
You're right; I *am* happy to see this :)

lukebjerring:
@Hexcles - re: https://github.com/web-platform-tests/wpt.fyi/pull/83 - rebased + pushed.

webapp/api_handlers.go
```diff
@@ -48,6 +49,9 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	labels := ParseLabelsParam(r)
```
mdittmer: I think we should have comments somewhere discussing the binding between `unstable` in one context and `-experimental` in another.

webapp/params.go
```diff
@@ -190,3 +190,30 @@ func ParsePathsParam(r *http.Request) (paths mapset.Set) {
 	}
 	return paths
 }
+
+// ParseLabelsParam returns a set list of test-run labels to include, or nil if
+// no labels are provided.
+func ParseLabelsParam(r *http.Request) (labels mapset.Set) {
+	labelParams := r.URL.Query()["label"]
```
mdittmer: Do we do this "singular or plural" work elsewhere? Just thinking about the convenience/complexity tradeoff here, but that's moot if we have established a pattern already.

webapp/api_handlers.go
```diff
@@ -48,6 +49,9 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
...
+	unstable := labels != nil && labels.Contains("unstable")
```
foolip: I'm reminded of https://github.com/w3c/web-platform-tests/issues/5956 and it doesn't seem far-fetched that it might be mistaken to mean flakiness on wpt.fyi as well, or that the label name makes its way back to wpt PRs via https://github.com/w3c/web-platform-tests/issues/10503 happens.

But the only counter-proposal I can come up with which accurately describes Chrome Dev/Canary, Edge Insider Preview, Firefox Nightly and Safari Technology Preview is "latest". But if you don't think that's any better, then I surrender to "unstable" :)

webapp/test_run_handler.go
```diff
@@ -31,6 +31,13 @@ func handleTestRunGet(w http.ResponseWriter, r *http.Request) {
 	}
 	sourceURL := fmt.Sprintf(`/api/runs?max-count=%d`, maxCount)
 
+	labels := ParseLabelsParam(r)
+	if labels != nil {
+		for label := range labels.Iterator().C {
+			sourceURL = sourceURL + "&label=" + label.(string)
```
mdittmer: Is `.(string)` a type assertion? Could it fail?

webapp/params.go
```diff
@@ -190,3 +190,30 @@ func ParsePathsParam(r *http.Request) (paths mapset.Set) {
...
+	labelsParam := r.URL.Query().Get("labels")
```
foolip: Can you exercise labels=foo,bar in tests as well? And label=foo&label=bar if that is intended to work, which it seems like it is. (Not testing labels=foo,bar&labels=baz seems fine.)

webapp/test_results_handler.go
```diff
@@ -119,7 +119,13 @@ func getTestRunsAndSources(r *http.Request, runSHA string) (testRunSources []str
 			testRunSources = append(testRunSources, fmt.Sprintf(singleRunURL, afterSpec.Revision, afterSpec.Platform))
 		}
 	} else {
-		const sourceURL = `/api/runs?sha=%s`
+		var sourceURL = `/api/runs?sha=%s`
+		labels := ParseLabelsParam(r)
+		if labels != nil {
+			for label := range labels.Iterator().C {
+				sourceURL = sourceURL + "&label=" + label.(string)
```
mdittmer: ditto (type assertion)

webapp/api_handlers.go
lukebjerring: @jugglinmike - can you weigh in on your preference or `stable` or `experimental` as the label? You chose `-experimental` as the browser suffix, so perhaps you prefer that? I'm not fussed - unstable does, as @foolip points out, kinda sound test-related.

webapp/params.go
lukebjerring: Yes, this was a copy-paste of some code above.
I'll do you one better and refactor it.

webapp/test_run_handler.go
lukebjerring: Yes it's an assertion; no it can't fail unless the set items aren't strings (which, they are).

webapp/params.go
lukebjerring: Done.

webapp/api_handlers.go
jugglinmike: Yup, I agree with @foolip's advice, and since he made the improvement to WPT, using the same convention here will promote consistency in terminology.

lukebjerring: OK - `unstable` => `experimental`

PR #83, opened 2018-04-23
Don't run deploy on forks

lukebjerring:
Forks can't access Travis CI secrets (by design), so the creds needed for a staging deploy are not available. Bypass the task.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,8 +90,11 @@ install:
     fi
 
 script:
-  - | # Deploy PR to staging environment
-    if [ "${MAKE_DEPLOY_TARGET}" != "" ]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -91,7 +91,7 @@ install:
 
 script:
   - | # Deploy PR to staging environment
```
Hexcles: Nit: a comment would be great. (e.g. "... only when Travis secrets are available")

lukebjerring:
Looks like the [breakages](https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/370396674) might be unrelated, potentially caused by 
https://github.com/golang/lint/commit/c363707d68842c977f911634e06201907b60ce58

Issue #71, opened 2018-04-18
Log or surface invalid params


jugglinmike:
*From @lukebjerring on Nov 21, 2017 2:12*

We currently silently fall back to 'latest' - would be good to at least log the issue.

*Copied from original issue: https://github.com/web-platform-tests/results-collection/issues/283*

jugglinmike:
*From @lukebjerring on Nov 21, 2017 2:14*

Not just ?sha, but ?browser too, so made the issue more general.

mdittmer:
FWIW I would prefer an error response on invalid params.

//Mark

PR #86, opened 2018-04-24
Throw errors for invalid params

lukebjerring:
## Description
Changes the silent-fallback behaviour of `?browser=invalid-name` and `?sha=invalid-sha` to throw 400's instead of falling back to default values.

Resolves https://github.com/web-platform-tests/wpt.fyi/issues/71

Running at https://lukebjerring-error-bad-params-dot-wptdashboard.appspot.com
Examples:
- https://lukebjerring-error-bad-params-dot-wptdashboard.appspot.com/results?sha=not-a-sha
- https://lukebjerring-error-bad-params-dot-wptdashboard.appspot.com/api/runs?browser=not-a-browser

### DIFF
--- a/webapp/params.go
+++ b/webapp/params.go
@@ -39,7 +39,10 @@ func ParseSHAParam(r *http.Request) (runSHA string, err error) {
 	}
 
 	runParam := params.Get("sha")
-	if SHARegex.MatchString(runParam) {
+	if runParam != "" && runParam != "latest" {
+		if !SHARegex.MatchString(runParam) {
+			return "", fmt.Errorf("Invalid sha param value: %s", runParam)
+		}
 		runSHA = runParam[:10]
 	}
 	return runSHA, err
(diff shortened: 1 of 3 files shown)


webapp/params.go
```diff
@@ -77,10 +80,14 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
 	// Otherwise filter to valid browser names.
 	for i := 0; i < len(browsers); {
 		if !IsBrowserName(browsers[i]) {
-			// 'Remove' browser by switching to end and cropping.
-			browsers[len(browsers)-1], browsers[i] = browsers[i], browsers[len(browsers)-1]
-			browsers = browsers[:len(browsers)-1]
-			continue
+			if browsers[i] == "" {
+				// 'Remove' empty browser by switching to end and cropping.
```
mdittmer: Vein curiosity: Is this a golang idiom that yields better performance (because you're not splicing the slice)?

lukebjerring: It's not for efficiency; you just can't easily remove an item... I think `append(browsers[0:i], ...browsers[i+1:])` would actually do the same in one line, but definitely less efficiently.

Optimizing for a list of size 4: 👍

lukebjerring:
Yeah, IMO if users wanna throw ?foo=bar for `foo` that we don't do anything with, let 'em.

Issue #6, opened 2018-04-10
Use canonical case for browser and OS names.


lukebjerring:
_From @foolip on February 25, 2018 7:49_

For example, instead of "safari 11.0 macos 10.12" it should say "Safari 11.0 macOS 10.12".

_Copied from original issue: w3c/wptdashboard#493_

PR #87, opened 2018-04-24
Add display names for platforms

lukebjerring:
## Description
Add a map of the dataset names to Canonical Case names, and use them in the UI.

Resolves https://github.com/web-platform-tests/wpt.fyi/issues/6

## Review Information
Running at https://lukebjerring-camel-case-names-dot-wptdashboard.appspot.com

### DIFF
--- a/webapp/components/test-run.html
+++ b/webapp/components/test-run.html
@@ -66,6 +66,7 @@ See models.go for more details.
       m.set('edge', 'Edge');
       m.set('linux', 'Linux');
       m.set('macos', 'macOS');
+      m.set('safari', 'Safari');
       m.set('windows', 'Windows');
       return m;
     })();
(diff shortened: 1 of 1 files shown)


webapp/components/test-run.html
```diff
@@ -49,16 +49,27 @@
...
     </div>
   </template>
 
   <script>
+    const DISPLAY_NAMES = (() => {
+      let m = new Map();
+      m.set('chrome', 'Chrome');
+      m.set('firefox', 'Firefox');
+      m.set('edge', 'Edge');
+      m.set('linux', 'Linux');
```
foolip: Is the omission of Safari deliberate? If so can you add a comment?

lukebjerring: Nope, an oversight. Added.

PR #95, opened 2018-04-26
Confirm push if it's web-platform-tests/master

lukebjerring:
## Description
Confirms intent to push the branch if the branch is `web-platform-tests/master`.
Generally, we squash and merge through the GitHub UI, so this is an attempt to prevent accidental (potentially unreviewed) pushes.


lukebjerring:
Running at https://confirm-push-master-dot-wptdashboard.appspot.com

git/hooks/pre-push
```diff
@@ -53,4 +53,13 @@ else
   error "Pre-push exit status: ${FINAL_STATUS}"
 fi
 
+# Confirm if pushing to web-platform-tests/master
+BRANCH_NAME=${BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"}
+USER="$(git remote -v get-url origin | sed -E 's#(https?:\/\/|git@)github.com(\/|:)##' | sed 's#/.*$##')"
```
Hexcles: It's not very robust to get the branch and user (org) this way. For example, I usually use `upstream` as the remote name of the original repo, while `origin` points to my fork. Such setup is commonly seen if the local checkout is cloned from a fork first and the upstream is added later. And `git push upstream master` can't be caught by this check.

In fact, git passes the URL of the remote as the second argument (`$2`) to pre-hook scripts. See [doc](https://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) and [example](https://github.com/git/git/blob/master/templates/hooks--pre-push.sample).

The more bullet-proof way to get `$BRANCH_NAME` is to parse the stdin of pre-hook scripts, because the local branch name isn't necessarily the same as the target push ref (e.g. `git push origin local_foo:master`), and there can be multiple target refs. Again, follow the [example](https://github.com/git/git/blob/master/templates/hooks--pre-push.sample) (I think we only need to check `$remote_ref` inside the while loop).

lukebjerring:
Yep, intentional, and (IMHO) useful.

Issue #49, opened 2018-04-13
Use full SHAs as the identifier


Hexcles:
I'm quite nervous about the current practice of using only 10 characters of the full SHAs as identifiers.

To be future-proof, I propose to gradually switch to full SHAs, at least internally (GCS paths, Datastore entities, etc.), and make all SHA lookup code capable of handling both shortened and full SHAs (and perhaps giving a warning when multiple results are found in a lookup).

lukebjerring:
Sure. Note that https://github.com/web-platform-tests/wpt.fyi/pull/46 was simply to prevent a broken UI for the current behaviour - not advocating for short-SHA-only.

lukebjerring:
https://github.com/web-platform-tests/wpt.fyi/pull/164 added `FullRevisionHash`

lukebjerring:
This is fixed - `?sha is a prefix of the FullRevisionHash`.

PR #101, opened 2018-04-27
Add /api/manifest endpoint

lukebjerring:
## Description
WIP - request for feedback

Adds an endpoint for getting a manifest for a given SHA.
To achieve this, it does:
- GitHub search for the PR associated with the SHA
- GitHub fetch for the release by tag-name `merge_pr_[PR number]`
- GitHub download of the release's asset with the name `MANIFEST-[sha].json.gz`

@gsnedders @jgraham if there is a better way to do this (without a local git clone) I'd be happy to know about it

@tabatkins - Haven't done `latest` behaviour yet, needs to be given a SHA param, but it's running here if you want to play around and point out any major flaws. AppEngine seems to behave nicely if you allow gzip encoding, but defaults to raw JSON @ ~20MB.

https://manifest-api-dot-wptdashboard.appspot.com/api/manifest?sha=ae34fb2b39f8b2cf9c3989a566ab938cde45dec0

## Context
There are some UI features (e.g. linking to w3c-test.org) which require manifest context, so we'd like to eventually cache manifests by SHA and support lookup of manifest info for a single test file.

### DIFF
--- a/util/populate_dev_data.go
+++ b/util/populate_dev_data.go
@@ -41,7 +41,7 @@ func main() {
 		log.Fatal(err)
 	}
 
-	emptySecretToken := []interface{}{&base.Token{}}
+	emptySecretToken := &base.Token{}
 	staticDataTime, _ := time.Parse(time.RFC3339, "2017-10-18T00:00:00Z")
 
 	// Follow pattern established in run/*.py data collection code.
@@ -134,15 +134,17 @@ func main() {
 		},
 	}
 
-	tokenKindName := "Token"
 	testRunKindName := "TestRun"
 	passRateMetadataKindName := metrics.GetDatastoreKindName(
 		metrics.PassRateMetadata{})
 	failuresMetadataKindName := metrics.GetDatastoreKindName(
 		metrics.FailuresMetadata{})
 
+	log.Print("Adding local (empty) secrets...")
+	addSecretToken(ctx, "upload-token", emptySecretToken)
+	addSecretToken(ctx, "github-api-token", emptySecretToken)
+
 	log.Print("Adding local mock data (static/)...")
-	addData(ctx, tokenKindName, emptySecretToken)
 	addData(ctx, testRunKindName, staticTestRunMetadata)
 	addData(ctx, passRateMetadataKindName, staticPassRateMetadata)
 	addData(ctx, failuresMetadataKindName, staticFailuresMetadata)
@@ -162,6 +164,14 @@ func main() {
 		latestProductionTestRunMetadata[i] = &prodTestRuns[i]
 	}
 	addData(ctx, testRunKindName, latestProductionTestRunMetadata)
+}
+
+func addSecretToken(ctx context.Context, id string, data interface{}) {
+	key := datastore.NewKey(ctx, "Token", id, 0, nil)
+	if _, err := datastore.Put(ctx, key, data); err != nil {
+		log.Fatalf("Failed to add %s secret: %s", id, err.Error())
+	}
+	log.Printf("Added %s secret", id)
 }
 
 func addData(ctx context.Context, kindName string, data []interface{}) {
--- a/webapp/api_handlers.go
+++ b/webapp/api_handlers.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
+	"google.golang.org/appengine/urlfetch"
 )
 
 const experimentalLabel = `experimental`
@@ -367,4 +368,46 @@ func handleAPIDiffPost(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	w.Write(bytes)
+}
+
+func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
+	sha, err := ParseSHAParam(r)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := appengine.NewContext(r)
+	if manifest, err := getManifestForSHA(ctx, sha); err != nil {
+		panic(manifest)
+	}
+}
+
+func getManifestForSHA(ctx context.Context, sha string) (manifest interface{}, err error) {
+	const githubSearch = `https://api.github.com/search/issues?SHA=%s+user:w3c+repo:web-platform-tests`
+	url := fmt.Sprintf(githubSearch, sha)
+	client := urlfetch.Client(ctx)
+	var resp *http.Response
+	if resp, err = client.Get(url); err != nil {
+		panic("hi")
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var body []byte
+	if body, err = ioutil.ReadAll(resp.Body); err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != 200 {
+		return nil, fmt.Errorf("%s returned HTTP status %d:\n%s", url, resp.StatusCode, string(body))
+	}
+	panic(string(body))
+
+	type Issue struct {
+		Number int `json:"number"`
+	}
+	var issues []Issue
+	if err = json.Unmarshal(body, &issues); err != nil {
+		return nil, err
+	}
+	panic(len(issues))
 }
--- a/webapp/api_handlers.go
+++ b/webapp/api_handlers.go
@@ -20,7 +20,6 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
-	"google.golang.org/appengine/urlfetch"
 )
 
 const experimentalLabel = `experimental`
@@ -388,6 +387,14 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func gitHubSHASearchURL(sha string) string {
+	return fmt.Sprintf(`https://api.github.com/search/issues?q=SHA:%s+user:w3c+repo:web-platform-tests`, sha)
+}
+
+func gitHubReleaseURL(tag string) string {
+	return fmt.Sprintf(`https://api.github.com/repos/w3c/web-platform-tests/releases/tags/%s`, tag)
+}
+
 type gitHubClient interface {
 	fetch(url string) ([]byte, error)
 }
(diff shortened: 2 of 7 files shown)


lukebjerring:
Running at https://manifest-api-dot-wptdashboard.appspot.com

webapp/api_handlers.go
```diff
@@ -374,3 +377,122 @@ func handleAPIDiffPost(w http.ResponseWriter, r *http.Request) {
...
+
+	var queryResults map[string]*json.RawMessage
+	if err = json.Unmarshal(body, &queryResults); err != nil {
+		return nil, err
+	}
+	var issues []map[string]*json.RawMessage
+	if err = json.Unmarshal(*queryResults["items"], &issues); err != nil {
+		return nil, err
+	}
+	if len(issues) < 1 {
```
Hexcles: I'd like to assert len(issues)==1 and perhaps log all anomalies.

webapp/api_handlers.go
```diff
@@ -18,6 +20,7 @@ import (
 	"golang.org/x/net/context"
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
+	"google.golang.org/appengine/urlfetch"
```
Hexcles: FWIW, if we decide to go ahead with #107 , we won't be using `urlfetch` anymore. `urlfetch` isn't used extensively in this PR yet, but I still suggest creating a wrapper for `urlfetch` so that we can switch to generic network I/O later.

webapp/params.go
```diff
@@ -28,9 +28,19 @@ const MaxCountMinValue = 1
 // SHARegex is a regex for SHA[0:10] slice of a git hash.
 var SHARegex = regexp.MustCompile("[0-9a-fA-F]{10,40}")
 
-// ParseSHAParam parses and validates the 'sha' param for the request.
-// It returns "latest" by default (and in error cases).
+// ParseSHAParam parses and validates the 'sha' param for the request,
+// cropping it to 10 chars. It returns "latest" by default. (and in error cases).
```
foolip: Why is the cropping introduced? It looks like the longest ambiguous shas in wpt are 7 characters long (93d383f, d0b03af, e54fe8f) and it's fairly unlikely we will have a collision, but using full hashes everywhere internally and only accepting shorter ones as input seems like a good idea. But maybe this is necessary because 10-char hashes are parts of the filenames?

webapp/api_handlers.go
```diff
@@ -374,3 +377,122 @@ func handleAPIDiffPost(w http.ResponseWriter, r *http.Request) {
...
+func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
+	sha, err := ParseSHAParamFull(r)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	ctx := appengine.NewContext(r)
+	if manifest, err := getManifestForSHA(ctx, sha); err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
+	} else if manifest == nil {
```
foolip: Letting both `manifest` and `err` be nil seems like an odd API, should `getManifestForSHA` return an error at its final line instead?

webapp/api_handlers.go
foolip: Assertions should be for things that the programmer thinks is impossible, but logging which requests don't find a manifest SGTM. Except, do we have any server side logs at all currently? The less we log the less we need to think about PII and that kind of thing.

webapp/api_handlers.go
```diff
@@ -374,3 +377,122 @@ func handleAPIDiffPost(w http.ResponseWriter, r *http.Request) {
...
+		return nil, err
+	}
+	var assets []map[string]*json.RawMessage
+	if err = json.Unmarshal(*release["assets"], &assets); err != nil {
+		return nil, err
+	}
+	if len(assets) < 1 {
+		return nil, fmt.Errorf("No assets found for release %s", releaseTag)
+	}
+	// Get (and unzip) the asset with name "MANIFEST-{sha}.json.gz"
```
foolip: Do we need to unzip it, or could we set the transfer encoding and just pass along the data?

webapp/api_handlers.go
```diff
@@ -374,3 +377,122 @@ func handleAPIDiffPost(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusNotFound)
+	} else if manifest == nil {
+		http.NotFound(w, r)
+	} else {
+		w.Header().Add("content-type", "application/json")
+		w.Write(manifest)
+	}
+}
+
+func getManifestForSHA(ctx context.Context, sha string) (manifest []byte, err error) {
```
foolip: Can you add some integration tests for this API? Do we have any tests that just hit the API endpoints, as opposed to invoking `getManifestForSHA` as we'd do here?

And... what do we do about GitHub API reliability in such tests? Mock out the GitHub API?

foolip:
I tried this:
```bash
for tag in `git tag --list 'merge_pr_1*'`; do
  curl "https://manifest-api-dot-wptdashboard.appspot.com/api/manifest?sha=$(git rev-parse $tag)" > MANIFEST.$tag.json
done
```

That's 520 tags/files. 32 of them didn't return a manifest, which is a bit much for comfort. These are the ones:
```
merge_pr_10005
merge_pr_10014
merge_pr_10113
merge_pr_10158
merge_pr_10159
merge_pr_10164
merge_pr_10180
merge_pr_10191
merge_pr_10194
merge_pr_10202
merge_pr_10205
merge_pr_10215
merge_pr_10216
merge_pr_10218
merge_pr_10219
merge_pr_10222
merge_pr_10224
merge_pr_10235
merge_pr_10237
merge_pr_10238
merge_pr_10240
merge_pr_10247
merge_pr_10267
merge_pr_10268
merge_pr_10276
merge_pr_10278
merge_pr_10280
merge_pr_10286
merge_pr_10294
merge_pr_10445
merge_pr_10605
merge_pr_10663
```

What do we do with this unreliability? Backfill releases for existing tags? Put in place some kind of monitoring going forward?

webapp/api_handlers.go
Hexcles: AppEngine has built-in logging support. We can just use the standard logging facility provided by the language, and all the logs will go to Google Cloud Stackdriver which has a fancy UI for filtering.

webapp/params.go
Hexcles: I think Luke's aware of #49 , but many existing code doesn't take full SHAs so I assume this is a transient solution. At least this endpoint now takes a full SHA (despite chopping it internally), which is good.

webapp/api_handlers.go
Hexcles: Regarding PII, AppEngine logs requests already. We won't and can't add extra PII if we log some error messages.

webapp/api_handlers.go
Hexcles: We played with this a bit. The short answer is that we can't, unless (potentially) we use some dirty hacks.

Basically, GitHub responds a 302 for the download URL to a long AWS S3 URL that has a bunch of query parameters to tell S3 to force a browser download, which means S3 will disregard the `Accept-Encoding` in request headers, and always return the raw octet stream of the file (which happens to be a gz here).

If we absolutely want to to do it, I guess we can intercept 302 and modify the query parameters, but I guess we don't :)

lukebjerring:
Well, abstractly speaking, the API is an HTTP endpoint for finding the manifest for a SHA.

Whether or not there actually was one available is a reliability issue with a separate service (the service for tagging PRs). Please file an issue on the WPT repo so we can discuss there :)

util/populate_dev_data.go
```diff
@@ -125,15 +125,17 @@ func main() {
...
 	}
 
-	tokenKindName := "Token"
 	testRunKindName := "TestRun"
 	passRateMetadataKindName := metrics.GetDatastoreKindName(
 		metrics.PassRateMetadata{})
 	failuresMetadataKindName := metrics.GetDatastoreKindName(
 		metrics.FailuresMetadata{})
 
+	log.Print("Adding local (empty) secrets...")
```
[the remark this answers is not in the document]
lukebjerring: I didn't manage to exceed it in all of my trials, so I think I'll wait until I see a problem before I solve it.

webapp/params.go
lukebjerring: Note: This PR doesn't introduce the cropping, it bypasses it (since we want to use the full hash here).

See https://github.com/web-platform-tests/wpt.fyi/pull/46 for the temporary 'fix' as Hexcles describes.

webapp/api_handlers.go
lukebjerring: We want to unzip it anyway, once we get to dumping it in the cache and allowing path filtering, etc.

webapp/api_handlers.go
lukebjerring: Will save that pain for another day.

webapp/api_handlers.go
lukebjerring: Done.

webapp/api_handlers.go
lukebjerring: Done, somewhat. IMHO abstracting to a `client` and asserting that we fetch a URL is as far as we should bother testing, otherwise you're writing tests that really only test how well you write your mocks.

foolip:
> Whether or not there actually was one available is a reliability issue with a separate service (the service for tagging PRs). Please file an issue on the WPT repo so we can discuss there :)

Poking you on https://github.com/w3c/web-platform-tests/issues/10572

util/populate_dev_data.go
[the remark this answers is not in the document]
foolip: OK, perhaps the 403s one easily gets with a few fast reloads of https://foolip.github.io/ecosystem-infra-rotation/ are a different kind of problem that you don't get with just a single request, maybe this: https://developer.github.com/v3/#abuse-rate-limits

webapp/params.go
foolip: Oh... there was already a `runParam[:10]` before and this introduces a `ParseSHAParamFull`. Sorry, didn't finish reading before reacting :/

webapp/api_handlers.go
foolip: That makes sense, thanks!

webapp/api_handlers_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// 2) Correct asset picked when first asset is some other asset.
+	releasesJSON["assets"] = []object{
+		object{
+			"name":                 "Some other asset.txt",
+			"browser_download_url": "http://huh.com?",
+		},
+		releasesJSON["assets"].([]object)[0],
+	}
+	client.Responses[gitHubReleaseURL("merge_pr_123")] = unsafeMarshal(releasesJSON)
```
foolip: It wouldn't affect the test, but checked that https://github.com/w3c/web-platform-tests/pull/123 was merged very long ago so that we won't be creating a real tag called merge_pr_123 :)

PR #106, opened 2018-05-01
Add support for browser name labels

lukebjerring:
## Description
Tweaks + expands on https://github.com/web-platform-tests/wpt.fyi/pull/82 which added the `label=experimental` behaviour, by supporting `label=[browser]` to show the stable + experimental results next to each other.

Example (Chrome):
https://browser-labels-dot-wptdashboard.appspot.com/results/?label=chrome


lukebjerring:
Running at https://browser-labels-dot-wptdashboard.appspot.com

webapp/params.go
```diff
@@ -61,23 +61,15 @@ func ParseBrowserParam(r *http.Request) (browser string, err error) {
...
-	// If no params found, return the default.
-	var browserNames []string
-	if browserNames, err = GetBrowserNames(); err != nil {
-		return nil, err
-	}
-	if len(browsers) == 0 {
-		return browserNames, nil
-	}
-	// Otherwise filter to valid browser names.
+	// Otherwise validate browser names.
```
foolip: This comment doesn't make sense now that there's no return in the preceding if clause.

webapp/params.go
```diff
@@ -95,6 +87,53 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
...
+		}
+
+		browserLabel := ""
+		for _, name := range browserNames {
+			if !labels.Contains(name) {
+				continue
+			}
+			// If we already encountered a browser name, nothing is two browsers (return empty set).
+			if browserLabel != "" {
+				browsers = nil
```
foolip: Don't know if this is the case hit with https://browser-labels-dot-wptdashboard.appspot.com/results/?labels=edge,experimental but when no browsers remain it would be good if there were some kind of error, maybe pointing to the documentation of the labels param?

webapp/params.go
```diff
@@ -95,6 +87,53 @@ func ParseBrowsersParam(r *http.Request) (browsers []string, err error) {
...
+				break
+			}
+			browserLabel = name
+			browsers = []string{name}
+			// For a browser label (e.g. "chrome"), we also include experimental, unless we explicitly only
```
foolip: I think all of this logic works, but it does make it very easy/tempting to infer some labels from others, treat combinations of labels differently, and so on.

To avoid intentional or accidental things like that, would it make sense to instead compute a set of labels for each run/runner, and then filter those by labels in a label-name-agnostic way "here?" That does require the frontend to know about all the runs/runners that exist, which perhaps it doesn't, but wouldn't that make sense?

I ask that partially because I think there's a limitation in labels, namely that filtering to Edge Insider Preview + Chrome Stable won't work, and those may well be interesting to compare. Don't know what the fix is, but suspect it'll get weird if we try to add it to this structure.

webapp/params.go
lukebjerring: Done.

webapp/params.go
lukebjerring: Not sure I'd block this PR on documentation of an evolving fake-feature (fake in that we don't have a real label concept in the data model just yet).
Yes, when labels are added by @Hexcles collection endpoint, it'd be good to error and help the user as to what labels should be used.

webapp/params.go
lukebjerring: `TestRun` will eventually include a `Labels []string` concept, which the collection endpoint will leverage (things like browser, os, stable/experimental, and then manually flagged (bad/incomplete) etc.)

For the browser v browser use-case, there's a (repeated) `browser` param in the API, e.g. https://wpt.fyi/api/runs?browsers=chrome,firefox - I'll put together a PR to leverage that in (bubble it from) the wpt.fyi UI.

All that is to say, the hacky-looking logic here will die in a fire when _actual_ labels exist. But the URLs/usage will stay the same :)

Your comment, surmisable as roughly _should multiple labels be AND combined, or OR combined?_ is something worth considering. Or-wise searching is probably not something we need to worry about, though, since the labels can be thought of as 'groups' (stable browsers. chrome browsers. mobile browsers).

webapp/params.go
foolip: SGTM, let the first person to be confused for realz file an issue :)

webapp/params.go
foolip: Yes, that is roughly the shape of my question. Note that if there's a way to refer to a specific runner's latest run or a specific run, then this concern also goes poof, since what I'm really after is comparing Edge Insider Preview to Chrome Stable.

OK to let this be until real labels exist, and probably even then.

PR #108, opened 2018-05-01
Support ?diff param for /results

lukebjerring:
## Description
Adds support for (more simply) diffing in the UI when the query has 2 runs.

- Defer diff fetch from the Go code into to the UI
- Support ?diff param

When combined with https://github.com/web-platform-tests/wpt.fyi/pull/106, we can diff chrome and chrome-experimental with `?label=chrome&diff`

Example URL (diffing the 2 experimental browsers):
https://diff-param-dot-wptdashboard.appspot.com/results/?label=experimental&diff


lukebjerring:
Running at https://diff-param-dot-wptdashboard.appspot.com

webapp/components/wpt-results.html
```diff
@@ -227,13 +233,23 @@
           this.refreshDisplayedNodes();
         };
 
+        // Add a diff testRun spoof, if needed.
+        if (this.diff && this.testRuns && this.testRuns.length == 2) {
```
foolip: If there aren't exactly two runs and the diff parameter is used, can some kind of message be shown? It could be that a URL that previously worked no longer does because a new configuration matches the labels, and that'd be a bit confusing for someone who doesn't know how the diff param works.

webapp/components/wpt-results.html
```diff
@@ -227,13 +233,23 @@
           this.refreshDisplayedNodes();
         };
 
+        // Add a diff testRun spoof, if needed.
```
foolip: What's being spoofed here? Looks like it's just fetching the diff if it's needed?

A comment on the UX of this view. The diff column isn't shown at all until the data is loaded, which takes at least a second. Would it make sense to show at least the header first so that it's clear something is loading?

webapp/components/wpt-results.html
```diff
@@ -227,13 +233,23 @@
...
         // Fetch the results JSON for the TestRuns
+        await this.fetchResults(this.testRuns);
+        this.refreshDisplayedNodes();
+      }
+
+
+      async fetchResults(testRuns) {
         const testFileResults = await Promise.all(
-          this.testRuns.map(async json => this.fetchResults(json.results_url))
+          testRuns.map(async json => this.fetchResultsJSON(json.results_url))
```
foolip: Nit: the `async` here is a bit odd since there's no `await`ing within. It does mean that it'll work even if `fetchResultsJSON` doesn't return a promise, but I still wonder if this isn't a typo/copypasta.

webapp/components/wpt-results.html
```diff
@@ -242,16 +258,34 @@
...
-        this.refreshDisplayedNodes();
       }
 
-      async fetchResults(resultsURL) {
+      async fetchResultsJSON(resultsURL) {
         const response = await window.fetch(resultsURL);
-        const testFiles = response.ok ? await response.json() : {};
+        let testFiles = {};
+        try {
+          testFiles = response.ok ? await response.json() : {};
```
foolip: Sent https://github.com/web-platform-tests/wpt.fyi/pull/117 when I saw this line. Would it make sense to log non-OK statuses here too?

webapp/components/wpt-results.html
```diff
@@ -188,7 +188,13 @@
           displayedNodes: {
             type: Array,
             value: []
-          }
+          },
+          // Show a diff column (if there are 2 testRuns).
+          diff: Boolean,
+          filter: {
```
foolip: I tried to use the filter param as https://diff-param-dot-wptdashboard.appspot.com/results/?label=experimental&diff&filter=R but couldn't make it do anything. Is the intention that this PR should be enough to wire it up?

webapp/components/wpt-results.html
lukebjerring: Added a toast notification.

webapp/components/wpt-results.html
lukebjerring: filter is a param that's handled by `/api/diff` but not yet forwarded in the UI. Maybe file an issue if you think that's a feature you want!

webapp/components/wpt-results.html
lukebjerring: The `TestRun` object doesn't exist in the datastore, so we add it here. Maybe, but adding a nice async UI is probably overkill for first cut :)

webapp/components/wpt-results.html
lukebjerring: Added a notification.

webapp/components/wpt-results.html
lukebjerring: I'll leave it alone for this PR (separate PR to fix welcome :)

webapp/components/wpt-results.html
lukebjerring: Might be nice to retry or log, but not something I'd add this PR.

lukebjerring:
@foolip PTAL, specifically at 
https://diff-param-dot-wptdashboard.appspot.com/results/?diff

PR #111, opened 2018-05-02
Remove deprecated changes from deploy.sh

lukebjerring:
## Description
Removes some debugging flags that were added while trying to diagnose the deploy in Travis.
Also, fixes some bugs with `exit` on cancellation (was only working because of `set -e` and a nested exit)

Note that `-q` is taken from `gcloud app deploy` for consistency.

> I don't approve this flag name.

_Me, just now_

### DIFF
--- a/util/deploy.sh
+++ b/util/deploy.sh
@@ -2,22 +2,16 @@
 
 # Helper script for using a standardized version flag when deploying.
 
-set -e
-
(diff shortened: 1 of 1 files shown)


lukebjerring:
Running at https://deploy-script-dot-wptdashboard.appspot.com

Issue #115, opened 2018-05-02
Color of diff column is inverted?


foolip:
In https://master-dot-wptdashboard.appspot.com/results/?before=chrome@latest&after=safari@latest, in the diff column (the third column) the colors don't make sense. No differences (0 / N) is probably good and should be green or at least not-red, while many differences is probably bad.

Highlighting the differences somehow would be good. Maybe a color scheme similar to the interop view?

PR #116, opened 2018-05-02
Improve appearance of diff column

lukebjerring:
## Description
Changes the diff column appearance to be more subtle, and the numbers used to be the deltas instead of totals (i.e. instead of `total changes / total tests` we have `passing delta / total delta`).

Indirectly fixes https://github.com/web-platform-tests/wpt.fyi/issues/115

## Review
Example URL:
https://diff-colors-dot-wptdashboard.appspot.com/results/?before=chrome@06179b0154&after=chrome@latest


webapp/components/wpt-results.html
```diff
@@ -63,6 +64,26 @@
...
+      .passing-60 { background-color: hsl(0, 85%, 80%); }
+      .passing-80 { background-color: hsl(0, 85%, 85%); }
+      .passing-100 { background-color: hsl(129, 85%, 65%); }
+
+      span.delta.negative {
+        color: var(--paper-red-700);
+      }
+      span.delta.positive {
+        color: var(--paper-green-700);
+        font-weight: bold;
```
foolip: I found it a bit odd that only the green bits are bold. Is this intentional?

Also, could some kind of background be given to the non-empty cells so that they stand out a bit more when scrolling?

Finally, when there aren't any differences at all, should that be highlighted in some way? Otherwise I'd just think the view is broken :)

lukebjerring: De-boldened, grey'd the background, but didn't do anything for no differences - not likely, and confident the user will notice that all the numbers are the same.

PR #117, opened 2018-05-03
Use Fetch's response.ok instead of checking for status 200

foolip:
Also log the non-OK fetches in a consistent way.

Note that in runs.html the resolved value changes from undefined to
null, but that doesn't matter as both are falsy and filter soon after
into `nonEmpty`.


lukebjerring:
Running at https://response-ok-dot-wptdashboard.appspot.com

PR #120, opened 2018-05-04
run.sh: Re-use already running image

lukebjerring:
## Description
Re-uses the existing wptd-dev-instance if it exists. We sometimes have old instances hanging around.


lukebjerring:
Running at https://run-script-dot-wptdashboard.appspot.com

Hexcles:
Hmm, so it seems to me the script *tries* to kill and delete the running Docker instance on exit, which of course can fail sometimes, but in that case I think a good practice (or a common pattern) is to kill and remove the instance and start a new one instead of reusing, as we can't be sure what's the state inside that Docker instance anymore.

lukebjerring:
Kill and restart instead. PTAL

PR #121, opened 2018-05-04
Fix read nesting bugs in pre-push

lukebjerring:
## Description
When a read occurs before the read for the pre-push params, it causes an infinite loop. Similarly, when we use the `confirm` logging helper, which reads file desc 1, we get caught in the loop.


lukebjerring:
Running at https://pre-push-fixes-dot-wptdashboard.appspot.com

PR #122, opened 2018-05-07
Don't deploy to staging unless webapp folder changed

lukebjerring:
Prevents unnecessary bloat of the versions in appengine.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -82,7 +82,7 @@ script:
         && [ "${TRAVIS_PULL_REQUEST_BRANCH}" != "" ];
     then
       # Skip if webapp isn't modified.
-      git diff --name-only FETCH_HEAD...$TRAVIS_BRANCH | grep ^webapp/ || {
+      git diff --name-only FETCH_HEAD...${TRAVIS_BRANCH} | grep ^webapp/ || {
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -87,6 +87,12 @@ script:
         && [ "${TRAVIS_SECURE_ENV_VARS}" != "false" ] \
         && [ "${TRAVIS_PULL_REQUEST_BRANCH}" != "" ];
     then
+      # Skip if webapp isn't modified.
+      git diff --name-only FETCH_HEAD...$TRAVIS_BRANCH | grep ^webapp/ || {
```
mdittmer: nit: Consider `"FETCH_HEAD...${TRAVIS_BRANCH}"`. The `${}` is consistent with surrounding code and the quoting will deal with branches that contain funky characters (like spaces) in the branch name.

lukebjerring:
@Hexcles @mdittmer - I think this actually broke staging deployment (`${TRAVIS_BRANCH}` != `${TRAVIS_PULL_REQUEST_BRANCH}`). Notable, though, is the fact that `exit 0` during a `script` execution seems to be the cause of the truncated logs. Might be worth having an if-else branch instead.

PR #124, opened 2018-05-07
Migrate api_handlers and related code to api go package

lukebjerring:
## Description
Splits a way-too-big file and creates a dir where we can expand and document more cleanly.

Note that file resources and working directories at runtime are problematic in Golang, so we move `browsers.json` to `browsers.go` and parse a multi-line string instead.

Deployed @ https://api-package-dot-wptdashboard.appspot.com


lukebjerring:
Running at https://api-package-dot-wptdashboard.appspot.com

api/browsers.go
```diff
@@ -1,3 +1,6 @@
+package api
+
+const browsersJSON = `
```
Hexcles: This change is somewhat important, but it's not in the PR description. Could you make sure to explain it in both the PR description and the final merge commit?

Issue #107, opened 2018-05-01
Proposal: Towards a more consistent deployment/runtime environment


mdittmer:
Conversations about #78, and the failure of #104 to actually fix the use of golang 1.9 APIs got me and @Hexcles talking about our overall deployment strategy.

# Context

Fixing #78 involves getting rid of `Dockerfile.jenkins` which begs the question: Should we still have multiple `Dockerfile.*`s? In discussing this, with @Hexcles I mentioned that AppEngine Standard (the current deployment model for `webapp/`) uses golang 1.8, whereas AppEngine Flex (by default) will use golang 1.10. That would mean two `Dockerfile`s with different golang versions in order to adequately test the two properties. (The latter issue--golang version--concerns me more than having multiple `Dockerfile`s).

# Proposal

@Hexcles and I think we should use AppEngine Flex everywhere. This provides consistency. It also improves portability because we can wrap code for each service up in its own container (that could be deployed everywhere).

*Aside*: This (containerization) isn't possible with golang AppEngine Standard unless we check in dead code as a program entry point; golang AppEngine Standard provides its own program entry point and side-loads our HTTP handlers.

This would entail the following changes:

- Add program entry point that starts a HTTP server to `webapp/`
- Create separate `Dockerfile` in `webapp/`, `announcer/`, and `receiver/` based on [AppEngine Flex base images](https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build#configuring_the_dockerfile)
- Update Travis config to use these files for testing each service's code separately
- Delete root directory `Docker.*` files

Another benefit here is that we will have more consistency between dev, CI, (staging,) and production environments because each of those environments can run multiple containers to deploy all services.

This shouldn't impose significant maintenance burden on us because things like security patches for the underlying system should be shipped in the base images we extend.

@foolip @lukebjerring @kereliuk WDYT?

Hexcles:
Thanks for filing this, Mark!

One extra point: we will still use AppEngine Standard when we have to (e.g. to create tasks in TaskQueue). For everything else, we can just unify to Flex, which provides major benefits in consistency (same Docker for development, Travis CI, and prod).

> Create separate Dockerfile in webapp/, announcer/, and receiver/ based on AppEngine Flex base images

[...]

PR #126, opened 2018-05-07
Add a gcloud update to Dockerfile.dev

lukebjerring:
## Description
The docker base image is v167 of gcloud SDK, which contains some bugs about resolving the Golang version to use for the separated `wpt.fyi/api` package.

This PR adds a `gcloud components update` which bumps us up to v200 (as at time of PR).

We are going to consolidate the docker files/images anyway (see #107), but this is a quick workaround until then.


lukebjerring:
Running at https://docker-gcloud-dot-wptdashboard.appspot.com

PR #127, opened 2018-05-07
Set up the processor service of results-receiver

Hexcles:
This PR lays the ground for the [processor service of results receiver](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork/edit?ts=5ade260c#heading=h.69zbikjaf3dy). The service will be written in Python 3 and run on AppEngine Flex (see the doc for rationales).

A **custom** Python runtime is used (hence the `Dockerfile`) so that we can install gcloud SDK and other binaries in the future. The [official Python Flex image](https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build#create_a_dockerfile) is used as the base image for ease of maintenance. The gcloud SDK is useful because I plan to shell out and call `gsutil` to efficiently upload a directory recursively in the processor, which is tricky to implement with the official GCS Python library. Besides, setting up the custom docker early will give us great reproducibility and flexibility in the future, with minimal overhead (AppEngine Flex uses Docker anyways).

The PR also adds the service to the staging deployment on Travis.


Hexcles:
@lukebjerring the staging deployment failed because:
```
ERROR: (gcloud.app.deploy) User [travis-ci-deployment@wptdashboard.iam.gserviceaccount.com] does not have permission to access project [wptdashboard] (or it may not exist): The caller does not have permission
```

Perhaps some extra permission is needed to deploy Flex. I'll investigate tomorrow. Meanwhile, please review the basic setup, naming, etc.

Hexcles:
I deployed one manually: https://robertma-dev-dot-processor-dot-wptdashboard.appspot.com/api/hello

util/deploy.sh
```diff
@@ -59,7 +59,7 @@ then
 else
     QUIET_FLAG=""
 fi
-COMMAND="gcloud app deploy ${PROMOTE} ${QUIET_FLAG} --version=${VERSION} ${WPTD_PATH}/webapp"
+COMMAND="gcloud app deploy ${PROMOTE} ${QUIET_FLAG} --version=${VERSION} ${WPTD_PATH}/webapp ${WPTD_PATH}/results-processor"
```
lukebjerring: On second thought, let's make this deploy the passed-in path, which can be a required param.
`deploy.sh [-p] [-q] [-b] [-h] [app path]`

That way, we can have 2 different travis tasks and bypass them when nothing for that service has changed (see https://github.com/web-platform-tests/wpt.fyi/pull/122)

Hexcles: I'm gonna back off my change in `deploy.sh` (i.e. not deploying `results-processor` to staging for now) in this PR. I'll fix the permission and change `deploy.sh` as you suggested in another PR.

PR #128, opened 2018-05-07
Add support for latest manifest

lukebjerring:
## Description
Special-case `latest` as a SHA (which is the default when no SHA is provided) and fetch the latest release using [GitHub's latest release API](https://developer.github.com/v3/repos/releases/#get-the-latest-release)

@tabatkins - as I understand it, this is a step in the direction you wanted.

## Review Information
See https://manifest-latest-dot-wptdashboard.appspot.com/api/manifest

### DIFF
--- a/api/manifest.go
+++ b/api/manifest.go
@@ -119,7 +119,7 @@ func getGitHubReleaseAssetForSHA(client gitHubClient, sha string) (manifest []by
 	// Get (and unzip) the asset with name "MANIFEST-{sha}.json.gz"
 	shaMatch := sha
 	if sha == "" || sha == "latest" {
-		shaMatch = "[0-9a-f]{1,40}"
+		shaMatch = "[0-9a-f]{40}"
 	}
 	assetRegex := regexp.MustCompile(fmt.Sprintf("MANIFEST-%s.json.gz", shaMatch))
 	for _, asset := range assets {
--- a/api/manifest.go
+++ b/api/manifest.go
@@ -12,7 +12,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"strings"
+	"regexp"
 
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine"
@@ -42,6 +42,8 @@ func gitHubReleaseURL(tag string) string {
 	return fmt.Sprintf(`https://api.github.com/repos/w3c/web-platform-tests/releases/tags/%s`, tag)
 }
 
+const gitHubLatestReleaseURL = `https://api.github.com/repos/w3c/web-platform-tests/releases/latest`
+
 type gitHubClient interface {
 	fetch(url string) ([]byte, error)
 }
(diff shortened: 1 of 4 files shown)


lukebjerring:
Running at https://manifest-latest-dot-wptdashboard.appspot.com

tabatkins:
Yes, this looks like exactly what I need.

(I just wanted something that I can poll every 10 minutes to assemble the current list of active test files.)

api/manifest.go
```diff
@@ -42,6 +42,8 @@ func gitHubReleaseURL(tag string) string {
 	return fmt.Sprintf(`https://api.github.com/repos/w3c/web-platform-tests/releases/tags/%s`, tag)
 }
 
+const gitHubLatestReleaseURL = `https://api.github.com/repos/w3c/web-platform-tests/releases/latest`
```
foolip: Ah, excellent, I thought we'd have to figure out what the master sha was and just hope there was a manifest. That wouldn't have been as reliable as this :)

Are the backticks here meaningful, or just copypasta?

api/manifest.go
```diff
@@ -60,55 +62,74 @@ func getManifestForSHA(ctx context.Context, sha string) (manifest []byte, err er
...
 		return nil, err
 	}
 	if len(assets) < 1 {
-		return nil, fmt.Errorf("No assets found for release %s", releaseTag)
+		return nil, fmt.Errorf("No assets found for %s release", releaseTag)
 	}
 	// Get (and unzip) the asset with name "MANIFEST-{sha}.json.gz"
+	shaMatch := sha
+	if sha == "" || sha == "latest" {
+		shaMatch = "[0-9a-f]{1,40}"
```
foolip: This doesn't matter much, but it seems like the shortest sha that git will even try to match on the command line is 4 characters, so if shorter things show up here it would be very surprising. Do you think it's too conservative to assume exactly 40 chars and otherwise not match?

api/manifest_test.go
```diff
@@ -71,34 +71,41 @@ func TestGetGitHubReleaseAssetForSHA(t *testing.T) {
...
 	// 1) Data is unzipped.
-	manifest, err := getGitHubReleaseAssetForSHA(&client, sha)
+	manifest, err := getGitHubReleaseAssetForSHA(&client, fullSHA)
 	assert.Nil(t, err)
 	assert.Equal(t, []byte("magic data"), manifest)
 
+	// 1a) Release by empty SHA or "latest" match.
+	latestManifest, _ := getGitHubReleaseAssetForSHA(&client, "")
+	assert.Equal(t, manifest, latestManifest)
+	latestManifest, _ = getGitHubReleaseAssetForSHA(&client, "latest")
```
foolip: I think this works as a test, but let me check my thinking. This shows that "" and "latest" are handled specially, because "foobar" would result in an error when we try to interpret it as a sha. But, since a manifest with fullSHA is in the (mock) response for any input, it doesn't show that "latest" and some other sha that isn't latest behave differently, that depends on https://api.github.com/repos/w3c/web-platform-tests/releases/latest being used as the URL, and we don't make any assertions about that. (That's all fine, not arguing for testing deeper.)

api/manifest.go
lukebjerring: Some editors use it as a convention when defining consts, but it's only ever different for some chars. No effect in this case.

api/manifest_test.go
lukebjerring: lol. I'll split the cases, instead of being lazy.

foolip:
@tabatkins FYI I don't think this API includes the sha that the manifest is for, and per our discussion that might be a problem. File an issue maybe? :)

tabatkins:
@foolip Done, in #133.

PR #129, opened 2018-05-08
Run build + test on prepush

lukebjerring:
## Description
Adds `go build` and `go test` during prepush, since they are super fast to execute, and can catch silly errors when prepping for a PR.

Note that `make build` was previously only grabbing the `go_deps`, but not actually building.


lukebjerring:
Running at https://prepush-tests-dot-wptdashboard.appspot.com

PR #130, opened 2018-05-08
Cache manifest fetches in memcache

lukebjerring:
## Description
Prevent redundant re-fetches of recently-fetched manifest files by caching them in memcache.
Expiry is explicitly 20m for `latest` as we don't want to go stale for too long.

@Hexcles as discussed, I've tagged the memcache test which uses aetest (relies on `dev_appserver.py`) with ` +build medium. Will follow that up at a later date by tagging all test files as small/medium/large and a lint rule to make sure all tests are tagged.


api/manifest.go
```diff
@@ -58,9 +62,42 @@ func getManifestForSHA(ctx context.Context, sha string) (manifest []byte, err er
...
+		body = cached.Value
+	} else {
+		if body, err = getGitHubReleaseAssetForSHA(client, sha); err != nil {
+			return nil, err
+		}
+		item := &memcache.Item{
+			Key:   sha,
+			Value: body,
+		}
+		// Shorter expiry for latest SHA, to keep it current.
```
Hexcles: Shall we cache latest at all? If we do want to cache it, 20 minutes seem a bit too long (at least based on my observation during North American work hours). I'd suggest 5 minutes or so.

lukebjerring: Changed to 5 min.

PR #135, opened 2018-05-09
Add support for ?path param in manifest API

lukebjerring:
## Description
Filters each of the `items` values maps in the manifest down to ones where the value's URL (first array entry) starts with (one of) the paths provided.

Example:
https://manifest-paths-dot-wptdashboard.appspot.com/api/manifest?path=/css/css-images/


shared/models.go
```diff
@@ -40,3 +44,56 @@ type Browser struct {
 type Token struct {
 	Secret string `json:"secret"`
 }
+
+type Manifest struct {
+	Items   ManifestItems `json:"items,omitempty"`
+	Version *int          `json:"version,omitempty"`
```
Hexcles: Why is this a pointer?

shared/models.go
```diff
@@ -40,3 +44,56 @@ type Browser struct {
...
+	if result.Items.TestHarness, err = m.Items.TestHarness.FilterByPath(paths); err != nil {
+		return result, err
+	}
+	return result, nil
+}
+
+type ManifestItems struct {
+	Manual      *ManifestItem `json:"manual,omitempty"`
+	Reftest     *ManifestItem `json:"reftest,omitempty"`
+	TestHarness *ManifestItem `json:"testharness,omitempty"`
```
Hexcles: `ManifestItem` is a map, right? [Maps are already references](https://golang.org/doc/faq#references), so we don't need to make a pointer of it again.

shared/models.go
```diff
@@ -40,3 +44,56 @@ type Browser struct {
 type Token struct {
 	Secret string `json:"secret"`
 }
+
+type Manifest struct {
+	Items   ManifestItems `json:"items,omitempty"`
+	Version *int          `json:"version,omitempty"`
+}
+
+func (m Manifest) FilterByPath(paths mapset.Set) (result Manifest, err error) {
```
Hexcles: Should the receiver type be a pointer for consistency and to avoid copying? Though in this case the cost of copying isn't really a problem as all children types are mostly pointers. Yet using a pointer receiver type for struct seems to be a more consistent default. https://github.com/golang/go/wiki/CodeReviewComments#receiver-type

BTW, perhaps we can simply modify in place in this method, but I don't really have a preference for that.

api/manifest.go
```diff
@@ -25,13 +26,21 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
-	if manifest, err := getManifestForSHA(ctx, sha); err != nil {
+	manifestBytes, err := getManifestForSHA(ctx, sha)
+	if err != nil {
 		http.Error(w, err.Error(), http.StatusNotFound)
-	} else {
-		w.Header().Add("content-type", "application/json")
-		w.Write(manifest)
+		return
 	}
+	w.Header().Add("content-type", "application/json")
```
Hexcles: Nit: Although I think Go probably handles the capitalization already, the standard key should be `Content-Type` here.

lukebjerring:
Running at https://manifest-paths-dot-wptdashboard.appspot.com

shared/models.go
lukebjerring: JSON marshalling in Go will only drop a field if it's a pointer, otherwise it's still marshalled as the empty/default value of the struct. 

Note that unmarshalling in this fashion is unaffected.

shared/models.go
lukebjerring: Done.

shared/models.go
lukebjerring: Done (modified in-place by copying the input).

api/manifest.go
lukebjerring: Done.

lukebjerring:
PTAL.

Issue #3, opened 2018-04-10
Change api/test-runs to use a time-based query param


lukebjerring:
_From @foolip on March 26, 2018 14:19_

https://wpt.fyi/test-runs internally uses https://wpt.fyi/api/runs?max-count=100 and fills in the rows using that data.

It seems like the 100 limit is per browser (or something) which leads to funny results in the older part of the data. Right now, it looks like b0c75cd885 has only Edge and Safari runs, but in https://wpt.fyi/?sha=b0c75cd885 one can see Chrome and Firefox as well. They also show up in https://wpt.fyi/api/runs?max-count=1000.

I noticed this when working on https://github.com/foolip/ecosystem-infra-stats/issues/6, because when I calculated which runs were "complete", the last few ones didn't look complete on https://wpt.fyi/test-runs.

Fix? Perhaps to not use any runs older than the oldest available run for any individual browser when presenting the data, or do change the API to avoid the problem.

_Copied from original issue: w3c/wptdashboard#524_

jugglinmike:
The "Recent Runs" page (https://wpt.fyi/test-runs) states

> Showing the last 100 runs per browser. 

And this is reflected in the URL of the data it requests: https://wpt.fyi/api/runs?max-count=100

That data does contain 100 results for Chrome and Firefox:



    $ curl 'https://wpt.fyi/api/runs?max-count=100' | python -c 'import sys, json; print len(filter(lambda x: x["browser_name"]=="chrome", json.loads(sys.stdin.read())))'
    100
    $ curl 'https://wpt.fyi/api/runs?max-count=100' | python -c 'import sys, json; print len(filter(lambda x: x["browser_name"]=="firefox", json.loads(sys.stdin.read())))'
    100

However, the table does not display all of the data:

![screenshot from 2018-05-09 13-53-31](https://user-images.githubusercontent.com/677252/39831069-521214d6-5391-11e8-83b6-6b62f501ab6b.png)

<details>
  <summary>(script used to annotate the cells</summary>

```js
[2,4].forEach(col => {
  Array.from(document.querySelector('wpt-runs').shadowRoot.querySelectorAll('.present:nth-child('+col+')'))
    .forEach((e, i) => {
      e.innerText=i;
      e.style.color ='#fff'
    });
})
```

</details>

The data *is* present from the visualization for the "test run", though:

[...]

PR #136, opened 2018-05-09
Cull any runs older than the youngest 100th run

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/3

If multiple browsers return `max-count` / `limit` runs, find out which has the youngest `limit`th run, and cull anything older than that, so that the number of SHAs (i.e. "rows" in the UI) matches `max-count`.

## Review Information
https://runs-dot-wptdashboard.appspot.com/test-runs

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -75,21 +75,26 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		testRuns = append(testRuns, testRunResults)
 	}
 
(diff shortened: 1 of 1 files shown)


lukebjerring:
Chatted with @Hexcles about this offline. "Fixing" this at the data layer is a bad call.

`max-count` is a poor API for this use-case. `from` & `to` timestamps is a way better idea.

PR #137, opened 2018-05-10
Add a basic FF webdriver test

lukebjerring:
## Description
Adds a single webdriver integration test for the search functionality.

`install.sh` is a simple script for fetching selenium, gecko driver and firefox 58. It works inside and outside docker.

## Review Information
Running `make go_webdriver_test` locally will execute the test. See `webdriver/README.md` for more info.


Hexcles:
Sorry misclicked the close button...

Another high-level question: what about making all the paths (Selenium, Firefox) mandatory options in the Go program, so that we don't need to hard code the default values there when we already have them in the shell script? It'd be great to have as few hardcoded version strings as possible.

lukebjerring:
Removed the defaults from Golang, and put into the Makefile.

Makefile
```diff
@@ -16,6 +16,11 @@ export GOPATH=$(shell go env GOPATH)
 # WPTD_PATH will have a trailing slash, e.g. /home/jenkins/wpt.fyi/
 WPTD_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 WPTD_GO_PATH ?= $(GOPATH)/src/github.com/web-platform-tests/wpt.fyi
+WEBDRIVER_PATH ?= $(WPTD_GO_PATH)/webdriver
+BROWSERS_PATH ?= $(HOME)/browsers
+SELENIUM_PATH ?= $(BROWSERS_PATH)/selenium-server-standalone-3.8.1.jar
```
kereliuk: Do we want to include the versions here? Might make maintenance easier if we rename all of these to not include the version number and then this script can be update invariant.

webdriver/firefox.go
```diff
@@ -0,0 +1,80 @@
...
+
+// FirefoxWebDriver starts up a Firefox WebDriver.
+// Make sure to close both the service and the WebDriver instances, e.g.
+//
+// s, d, e := FirefoxWebDriver()
+// if e != nil {
+//   panic(e)
+// }
+// defer s.Stop()
+// defer wd.Quit()
```
kereliuk: This is probably just my lack of context/knowledge in go, but where does this wd variable come from in the example? I assume wd is webdriver, but from where? If its the selenium WebDriver instance does it make sense to also include that definition in the example here?

Also example comments like these are awesome :)

Makefile
lukebjerring: Removed.

webdriver/firefox.go
lukebjerring: copy pasta, fixed!

kereliuk:
Code looks good to me, but I won't press the approve button (unless you think I should), since I feel I lack context with go and the greater context at the moment.

I think we should have the README explain what things you need and how/which ones (like geckodriver and firefox), to include as arguments or if they can be in the path. The use case I'm thinking of here is someone who doesn't want to use the install script, or when if the install script is broken. They should still be able to know exactly what they need and manually configure things so they aren't dependent solely on deciphering the install script.

WYT?

lukebjerring:
@kereliuk - See `webdriver/README.md` for my stab at helpful instructions.

Issue #133, opened 2018-05-09
Need a link between manifest and the SHA it came from


tabatkins:
In #128 y'all added a `/latest` manifest endpoint, so my tooling can just regularly grab the most recent list of tests. Yay!

However, I'll eventually need to consume more information, like test results, and these are all keyed to SHA as well. So I'll need some way to get the SHA that the "latest metadata" corresponds to, so I can grab the test results for that run and have consistent data sets.

Presumably this can just be inserted into the returned JSON?

tabatkins:
That works for me. :+1:

PR #138, opened 2018-05-10
Return an HTTP header with the manifest's SHA

lukebjerring:
## Description
Adds a header `x-wpt-sha` to the response which contains the full SHA of the manifest that is returned.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/133

cc: @tabatkins 

## Review
Header is present for requests to
https://sha-header-dot-wptdashboard.appspot.com/api/manifest?sha=latest


lukebjerring:
Will document as part of #135 which introduces docs for the endpoint.

Issue #61, opened 2018-04-17
"View source on GitHub" is broken for .any.js tests


foolip:
Moved from https://github.com/web-platform-tests/results-collection/issues/215.

Example: https://wpt.fyi/console/console-count-label-conversion.any.html

MANIFEST.json could probably be used to fix this, as it contains this:
```JSON
   "console/console-count-label-conversion.any.js": [
    [
     "/console/console-count-label-conversion.any.html",
     {}
    ],
    [
     "/console/console-count-label-conversion.any.worker.html",
     {}
    ]
   ],
```

See also https://github.com/web-platform-tests/wpt.fyi/issues/56.

lukebjerring:
Stepping towards supporting this with #101 - need path filtering support next.

CYBAI:
It seems the `View Source on GitHub` is still broken for `.any.js` tests 👀?

I can reproduce 404 GitHub page while clicking `View Source on GitHub` in https://wpt.fyi/results/xhr/send-usp.any.html .

lukebjerring:
You can enable the fix if you visit wpt.fyi/flags and enable the manifest fetching feature. Without the manifest we can't be sure what the source file is.

Issue #14, opened 2018-04-10
Don't display "Run in your browser on w3c-test.org" for tests that can't run in a browser


lukebjerring:
_From @rwaldron on January 18, 2018 19:21_

For example... https://wpt.fyi/webdriver/tests/element_retrieval/get_active_element.py

_Copied from original issue: w3c/wptdashboard#373_

lukebjerring:
@gsnedders - is there something concrete that we can use to determine what can be run on w3c-test.org and what can't?

PR #139, opened 2018-05-10
Use manifest test source for source and w3c-test links

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/61 and https://github.com/web-platform-tests/wpt.fyi/issues/14 using the `/api/manifest` endpoint.

## Review Information
This example is from an `any.js` source-file:
https://manifest-test-source-dot-wptdashboard.appspot.com/results/console/console-counting-label-conversion.any.html

This example is from a `wdspec` python test:
https://manifest-test-source-dot-wptdashboard.appspot.com/webdriver/tests/element_retrieval/get_active_element.py

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -181,9 +181,8 @@ found in the LICENSE file.
   </template>
 
   <script>
+    const TEST_TYPES = ['manual', 'reftest', 'testharness', 'wdspec'];
 
-    const TEST_TYPES = ["manual", "reftest", "testharness", "wdspec"];
-    
     /* global TestRunsBase */
     class WPTResults extends TestRunsBase {
       static get is() {
@@ -294,7 +293,7 @@ found in the LICENSE file.
         for (const type of TEST_TYPES) {
           const items = manifest.items[type];
           const test = Object.keys(items).find(k => items[k].find(i => i[0] === path));
-          if (test && type == "wdspec") {
+          if (test && type === 'wdspec') {
             return false;
           }
         }
@@ -376,7 +375,7 @@ found in the LICENSE file.
         this.manifest = null;
         const that = this;
         fetch(`/api/manifest?path=${path}`).then(async r => {
-            that.manifest = await r.json()
+          that.manifest = await r.json();
         });
       }
 
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -131,8 +131,10 @@ found in the LICENSE file.
     <template is="dom-if" if="{{ pathIsATestFile }}">
       <div class="links">
         <ul>
-          <li><a href$="https://github.com/w3c/web-platform-tests/blob/master[[path]]" target="_blank">View source on GitHub</a></li>
-          <li><a href$="[[scheme]]://w3c-test.org[[path]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+          <li><a href$="https://github.com/w3c/web-platform-tests/blob/master[[sourcePath]]" target="_blank">View source on GitHub</a></li>
+          <template is="dom-if" if="{{ testCanRunOnW3C }}">
+            <li><a href$="[[scheme]]://w3c-test.org[[path]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+          </template>
         </ul>
       </div>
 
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-results.html
```diff
@@ -131,8 +131,10 @@
     <template is="dom-if" if="{{ pathIsATestFile }}">
       <div class="links">
         <ul>
-          <li><a href$="https://github.com/w3c/web-platform-tests/blob/master[[path]]" target="_blank">View source on GitHub</a></li>
-          <li><a href$="[[scheme]]://w3c-test.org[[path]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+          <li><a href$="https://github.com/w3c/web-platform-tests/blob/master[[sourcePath]]" target="_blank">View source on GitHub</a></li>
```
foolip: Aside: Would it be straightforward to use the SHA here to get to exactly the right version? Especially when looking at old results the chance for confusion is otherwise there. (Can't fix it for w3c-test.org of course.)

webapp/components/wpt-results.html
```diff
@@ -326,6 +368,17 @@
         this.refreshDisplayedNodes();
       }
 
+      async fetchManifest(path) {
+        if (!this.computePathIsATestFile(path)) {
+          return;
```
foolip: Shouldn't this throw so that the caller doesn't need to check the return value? (Looks like it's only used for an observer and I don't know how that works.)

webapp/components/wpt-results.html
```diff
@@ -326,6 +368,17 @@
...
+        }
+        this.manifest = null;
+        const that = this;
```
foolip: Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions, `that = this` ought not be needed. And because we're in an async function this can use `await` instead of `then()` to not have the arrow function at all, avoiding the issue.

webapp/components/wpt-results.html
```diff
@@ -326,6 +368,17 @@
         this.refreshDisplayedNodes();
       }
 
+      async fetchManifest(path) {
+        if (!this.computePathIsATestFile(path)) {
+          return;
+        }
+        this.manifest = null;
```
foolip: Given the early return I wonder if this is necessary, since `this.manifest` won't always be set anyway. Initialize it in a constructor or something if it matters?

webapp/components/wpt-results.html
```diff
@@ -258,6 +271,35 @@
...
+      computeSourcePath(path, manifest) {
+        if (!this.computePathIsATestFile(path) || !manifest) {
+          return path;
+        }
+        const itemSets = TEST_TYPES.map(t => manifest.items[t]).filter(i => i);
+        for (const items of itemSets) {
+          const key = Object.keys(items).find(k => items[k].find(i => i[0] === path));
+          if (key) {
+            // Ensure leading slash.
+            return key.indexOf('/') === 0 ? key : `/${key}`;
```
foolip: Nit: `key.startsWith('/')` should be a teeny bit faster and looks nice.

webapp/components/wpt-results.html
```diff
@@ -258,6 +271,35 @@
         return /(\.(html|htm|py|svg|xhtml|xht|xml)$)/.test(path);
       }
 
+      computeSourcePath(path, manifest) {
+        if (!this.computePathIsATestFile(path) || !manifest) {
+          return path;
+        }
+        const itemSets = TEST_TYPES.map(t => manifest.items[t]).filter(i => i);
```
foolip: Can you document that this is because the manifest can be missing some test types entirely? (Took me a while to work that out, and still not entirely sure.)

webapp/components/wpt-results.html
lukebjerring: Yes, it would; I'll follow up this PR with that!

webapp/components/wpt-results.html
lukebjerring: `fetchManifest` is an observer - it's just a callback for when something changes

webapp/components/wpt-results.html
lukebjerring: It's for (instantly) clearing out previous values (while the new value loads).

webapp/components/wpt-results.html

webapp/components/wpt-results.html

webapp/components/wpt-results.html
foolip: Oh, so this is async in order use await, not to return a promise. Got it.

PR #141, opened 2018-05-11
Add lint for Golang test size tags

lukebjerring:
## Description
Lints each of the `*_test.go` files for a test size tag, e.g. `// +build small`

This allows us to more-generally and simply separate the integration tests, rather than having to do file-level separation.


Hexcles:
Actually, I'm expecting Travis to fail, because my #142 has landed which has some tests without build tags.

It happens to be a good chance to see what Travis logs look like when the lint fails.

Hexcles:
@lukebjerring turns out the previous implementation couldn't find tests in deeper directories. I fixed the Makefile by using `$(shell find ...)` instead of `$(wildcard ...)`, and added build tags to the previously missed test files.

Issue #144, opened 2018-05-14
Change Sauce to run Edge 17 


lukebjerring:
Edge 17 now available, but 16 is shown on wpt.fyi.
https://wiki.saucelabs.com/pages/viewpage.action?pageId=72009845

lukebjerring:
Now running @17 (e.g. https://wpt.fyi/results/?sha=7f56040b8f)

PR #145, opened 2018-05-14
Add v 16 + 17 to Edge browser entries

lukebjerring:
See https://github.com/web-platform-tests/wpt.fyi/issues/144

### DIFF
--- a/shared/browsers.go
+++ b/shared/browsers.go
@@ -43,10 +43,28 @@ const browsersJSON = `
         "os_version": "*"
     },
     "edge-15-windows-10-sauce": {
+        "initially_loaded": false,
+        "currently_run": false,
+        "browser_name": "edge",
+        "browser_version": "15",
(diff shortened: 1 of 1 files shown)


lukebjerring:
Yup - sorry you went to the hassle of clarifying all of this (as usual.. great description!)

This change will be necessary eventually; I'm currently working on supporting `/api/runs?browser=edge-16` type params correctly
(currently just `edge` "works" e.g. https://wpt.fyi/api/runs?max-count=10&browser=edge)

We also have `?before=platform@sha&after=platform@sha` for specifying specific platform __and__ run, for diffing purposes. I'll need it if we want to be supporting edge-16@sha & edge-17@sha when we deliberately overlap for one or more SHAs when switching sauce's config.

jugglinmike:
I'd merge for you, but I don't have write access to this repository.

Issue #150, opened 2018-05-15
Show interop score in interop view


zcorpan:
Here's a mockup I did today

<img width="1033" alt="screen shot 2018-05-15 at 14 12 33" src="https://user-images.githubusercontent.com/244772/40057224-f00c7dd2-584d-11e8-8696-8df2823db24d.png">

The "0 / 4" and "4 / 4" have a multiplier of 1, the "1 / 4" and "3 / 4" have a multiplier of 0.5, and "2 / 4" has a multiplier of 0, and those scores are then added together, setting the <progress>.value to something between 0 and 1. I made the colors of the columns represent these weights. The idea is that "2 / 4" is the worst case, interop-wise.

There could be other ways to calculate the score that may be better; after all 0/4 really means that the interop score is "unknown" since tests can fail for different reasons (but the feature under test is most probably not usable by web developers). An alternative way to calculate is N/M, i.e., "0 / 4" has a multiplier of 0, "1 / 4" has a multiplier of 0.25, etc.

Thoughts @lukebjerring @foolip @isaacdurazo @boazsender

foolip:
Also cc @mdittmer.

Originally we didn't compute a score and went with purple across all columns to avoid deciding what we think is "good" or "bad". See [WPT Dashboard Interop Stats PRD (comment)](https://docs.google.com/document/d/1g22NuZ82RqPUYEOUzihc3quCuWqZ2NcearOGOIYlF20/edit?disco=AAAABhioaHU) for some of that background.

However, if we just accept that we want tests to pass, then it's straightforward to say that more browser passing is better. Until we have some semi-robust notion of coverage I'm still a bit hesitant about scores, but keen to hear what others think. It'd be interesting to see which directories would have the highest and lowest scores, to see if that matches our hunches about what's interoperable and not.

In https://github.com/web-platform-tests/wpt.fyi/issues/132 I suggested another UI change to make it easier to find problems.

zcorpan:
<img width="1033" alt="screen shot 2018-05-15 at 15 45 05" src="https://user-images.githubusercontent.com/244772/40060493-0dd3c0e2-5857-11e8-9caf-104aa73bb569.png">

This has N/M scores.

foolip:
I think that looks pretty nice. Apart from the score, is the only difference what the color is which is faded in/out for each column? Do the intensities still add up to 1?

zcorpan:
[...]

PR #154, opened 2018-05-15
Change interop colors

lukebjerring:
## Description

Part of https://github.com/web-platform-tests/wpt.fyi/issues/150

Uses (equivalent rgb values of) the `--paper-{grey, red, deep-orange, orange, light-green}-500` material colors.

## Review

https://interop-colors-dot-wptdashboard.appspot.com/interop


zcorpan:
Did you intend to match the first or second mockup, or try something else?

zcorpan:
The first mockup has red in the middle  (2/4).

The second mockup has red on the left (0/4), going towards green (4/4).

It seemed like @foolip liked the second approach better.

foolip:
I do like the red&rarr;green better than red in the middle. However, I think maybe 0/4 should be reddest of them all and not gray.

lukebjerring:
lol - didn't notice! Disclaimer: I struggle with https://colormax.org/color-blind-test/, which is why I defer to the material palette and avoid color decisions where possible!

lukebjerring:
@foolip - well... have to say, 0/4 is _technically_ interoperable... hence grey made sense to me.

Happy to change to super-duper red if we are all in agreement there.

zcorpan:
red->green is OK with me :)

lukebjerring:
Re-deployed with red, deep-orange, orange, yellow, green.. thoughts?

zcorpan:
Is it the same green as is used in results/? https://interop-colors-dot-wptdashboard.appspot.com/results/2dcontext the green seems brighter here I think.

Otherwise LGTM

PR #155, opened 2018-05-15
Fix + prettify the upload form

lukebjerring:
## Description
Corrects the `action` attribute of the form, and re-jigs things to use MD/paper Polymer elements.

TODO: Fix the way we include/deploy bower_components


lukebjerring:
@Hexcles note that the deployment from Travis seems to have failed and the logs are truncated again...

Issue #146, opened 2018-05-14
Process and persist the received results


Hexcles:
This is the [processor service](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork/edit?ts=5ade260c#heading=h.69zbikjaf3dy) part of the results receiver #55 , which will be implemented in Python (3) as a Flex service.

The MVP is to:
1. Extract metadata and store the results in a permanent place (*).
2. Split the monolithic JSON report into individual test reports compatible with the current frontend architecture.
3. Call the test run creation API to inform the frontend of the new results.

(*) Since we decided to transfer metadata in band (i.e. inside `wptreport`), the receiver can't immediately decide the permanent file path for the results (which requires browser, OS, sha, etc. from the `wptreport`). The receiver hence temporarily stores the results in a buffer bucket and pass them to the processor, which will then move the results to the permanent location.

PR #157, opened 2018-05-17
[processor] Implement WPTReport and tests

Hexcles:
WPTReport abstracts wptreport.json and provides some helper functions to
summarize and split the full report (the split output is needed by the
current frontend).

The code is largely based on
https://github.com/web-platform-tests/results-collection/blob/master/src/scripts/upload-wpt-results.py
but without any sharding support.

wptreport.py is also an executable (requires Python 3), a handy script
to summarize and/or split a wptreport JSON locally. See its help message
for detailed usage.

The plan is to run unit tests via `tox`, which takes care of creating
virtualenv and installing dependencies. The setup currently works
locally as long as Python 3 and tox are available. It hasn't been
integrated into Travis or the project Makefile yet; the integration will
happen soon later.

### DIFF
--- a/results-processor/tox.ini
+++ b/results-processor/tox.ini
@@ -1,0 +1,8 @@
+[tox]
+envlist = py3
+# We don't have or need setup.py for now.
+skipsdist=True
+
+[testenv]
+deps = -rrequirements.txt
+commands = python -m unittest discover . "*_test.py"
--- a/results-processor/tox.ini
+++ b/results-processor/tox.ini
@@ -5,4 +5,6 @@ skipsdist=True
 
 [testenv]
 deps = -rrequirements.txt
-commands = python -m unittest discover . "*_test.py"
+commands =
+    flake8 *.py
+    python -m unittest discover . "*_test.py"
--- a/results-processor/.gitignore
+++ b/results-processor/.gitignore
@@ -1,0 +1,3 @@
+env/
+.tox/
+__pycache__/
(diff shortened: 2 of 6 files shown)


Hexcles:
@lukebjerring FYI, this achieves the same purpose as your https://github.com/web-platform-tests/results-collection/pull/551

As I mentioned to you earlier, since this class is needed by my processor anyway, it probably makes more sense for me to make the class an executable script instead of adding another script to `results-collection`. Here's a more elaborated rationale: our use cases are exactly the same, whereas in `results-collection` there are some extra complexities because of the sharding support and BuildBot integration, so we can have a more concise, and perhaps better tested/exercised code path here, and in the meantime give `results-collection` more flexibility in case @jugglinmike wants to change the upload script.

Run `results-processor/wptreport.py -h` to try it out.

results-processor/tox.ini
```diff
@@ -0,0 +1,10 @@
+[tox]
+envlist = py3
+# We don't have or need setup.py for now.
+skipsdist=True
+
+[testenv]
+deps = -rrequirements.txt
```
lukebjerring: This look suspiciously like a typo?

Hexcles: Actually, no, though I agree it's weird... It's in fact equivalent to `-r requirements.txt` in the command line but somehow `tox` doesn't like the space (the syntax is documented in one of its basic examples).

Hexcles:
Note: this is part of #146

PR #158, opened 2018-05-17
Docker cleanup

mdittmer:
## Description

Reduce root directory Dockerfile down to one "unified dev environment" Dockerfile. Push as many install steps to Makefile as possible.


mdittmer:
@Hexcles and @lukebjerring PTAL. Minimizing Dockerfile and maximizing Makefile makes `docker build` fast, but has the unfortunate side-effect of a high cost whenever the instance has to be restarted (because all tools need to be reinstalled). Is this the right approach, or should some of the system-level tools still be installed in the Dockerfile?

mdittmer:
... perhaps the restart issue can be fixed in `run.sh`? I noticed that it often stops and removes any existing image. We could prompt the user to remove or preserve. WDYT?

mdittmer:
I pushed a change to manage each case:

- Image may or may not exist
- Image may or may not be running

It prompts to preserve or remove existing images and supports `-q` for quiet mode to preserve by default.

Hexcles:
I originally suggested to remove previous Docker instances, especially if the script is terminated unexpectedly last time, to prevent getting into an unknown or stale state in Docker. And it wasn't much of a performance issue back then because most setup was done in Dockerfile and Docker has good caching.

But now since we are moving a whole bunch of stuff into runtime, the performance penalty of restarting an instance is much higher, so I think prompting users to choose whether to recreate the instance is a good compromise.

Makefile
```diff
@@ -82,15 +68,70 @@ go_webdriver_test: go_webdriver_deps bower_components
...
+			gsutil \
+			app-engine-python; \
+		gcloud config set disable_usage_reporting false; \
+	fi
+	if [[ "$$(which nodejs)" == "" ]]; \
+	then \
+		curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -; \
+		sudo apt-get install --assume-yes --no-install-suggests nodejs; \
+	fi
+	if [[ "$$(which npm)" == "" ]]; \
```
Hexcles: The `nodejs` package from `deb.nodesource.com` [already provides `npm`](https://deb.nodesource.com/node_8.x/dists/sid/main/binary-amd64/Packages), so there's no need to detect and install `npm` seperately (in fact, `apt-get install npm` will try to install `nodejs` again and hence be a no-op).

Makefile
```diff
@@ -82,15 +68,70 @@ go_webdriver_test: go_webdriver_deps bower_components
...
+
 go_webdriver_deps: go_deps webdriver_deps
 
 webdriver_deps:
+	sudo apt-get install --assume-yes --no-install-suggests default-jdk wget xvfb $$(apt-cache depends firefox-esr chromedriver |  grep Depends | sed "s/.*ends:\ //" | tr '\n' ' ')
+	cd $(WPTD_PATH)webapp; npm install web-component-tester --unsafe-perm
 	cd $(WEBDRIVER_PATH); ./install.sh $(BROWSERS_PATH)
 
-go_deps: $(find .  -type f | grep '\.go$' | grep -v '\.pb.go$')
+go_deps: sys_deps $(find .  -type f | grep '\.go$' | grep -v '\.pb.go$')
```
Hexcles: Does `$(find ...)` work in `Makefile`? I think we need `$(shell find ...)`.

Besides, can we simply use `$(GO_FILES)` here? (`$(GO_FILES)` doesn't filter out `.pb.go`, but I don't think we have any of those, do we?).

PR #159, opened 2018-05-17
Remove bower_components from repo, install as part of deploy/build

lukebjerring:
## Description
Deletes the `bower_components` dir from the repo, and makes it an as-needed `make` rule step.


Hexcles:
Webdriver tests failed.

Issue #161, opened 2018-05-17
Figure out (automatic) staging & prod deployment


lukebjerring:
Currently, we deploy the `webapp` project automatically, using `util/deploy.sh`. This issue requires:

 - Generalizing (parameterizing) the deployment script, to apply to other projects (i.e. `results-processor`)
 - Updating the Travis CI client-secret.json.enc to have the needed permissions.

lukebjerring:
I'm going to get started on the deploy script.

@mdittmer - you recently wrestled with permissions for a Flex environment. Can you update the secret?

mdittmer:
I double checked the roles for the account that appears to be associated with Travis deployment. It should now have all the permissions it needs.

lukebjerring:
After a large number of PRs (#183, #181, #179, #178, #171, #162, #159, #143, #122, #111, #83, ...), this behaviour is looking stable.

Hexcles:
Apart from a rare race condition (#191 ), this issue seems to be fixed. Closing.

PR #162, opened 2018-05-17
Generalize deploy.sh for multiple apps

lukebjerring:
## Description
Working towards https://github.com/web-platform-tests/wpt.fyi/issues/161

Rather than have to duplicate the ~10 lines in the `.travis.yml` file, I've pulled that out into `travis-deploy-staging.sh` which logs a little nicer (and is a subshell, so hopefully we see less logs issues).

### DIFF
--- a/util/deploy-comment.sh
+++ b/util/deploy-comment.sh
@@ -5,8 +5,8 @@
 
 STAGING_URL="$1"
 
-REPO_DIR="$(dirname $0)"
-source "${REPO_DIR}/logging.sh"
+UTIL_DIR="$(dirname $0)"
+source "${UTIL_DIR}/logging.sh"
 
 if [[ -z "${STAGING_URL}" ]];
 then fatal "Deployed URL is required";
--- a/util/travis-deploy-staging.sh
+++ b/util/travis-deploy-staging.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# Helper script for posting a GitHub comment pointing to the deployed environment,
+# from Travis CI. Also see deploy.sh
+
+APP_PATH=$1
+
+UTIL_DIR="$(dirname $0)"
+source "${UTIL_DIR}/logging.sh"
+
+if [ ${TRAVIS_SECURE_ENV_VARS} == false ]; then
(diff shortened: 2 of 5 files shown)


util/deploy-comment.sh
```diff
@@ -5,8 +5,8 @@
 
 STAGING_URL="$1"
 
-REPO_DIR="$(dirname $0)"
-source "${REPO_DIR}/logging.sh"
+UTIL_DIR="$(dirname $0)"
```
mdittmer: nit: here and elsewhere: prefer `$(dirname ${BASH_SOURCE[0]})` in case of varied invocation patterns.

util/travis-deploy-staging.sh
```diff
@@ -0,0 +1,32 @@
...
+  exit 0
+fi
+
+if [ -z "${TRAVIS_PULL_REQUEST_BRANCH}"]; then
+  info "Not on a PR. Skipping ${APP_PATH} deployment."
+  exit 0
+fi
+
+# Skip if webapp isn't modified.
+git diff --name-only FETCH_HEAD...${TRAVIS_PULL_REQUEST_BRANCH} | grep ^$APP_PATH/ || {
```
mdittmer: nit: `"^$APP_PATH/"`

util/travis-deploy-staging.sh
```diff
@@ -0,0 +1,32 @@
...
+
+# Helper script for posting a GitHub comment pointing to the deployed environment,
+# from Travis CI. Also see deploy.sh
+
+APP_PATH=$1
+
+UTIL_DIR="$(dirname $0)"
+source "${UTIL_DIR}/logging.sh"
+
+if [ ${TRAVIS_SECURE_ENV_VARS} == false ]; then
```
mdittmer: nit: double-square-brackets preferred, because [reasons](https://serverfault.com/questions/52034/what-is-the-difference-between-double-and-single-square-brackets-in-bash). That said, I think we have more single than double at this point in this project ¯\_(ツ)_/¯

lukebjerring:
Merging as webdriver/wct failures are independent.

PR #164, opened 2018-05-18
Extend TestRun to store complete hash and consolidated results URL

mdittmer:
## Description

Add additional (optional) fields to `TestRun` golang type. These fields will be filled out initially by infrastructure that consolidates sharded run results under the current results collection scheme. When the results receiver that accepts consolidated results launches it will also fill out these fields (@Hexcles FYI).

## Changes

Additional fields:

```
	RevisionHash  string   `json:"revision_hash"`
	RawResultsURL string   `json:"raw_results_url"`
```


mdittmer:
Initial tangible use case: https://github.com/mdittmer/results-analysis/blob/migration/migration/run/migrate_runs.go#L319

shared/models.go
```diff
@@ -69,7 +69,9 @@ type TestRun struct {
 	// Results URL
 	ResultsURL string `json:"results_url"`
 
-	CreatedAt time.Time `json:"created_at"`
+	CreatedAt     time.Time `json:"created_at"`
+	RevisionHash  string    `json:"revision_hash"`
```
lukebjerring: This belongs in `ProductAtRevision` above (since it'll be wanted whenever `Revision` is wanted).

Also, let's use `FullRevisionHash` to be nice and explicit about the difference between the fields.

shared/models.go
```diff
@@ -54,8 +54,13 @@ type Version struct {
...
 	Product
 
-	// The first 10 characters of the SHA1 of the tested WPT revision
+	// The first 10 characters of the SHA1 of the tested WPT revision.
+	//
+	// Deprecated: The authoritative git revision indicator is FullRevisionHash.
 	Revision string `json:"revision"`
+
+	// The complete SHA1 hash of the tested WPT revision.
+	FullRevisionHash string `json:"full_revision_hash"`
```
lukebjerring: Might be worth adding `omitempty`

PR #166, opened 2018-05-18
[processor] Implement GCS upload & task handler

Hexcles:
This PR is part of #146 and implements all basic functionalities except creating the task run, namely:

1. `wptreport` can now create a full results directory that contains both the summary and split test results (both gzipped), which is ready to be uploaded to the GCS bucket that backs the current frontend.
2. Create a new `gsutil` module that abstracts external calls out to `gsutil` which does all the parallel recursive upload. It's much simpler than implementing ourselves using the Python GCS library.
3. We now have a simple Flask-based web server which handles the tasks from TaskQueue. The handler currently uploads the processed results to a temporary dev bucket.

The next step is to upload to the prod bucket, and create test runs by calling a HTTP API on wpt.fyi.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import logging
 import re
 import shutil
 import tempfile
@@ -13,6 +14,7 @@ import gsutil
 APPENGINE_INTERNAL_IP = '10.0.0.1'
 
 
+logging.basicConfig(level=logging.INFO)
 app = flask.Flask(__name__)
 
 
@@ -75,4 +77,5 @@ def task_handler():
 
 # Run the script directly locally to start Flask dev server.
 if __name__ == '__main__':
+    logging.basicConfig(level=logging.DEBUG)
     app.run(debug=True)
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -57,8 +57,10 @@ def task_handler():
 
     resp = "{} results loaded from {}".format(len(report.results), gcs_path)
 
-    gsutil.copy(gcs_path, 'gs://wptd-results/{}/{}/report.json'.format(
-        revision, browser))
+    gsutil.copy(
+        'gs:/' + gcs_path,
+        'gs://wptd-results/{}/{}/report.json'.format(revision, browser)
+    )
 
     tempdir = report.populate_upload_directory(
         browser=browser, revision=revision)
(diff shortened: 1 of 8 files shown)


Hexcles:
(Don't review it yet. I'm actually just gonna put the "next step" in this PR, too.)

lukebjerring:
Looks to me like Travis doesn't actually execute the tests you've added? Correct me if I'm wrong, or add a matrix entry for Python tests.

results-processor/main.py
```diff
@@ -0,0 +1,63 @@
...
+    gcs_path = params['gcs']
+    result_type = params['type']
+    # TODO(Hexcles): Support multiple results.
+    assert result_type == 'single'
+    # Optional fields (to be deprecated once everyone has embedded metadata):
+    browser = params.get('browser')
+    revision = params.get('revision')
+
+    match = re.match(r'/([^/]+)/(.*)', gcs_path)
+    assert match
```
lukebjerring: Forgive my lack of Python, but what is the outcome of a failed assert?

results-processor/main.py
```diff
@@ -0,0 +1,63 @@
...
+    bucket = gcs.get_bucket(bucket_name)
+    blob = bucket.blob(blob_path)
+
+    with tempfile.NamedTemporaryFile(suffix='.json') as temp:
+        blob.download_to_file(temp)
+        temp.seek(0)
+        report = wptreport.WPTReport()
+        report.load_json(temp)
+
+    resp = "{} results loaded from {}".format(len(report.results), gcs_path)
```
lukebjerring: Is `"{} and {}".format(foo, bar)` somehow preferable to `"%s and %s" % (foo, bar)` ?

results-processor/main.py
Hexcles: This will throw a HTTP 500, and traceback will be logged.

This assert really shouldn't fail as the GCS path is generated by the receiver. If this fails, there's a programming error, which is why this is an assert instead of some more elegant error handling.

results-processor/main.py
Hexcles: Not really in this case, but `"{}".format` has some more advanced features (like named arguments) which aren't used here.

lukebjerring:
Running at https://results-processor-dot-wptdashboard.appspot.com

Hexcles:
The PR is ready for review. I've done some manual end-to-end integration tests (which involves TaskQueue) using the staging deployment. Everything seems to WAI.

(In fact, @lukebjerring has already reviewed it. Thanks!)

PR #171, opened 2018-05-22
Fix deploy.sh issues

lukebjerring:
## Description
Pushing to prod highlighted some bugs in the script.


Hexcles:
Actually one nit: perhaps remove the recipe around L175 which isn't used anymore?

```
 gcloud-%: gcloud
 	gcloud components install --quiet $*
```

PR #174, opened 2018-05-23
Test and implement Announcer interface

mdittmer:
## Description

Announcer interface for component responsible for accepting queries for announced revisions.


revisions/announcer/announcer.go
```diff
@@ -0,0 +1,293 @@
...
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage/memory"
+)
+
+const mergedPrTagPrefix = "refs/tags/merge_pr_"
+
+var errNotAllEpochsConsumed = errors.New("Not all epochs consumed")
+var errNilRepo = errors.New("Repository may not be nil")
+var errVacuousEpochs = errors.New("[]epoch.Epoch slice is vacuous: contains no epochs")
```
lukebjerring: Nit: Simpler naming is better (`empty` vs `vacuous`)

revisions/announcer/announcer.go
```diff
@@ -0,0 +1,293 @@
...
+		return nil, err
+	}
+
+	prIter, err := agit.NewMergedPRIter(tagsIter, repo)
+	if err != nil {
+		log.Printf("ERRO: Failed to create git remote reference iter: %v", err)
+		return nil, err
+	}
+
+	return agit.NewStopReferenceIter(agit.NewStartReferenceIter(prIter, func(ref *plumbing.Reference) bool {
```
lukebjerring: nit: declare + name the 2 func parameters here, above, so this call is more readable.

revisions/announcer/announcer.go
```diff
@@ -0,0 +1,293 @@
...
+	// Surface error if not all epochs have a revision.
+	if numChangesFound != numChanges {
+		return revs, errNotAllEpochsConsumed
+	}
+
+	return revs, nil
+}
+
+// Update performs a fetch on the underlying repository. Subsequent calls to GetRevisions() will incorporate any newly fetched revisions.
+func (a *gitRemoteAnnouncer) Update() (err error) {
```
lukebjerring: Nit: Given our dependency on git, naming the function to match the git behaviour is probably easier on the reader.
i.e. call this Fetch

revisions/announcer/announcer_test.go
```diff
@@ -0,0 +1,606 @@
...
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+	"gopkg.in/src-d/go-git.v4/plumbing/object"
+	"gopkg.in/src-d/go-git.v4/plumbing/storer"
+	"gopkg.in/src-d/go-git.v4/storage"
+)
+
+type Fake struct{}
+
+var errFake = errors.New("Implementation is fake")
```
lukebjerring: You're better off using [`mockgen`](https://github.com/golang/mock) here, since that throws when unexpected calls occur and the tests can more easily see what's being (mocked) returned.
See https://github.com/web-platform-tests/wpt.fyi/blob/master/api/receiver/results_receiver_test.go for a recent example in our codebase.

mdittmer: I see invocations of `NewMockAppEngineAPI`, but no implementation.

revisions/announcer/announcer.go
mdittmer: Done.

revisions/announcer/announcer.go
mdittmer: Done.

revisions/announcer/announcer.go
mdittmer: Done.

revisions/announcer/announcer_test.go
mdittmer: Done.

revisions/announcer/announcer_test.go
```diff
@@ -0,0 +1,600 @@
...
+	a, err := announcer.NewGitRemoteAnnouncer(announcer.GitRemoteAnnouncerConfig{
+		EpochReferenceIterFactory: mockFactory,
+		Git: repo,
+	})
+	assert.True(t, a != nil)
+	assert.True(t, err == nil)
+
+	revs, err := a.GetRevisions(make(map[epoch.Epoch]int), announcer.Limits{})
+	assert.True(t, revs == nil)
+	assert.True(t, err == errFake)
```
lukebjerring: `assert.Equal` here, and elsewhere, (and `NotEqual`)

Hexcles:
Quick question: the whole `revision/` will end up in a Flex Go service with a `Dockerfile` and `app.yaml`, right?

mdittmer:
@Hexcles that's right.

PR #175, opened 2018-05-23
Add HSTS headers to all requests

### LINKED ISSUE #58: Redirect http://wpt.fyi/ to https://wpt.fyi/ and use HSTS

lukebjerring:
Clone of #77

This adds an HSTS header to every request that expires after 1 year.
It also applies the rule to all subdomains (we should check if this is
feasible) and opts-in to the HSTS preload list.

Resolves #58

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,9 @@ services:
 
 matrix:
   include:
-    # Integration tests are broken - see #170
-    # - language: go
-    #  env:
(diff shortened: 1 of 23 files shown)


Hexcles:
We also need to add HSTS headers to `api/routes.go`

PR #180, opened 2018-05-24
Handle version strings as prefix

lukebjerring:
## Description
Support filtering by the prefix of a version string (instead of a complete match).

@Hexcles - note that we can't actually do this cleanly; searching by prefix requires primary sort on the field that we filter by, which means that multiple versions which match the prefix will be ordered by the version, not the run `CreatedAt`. I've mitigated this by using `-BrowserVersion` (descending sort) - we'll almost never be creating newer results for an older version.


lukebjerring:
@Hexcles @mdittmer as mentioned, the strange UTF-8 proto error happens in Docker (but not on macOS, for the same test execution!)
https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/383427308

api/test_run_medium_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	r, _ = i.NewRequest("GET", "/api/run?product=chrome-66.0.3359.139", nil)
+	resp = httptest.NewRecorder()
+	apiTestRunGetHandler(resp, r)
+	body, _ = ioutil.ReadAll(resp.Result().Body)
+	assert.Equal(t, http.StatusOK, resp.Code)
+	json.Unmarshal(body, &result)
+	assert.Equal(t, "66.0.3359.139", result.BrowserVersion)
+
+	r, _ = i.NewRequest("GET", "/api/run?product=chrome-68", nil)
```
foolip: Can you add tests that ensure that "/api/run?product=chrome-6" doesn't match any run? I can't find where the prefix matching avoids this problem, so maybe some code changes required too?

shared/params.go
```diff
@@ -110,19 +110,28 @@ func ParseProduct(product string) (result Product, err error) {
 // ParseVersion parses the given version as a semantically versioned string.
 func ParseVersion(version string) (result *Version, err error) {
```
foolip: Is the Version struct and ParseVersion need at all now? I was expecting them to disappear with this change, not grow :)

It looks like before this change they were only used to validate and throw an error in ParseProduct, the struct isn't used. After the change it also looks like this is the only purpose? Since these strings won't comply with semver and will have varying number of components, do we need to parse them at all? Just checking that they match a regex seems like less code, and clearer that only validation is going on.

shared/util.go
```diff
@@ -74,3 +78,16 @@ func wrapHSTS(h http.HandlerFunc) http.HandlerFunc {
...
+// QueryPrefix returns the given query with a prefix filter on the given
+// field name, using the >= and < filters.
+func QueryPrefix(query *datastore.Query, fieldName, prefix string, desc bool) *datastore.Query {
+	order := fieldName
+	if desc {
+		order = "-" + order
+	}
+	return query.
+		Order(order).
+		Filter(fieldName+" >=", prefix).
```
foolip: (I assume that `>=` and `<` compare strings by unicode values and doesn't do anything numeric.)

I think this is where the logic needs to be different to let "68" and "68.0" match "68.0.3432.3" but not "6" or "68.0.3".

shared/params.go
lukebjerring: I intend to use them properly at a later date - this is validation + prepwork mixed together :)

api/test_run_medium_test.go
lukebjerring: Added the test (which fails). TODO: Fix it.

shared/util.go
lukebjerring: It's lexographically comparing, which is the only way to do a prefix in Golang :angry:

lukebjerring:
Running at https://versions-dot-wptdashboard-staging.appspot.com

foolip:
Something seems to be amiss, https://versions-dot-wptdashboard-staging.appspot.com/api/run?product=chrome-66 says:
```
API error 4 (datastore_v3: NEED_INDEX): no matching index found. recommended index is:
- kind: TestRun
  properties:
  - name: BrowserName
  - name: BrowserVersion
    direction: desc
  - name: CreatedAt
    direction: desc
```

I actually first tried https://versions-dot-wptdashboard-staging.appspot.com/results/?product=chrome-66 but that doesn't do anything, wondering if it should eventually?

shared/params.go
foolip: OK, will take another look when it's used for realz :)

(I'm guessing funny things will happen when the number of components is less than 4 or when some part isn't entirely numerical, but will see.)

shared/util.go
foolip: Lexicographically, right. To fix the "/api/run?product=chrome-6" test I guess it'll be necessary to complicate this by requiring either a full match, or the `>=` + `<` trick with the input plus a `.`. In other words, for chrome-6, version either has to be "6", or start with "6."

api/test_run_medium_test.go
```diff
@@ -0,0 +1,82 @@
...
+		CreatedAt: now.AddDate(0, 0, -1),
+	}
+	datastore.Put(ctx, datastore.NewIncompleteKey(ctx, "TestRun", nil), &chrome)
+
+	// Today, revision increased and an experimental run
+	chrome.BrowserVersion = "66.0.3359.181"
+	chrome.CreatedAt = now
+	datastore.Put(ctx, datastore.NewIncompleteKey(ctx, "TestRun", nil), &chrome)
+
+	// Also today, a v68 run.
```
foolip: Make this a TODO?

Makefile
```diff
@@ -64,7 +64,9 @@ go_small_test: go_deps
 	cd $(WPTD_GO_PATH); go test -tags=small -v ./...
 
 go_medium_test: go_deps dev_appserver_deps
-	cd $(WPTD_GO_PATH); go test -tags=medium -v ./...
+	# Hack to work around https://github.com/golang/appengine/issues/136
```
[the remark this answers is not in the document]
lukebjerring: The particular test we have for this PR triggers the linked bug (and the test fails), so we are fixing the version of `aetest` prior to the breakage until the issue is fixed.

api/test_run_medium_test.go
lukebjerring: Clarified comment's meaning better.

PR #182, opened 2018-05-24
Add Python tests to Travis

Hexcles:
Integrate Python unit tests and lints (both managed by `tox`) into Makefile & Travis.

Also, since we are running everything inside Docker now, we don't need to specify `language`, which supposedly saves some time installing the language-specific dependencies on Travis outside Docker.


lukebjerring:
👍

PR #184, opened 2018-05-25
Announcer  api types

mdittmer:
## Description

Test and implement HTTP API request/response golang types and conversion functions between internal API types.


revisions/api/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+	if len(rs) < len(epochs) {
+		return latest, errMissingRevision
+	}
+
+	return latest, nil
+}
+
+// EpochsRequest models a request for the epochs supported by the service.
+type EpochsRequest struct{}
```
lukebjerring: What's this used by?

revisions/api/types.go
```diff
@@ -0,0 +1,178 @@
...
+
+// EpochsResponse models a response for the epochs supported by the service.
+type EpochsResponse []Epoch
+
+// RevisionsRequest models a request for the announced revisions.
+type RevisionsRequest struct {
+	Epochs       []epoch.Epoch `json:"epochs,omitempty"`
+	NumRevisions int           `json:"num_revisions,omitempty"`
+	Now          time.Time     `json:"now,omitempty"`
```
lukebjerring: I don't know about `Now` as a field passed by the client. Maybe `Time`, `From` or `At` ?

revisions/api/types.go
mdittmer: Nothing. It's leftover from when I was playing with abstract validation of requests and needed a struct for params of all request types.

revisions/api/types.go
mdittmer: Changed to `At` (along with announcer layer field by the same name).

revisions/api/types.go
mdittmer: Deleted.

PR #186, opened 2018-05-25
Push webapp if api/ changes

lukebjerring:


Hexcles:
Question: what about `shared`? Will it also affect the webapp?

Issue #177, opened 2018-05-23
Set up health checks


mdittmer:
According to this:

https://cloud.google.com/appengine/docs/flexible/custom-runtimes/configuring-your-app-with-app-yaml#health_checks

A project-wide setting is used to enable/disable "split health checks" (as opposed to "legacy health checks" which will go away this September).

Right now, `wptdashboard` is set to the default "legacy health checks". We should probably switch to split health checks with appropriate configurations on all services. From what I understand this would look like:

1. Land changes to all `app.yaml` files that appropriately configure split health checks;
1. Switch, project-wide to split health checks with `gcloud app update --split-health-checks --project wptdashboard`;
1. Redeploy all services to ensure that they are configured appropriately with respect to the project-wide health check configuration.

@Hexcles and @lukebjerring do you want to own config changes for receiver and webapp (resp.), and I'll handle the announcer?

Hexcles:
Yeah I can take care of my Flex service (processor).

It's relatively low priority since the default legacy health check seems to meet my needs so far.

Hexcles:
When @mdittmer discussed this with me offline a couple of days ago, I said that I didn't believe there was a project-level setting for health checks. I was not exactly correct. There's indeed a project level setting as linked at the top, but the setting doesn't really do much: it will (supposedly) give an error if custom legacy health checks are used in the project; otherwise, it simply switches from default legacy health checks to default split health checks (which isn't really noticeable).

And even without tweaking that project-level setting, uploading an `app.yaml` with custom split health checks will turn on split health checks for the corresponding service.

Both the receiver (#187) and the announcer (#262) are already using split health checks. There's not much need to switch the webapp to split health checks for now. So I'd consider all work done here! :)

PR #187, opened 2018-05-25
Make results-processor more robust

Hexcles:
## Description

The results processor was suffering from concurrency issues, namely:

1. If only one worker is used, the processor cannot respond to the health checks when processing a result, which causes AppEngine to kill the instance.
2. If we simply increase the number of workers to 2, there's nothing preventing the TaskQueue (and admins via the dashboard) to send more more than 1 task to the processor at the same time, which will occupy the second worker and we are back to # 1.

The solution is to use a file lock to ensure each processor instance only processes one result file at a time, and the second worker can always respond to health checks. The new-style [health checks](https://cloud.google.com/appengine/docs/flexible/python/configuring-your-app-with-app-yaml#updated_health_checks) are also implemented along the way, which are actually trivial now that we have the file lock. (This fixes the results-processor part of #177 .)

Later if we need to scale the performance, we can scale linearly simply by increasing the number of instances.

## Change

1. Set the number of Gunicorn workers to 2.
2. Put the processor behind a file lock.
3. Implement health check endpoints and add them to `app.yaml`.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -78,7 +78,7 @@ def _internal_only(func):
 def liveness_check():
     lock = filelock.FileLock(LOCK_FILE)
     try:
-        lock.acquire(timeout=1)
+        lock.acquire(timeout=0.1)
         lock.release()
     except filelock.Timeout:
         try:
@@ -93,9 +93,14 @@ def liveness_check():
 
 
 @app.route('/_ah/readiness_check')
-@_serial_task
 def readiness_check():
-    return 'Ready to process results'
+    lock = filelock.FileLock(LOCK_FILE)
+    try:
+        lock.acquire(timeout=0.1)
+        lock.release()
+    except filelock.Timeout:
+        return 'A result is currently being processed.', 503
+    return 'Service alive'
 
 
 # Check request origins before acquiring the lock.


results-processor/Dockerfile
```diff
@@ -20,4 +20,7 @@ ADD requirements.txt /app/
 RUN pip install -r requirements.txt
 
 ADD . /app/
-CMD exec gunicorn -b :$PORT -t 3600 main:app
+# The number of workers should always be 2: one for processing tasks, the other
```
mdittmer: Doesn't this make health checks pretty useless? The worker responsible for servicing "real" requests could be unhealthy and we would never find out.

results-processor/main.py
```diff
@@ -12,13 +13,40 @@
 
 
 APPENGINE_INTERNAL_IP = '10.0.0.1'
+LOCK_FILE = '/tmp/results-processor.lock'
```
mdittmer: Should the `Dockerfile` contain any explicit permissions setting to ensure that the user running `main.py` has permission to write to this path?

Hexcles: Every user should be able to access `/tmp` on Linux. (I think it's required by [LSB](https://en.wikipedia.org/wiki/Linux_Standard_Base) but can't find the section right now.)

results-processor/Dockerfile
Hexcles: We kinda have to have two workers (to clarify, they are simply two processes running the same code) because of the way WSGI works (WSGI is very similar to the traditional CGI and has a synchronous design). I could rewrite the whole thing from ground up using `asyncio` to make it asynchronous, but that'd make it substantially more complicated.

The health checks are indeed not ideal. One of the workers can die or hang while processing the results without releasing the lock and the other worker will forever return busy, leaving the instance stuck at this state.

We don't need to rewrite the whole thing asynchronous to solve this problem. As suggested by @lukebjerring , in liveness checks, we can respond an error if the lock hasn't been modified for a while, and AppEngine will restart the instance. As for the threshold, we can use 1 hour as a starter (which should be enough to process one full report); later we can make some time-consuming external calls (namely `gsutil rsync`) non-blocking and touch the file every few minutes while waiting for the calls to complete, so that we can use a smaller threshold. (I'd like to leave the latter improvement in a future PR.)

results-processor/main.py
```diff
@@ -36,6 +43,15 @@ def decorated_func(*args, **kwargs):
...
 def liveness_check():
+    lock = filelock.FileLock(LOCK_FILE)
+    try:
+        lock.acquire(timeout=1)
+        lock.release()
+    except filelock.Timeout:
+        with open(lock.lock_file, 'rt') as f:
+            last_locked = float(f.readline().strip())
+        if time.time() - last_locked > TIMEOUT:
+            return 'The current result processing has taken too long.', 500
```
Hexcles: @lukebjerring @mdittmer PTAL

Hexcles:
@lukebjerring I figured out a way to avoid the read-while-writing race condition: POSIX guarantees the renaming operation to be atomic, so I write the timestamp to a temporary file first and rename the file. As a side effect, the timestamp now lives in a separate file as we can't replace the lock file (replacing a file would change its inode and another process can acquire the lock on the new inode).

Hexcles:
I'm admin-merging the PR because of #191

PR #190, opened 2018-05-28
Use 'from' param for /api/runs & /test-runs

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/3

Currently, we load 100 of each browser and try to 'match' them in the UI, which causes a misleading appearance after the first 100 runs of the most-frequent browsers.

This PR changes it to be all runs in the last 3 months.

## Review Information
https://runs-after-time-dot-wptdashboard.appspot.com/test-runs

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/deckarep/golang-set"
 )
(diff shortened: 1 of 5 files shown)


lukebjerring:
TODO: add some tests

lukebjerring:
Running at https://runs-after-time-dot-wptdashboard.appspot.com

foolip:
Looks like this doesn't quite work right yet. The last row in https://runs-after-time-dot-wptdashboard.appspot.com/test-runs has two blue boxes and leads to https://runs-after-time-dot-wptdashboard.appspot.com/results/?sha=a200a0fb41 which has runs for four browsers.

foolip:
> Looks like this doesn't quite work right yet.

I guess the problem is that the dates are run submission dates and not commit dates. It's probably going to look a bit funny at the very edge. Unfortunately, fixing this probably isn't straightforward right now, as the commit date isn't available anywhere...

shared/params.go
```diff
@@ -269,28 +270,43 @@ func GetProductsForRequest(r *http.Request) (products []Product, err error) {
...
+		return &count, nil
+	}
+	return nil, nil
+}
+
+// ParseMaxCountParamWithDefault parses the 'max-count' parameter as an integer, or returns the
+// default when no param is present, or on error.
+func ParseMaxCountParamWithDefault(r *http.Request, defaultValue int) (count int, err error) {
+	if maxCountParam, err := ParseMaxCountParam(r); maxCountParam != nil || err != nil {
+		return *maxCountParam, err
```
foolip: Won't this "segfault" if `err` is non-nil but `maxCountParam` is nil, as in the `return nil, err` case above? Is there a test for that case showing it not crashing? "http://wpt.fyi/?max-count=billions" I guess would do it?

lukebjerring: Fixeded

Issue #192, opened 2018-05-28
Prepare to switch to results-receiver in prod


Hexcles:
#146 is only one step away (calling the test run creation API) and will be done soon, so it's time to prepare for production switchover. With @lukebjerring 's recent effort of setting up a completely isolated staging project, the process is actually quite simple:

0. Deploy results-receiver to both the staging and prod projects. In each project, the receiver stores results to the main GCS bucket and Datastore respectively.
1. Generate and distribute tokens (also decide usernames) for runners.
2. Ask runners to send the full reports to the **staging** deployment of the new API **in addition to** the existing procedure.
3. Verify everything works.
4. Runners send results to **prod** API and **in the meantime** stop using the legacy upload process.

lukebjerring:
@mikewest - to clarify, this will mean that we are hoping you can upload the results twice (keeping the current behaviour, but also uploading the full report to staging.wpt.fyi).

We feel (strongly) that this makes more sense than trying to replicate existing client behaviour server-side and switching your client to staging-only uploads.

Hexcles:
Sorry, wrong Mike. It's @jugglinmike .

Hexcles:
All done :)

PR #196, opened 2018-05-29
Processor creates test runs after uploading processed results

Hexcles:
## Description

This PR implements the final piece of #146 , i.e. creating test runs.

We now have a fully isolated staging project (which gives us a separate Datastore) so we don't need to worry about polluting prod data. The processor now decides which GCS buckets to upload the processed results to depending on which project it's running in. And then the it will send a POST request to the `/api/run` endpoint of the *current* project, which is handled by the default Go app and will eventually create a test run in the Datastore of the project.

Also see #192 for the migration plan.

### GCS buckets Permissions

Since GCS buckets are all in a global namespace, we need to create separate buckets for the staging deployment.

* `wptd-staging` (in `wptdashboard-staging` project) is the staging counterpart of `wptd`, and is similarly publicly readable.
* `wptd-results-staging` (in `wptdashboard-staging` project) is the staging counterpart of `wptd-results`, and is similarly publicly readable.
* `wptd-results-buffer` (in **`wptdashboard`** project) is still used as the results buffer (when a task is in the queue) for the staging deployment: since we only need to append new data to and read existing data from the bucket, it's safe to share the bucket across prod and staging projects. The service account for `wptdashboard-staging` is given create-and-read permissions (no delete) to the bucket.

## Review Information

The PR is ready for review now. Try uploading results at https://wptdashboard-staging.appspot.com/admin/results/upload and the whole pipeline should work. You should see new runs at https://wptdashboard-staging.appspot.com/test-runs eventually (after ~20 minutes).

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -14,6 +14,7 @@ import tempfile
 
 import requests
 
+import config
 import gsutil
 
 
@@ -282,11 +283,8 @@ def create_test_run(report, secret, results_gcs_path, raw_results_gcs_path):
     payload['results_url'] = GCS_PUBLIC_DOMAIN + results_gcs_path
     payload['raw_results_url'] = GCS_PUBLIC_DOMAIN + raw_results_gcs_path
 
-    # If we are running in cloud, post to the current project.
-    project = os.getenv('GOOGLE_CLOUD_PROJECT') or DEFAULT_PROJECT
-
     response = requests.post(
-        "https://%s.appspot.com/api/run" % project,
+        config.project_baseurl() + '/api/run',
         params={'secret': secret},
         data=json.dumps(payload)
     )
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -17,6 +17,7 @@ import requests
 import gsutil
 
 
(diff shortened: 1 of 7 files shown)


lukebjerring:
Running at https://create-test-run-dot-wptdashboard-staging.appspot.com

results-processor/wptreport.py
```diff
@@ -256,18 +259,34 @@ def test_run_metadata(self):
...
-    # TODO(Hexcles): Do not hardcode the URLs.
     payload = report.test_run_metadata
-    payload['results_url'] = "https://storage.googleapis.com/wptd/%s".format(
-        report.sha_summary_path
-    )
+    payload['results_url'] = GCS_PUBLIC_DOMAIN + results_gcs_path
+    payload['raw_results_url'] = GCS_PUBLIC_DOMAIN + raw_results_gcs_path
+
+    # If we are running in cloud, post to the current project.
+    project = os.getenv('GOOGLE_CLOUD_PROJECT') or DEFAULT_PROJECT
```
lukebjerring: Falling back to production seems dangerous; maybe use staging as the default?

lukebjerring: Also, maybe make this a function in `config`

PR #200, opened 2018-05-29
Preserve the label param

### LINKED ISSUE #188: Query params not preserved when navigating
In https://wpt.fyi/results/?label=experimental, the first link goes to https://wpt.fyi/results/2dcontext, which initially works but upon reload gets you the stable browsers. Query params need to be preserved.

lukebjerring:
## Description
Resolves #188 for the `label` param.

Pipe through the param and append it when navigating.

## Review Information
https://label-param-persist-dot-wptdashboard-staging.appspot.com/results?label=chrome

### DIFF
--- a/shared/util.go
+++ b/shared/util.go
@@ -9,6 +9,8 @@ import (
 	"net/http"
 	"sort"
 	"strings"
+
+	"github.com/deckarep/golang-set"
 )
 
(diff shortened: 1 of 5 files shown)


lukebjerring:
Running at https://label-param-persist-dot-wptdashboard-staging.appspot.com

Issue #56, opened 2018-04-17
Use MANIFEST.json to show manual tests and tests with no results


foolip:
Moved from https://github.com/web-platform-tests/results-collection/issues/123.

Two related problems:
1. Some test suites have a lot of manual tests, and making this clear would be useful. Example: https://wpt.fyi/fullscreen/ vs. https://github.com/w3c/web-platform-tests/tree/master/fullscreen
2. Not all runners will run all test types (especially wdspec) and may have bugs that cause certain tests to have no recorded results at all.

These could be solved by somehow using MANIFEST.json for each commit, or information derived from it. That could also be used to do something more useful for https://wpt.fyi/does/not/exist/, although that's probably already doable, just not implemented.

foolip:
The discussion in https://github.com/web-platform-tests/results-collection/issues/98 might also be relevant to this, although we could at most align the tests, not the subtests.

lukebjerring:
From https://github.com/web-platform-tests/results-collection/issues/123, @gsnedders said:
> That relies on everyone implementing runners correctly. I think there's definitely an argument that it's the collection side that should handle it?

There's an imbalance here with WPT run not working cross-platform that means we really are likely to have multiple runner implementations (already do for Edge, Content Shell). I feel like if we can't necessarily trust the clients/runners to do the right thing, then the step for fleshing out "missing"/manual/un-run tests as per manifest can be shoved server-side, but I still think it needs to be (implicit) in the results JSON blobs, not implemented by storing a copy of the as-was manifest for every run (or retrospectively loading it from the repo, or other bug-prone approaches).

i.e. It should be part of the post-processing of the [results receiver](https://docs.google.com/document/d/1F3GWJ6lA8C3S5WgyJgXxmTFzu92mqF-1ddGije67ork/edit#).

foolip:
@Hexcles, WDYT, would it make sense for the results receiver to combine information from the run and the manifest to produce the final form suitable for wpt.fyi?

Hexcles:
I agree with @gsnedders and @lukebjerring here.

The benefits of implementing this inside the results-receiver is two-fold:
* Reducing the total complexity (implement once as opposed to in all the runners; and we know there are already >1 runners.)
* Not requiring runners to be 100% reliable (results-receiver would be the final catch-all check; also, fewer implementations lead to lower possibility of bugs and faster bug fixes.)

[...]

PR #206, opened 2018-05-30
Use manifest as source of test paths

lukebjerring:
## Description
Work towards https://github.com/web-platform-tests/wpt.fyi/issues/56

Asynchronously loads the latest manifest and uses that as the source-of-truth on the exhaustive list of tests that execute.

## Review Information
See homepage of auto-deployed environment.

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -503,6 +503,16 @@ found in the LICENSE file.
               };
               this.fetchDiff();
             }
+
+            // Load a manifest.
+            if (this.fetchManifestForTestList) {
+              const shas = (runs || []).reduce((set, run) => {
+                set.add(run.revision);
+                return set;
+              }, new Set());
+              const sha = shas.size == 1 ? Array.from(shas)[0] : 'latest';
+              this.fetchManifestForSHA(sha);
+            }
           }),
           () => {
             this.resultsLoadFailed = true;
@@ -544,7 +554,9 @@ found in the LICENSE file.
             this.searchResults = json.results;
             this.refreshDisplayedNodes();
           }),
-          () => {
+          (e) => {
+            // eslint-disable-next-line no-console
+            console.log(`Failed to load: ${e}`);
             this.resultsLoadFailed = true;
           }
         );
@@ -569,19 +581,33 @@ found in the LICENSE file.
         );
       }
 
+      fetchManifestForSHA(sha) {
+        const url = new URL('/api/manifest', window.location);
+        const isSpecificSHA = sha && !this.computeIsLatest(sha);
+        if (isSpecificSHA) {
+          url.searchParams.set('sha', sha);
+        }
+        this.load(
+          fetch(url).then(
+            async r => {
+              if (!r.ok) {
+                // Fall back to the latest manifest if we 404 for a specific SHA.
+                return r.status === 404
+                  && isSpecificSHA
+                  && this.fetchManifestForSHA('latest')
+              }
+              let manifest = await r.json();
+              manifest.sha = sha || r.headers && r.headers['x-wpt-sha'];
+              this.manifest = manifest;
+              this.refreshDisplayedNodes();
+            }
+          )
+        );
+      }
+
       pathUpdated(path) {
         super.pathUpdated(path);
         this.refreshDisplayedNodes();
-        this.fetchManifest(path);
-      }
-
-      async fetchManifest(path) {
-        if (!this.computePathIsATestFile(path)) {
-          return;
-        }
-        path = this.encodeTestPath(path);
-        this.manifest = null; // Clear ASAP.
-        this.manifest = await fetch(`/api/manifest?path=${path}`).then(r => r.json());
       }
 
       nodeSort(a, b) {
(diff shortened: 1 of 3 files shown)


lukebjerring:
Running at https://manifest-test-list-dot-wptdashboard-staging.appspot.com

webapp/components/wpt-results.html
```diff
@@ -372,13 +377,13 @@
...
 
       async fetchManifest(path) {
-        if (!this.computePathIsATestFile(path)) {
-          return;
-        }
-        this.manifest = null; // Clear ASAP.
         this.manifest = await fetch(`/api/manifest?path=${path}`).then(r => r.json());
       }
 
+      computeManifest(manifests, path) {
```
foolip: Looks like `computeManifest` is unused. In case it isn't: Is the plural `manifests` here a typo? Or is it slicing things out of a bunch of manifests, one per run?

webapp/components/wpt-results.html
```diff
@@ -372,13 +377,13 @@
       }
 
       async fetchManifest(path) {
-        if (!this.computePathIsATestFile(path)) {
-          return;
-        }
-        this.manifest = null; // Clear ASAP.
         this.manifest = await fetch(`/api/manifest?path=${path}`).then(r => r.json());
```
foolip: Before the manifest was used only in `computeSourcePath` and `computeTestCanBeRunOnW3C` which made a revision mismatch prettty harmless. Now, https://manifest-test-list-dot-wptdashboard-staging.appspot.com/results/?sha=82bf967918 ends up looking quite different to https://wpt.fyi/results/?sha=82bf967918, including async-local-storage and other directories which didn't exist at the time.

How to fix? I guess the only way is to include the sha here, and do something (what?) if showing mixed results involving multiple wpt shas.

webapp/components/wpt-results.html
```diff
@@ -401,16 +406,30 @@
...
         };
 
-        Object.keys(this.testFiles).forEach(testFileName => {
+        let paths = Object.keys(this.testFiles);
+        if (this.manifest) {
+          paths = [];
+          for (const files of TEST_TYPES.map(i => this.manifest.items[i]).filter(i => i)) {
+            for (const items of Object.values(files)) {
+              for (const item of items) {
+                paths.push(item[0]); // First item array value is url
```
foolip: Yes, this is the good stuff! https://manifest-test-list-dot-wptdashboard-staging.appspot.com/results/ now includes accname and other directories with only manual tests, and all my manual tests in https://manifest-test-list-dot-wptdashboard-staging.appspot.com/results/fullscreen/api.

Visually, the missing results are just white boxes, and I think it becomes important with this change to show the not-run tests differently somehow. Maybe just count directories with no tests as 0 / 0? At the file level, I guess "not run"?

The subtest level is a pre-existing problem. NOTRUN is a possible status for tests found but not run, so maybe call it "missing", like the stability checker does? I suppose that "missing" at the other levels would also work.

WDYT?

webapp/components/wpt-results.html
lukebjerring: I'll follow up this PR with forwarding the SHA to the manifest request (since we now use complete runs).

webapp/components/wpt-results.html
lukebjerring: I was toying with keeping the manifests when changing paths, but decided that was overly complex.

webapp/components/wpt-results.html
```diff
@@ -401,16 +406,30 @@
...
+              }
+            }
+          }
+        }
+        console.log(paths);
```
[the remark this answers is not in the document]

webapp/components/wpt-results.html
lukebjerring: Happy to tweak appearance. Could you whip up some mocks?

webapp/components/wpt-results.html
foolip: Complete runs yay! Not deploying this change to https://wpt.fyi/results/?sha=82bf967918 before that's fixed would be good, since it'll make a mess of existing links. And that does require handling the case where the manifest isn't available, but maybe that's already fine?

foolip:
Mockup for directory with no tests run, accname here:
screenshot from 2018-05-31 15-24-09

For the denominator, the choice will be to either include unrun tests or not. Thinking about it, I think we probably should. So accname should ideally say "0 / 80" or something.

foolip:
At the test and subtest level, I suggest gray and the word "MISSING" (no screenshot because I have to run to a meeting)

lukebjerring:
OK on second though, the appearance issue can be separated from this PR, since we should probably also highlight things like test types. More thought required. #211

lukebjerring:
Re:
> For the denominator, the choice will be to either include unrun tests or not. Thinking about it, I think we probably should. So accname should ideally say "0 / 80" or something.

This PR should address that (i.e. the Manifest should be used as the 'total' count here, too).

foolip:
> Discussion with @Hexcles - we don't know how many subtests (will) run, so the total would have to be the number of top-level tests, which isn't necessarily valuable. Additionally, we use "0 / N" as pass rate elsewhere in the screen, so showing `0 / 80` as you suggested would be confused with "0 / 80 tests pass".

Right, this suggestion only really makes sense together with also changing the numbers across the board to reflect manifest-level tests and not subtests. Whether that would be an improvement or not is hard to say, but I'm cautiously optimistic.

lukebjerring:
Update: I'd considered dropping this PR, but now that we support feature-flags, I might bring it up to speed with `master` and revive the thread.

lukebjerring:
Ok, @foolip I've changed this PR to be behind a feature flag on the client, which means we can happily merge it as some kind of debug/beta feature.

lukebjerring:
Re: The appearance of `0/0` outcomes: That's highlighted by this PR, but actually already an existing problem for some bad runs, and we should discuss + fix it in another PR.

webapp/components/wpt-results.html
```diff
@@ -503,6 +503,16 @@ <h3>
               };
               this.fetchDiff();
             }
+
+            // Load a manifest.
+            if (this.fetchManifestForTestList) {
+              const shas = (runs || []).reduce((set, run) => {
```
foolip: Unless this is an optimization, then something like `const shas = runs ? new Set(runs.map(run = > run.revision)) : null` would be easier to follow.

webapp/components/wpt-results.html
```diff
@@ -569,19 +581,33 @@ <h3>
...
+        const url = new URL('/api/manifest', window.location);
+        const isSpecificSHA = sha && !this.computeIsLatest(sha);
+        if (isSpecificSHA) {
+          url.searchParams.set('sha', sha);
+        }
+        this.load(
+          fetch(url).then(
+            async r => {
+              if (!r.ok) {
+                // Fall back to the latest manifest if we 404 for a specific SHA.
```
foolip: Can you log a warning in this case?

webapp/components/wpt-results.html
```diff
@@ -569,19 +581,33 @@ <h3>
...
+                return r.status === 404
+                  && isSpecificSHA
+                  && this.fetchManifestForSHA('latest')
+              }
+              let manifest = await r.json();
+              manifest.sha = sha || r.headers && r.headers['x-wpt-sha'];
```
foolip: Drop `x-` prefix in header.

webapp/components/wpt-results.html

webapp/components/wpt-results.html
lukebjerring: kept `(runs || [])` which is more succinct, but took the ctor suggestion.

PR #208, opened 2018-05-30
Add documentation for results receiver API

Hexcles:
Please use the rich diff to review: https://github.com/web-platform-tests/wpt.fyi/pull/208/files?short_path=d4b47e8#diff-d4b47e85ce206fdbe458cb599edf2a90


Hexcles:
cc @jugglinmike as you requested earlier, here's the doc with mandatory fields explicitly listed.

lukebjerring:
Running at https://receiver-doc-dot-wptdashboard-staging.appspot.com

api/README.md
```diff
@@ -100,3 +100,55 @@ The high-level structure of the `v4` manifest is as follows:
...
+    ...
+  }
+}
+```
+
+Note that `run_info.{revision,product,browser_version,os}` are required, and should be automatically
+generated by `wpt run`. If for some reason the report does not contain these fields (e.g. old WPT
+version, Sauce Labs, or custom runners), they can be overridden with the following *optional*
+parameters in the POST payload (this is NOT recommended; please include metadata in the reports
+whenever possible):
```
jugglinmike: I can feel your distaste in these words :) If this is only for the sake of the current results reporter, remove it! It's looking like it's easier for me to insert this information into the report itself, anyway

Hexcles: I provided this override mostly for debugging (before I added the metadata to wptreport). And as you can see, the inconsistency of names is not great...

Yeah it should be easy for a custom wptrunner to add these fields. The only concern for now is the Sauce Labs. Ideally, reports generated by the official `wpt run` should be compatible with the receiver. Let me see if I can fix it from wptrunner.

jugglinmike: > Let me see if I can fix it from wptrunner.

Only if you're feeling bored. Remember that consolidating multiple reports, so I'm already responsible for manually building this data structure. It's really simple for me to insert the fields that are missing--honest!

PR #209, opened 2018-05-30
Try to fix Travis staging deployment

Hexcles:
Something was wrong with the conditions. The two staging deployment should never run at the same time.

Travis docs: https://docs.travis-ci.com/user/conditional-builds-stages-jobs/

// Not sure if this will fix the problem. Let's wait for the Travis build.


lukebjerring:
Running at https://travis-deployment-dot-wptdashboard-staging.appspot.com

PR #210, opened 2018-05-31
Announcer API: Shared controller and HTTP handlers (+ tests)

mdittmer:
## Description

Shared state controller (owner of `Announcer` instance) and handlers for public announcer API.


mdittmer:
Note: https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/386222196#L7402 is not a problem with the code or tests.

lukebjerring:
Can you rebase on master so the behaviour of Travis is current (and probably passing)?

revisions/api/handlers/latest.go
```diff
@@ -0,0 +1,57 @@
...
+		At:    now,
+		Start: now.Add(-2 * epochs[0].GetData().MaxDuration),
+	})
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write(a.ErrorJSON(err.Error()))
+		return
+	}
+
+	response, err := api.LatestFromEpochs(revs)
```
lukebjerring: nit: One-line the `if`, and use `http.Error` (might need a `[]byte` cast)
    if response, err := ...; err != nil {
        http.Error(w, a.ErrorJSON(err.Error()), http.StatusInternalServerError)
        return
    }

revisions/api/handlers/list.go
```diff
@@ -0,0 +1,150 @@
...
+	if len(epochs) == 0 {
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write(a.ErrorJSON("No epochs"))
+		return
+	}
+
+	q := r.URL.Query()
+
+	numRevisions := 1
+	if nr, ok := q["num_revisions"]; ok {
```
lukebjerring: - `Query` has a `Get("num_revisions")` method which returns an empty string if missing.
- Probably just parse as an int, then assert size limits.
- Move the `num_revisions` param parsing to the `params.go` file

revisions/api/handlers/latest.go
mdittmer: Done (push forthcoming).

revisions/api/handlers/list.go
mdittmer: I don't quite follow on this one. Let's sync offline.

lukebjerring: Ignore the last 2 items there - I misread the `len(1)` assertion as being the number of chars lol.
But yeah, no need to fail for multiple copies - `Get` will only give you the last one.

Hexcles:
Hooray!

Issue #213, opened 2018-05-31
Migrate from nested TestRun models to TestRun ids


lukebjerring:
Blocking https://github.com/web-platform-tests/wpt.fyi/issues/212

TestRun metadata is duplicated in the metrics models, which also causes a slice of a slice when trying to add `Labels []string` to `TestRun`.

To work around this, we should be storing the IDs of the TestRun objects that generated the interop data (i.e. `PassRateMetadata`) instead of the metadata itselt. This requires `GET` by id in `/api/run`.

PR #215, opened 2018-05-31
Migrate webapp to use a gorilla/mux Router

lukebjerring:
## Description
Migrate the webapp's routing to leverage gorilla/mux, a more powerful routing library. Unblocks #213 by supporting an `{id}` in the path.

@foolip - note that this highlighted a poor choice of nesting `/interop/anomalies` and thus migrates it to `/anomalies`

## Review Information

### DIFF
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -7,12 +7,20 @@ package webapp
 import (
 	"html/template"
 
+	"github.com/web-platform-tests/wpt.fyi/api"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
 var templates = template.Must(template.ParseGlob("templates/*.html"))
 
 func init() {
+	// webapp.RegisterRoutes has a catch-all, so needs to go last.
+	api.RegisterRoutes()
+	RegisterRoutes()
+}
+
+// RegisterRoutes adds the route handlers for the webapp.
+func RegisterRoutes() {
 	// About wpt.fyi
 	shared.AddRoute("/about", "about", aboutHandler)
 
@@ -27,10 +35,11 @@ func init() {
 	shared.AddRoute("/test-runs", "test-runs", testRunsHandler)
 
 	// Admin-only manual results upload.
-	shared.AddRoute("/admin/results/upload", "upload", adminUploadHandler)
+	shared.AddRoute("/admin/results/upload", "admin-results-upload", adminUploadHandler)
 
 	// Test run results, viewed by browser (default view)
 	// For run results diff view, 'before' and 'after' params can be given.
+	shared.AddRoute("/results/", "results", testResultsHandler)
 	shared.AddRoute("/results/{path}", "results", testResultsHandler)
 
 	// Legacy wildcard match
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -20,7 +20,6 @@ func init() {
 	shared.AddRoute("/anomalies", "anomaly", anomalyHandler)
 
 	// Test run results, viewed by pass-rate across the browsers
-	shared.AddRoute("/interop", "interop", interopHandler)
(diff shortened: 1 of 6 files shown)


lukebjerring:
Running at https://mux-routes-dot-wptdashboard-staging.appspot.com

foolip:
> @foolip - note that this highlighted a poor choice of nesting /interop/anomalies and thus migrates it to /anomalies

That makes more sense to me! Still nothing to see at /anomalies, but maybe one day :)

webapp/routes.go
```diff
@@ -13,24 +13,27 @@ import (
...
-	// Test run results, viewed by pass-rate across the browsers
-	shared.AddRoute("/interop/", interopHandler)
+	shared.AddRoute("/about", "about", aboutHandler)
 
 	// Lists of test run results which have poor interoperability
-	shared.AddRoute("/interop/anomalies", anomalyHandler)
+	shared.AddRoute("/anomalies", "anomaly", anomalyHandler)
+
+	// Test run results, viewed by pass-rate across the browsers
+	shared.AddRoute("/interop/", "interop", interopHandler)
```
Hexcles: Why is this the only route that has a trailing slash?

webapp/routes.go
```diff
@@ -13,24 +13,27 @@ import (
...
+	shared.AddRoute("/interop/{path}", "interop", interopHandler)
 
 	// List of all test runs, by SHA[0:10]
-	shared.AddRoute("/test-runs", testRunsHandler)
+	shared.AddRoute("/test-runs", "test-runs", testRunsHandler)
 
 	// Admin-only manual results upload.
-	shared.AddRoute("/admin/results/upload", adminUploadHandler)
+	shared.AddRoute("/admin/results/upload", "upload", adminUploadHandler)
```
Hexcles: Should this be named "admin-results-upload" for consistency?

webapp/routes.go
```diff
@@ -13,24 +13,27 @@ import (
...
+
+	// Test run results, viewed by browser (default view)
+	// For run results diff view, 'before' and 'after' params can be given.
+	shared.AddRoute("/results/{path}", "results", testResultsHandler)
+
+	// Legacy wildcard match
+	shared.AddRoute("/", "results-legacy", testResultsHandler)
+	shared.AddRoute("/{path}", "results-legacy", testResultsHandler)
```
Hexcles: Could you explain to me (or point me to a doc) how Gorilla mux handles multiple matches? For example here, `"/{path}"` can also match `/interop`, `/test-runs`, etc. Is there a priority?

lukebjerring: https://github.com/gorilla/mux#matching-routes

> Routes are tested in the order they were added to the router. If two routes match, the first one wins

webapp/routes.go
lukebjerring: Sure, though the name's not really needed for anything other than the unit tests.

webapp/routes.go
Hexcles: Thanks for the pointer!

Since we set up routes in `init()`, how do we know for sure which `init()` runs first? For example, if `webapp/routes.go` sets up before `api/routes.go`, would this catch-all rule render all the routes in `api/routes.go` useless?

webapp/routes.go
lukebjerring: A trailing slash and `StrictSlash(true)` mean that the slash is optional, and a redirect is added from the non-slash to the slash.

The test `path` part of the url after `/results` or `/interop` is expected in the UI, so a trailing slash is used in these paths.

webapp/routes.go
lukebjerring: Good point. I think this has been a potential issue the whole time (not just before gorilla/mux), so I've made the calling order explicit during the `init()` call of `webapp`

PR #216, opened 2018-06-01
[processor] Remove the cached gce flag in Docker

### LINKED ISSUE #203: [processor] AppEngine default credentials don't show up immediately after a Flex instance starts
Not sure why, but it seems to take a few minutes for the application default creds to show up inside the Docker container for a Flex service after it's deployed.

@mdittmer suggested to use a dedicated service account and configure the credentials explicitly. I'll look into that.

Hexcles:
This fixes #203. See the comment in Dockerfile for details.


lukebjerring:
Running at https://fix-service-account-dot-wptdashboard-staging.appspot.com

Hexcles:
FYI @mdittmer I asked you about this last week. This might also be useful knowledge to you since you also use Flex.

PR #218, opened 2018-06-04
Add an ID field to the TestRun model.

lukebjerring:
Working towards #213 

## Description

Adds an `ID` field to the `TestRun` struct, which is manually populated after loading from datastore, and ignored by datastore (to avoid duplicating the field).

## Review Information
Load https://test-run-id-field-dot-wptdashboard-staging.appspot.com/api/runs and witness `id` in the JSON.


lukebjerring:
Running at https://test-run-id-field-dot-wptdashboard-staging.appspot.com

Issue #220, opened 2018-06-04
/api/manifest/?sha=latest is unreliable, 404s often


tabatkins:
It looks like there's a race condition in the reporting of the "latest" manifest - sometimes hitting the endpoint works great, other times it returns a 404 and reports that there's no results for a particular sha.

I suspect that while it's building a new manifest, the new sha is available in such a way that `?sha=latest` can grab it, but the actual data isn't ready yet.  We should be exposing only the latest *available* manifest data.

tabatkins:
Hm, and even when it *does* come up in some places (currently, the latest SHA *is* coming up for me in my browser) it doesn't come up in others (the exact same URL is failing in `bikeshed update`).

tabatkins:
What really confuses me is that the error message I get is

```
No search results found for SHA 4f7d09666f738d1db19ba49721b2db4330bf4e82
```

...which is only emitted by [/api/manifest.go#L157](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/manifest.go#L157), but that particular code path looks like it's only hit when you call the function with an *actual* sha, *not* "latest" or "".

(And now it won't load in the browser, but it did load, once, for `bikeshed update`, but not it's back to not loading in the console... SHA is unchanged, so I dunno *what* is going on.)

Hexcles:
@lukebjerring when I was investigating the 500s in my results receiver just now, I found it was caused by OOM (#221 ). On a hunch, I thought it could also cause manifest API to fail. And after examining the logs, the flakiness of this API is indeed caused by OOM. Example:

```
GET /api/manifest?path=/css/css-flexbox/auto-margins-001.html
Exceeded soft private memory limit of 128 MB with 136 MB after servicing 1553 requests total
After handling this request, the process that handled this request was found to be using too much memory and was terminated. This is likely to cause a new process to be used for the next request to your application. If you see this message frequently, you may have a memory leak in your application.
```

I don't know how much memory the manifest API needs. Instead of spending time on trial and error, shall we just go for the largest instance (1GB memory)? After monitoring the memory usage for a while, we can scale it down if we don't need that much memory.

Hexcles:
BTW, inside the API, "latest" gets translated to a full hash which is (or close to, due to caching) the latest commit on WPT master. This is where the SHA comes from. @lukebjerring can tell you more.

To clarify, I'm not sure OOM is the root cause here. It's definitely an issue, but there might be others.

[...]

PR #221, opened 2018-06-04
Increase the memory limit of webapp

Hexcles:
## Description

Both the results receiver and the manifest API need a lot more memory than the default 128MB:
* The receiver needs to fetch wptreport from a remote URL to memory if the runner provides a URL instead of uploading a file.
* The manifest API needs to deserialize the WPT manifest JSON, which is also pretty big.

1GB is the largest memory AppEngine offers. We might not need that much, but we can scale down later based on monitoring.

This should fix or alleviate #220 .

## Review Information

Once the staging deployment is up, try the manifest API a few times and examine the logs. There should be no OOM.

### DIFF
--- a/webapp/app.yaml
+++ b/webapp/app.yaml
@@ -4,6 +4,7 @@
 
 runtime: go
 api_version: go1
+instance_class: F4_1G
 
 builtins:
 - remote_api: on


lukebjerring:
Running at https://increase-memeory-limit-dot-wptdashboard-staging.appspot.com

PR #223, opened 2018-06-04
Silence gloud install script

Hexcles:
The script prints out a few thousand lines of logs. They were put inside
a pair of Travis folding marks, but are still too noisy (e.g. folding
doesn't take effect when logs are still being streamed, or when Makefile
is used locally).

In the Dockerfile of results receiver, the stdout of this script is
directed to /dev/null, which seems to be a good compromise so far as
some of the more useful information is in stderr, so let's do this in
the project Makefile, too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,15 +158,13 @@ node: curl gpg
 	fi
 
 gcloud: python curl gpg
-	@ echo "travis_fold:start:gcloud_install"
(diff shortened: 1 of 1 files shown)


lukebjerring:
Running at https://silence-gcloud-install-dot-wptdashboard-staging.appspot.com

Issue #224, opened 2018-06-05
/api/results is undocumented


tabatkins:
The `/api/README` explains which endpoints are available, but it (a) has incorrect parameters in several places (it claims `browser` in several places, but actually wants `product`), and (b) gives zero explanation as to what the returned JSON's structure will be.

For example, `/api/results?product=chrome` returns an object with entries like `"/css/css-text/i18n/css3-text-line-break-opclns-213.html": [1, 1]`. I have no idea what that array means!

I also don't see a way to get a list of products, so I don't know how to query the results per product. Is that information contained in one of the other endpoints' data?

foolip:
https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md is quite extensive now, but /api/results is till undocumented. Leaving issue open for that.

lukebjerring:
There is (now) a basic description, so closing this.

PR #226, opened 2018-06-05
Improve API docs

lukebjerring:
PR aims to address some of the issues listed in https://github.com/web-platform-tests/wpt.fyi/issues/224

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -21,13 +21,55 @@ __Parameters__
 
 __`sha`__ : SHA[0:10] of the runs to get, or the keyword `latest`. Defaults to `latest`.
 
+__`product`__ : Product to include (repeated param), e.g. `chrome` or `firefox-60`
+
 __`from`__ : RFC3339 timestamp, for which to include runs that occured after the given time.
 
 __`max-count`__ : Maximum number of runs to get (for each browser). Only relevant when `sha` is `latest`. Maximum of 500.
 
+#### Examples
+
+- https://wpt.fyi/api/runs?product=chrome&product=safari
+- https://wpt.fyi/api/runs?product=chrome&from=2018-01-01T00:00:00Z&max-count=10
+
+__Example JSON__
+
+    [
+      {
+        "browser_name": "chrome",
+        "browser_version": "67.0.3396.62",
+        "os_name": "linux",
+        "os_version": "4.4",
+        "revision": "2bd11b91d4",
+        "full_revision_hash": "2bd11b91d490ddd5237bcb6d8149a7f25faaa101",
+        "results_url": "https://storage.googleapis.com/wptd/2bd11b91d4/chrome-stable-linux-summary.json.gz",
+        "created_at": "2018-06-05T08:27:30.627865Z",
+        "raw_results_url": "https://storage.googleapis.com/wptd-results/2bd11b91d490ddd5237bcb6d8149a7f25faaa101/chrome_67.0.3396.62_linux_4.4/report.json"
+      }
+    ]
+
 ### /api/runs/{id}
 
 Gets a specific (single) TestRun metadata by its datastore ID.
+
+#### Example
+
+https://wpt.fyi/api/runs/5164888561287168
+
+__Example JSON__
+
+    {
+      "id": "5164888561287168",
+      "browser_name": "chrome",
+      "browser_version": "67.0.3396.62",
+      "os_name": "linux",
+      "os_version": "4.4",
+      "revision": "2bd11b91d4",
+      "full_revision_hash": "2bd11b91d490ddd5237bcb6d8149a7f25faaa101",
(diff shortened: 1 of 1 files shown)


lukebjerring:
Running at https://docs-dot-wptdashboard-staging.appspot.com

api/README.md
```diff
@@ -21,14 +21,56 @@ __Parameters__
 
 __`sha`__ : SHA[0:10] of the runs to get, or the keyword `latest`. Defaults to `latest`.
 
+__`product`__ : Product to include (repeated param), e.g. `chrome` or `firefox-60`
+
 __`from`__ : RFC3339 timestamp, for which to include runs that occured after the given time.
 
 __`max-count`__ : Maximum number of runs to get (for each browser). Only relevant when `sha` is `latest`. Maximum of 500.
```
tabatkins: What's the relevance of `max-count`, and why does it only have an affect for `latest`?

api/README.md
```diff
@@ -62,12 +123,34 @@ __`filter`__ : Differences to include in the summary.
...
+    {
+      "/path/to/test.html": [1, 1],
+    }
+
+Where the array contains [`number of passes`, `total tests`].
+
 __Parameters__
 
-__`browser`__ : Browser to fetch the results for, e.g. `chrome`
+__`product`__ : Product to fetch the results for, e.g. `chrome-66`
```
tabatkins: This one has a version specifier, but previous `product` examples just had an unversioned `chrome`. How do I know what "products" exist, and what I should be asking for?

api/README.md
```diff
@@ -81,6 +164,8 @@ NOTE: The full SHA of the fetched manifest is returned in the HTTP response head
 
     x-wpt-sha: abcdef0123456789abcdef0123456789abcdef01
 
+__Response format__
+
```
tabatkins: This example JSON is missing several keys.

The high-level object has "items", "paths", "url_base", and "version".  Inside of "items", we have 8 keys: "manual", "reftest", "reftest_node", "stub", "support", "testharness", "visual", and "wdspec".  I *think* I can ignore reftest_node, stub, and support if I'm trying to just get at all the test paths?

api/README.md
```diff
@@ -21,14 +21,56 @@ __Parameters__
...
+      "browser_version": "67.0.3396.62",
+      "os_name": "linux",
+      "os_version": "4.4",
+      "revision": "2bd11b91d4",
+      "full_revision_hash": "2bd11b91d490ddd5237bcb6d8149a7f25faaa101",
+      "results_url": "https://storage.googleapis.com/wptd/2bd11b91d4/chrome-stable-linux-summary.json.gz",
+      "created_at": "2018-06-05T08:27:30.627865Z",
+      "raw_results_url": "https://storage.googleapis.com/wptd-results/2bd11b91d490ddd5237bcb6d8149a7f25faaa101/chrome_67.0.3396.62_linux_4.4/report.json"
+    }
+
```
tabatkins: What *is* a "run"? What does this data represent?

api/README.md
```diff
@@ -21,14 +21,56 @@ __Parameters__
 
 __`sha`__ : SHA[0:10] of the runs to get, or the keyword `latest`. Defaults to `latest`.
 
+__`product`__ : Product to include (repeated param), e.g. `chrome` or `firefox-60`
```
tabatkins: `product` appears to be repeatable here and elsewhere, but this isn't indicated.

api/README.md
lukebjerring: If you select a specific SHA, you'll (probably) be getting a single run... techincally, though, there can be multiple runs for the same SHA so this comment is not strictly true. I'll remove it.

api/README.md
lukebjerring: We don't (currently) have a way to fetch "all the products".

FWIW, your use case the products (unversioned) are what you'd want to call with, i.e.

- `chrome`
- `edge`
- `firefox`
- `safari`

Since it's ordered by date desc, that's the latest version of each.

api/README.md
lukebjerring: The docs here are limited deliberately since web-platform-tests/wpt would prefer to be able to freely change the manifest format and treat the python code as the source-of-truth and only handler.

api/README.md
lukebjerring: It's repeatable for `runs` but not for a single `run`.

api/README.md
lukebjerring: Added an overview, see `#test-run-entities`

api/README.md
tabatkins: All right, in that case the important missing bit is that "visual" is a test type but isn't in the list here.

lukebjerring: Good spotting. https://github.com/web-platform-tests/wpt.fyi/pull/251

PR #230, opened 2018-06-05
Mount volume that includes local Go code for results-analysis

lukebjerring:
Allows for dependent development in both packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,8 +37,9 @@ ENV PATH=$PATH:/usr/local/go/bin:$GOPATH/bin:${USER_HOME}/.local/bin:${GCLOUD_PA
 RUN apt-get update && apt-get install sudo make
 
 # Put wpt.fyi code in GOPATH
-RUN mkdir -p "${GOPATH}/src/github.com"
-RUN ln -s "${WPT_PATH}" "${GOPATH}/src/github.com/web-platform-tests"
+RUN mkdir -p "${GOPATH}/src/github.com/web-platform-tests"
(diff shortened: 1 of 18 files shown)


lukebjerring:
Running at https://analysis-docker-mount-dot-wptdashboard-staging.appspot.com

lukebjerring:
Good point. I'd gone of my structure, `~/[user]/[repo]/ so figured it isolated to `web-platform-tests`. Perhaps we should just explicitly add both (i.e. `../results-analysis` too). WDYT?

Hexcles:
I have `~/github/[repos]` on my machine. This would affect those who clone repos to HOME the most.

Your proposal sounds good to me. We can keep your new directory structure within Docker, but when starting the instance, we explicitly mount two volumes (wpt.fyi and results-analysis) instead of mapping their parent directory.

lukebjerring:
Done :)

Hexcles:
I think this unfortunately breaks master (and all pending PRs). [Example failure](https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/388459432):
```
cd /home/user/go/src/github.com/web-platform-tests/wpt.fyi; go get -t -tags="small medium large" ./...
package github.com/web-platform-tests/results-analysis/metrics: /home/user/go/src/github.com/web-platform-tests/results-analysis exists but /home/user/go/src/github.com/web-platform-tests/results-analysis/.git does not - stale checkout?
Makefile:123: recipe for target 'go_packages' failed
```

My understanding is that on Travis `results-analysis` isn't checked out so this PR would link `/home/user/go/src/github.com/web-platform-tests/results-analysis` to an empty directory. I don't have a good fix off the top of my head, so I'm reverting the change.

PR #231, opened 2018-06-05
Add TestRuns helper type

lukebjerring:
Alias for a `[]TestRun` with a helper method for getting IDs from the entities (in one line).


lukebjerring:
Running at https://testruns-array-dot-wptdashboard-staging.appspot.com

PR #233, opened 2018-06-05
Use full revision hash and include a short SHA of the report content in the results URL

### LINKED ISSUE #54: https://wpt.fyi/api/runs?max-count=100 has duplicate entries for some runs
Moved from https://github.com/web-platform-tests/results-collection/issues/528, examples of duplication there.

After @Hexcles results receiver is in place, this will be a wpt.fyi issue. It's possible we should just accept duplicate runs, but if so we need to store all of them and have a bucket structure that allows for it.

Hexcles:
## Description

This is a slightly risky change, intended to fix #54 and partly #49 . The results URL used to be:
`[GCS]/[short_revision]/[product-id]-summary.json.gz`
with this change they become:
`[GCS]/[full_revision]/[product-id]-[short SHA of the report]-summary.json.gz`
(where `[short SHA of the report]` is the first 10 hex digits of the sha1sum of the decompressed full report, serving as a unique yet deterministic identifier).

And the paths for split JSONs are changed accordingly.

The change should be backward compatible. The existing results URL should still work.

## Review Information

Please wait for the staging deployment. And then I will manually upload a test run in the staging deployment to verify everything works. I'll post more details later.

## Changes

Two pieces need to be changed:
* The frontend needs to stop parsing the URL. AFAIK, there's only one place parsing the results URL (`test-file-results.html`). Luckily, it turns out the code doesn't really need to parse the URL, so I simply removed it.
* The results processor was changed to produce the new paths. (A buffered hash helper was added to the module to calculate the hash the report.)


Hexcles:
This might interfere with @mdittmer 's legacy result migrator. Can you give me a pointer to the code so that I can check?

lukebjerring:
Running at https://full-sha-and-uid-dot-wptdashboard-staging.appspot.com

PR #234, opened 2018-06-05
Revert "Mount volume that includes local Go code for results-analysis (#230)"

Hexcles:
This reverts commit a2b3f48b6f72d6be876c56d420762ee01f974af1. (PR #230)


Hexcles:
@lukebjerring I tried to revert in the hope that it would quickly unblock master and other PRs, but given that Travis seems to have a hard time at the moment (long queue) and is unlikely to build my PR soon, we don't need to rush and can fix it properly tomorrow.

lukebjerring:
Reverting is always the quickest + cleanest approach :)

PR #238, opened 2018-06-06
Fix staging deployment and tweak some logging

### LINKED ISSUE #229: Travis shouldn't try to deploy staging for PRs coming from forks
This is from #219, a PR sent from a fork of the repo instead of a branch in the repo. Travis shouldn't try to deploy in such case; otherwise, the deployment would always fail because of lack of secrets.

We do have check in place: https://github.com/web-platform-tests/wpt.fyi/blob/master/util/travis-deploy-staging.sh#L30 , but looks like it's not working as intended.

Hexcles:
## Description

This PR tries to fix the staging deployment for `results-processor` (by quoting some variables that might contain spaces), as well as to fix #229 (by inverting the check for `TRAVIS_SECURE_ENV_VARS` so that it also works when the env var doesn't exist).

## Changes

I also made some drive-by changes to improve logging, mostly to make `curl` and `apt-get` quieter.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ script:
     # Deploy PR to staging environment (only when Travis secrets are available).
     # Note: Done here (in 'script', not 'deploy') because we need deploy to happen before staging webdriver test.
     if [ ${DEPLOY_STAGING} == true ]; then
-      if [ "${TRAVIS_BRANCH}" == "master" ]; then
+      if [ "${TRAVIS_PULL_REQUEST}" == "false" -a "${TRAVIS_BRANCH}" == "master" ]; then
         bash util/travis-deploy-staging.sh -f "webapp results-processor";
       elif [ -n "${TRAVIS_PULL_REQUEST_BRANCH}" ]; then
         bash util/travis-deploy-staging.sh webapp;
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -54,7 +54,7 @@ script:
     # Deploy PR to staging environment (only when Travis secrets are available).
     # Note: Done here (in 'script', not 'deploy') because we need deploy to happen before staging webdriver test.
     if [ ${DEPLOY_STAGING} == true ]; then
-      if [ "${TRAVIS_BRANCH}" == "master" ]; then
+      if [ "${TRAVIS_PULL_REQUEST}" == "false" -a "${TRAVIS_BRANCH}" == "master" ]; then
```
Hexcles: @lukebjerring I just made one more change here. `TRAVIS_BRANCH` has different meanings in push & PR builds. In PR, it means the *base* branch of the PR, which is usually `master`, so the old condition would match both cases. I added a check `TRAVIS_PULL_REQUEST` so that it only matches master pushes.

https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

Hexcles:
Build looks good! https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/388841365 Merging.

PR #239, opened 2018-06-06
Add a flag to run.sh that mounts the results-analysis dir too

lukebjerring:
Re-attempt at #230, which hides the behaviour in a flag (so as to not affect travis).


Makefile
```diff
@@ -186,6 +188,27 @@ xvfb:
 		sudo apt-get install --assume-yes --no-install-suggests xvfb; \
 	fi
 
+# symlinks the Go folder for the wpt.fyi project to (this) folder.
+wpt_fyi_symlink:
```
Hexcles: Where's this recipe used?

lukebjerring: Can be used locally, for development setup.

Issue #212, opened 2018-05-31
Add support for 'Labels' on TestRun metadata


lukebjerring:
When uploading a TestRun, Labels such as `chrome`, `experimental`, `taskcluster` should be added, as well as any arbitrary labels that the uploader requests.

PR #240, opened 2018-06-06
[processor] Send labels when creating test runs

Hexcles:
Currently, there is only one label: uploader.

The test run creation API doesn't handle labels at the moment, but extra params are safely ignored.

Part of #212 .

The JSON payload now has one extra field: `labels`, which is an array of strings.

### DIFF
--- a/api/test_run_medium_test.go
+++ b/api/test_run_medium_test.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


lukebjerring:
> ...,  but extra params are safely ignored.

I do believe you, but can you qualify this claim - has this explicitly been verified?

Hexcles:
@lukebjerring actually I just found out TestRun POST handler didn't have tests... Let me add one.

Hexcles:
@lukebjerring added a test to ensure extra params don't cause problems.

lukebjerring:
Running at https://send-labels-dot-wptdashboard-staging.appspot.com

PR #242, opened 2018-06-06
Add Labels to the TestRun entity

### LINKED ISSUE #212: Add support for 'Labels' on TestRun metadata

lukebjerring:
## Description
Resolves #212 

Adds the field `Labels []string` to the TestRun entity, for arbitrary label assignment that can be used to query with the `?label=foo` param in the `/api/runs` endpoint.

## Review Information
To verify, we'll have to combine #240 with the auto-deployed environment, uploading a run that includes a `labels` field (with an array of strings).

Then, `/api/runs?label=[uploader]` should display a result.

### DIFF
--- a/api/results_redirect_handler.go
+++ b/api/results_redirect_handler.go
@@ -55,18 +55,18 @@ func apiResultsRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	if (shared.TestRun{}) == run {
+	if run == nil {
 		http.Error(w, fmt.Sprintf("404 - Test run '%s' not found", runSHA), http.StatusNotFound)
 		return
 	}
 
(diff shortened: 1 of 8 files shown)


lukebjerring:
Running at https://labels-dot-wptdashboard-staging.appspot.com

api/results_redirect_handler.go
```diff
@@ -55,18 +55,18 @@ func apiResultsRedirectHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
-	if (shared.TestRun{}) == run {
+	if run == nil {
```
Hexcles: I like it! Much cleaner now.

PR #246, opened 2018-06-07
Fix Travis staging deployment of master

Hexcles:
As it turns out, PR #238 wasn't the right fix. There are a lot more
places that we need to add quotes; and in the end, util/deploy.sh
can't support multiple targets without some big changes. Passing strings
with spaces around in Bash is too brittle, so the easy and reliable fix
is to split the master deployment into two commands.

One downside of splitting it into two commands is increased overhead.
Reducing the dependency in util/deploy.sh from webserver_deps to
webapp_deps hopefully alleviates the problem.

Lastly, simplify the conditions in .travis.yml a bit.

@lukebjerring can you confirm `util/deploy.sh` only needs webapp_deps instead of webserver_deps?


Hexcles:
@lukebjerring r? Travis seems happy now.

Issue #245, opened 2018-06-07
Append 'stable' / 'experimental' labels to the uploaded TestRun entities


lukebjerring:
This would support wpt.fyi's default view switch to the 'stable' label, which means that arbitrary runs aren't shown on the dashboard by default.

lukebjerring:
Building off #212 and #240

lukebjerring:
cc @jugglinmike - this is information available during execution. While we need to continue to name the browsers `[browser]-experimental`, we also need to include the `experimental` label in the uploaded run (and `stable` for the non-experimental runs) to support the intended change to wpt.fyi's default query (?label=stable).

jugglinmike:
Got it. Can you explain how the label should be specified in each receiver system?

Hexcles:
In the new world of results receiver, the browser names in the uploaded reports *won't* contain the `-experimental` suffix, because the names are directly from the argument of `wpt run` (well, except Sauce runs).

I'm thinking the runner can add a `labels` param in the payload when uploading results, which is a comma-separated string of labels (the payload is a form instead of JSON so we don't have real arrays). The labels can be quite free-form. For this issue, we can simply add the [`browser_channel`](https://github.com/web-platform-tests/results-collection/blob/19aab09fe38e55c78d47e582e5599dd58de1076b/src/master/browsers.json) as a label.

(Note: runners don't need to change the legacy upload at all. Just keep the name spoofing as is for now. And we'll soon remove the whole legacy upload altogether.)

The receiver creates test runs with the vanilla browser names and add the 'stable'/'experimental' label. The webapp needs to be changed to filter experimental runs by: the browser name ends with `-experimental` **OR** the run has the `experimental` label (so that history experimental runs are still recognized).

Hexcles:
> The webapp needs to be changed to filter experimental runs by: the browser name ends with -experimental OR the run has the experimental label (so that history experimental runs are still recognized).

Sorry, this is already the case.

PR #249, opened 2018-06-07
Change the filtering of experimental runs

Hexcles:
## Description

This is part of #245. With this change, two kinds of runs will be considered as experimental and show up in `labels=experimental` queries: those with the `-experimental` suffix in the browser names, and those with the "experimental" label.

The implementation isn't the most efficient. Filtering happens in the app instead of in Datastore, because the union of the two conditions can't be expressed as Datastore queries. However, once we no longer create those spoofed `*-experimental` runs (i.e. once the results receiver is in prod), we can batch convert the existing `*-experimental` runs to have an `experimental` label and remove a whole bunch of special cases in the code.

## Review Information

Please verify in staging deployment that various label/browser params (especially `labels=experimental`) still filter runs correctly.

## Changes

I've cleaned up the branch history. See the commit titles.

### DIFF
--- a/shared/datastore_medium_test.go
+++ b/shared/datastore_medium_test.go
@@ -6,6 +6,8 @@ import (
 	"testing"
 	"time"
 
+	mapset "github.com/deckarep/golang-set"
+
 	"github.com/stretchr/testify/assert"
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/aetest"
@@ -19,7 +21,6 @@ func TestLoadTestRuns(t *testing.T) {
 				BrowserName:    "chrome",
 				BrowserVersion: "63.0",
 				OSName:         "linux",
-				OSVersion:      "3.16",
 			},
 			Revision: "1234567890",
 		},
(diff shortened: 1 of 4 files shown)


shared/datastore_medium_test.go
```diff
@@ -44,3 +44,68 @@ func TestLoadTestRuns(t *testing.T) {
...
+			ResultsURL: "/static/chrome-64.0-linux-summary.json.gz",
+			CreatedAt:  time.Now(),
+			Labels:     []string{"experimental"},
+		},
+	}
+
+	i, err := aetest.NewInstance(&aetest.Options{StronglyConsistentDatastore: true})
+	assert.Nil(t, err)
+	defer i.Close()
+	r, err := i.NewRequest("GET", "/api/run?product=chrome-66.0&labels=experimental", nil)
```
lukebjerring: Nit: These params aren't actually used, so I'd make the URL simpler to avoid confusion for the test reader.

Issue #214, opened 2018-05-31
?sha param isn't honored by /interop/


lukebjerring:
When switching between `interop` and `results` views for a specific `sha` param, the param is dropped after the interop page first loads, resulting in `latest` when the page is refreshed or the URL is shared.

Hexcles:
I think this has been fixed by #254.

PR #254, opened 2018-06-08
Persist /interop/ sha param

lukebjerring:
## Description
Works on #214 

Changes in-page navigation behaviour to be shared between `/interop/` and `/results/` for consistency, using inheritance. This means that specifying a SHA will persist.

Note that the SHA itself does not yet affect the datastore query.

## Review
- See auto-deployed environment
- Navigate to /interop?sha=1234567890
- Filter into a folder, e.g. 2dcontext
- Observe ?sha is still on the path
- Hit the back button
- Observe ?sha is still on the path


Hexcles:
This change looks good overall, but the integration test seems to be failing. @lukebjerring any idea why?

Hexcles:
ping @lukebjerring in case you missed it

Hexcles:
@lukebjerring why's the linter complaining about some go internal stuff?

```
# golang.org/x/tools/go/internal/gcimporter
../tools/go/internal/gcimporter/bimport.go:541:8: undefined: types.NewInterface2
../tools/go/internal/gcimporter/iimport.go:540:10: undefined: types.NewInterface2
```

lukebjerring:
re: /x/tools error: See https://github.com/web-platform-tests/wpt.fyi/pull/268

PR #255, opened 2018-06-11
update README

kereliuk:
## Description
Update the README for the summary file URLs.

## Review Information
Just a change to the docs, didn't include anything on getting summaries of specific directories so might need to add that?

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,14 +51,12 @@ All test result data is public. There are two types of gzipped JSON data files w
 
 ### Test run summary files
 
-These are of the pattern: `{sha[0:10]}/{platform_id}-summary.json.gz`
+These are of the pattern: `https://wpt.fyi/api/results?sha={sha[0:10]}&product={product}`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -51,14 +51,12 @@ All test result data is public. There are two types of gzipped JSON data files w
 
 ### Test run summary files
 
-These are of the pattern: `{sha[0:10]}/{platform_id}-summary.json.gz`
+These are of the pattern: `https://wpt.fyi/api/results?sha={sha[0:10]}&product={product}`
```
lukebjerring: The URLs are actually of the pattern described previously; I'd instead suggest adding a link to the API docs (@ `/api/README.md`) which cover the 'helper redirect' that you're linking to here.

Hexcles: I replaced the whole section with a simple reference to API docs. Let's get this long-standing PR merged.

PR #262, opened 2018-06-12
Deployment config for announcer

mdittmer:
## Description

Setup deployment for announcer service.


revisions/service/Dockerfile
```diff
@@ -0,0 +1,19 @@
+# Production deployment spec for revisions announcer service.
+
+# Base golang 1.10 image.
+FROM gcr.io/gcp-runtimes/go1-builder:1.10 as builder
+
+RUN apt-get update
+RUN apt-get install -qy --no-install-suggests git
+WORKDIR /go/src/app
+COPY main.go .
```
Hexcles: What about copying the whole directory to make it more future-proof? Otherwise it could be hard to figure out what goes wrong if a new file is added in some distant future.

revisions/service/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN /usr/local/go/bin/go get -d .
+RUN /usr/local/go/bin/go build -o app .
+
+# Application image.
+FROM gcr.io/distroless/base:latest
```
Hexcles: I don't understand what this does. Could you explain a bit?

revisions/service/Dockerfile
mdittmer: Sure. We don't want/need `Dockerfile` or `app.yaml`, but we could do `COPY *.go .`

revisions/service/Dockerfile
mdittmer: `gcr.io/gcp-runtimes/go1-builder` and `gcr.io/distroless/base` interpret binary code compatibly. The former contains a "proper Linux distro" for building golang code. The latter contains the bare minimum for running binaries. This line and the `COPY` + `CMD` that follow essentially copy the binary built in the "builder" image into a lean "distroless" image, and configure the image to run the binary when an instance of the resulting image starts.

PR #264, opened 2018-06-12
Sanitize special characters in report URLs

### LINKED ISSUE #259: Sanitize report URLs
This is related to #258 . We include some fields from the metadata in the URL for reports, and these fields may include spaces or other unsafe characters. We'll need to sanitize them (only in the URLs; the values in the Datastore fields should't be modified).

This is blocking the launch of results receiver.

Hexcles:
Fixes #259

Replace non-URL-safe characters with underscores in URLs. e.g. `chrome-69.0 dev-linux.json` will now become `chrome-69.0_dev-linux.json`.

I didn't use percent encoding because I'd like to keep the URLs short and easily readable. These fields can be found inside the report anyway, so we are not losing any information.

The fields in Datastore are kept intact, i.e. they may still contain spaces. Only the URLs (of the full report, summary, and sharded results) are sanitized.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -144,7 +144,7 @@ def task_handler():
 
     revision = report.run_info['revision']
     # For consistency, use underscores in wptd-results.
-    product = report.product_id('_')
+    product = report.product_id('_', sanitize=True)
 
     resp = "{} results loaded from {}".format(len(report.results), gcs_path)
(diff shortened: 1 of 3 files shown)


Hexcles:
Travis timed out because of some transient network issue. Admin-merged the PR.

PR #265, opened 2018-06-13
Announcer: Perform time-based logic in terms of UTC timestamps

mdittmer:
## Description

Announcer: Convert times to UTC before executing `IsEpochal(a, b)` logic.


mdittmer:
Note: This causes us to select the correct epochal revisions, but does not output timestamps in UTC. @foolip do you still want converted-to-UTC timestamps in the JSON?

foolip:
> @foolip do you still want converted-to-UTC timestamps in the JSON?

That would be nice, yes, if it's just a one-liner.

mdittmer:
Not quite a one-liner, but not too much code:
https://github.com/web-platform-tests/wpt.fyi/pull/266

PR #269, opened 2018-06-13
Expand integration test to cover /interop/ too

lukebjerring:
## Description
Runs the same 'search for 2dcontext and assert there's now only one row visible' webdriver test on the `/interop/` view.
This would have caught a recent regression in the datastore.

## Testing
Best bet is to patch this in locally and run using 

    make go_webdriver_test USE_FRAME_BUFFER=false

which should, in theory, install FF in the default location and all the needed deps (`node`, `selenium`, ...)

That will allow you to watch what happens and ensure that it's legit.


webdriver/webapp_server.go
```diff
@@ -311,10 +312,83 @@ func addStaticData(i DevAppServerInstance) (err error) {
...
 		key := datastore.NewIncompleteKey(ctx, "TestRun", nil)
-		if _, err := datastore.Put(ctx, key, &staticTestRuns[i]); err != nil {
+		if key, err = datastore.Put(ctx, key, &staticTestRuns[i]); err != nil {
 			return err
 		}
+		staticTestRuns[i].ID = key.IntID()
 		fmt.Printf("Added static run for %s\n", staticTestRuns[i].BrowserName)
 	}
+
+	log.Println("Adding static interop data...")
```
lukebjerring: Note: This is copied from `populate_dev_data.go` so doesn't require much scrutiny.

webdriver/search_test.go
```diff
@@ -47,14 +52,47 @@ func TestSearch(t *testing.T) {
...
-	text, err := results[0].Text()
+	assert.Lenf(t, pathParts, 1, "Expected exactly 1 '%s' search result.", query)
+	text, err := pathParts[0].Text()
 	if err != nil {
 		assert.Fail(t, err.Error())
 	}
 	assert.Equal(t, "2dcontext/", text)
 }
+
+// NOTE(lukebjerring): Firefox, annoyingly, throws a TypeError querying shadowRoot because of a
```
lukebjerring: As a team who works on interop, things like this make me die inside a little.

lukebjerring:
@Hexcles - after this lands, will follow-up to #297 with a test for `/anomalies`

Makefile
```diff
@@ -24,7 +24,9 @@ WPTD_GO_PATH ?= $(WPT_GO_PATH)/wpt.fyi
 NODE_SELENIUM_PATH ?= $(WPTD_PATH)webapp/node_modules/selenium-standalone/.selenium/
 SELENIUM_SERVER_PATH ?= $(NODE_SELENIUM_PATH)selenium-server/3.8.1-server.jar
 GECKODRIVER_PATH ?= $(NODE_SELENIUM_PATH)geckodriver/0.20.0-x64-geckodriver
+CHROMEDRIVER_PATH ?= $(NODE_SELENIUM_PATH)chromedriver/2.37-x64-chromedriver
```
Hexcles: Just to confirm: we are not downloading the binary ourselves, but instead they come with the node Selenium package, correct?

(The reason I ask is because this is not the most recent release of Chromedriver, but it's not a problem if they are provided by something indirectly.)

lukebjerring: Yes, we're just leveraging the install that's done by Selenium. I guess it will eventually break when the node package starts fetching a newer version, but we'll deal with that when it happens.

PR #291, opened 2018-06-18
Omit complete if specific sha requested

### LINKED ISSUE #283: `sha` param is broken on `/results/...` pages
Adding the `?sha=` parameter to result pages doesn't seem to do anything.

e.g. click on any sha from https://staging.wpt.fyi/test-runs and you'll always see the latest results.

lukebjerring:
Fixes #283 

`complete` should only be 'passed by default' when not requesting a specific SHA.


lukebjerring:
Example URL: https://params-dot-wptdashboard-staging.appspot.com/results/?sha=04a0711307

Issue #294, opened 2018-06-18
Warning "Polymer::Attributes: couldn't decode Array as JSON" on the front page


Hexcles:
Thrown from wpt-results.html:568

Nothing seems to be broken by this AFAICT

PR #301, opened 2018-06-20
Add conditional for polymer attributes

lukebjerring:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/294

Spitting out an empty string for the attribute doesn't correctly parse as an empty array.


lukebjerring:
See https://go-template-ifs-dot-wptdashboard-staging.appspot.com

PR #318, opened 2018-06-25
Write general handling for TestRun filter params

lukebjerring:
This fixes the `?complete=true` behaviour when combined with a `?label=experimental` flag.

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -29,9 +29,9 @@ type TestRunFilter struct {
 }
 
 // ToQuery converts the filter set to a url.Values (set of query params).
-// defaultQuery is whether the params should fall back to the default query
-// (on the homepage).
-func (filter TestRunFilter) ToQuery(defaultQuery bool) (q url.Values) {
+// complete is whether the params should fall back to a complete run
+// (the default on the homepage).
+func (filter TestRunFilter) ToQuery(complete bool) (q url.Values) {
 	u := url.URL{}
 	q = u.Query()
 	if !IsLatest(filter.SHA) {
@@ -47,7 +47,7 @@ func (filter TestRunFilter) ToQuery(defaultQuery bool) (q url.Values) {
 			q.Add("product", p.String())
 		}
 	}
-	if filter.Complete || (defaultQuery && len(q) == 0) {
+	if filter.Complete || (complete && len(q) == 0) {
 		q.Set("complete", "true")
 	}
 	if filter.MaxCount != nil {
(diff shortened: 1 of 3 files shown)


lukebjerring:
Example URL: https://label-param-dot-wptdashboard-staging.appspot.com/results/?label=lukebjerring&diff

Hexcles:
A unit test failed.

shared/params.go
```diff
@@ -18,6 +18,44 @@ import (
...
+	Labels   mapset.Set
+	Complete bool
+	MaxCount *int
+	Products []Product
+}
+
+// ToQuery converts the filter set to a url.Values (set of query params).
+// defaultQuery is whether the params should fall back to the default query
+// (on the homepage).
+func (filter TestRunFilter) ToQuery(defaultQuery bool) (q url.Values) {
```
Hexcles: IMHO instead of `defaultQuery` we should call it `complete`. And in the future if we have more "default" filtering on the homepage we should add separate boolean parameters for each one of them. WDYT?

shared/params.go
```diff
@@ -18,6 +18,44 @@ import (
...
+		q.Set("complete", "true")
+	}
+	if filter.MaxCount != nil {
+		q.Set("max-count", fmt.Sprintf("%v", *filter.MaxCount))
+	}
+	return q
+}
+
 // MaxCountDefaultValue is the default value returned by ParseMaxCountParam for the max-count param.
 const MaxCountDefaultValue = 1
```
Hexcles: Is this still used?

webapp/test_results_handler.go
```diff
@@ -149,17 +150,13 @@ func getTestRunsAndSources(r *http.Request, runSHA string) (testRunSources []str
...
 		}
-		labels := shared.ParseLabelsParam(r)
-		if labels != nil {
-			for label := range labels.Iterator().C {
-				sourceURL = sourceURL + "&label=" + label.(string)
-			}
-		}
-		testRunSources = []string{sourceURL}
+		sourceURL.RawQuery = f.ToQuery(true).Encode()
+		testRunSources = []string{sourceURL.String()}
```
Hexcles: (GitHub doesn't allow me to comment outside of the diff, so I'm commenting here.)

A nit about the comment for this function: it seems we are now (actually, have been) returning an API URL for getting the list of test runs, instead of the list of test runs directly in this function. The documentation suggests otherwise and is probably out of date.

shared/params.go
lukebjerring: Sure.

webapp/test_results_handler.go
lukebjerring: Tweaked, but it did say 'array of source urls'.

shared/params.go
lukebjerring: Don't think so. Removed.

Issue #319, opened 2018-06-26
webapp deployment is failing (but the build is still green)


Hexcles:
https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/397034462

```
ERROR: (gcloud.app.deploy) Error Response: [9] Deployment contains files that cannot be compiled: Compile failed:
2018/06/26 12:19:22 go-app-builder: Failed parsing input: parser: bad import "syscall" in google.golang.org/grpc/credentials/credentials_util_go19.go
```

This is happening on both master and PR jobs. cc @lukebjerring

Hexcles:
Still can't reproduce locally.

The first failure on master was https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/396550762 (eb318d50547f31b8b0d76da0584c67aea7d50b02), but that commit didn't change Go code at all and the PR staging deployment was successful. Perhaps something happened in the dependencies.

lukebjerring:
@mdittmer, is this the issue that you patched as part of the Makefile rule?

Hexcles:
Don't know why but the issue is gone...

PR #321, opened 2018-06-26
Refactor api methods to re-use the TestRunFilter struct

lukebjerring:
Follow-up for #318, re-using the generalized `TestRunFilter` code in the `/api/runs` endpoint.

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -6,9 +6,7 @@ package api
 
 import (
 	"encoding/json"
-	"fmt"
 	"net/http"
-	"time"
 
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine"
@@ -19,29 +17,14 @@ import (
 // URL Params:
 //     sha: SHA[0:10] of the repo when the tests were executed (or 'latest')
 func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
-	runSHA, err := shared.ParseSHAParam(r)
+	filters, err := shared.ParseTestRunFilterParams(r)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
 
-	var products []shared.Product
-	if products, err = shared.GetProductsForRequest(r); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
-	}
-
-	limit, err := shared.ParseMaxCountParam(r)
-	if err != nil {
-		http.Error(w, "Invalid 'max-count' param: "+err.Error(), http.StatusBadRequest)
-		return
-	}
-	var from *time.Time
-	if from, err = shared.ParseFromParam(r); err != nil {
-		http.Error(w, fmt.Sprintf("Invalid 'from' param: %s", err.Error()), http.StatusBadRequest)
-		return
-	}
-
+	limit := filters.MaxCount
+	from := filters.From
 	if limit == nil && from == nil {
 		// Default to a single, latest run when from & max-count both empty.
 		one := 1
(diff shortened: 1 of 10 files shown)


lukebjerring:
Example:
https://test-run-filter-dot-wptdashboard-staging.appspot.com/results/?products=chrome-69,safari
https://test-run-filter-dot-wptdashboard-staging.appspot.com/results/?products=chrome-69,safari&complete

I actually think prior to #318 we didn't even support the `product` param on the homepage?

lukebjerring:
@Hexcles - another interesting new behaviour, which at first I set out to prevent, but then realized has _some_ viable use-cases, is `max-count`.

Example: Comparing the most recent 2 Chrome 69 runs:
https://test-run-filter-dot-wptdashboard-staging.appspot.com/results/?product=chrome-69&max-count=2&diff

This is already accomodated indirectly, though, with the before & after params, so, WDYT, should we disable it (and, implicitly, always make it 1?)

Hexcles:
@lukebjerring hmm, did you do anything about #319 ? I wonder why the staging deployment for this PR succeeded..

shared/params.go
```diff
@@ -259,28 +263,23 @@ func ParseProductOrBrowserParams(r *http.Request) (products []Product, err error
...
 }
 
-// GetProductsForRequest parses the 'products' (and legacy 'browsers') params, returning
+// GetProductsOrDefault parses the 'products' (and legacy 'browsers') params, returning
 // the sorted list of products to include, or a default list.
-func GetProductsForRequest(r *http.Request) (products []Product, err error) {
-	products, err = ParseProductOrBrowserParams(r)
+func (filter TestRunFilter) GetProductsOrDefault() (products []Product) {
+	products = filter.Products
+	browserNames, err := GetBrowserNames()
```
Hexcles: I know we've been doing this for a while, but I'm just wondering why we need to create a copy of the `shared.browserNames` slice every time?

api/test_runs.go
```diff
@@ -68,11 +47,11 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 				return
 			}
 		}
```
Hexcles: It seems we need to also handle the `else` case here.

e.g. `https://test-run-filter-dot-wptdashboard-staging.appspot.com/results/?sha=61df0b90ef&complete` actuallly returns the latest run for each browser, i.e. neither `sha` nor `complete` is honored in this request.

In such case, we can either discard `complete` or return an error. WDYT?

shared/params.go
```diff
@@ -47,12 +48,15 @@ func (filter TestRunFilter) ToQuery(complete bool) (q url.Values) {
 			q.Add("product", p.String())
 		}
 	}
-	if filter.Complete || (complete && len(q) == 0) {
+	if filter.Complete || (completeIfDefault && len(q) == 0) {
```
Hexcles: On a second thought, this logic introduced in #318 was incorrect. We probably need to differentiate not setting `complete` and setting `complete=false` by using `*bool` instead of `bool`.

E.g. https://test-run-filter-dot-wptdashboard-staging.appspot.com/results/?complete=false , I'd expect to see the latest run from each browser (i.e. not necessarily complete runs). Currently the `complete=false` is simply overridden here.

shared/params.go
lukebjerring: Only way to make it 'immutable'.

shared/params.go

api/test_runs.go
lukebjerring: Switched the order so `sha` overrides it, and just ignore the complete param (erroring isn't necessary.)

PR #324, opened 2018-06-27
Extract shared test-search component

lukebjerring:
## Description
Consolidates the search behaivour by pulling it out into a re-used component. This reduces duplicated code and makes it easier to extend the search in the future.

## Review Information
Running at https://fuzzy-search-dot-wptdashboard-staging.appspot.com


Hexcles:
I have a high-level usability question:

This fuzzy matching seems counter-intuitive, and perhaps worse than prefix matching, given our current UI, namely the fact that when you start searching, the result list doesn't automatically expand. So say if you search for "fileapi", quite a few directories will show up even though they seem to have nothing to do with FileAPI (chances are somewhere in these directories there are files that contain the substring), which is a bit confusing IMO.

I also think this particular fuzzy matching algorithm may not suit our needs. For example, `2dcontext/fill-and-stroke-styles/2d.gradient.interpolate.multiple.html` is a match of `fileapi` because:
2dcontext/**fil**l-and-strok**e**-styles/2d.gr**a**dient.inter**p**olate.mult**i**ple.html
I think we need to treat slashes specially.

That said, the rest of the PR is great! Both the code and the UI are much cleaner. Shall we perhaps leave `fuzzysearch` as a placeholder (perhaps keep using prefix matching) in this PR, and iterate on the search algorithm in another PR?

lukebjerring:
OK, reverted the fuzzysearch change and keeping the rest as a refactor change. PTAL.

PR #325, opened 2018-06-27
make rule for deploying to production

mdittmer:
Using this make rule ensures that necessary file copying that is a part of announcer deployment gets cleaned up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -210,7 +210,7 @@ deploy_staging: gcloud webapp_deps package_announcer var-BRANCH_NAME var-APP_PAT
 	cd $(WPTD_PATH); util/deploy.sh -q -b $(BRANCH_NAME) $(APP_PATH)
 	rm -rf $(WPTD_PATH)revisions/service/wpt.fyi
 
-deploy_production: gcloud webapp_deps package_announcer var-BRANCH_NAME var-APP_PATH var-PROJECT $(WPTD_PATH)client-secret.json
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -210,6 +210,12 @@ deploy_staging: gcloud webapp_deps package_announcer var-BRANCH_NAME var-APP_PAT
 	cd $(WPTD_PATH); util/deploy.sh -q -b $(BRANCH_NAME) $(APP_PATH)
 	rm -rf $(WPTD_PATH)revisions/service/wpt.fyi
 
+deploy_production: gcloud webapp_deps package_announcer var-BRANCH_NAME var-APP_PATH var-PROJECT $(WPTD_PATH)client-secret.json
+	gcloud config set project $(PROJECT)
+	gcloud auth activate-service-account --key-file $(WPTD_PATH)client-secret.json
```
Hexcles: Nit: I *think* prod deploy will almost always be done by a human, so we probably don't need to use a service account (and don't need to depend on `client-secret.json`).

Hexcles:
Actually, @mdittmer could you go ahead and use this branch to do a prod deploy of `revisions/service` **only**?
`make deploy_production PROJECT=wptdashboard APP_PATH=revisions/service`

mdittmer:
@Hexcles sure thing. Starting process now.

Hexcles:
LGTM again and merging.

PR #332, opened 2018-06-28
Fix logos for experimental browsers

### LINKED ISSUE #331: Fix the experimental logos
Since the introduction of results receiver, the browser names no longer have the "-experimental" suffix which breaks the display of special "experimental" logos.

We should use labels to decide logos instead.

cc @lukebjerring

Hexcles:
Fixes #331 

## Description

This PR takes both the browser names and the labels into account when deciding the logos, because browser names in the recent runs (since the introduction of results receiver) no longer have the "-experimental" suffix.

## Review Information

See the homepage of the staging deployment. Chrome Dev logo should be used.

### DIFF
--- a/webapp/components/test-run.html
+++ b/webapp/components/test-run.html
@@ -48,7 +48,7 @@ See models.go for more details.
     </style>
 
     <div>
-      <div><img src="{{displayLogo(testRun.browser_name, testRun.labels)}}" /></div>
+      <div><img src="{{displayLogo(testRun)}}" /></div>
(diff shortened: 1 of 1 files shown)


webapp/components/test-run.html
```diff
@@ -48,7 +48,7 @@
     </style>
 
     <div>
-      <div><img src="/static/{{testRun.browser_name}}_64x64.png" /></div>
+      <div><img src="{{displayLogo(testRun.browser_name, testRun.labels)}}" /></div>
```
lukebjerring: Collapse these 2 params as one `testRun` param :)

Issue #335, opened 2018-06-28
`?labels=chrome` shows two Chromes in staging


Hexcles:
https://staging.wpt.fyi/results/?labels=chrome

It should only show one.

Most likely somewhere in the code `chrome` and the legacy `chrome-experimental` are considered as two separate products.

Hexcles:
I think this issue is fixed in staging.

Issue #311, opened 2018-06-21
Post-label cleanup


Hexcles:
This is the tracking issue to clean up the code base now that we have consistent experimental/stable labels for all runs (in staging).

Commits related to this PR must not be deployed to prod before we run the batch tagger in prod.

Hexcles:
FYI @lukebjerring @foolip prod data have also been labeled (browser names and release channels).

PR #336, opened 2018-06-28
Remove special case of browser name label

lukebjerring:
This no longer works as intended - you end up comparing `chrome-experimental` with (the latest) `chrome`, which is also experimental.

Related to #311


lukebjerring:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/335 but doesn't take us back to the old behaviour that supported comparing stable + experimental, so opened #337

PR #338, opened 2018-06-28
Add dispatch.yaml for the announcer

### LINKED ISSUE #326: Expose the revision announcer API on the canonical domains
Currently we have to use the service-specific domain to access the announcer API: https://announcer-dot-wptdashboard.appspot.com/api/revisions/latest

Since this is an external-facing API, it should be exposed at the canonical {staging.}wpt.fyi domains.

There are two ways to do this:
1. The flex service writes the JSON somewhere, and we add a handler to `api/` to serve that JSON at the desired URL.
2. Add a `dispatch.yaml` to route this URL to the announcer service.

Hexcles:
Fixes #326

I **think** URLs that don't match any of the rules will go to the default service, but I can't find the doc to prove it. https://cloud.google.com/appengine/docs/standard/python/reference/dispatch-yaml

### DIFF
--- a/webapp/dispatch.yaml
+++ b/webapp/dispatch.yaml
@@ -1,0 +1,3 @@
+dispatch:
+  - url: "*/api/revisions/*"
+    service: announcer
--- a/webapp/dispatch.yaml
+++ b/webapp/dispatch.yaml
@@ -1,3 +1,5 @@
 dispatch:
-  - url: "*/api/revisions/*"
+  - url: "*wpt.fyi/api/revisions/*"
+    service: announcer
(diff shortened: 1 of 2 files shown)


webapp/dispatch.yaml
```diff
@@ -0,0 +1,3 @@
+dispatch:
+  - url: "*/api/revisions/*"
```
mdittmer: There is an unlikely chance that we would ship `/something/else/api/revisions/*` in AppEngine Standard, and it wouldn't work, but I doubt it. We could have something like `*wpt.fyi/...` AND `*appspot.com/...` if we want to guard against that.

PR #342, opened 2018-07-04
Add test file results to interop view

mdittmer:
## Description

Add individual (sub)test results to bottom of interop view iff view is at the file level.


mdittmer:
Done, and done! @lukebjerring PTAL

PR #343, opened 2018-07-04
Use proper filters for the run fetch

lukebjerring:
## Description
Another case where non-consolidated query param handling causes problems. This causes `complete` to only be used when nothing that clashes with it is used (e.g. `?label=experimental`, used by results-analysis).

## Review Information
The current interop view @ https://staging.wpt.fyi/interop/?label=experimental was computed using this change locally.
The previous one, https://staging.wpt.fyi/interop/?label=experimental&sha=27406c9d23 was computed without the fix.

### DIFF
--- a/shared/fetch_runs.go
+++ b/shared/fetch_runs.go
@@ -7,7 +7,6 @@ package shared
 import (
 	"encoding/json"
 	"errors"
-	"fmt"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


Hexcles:
The change makes a lot of sense to me, but why does the given example (https://staging.wpt.fyi/interop/?label=experimental) show four stable browsers, seemingly disregarding the `label=experimental` param?

lukebjerring:
labels being ignored is addressed in https://github.com/web-platform-tests/wpt.fyi/pull/344

PR #344, opened 2018-07-04
Persist params when switching tabs

lukebjerring:
## Description
Ensures params are kept when switching between `interop` and `results`.

## Review Information
e.g. https://interop-tab-dot-wptdashboard-staging.appspot.com/interop/?label=experimental

### DIFF
--- a/webapp/components/results-navigation.html
+++ b/webapp/components/results-navigation.html
@@ -26,12 +26,12 @@
     </style>
     <paper-tabs selected='[[selected]]'>
       <paper-tab>
-        <a href="/results/[[query]]">
+        <a href="/results[[path]][[query]]">
           <h2>Test Results</h2>
         </a>
       </paper-tab>
       <paper-tab>
-        <a href="/interop/[[query]]">
+        <a href="/interop[[path]][[query]]">
           <h2>Interoperability</h2>
         </a>
       </paper-tab>
@@ -51,6 +51,10 @@
             type: Number,
             computed: 'computeSelectedTab(tab)',
             value: 0,
+          },
+          path: {
+            type: String,
+            value: window.location.pathname.replace(/^\/(interop|results)/, ''),
           },
           query: {
             type: String,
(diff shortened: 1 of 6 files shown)


webapp/components/results-navigation.html
```diff
@@ -51,6 +51,10 @@ <h2>Interoperability</h2>
             type: Number,
             computed: 'computeSelectedTab(tab)',
             value: 0,
+          },
+          query: {
+            type: String,
+            value: window.location.search,
```
mdittmer: Does every "virtual navigation" (using history API) instantiate a new `results-navigation`? If not, this may need to be a binding instead.

Hexcles:
I saw a couple of 500 errors on this branch:

```
runtime error: invalid memory address or nil pointer dereference
at panic (go/src/runtime/panic.go:491)
at main84340.interopHandler (interop_handler.go:37)
at github.com/web-platform-tests/wpt.fyi/shared.WrapHSTS.func1 (routing.go:37)
at net/http.HandlerFunc.ServeHTTP (server.go:1918)
at github.com/gorilla/mux.(*Router).ServeHTTP (mux.go:162)
at net/http.(*ServeMux).ServeHTTP (server.go:2254)
at appengine_internal.executeRequestSafely (api_prod.go:296)
at appengine_internal.(*server).HandleRequest (api_prod.go:232)
at reflect.Value.call (value.go:434)
```

Something is wrong with the route setup?

Hexcles:
Also visiting the example link shows: `API error 1 (datastore_v3: BAD_REQUEST): The query may not have more than 100 filters + sort orders + ancestor total. Currently there are 210 filters, 1 sort orders, and no ancestor filter.`

lukebjerring:
500s were because it's hard to test locally so was deploying as I went.

lukebjerring:
lol - need to cap the `TestRun` query to 1 run for each product, otherwise it tries to find an interop run with __all__ the TestRuns.

lukebjerring:
@mdittmer - I'm about to migrate the `interop` query to `/api/interop`, so will cover tests then.

webapp/components/results-navigation.html
lukebjerring: @mdittmer - path now updates when user 'navigates' inside the page.

PR #349, opened 2018-07-10
Sort test runs by start time

### LINKED ISSUE #310: Sort "Recent Runs" by commit date
Currently, [the "Recent Runs" page](https://wpt.fyi/test-runs) lists results grouped by revision and orders these groups according to the date of the most recent upload. However, for many visitors to wpt.fyi, the moment that a particular result set was uploaded is not as relevant as the date of the revision under test.

Because results may be uploaded out of order\*, this sorting method produces a listing which does not reflect the revision dates. This is why on wpt.fyi today, the results for the Safari and Edge browsers appear to have been collected sporadically:

screenshot from 2018-06-21 13-47-43

What would it take to update the UI to sort these rows according to revision date?

\* - There are two main causes of this in practice.

The first is due to differences in duration of results collection. Edge and Safari require approximately 13 hours, while Chrome and Firefox require less than 4. When Edge and Safari results are uploaded, the relevant row is moved to the top of the listing, above revisions that have been tested with Firefox and Chrome in the time since.

The second cause is recovery from intermittent failures. When a test collection attempt fails, we diagnose the issue and manually re-trigger the process at a later date. We report the results for these manually-rescued attempts long after subsequent collections have completed successfully. The "sort by upload time" heuristic causes these results sets to be displayed at the top of the listing.

Hexcles:
## Description

Fixes #310 by changing the sorting property from `CreatedAt` (when the test run is uploaded to wpt.fyi) to `TimeStart` (which the test run starts). Ideally we should order by the commit date of the revision, but `TimeStart` is a good approximation which already exists in the wptreport (as `time_start`) and saves us the trouble of consulting git.

For historical runs, `CreatedAt` is used to backfill `TimeStart` using [this script](https://github.com/web-platform-tests/data-migration/blob/master/add_time_start/add_time_start.go). I've already run the migration script against the staging project.

## Review Information

Because of the backfilling, there should be no observable change at the moment, but moving forward we should expect Edge & Safari runs to space more evenly in "recent runs".

## Changes

In addition to changing the argument of `query.Order` in a few places, this PR also refactors `api/results_redirect_handler.go` by the way to reuse `shared.ParseProductParam` & `shared.LoadTestRuns`. A monolithic test (`api/shas_medium_test.go`) is split into two.

`index.yaml` is updated accordingly. Besides, an unneeded index (AFAIK) is removed:
```yaml
- kind: TestRun
  properties:
  - name: BrowserName
  - name: CreatedAt
    direction: desc
  - name: Revision
```

### DIFF
--- a/api/results_redirect_handler.go
+++ b/api/results_redirect_handler.go
@@ -5,16 +5,13 @@
 package api
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
-	"net/url"
 	"regexp"
 	"strings"
 
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine"
-	"google.golang.org/appengine/datastore"
 )
 
 // apiResultsRedirectHandler is responsible for redirecting to the Google Cloud Storage API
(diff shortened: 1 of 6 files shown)


lukebjerring:
FWIW, `util/deploy.sh webapp` locally can get you the same outcome without a PR :)

Hexcles:
@lukebjerring yeah I was lazy. And the PR is now ready for review.

PR #352, opened 2018-07-11
Add line graphs for test history

mdittmer:
- Add `to` as time upper bound on run fetches
- Add `<test-results-chart>` that fetches results and graphs history
- Integrate `<test-results-chart>` into `<wpt-results>`


Hexcles:
@mdittmer I just merged my #349 so you'd need to do some non-trivial merging here.

Hexcles:
Go tests are failing but I don't know which one because Travis is truncating logs (again)...

api/test_runs.go
```diff
@@ -15,7 +15,7 @@ import (
 // apiTestRunsHandler is responsible for emitting test-run JSON for all the runs at a given SHA.
 //
 // URL Params:
-//     sha: SHA[0:10] of the repo when the tests were executed (or 'latest')
+//     sha: SHA[0:10] of the repo when the testsn the tests were executed (or 'latest')
```
Hexcles: Typo?

shared/params.go
```diff
@@ -373,9 +374,9 @@ func ParseMaxCountParamWithDefault(r *http.Request, defaultValue int) (count int
 	return defaultValue, nil
 }
 
-// ParseFromParam parses the "from" param as a timestamp.
-func ParseFromParam(r *http.Request) (*time.Time, error) {
-	if fromParam := r.URL.Query().Get("from"); fromParam != "" {
+// ParseDTParam parses the date/time param named "name" as a timestamp.
+func ParseDTParam(r *http.Request, name string) (*time.Time, error) {
```
Hexcles: Personally I'd prefer `ParseDateTimeParam`, unless `DT` is a well established abbreviation in Go.

shared/datastore.go
```diff
@@ -62,6 +63,9 @@ func LoadTestRuns(
 		if from != nil {
 			query = query.Filter("TimeStart >=", *from)
 		}
+		if to != nil {
+			query = query.Filter("TimeStart <", *to)
+		}
```
Hexcles: Please add some tests for this in `datastore_medium_test.go`.

Hexcles: Done.

webapp/test_runs_handler.go
```diff
@@ -25,9 +25,9 @@ func testRunsHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 func handleTestRunGet(w http.ResponseWriter, r *http.Request) {
-	from, err := shared.ParseFromParam(r)
+	from, err := shared.ParseDateTimeParam(r, "from")
```
Hexcles: I feel like we should also support `to` in `/test-runs`. And on a second thought, I think we should just use `ParseTestRunFilterParams` and `LoadTestRuns` (unless there's some reason that I'm not aware of), but that'd be a bigger change and I can send a follow-up PR to clean up this file.

Hexcles: Done.

PR #353, opened 2018-07-11
Remove the legacy "POST /api/run" endpoint

Hexcles:
## Description

This is a major cleanup after the launch of Results Receiver. The core change is to remove the legacy API to create test runs in Datastore (`POST /api/run`, and its legacy alias `POST /test-runs`), which serves two purposes:
1. Stop people from (accidentally) using the legacy upload process and make Results Receiver the canonical way. This will give us full control over the creation of test runs and prevent unexpected breakage.
2. Fix a major security vulnerability. It turns out we used to pass the "secret" token via URL parameters in this old API. URLs are not secure and are often logged -- there's a high chance that we have already exposed the secret token in public logs.

A new endpoint, `POST /api/results/create`, is created to add test runs to Datastore, with a similar authentication method as `/api/results/upload` (i.e. HTTP Basic Auth with credentials stored in Datastore; a special user "_processor" is used to authenticate the processor service). The payload is the same as the old `POST /api/run` (i.e. JSON-encoded `TestRun`). This API is only supposed to be used internally by the processor service.

## Review Information

Please see below for a breakdown of the changes. I'm also open to splitting this PR if it's a bit too hard to review. Once the staging deployment is up, I'll manually verify the whole result-upload pipeline.

## Changes

This PR also comes with some code-health refactoring/cleanup.

* Some files in `api/receiver` are renamed for consistency and readability: we now have `receive_results.go` (renamed from `results_receiver.go`) and `create_run.go`.
* `api.TestRunPostHandler` is extracted into `api/receiver/create_run.go`, which allows us to fold `apiTestRunGetHandler` into `apiTestRunHandler`. Tests are moved accordingly.
* Following the convention in `api/receiver`, an abstraction for adding `TestRun` to Datastore is added to `AppEngineAPI` so that `create_run.go` can be tested with mocks.
* A few more test cases are added to test the authentication.
* Python code in the processor service is updated to use the new API.
* Lastly, `api/README.md` is re-orged and ToC is updated (not really closely related to this PR).


lukebjerring:
Glad to see this change.

PR #355, opened 2018-07-12
Propagate TestRun filter params in the wpt components

lukebjerring:
## Description
Consolidates the querying behaviour of the components, by plumbing the params through to the components. This ensures that linkability is preserved and nested components are aware of updates to query params.

It also ditches the strange/buggy base64 behaviour of the ?diff param.


webapp/components/interop.html
```diff
@@ -269,16 +269,13 @@
       };
     }
 
-    async connectedCallback() {
-      await super.connectedCallback();
+    async ready() {
+      super.ready();
+
```
Hexcles: I'm not familiar with `connectedCallback` (and why we don't need it anymore). Could you explain a bit or give me a pointer?

webapp/test_results_handler.go
```diff
@@ -42,122 +53,106 @@ func testResultsHandler(w http.ResponseWriter, r *http.Request) {
...
 
 	data := struct {
-		TestRuns       string
-		TestRunSources string
-		SHA            string
-		Diff           bool
-		Filter         string
-		Labels         string
+		TestRuns      string
+		TestRunFilter testRunsUIFilter
```
Hexcles: The field name `TestRunFilter` collides with a commonly used type name in the `shared` package. It might be a bit confusing (especially in the template where it's not obvious which package the field comes from).

webapp/test_results_handler.go
```diff
@@ -42,122 +53,106 @@ func testResultsHandler(w http.ResponseWriter, r *http.Request) {
...
-			return nil, nil, errors.New("after param provided, but before param missing")
+			return filter, errors.New("after param provided, but before param missing")
 		} else if after == "" {
-			return nil, nil, errors.New("before param provided, but after param missing")
+			return filter, errors.New("before param provided, but after param missing")
 		}
 
 		const singleRunURL = `/api/run?sha=%s&product=%s`
 
 		if beforeDecoded, err := base64.URLEncoding.DecodeString(before); err == nil {
```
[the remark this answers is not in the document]
Hexcles: Offline chat with Luke: so it's only the client-side base64 transformation that's being removed; the backend still supports base64 params so that old URLs still work.

PR #357, opened 2018-07-13
Fix up result history chart UX

mdittmer:
- Fixed height
- When hidden, issue waning with 'Show anyway' button
- Reset 'Show anyway' state when tests, chunkSize, labels, or isFeasible changes

### DIFF
--- a/webapp/components/test-results-chart.html
+++ b/webapp/components/test-results-chart.html
@@ -10,14 +10,30 @@ found in the LICENSE file.
 <dom-module id="test-results-chart">
   <template>
     <style>
-      .chart {
-        display: none;
+      .container, google-chart {
+        height: 100%;
       }
-      .chart.true {
-        display: initial;
+      .container {
+        display: flex;
+        flex-direction: column;
+        align-items: center;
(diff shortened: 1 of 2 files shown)


Hexcles:
Thoughts after playing with the UI:

1. This is much better than before! Less jank, less magic (regarding when a graph would show up), more control. Awesome!
2. The graph is smaller than before. And the legend is now completely hidden in ellipses...
3. Can we fix the y-axis always to [0, 1]? Scale graphs that don't start from 0 are misleading; and what's worse is graphs that *sometimes* start from 0 but sometimes not. (I know this is not a new issue introduced by this PR.)

lukebjerring:
This is going to clash with #376

lukebjerring:
@Hexcles - re 3) above: we snuck that in, in #376

mdittmer:
@lukebjerring and @Hexcles is this basically unnecessary now that #376 merged?

lukebjerring:
The 'not feasible' case should probably still be displayed in the UI (currently the graph just doesn't load).

PR #358, opened 2018-07-13
Add param to populate_dev_data

mdittmer:
This allows for use cases like `./util/docker-dev/dev_data.sh -num_remote_runs=10`for 10 runs from each label configuration to be copied from remote (prod/staging) to local datastore.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ eslint: node-babel-eslint node-eslint node-eslint-plugin-html
 
 dev_data:
 	cd $(WPTD_GO_PATH)/util; go get -t ./...
-	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS) "-static_dir=$(WPTD_GO_PATH)/webapp/static"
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -202,7 +202,7 @@ eslint: node-babel-eslint node-eslint node-eslint-plugin-html
 
 dev_data:
 	cd $(WPTD_GO_PATH)/util; go get -t ./...
-	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS)
+	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS) "-static_dir=$(WPTD_GO_PATH)/webapp/static"
```
Hexcles: Nit: technically I think this should be:

`... -static_dir="$(WPTD_GO_PATH)/webapp/static"`

(In reality it probably doesn't matter.)

Hexcles: @mdittmer ahh sorry I was perhaps not clear enough. I meant the location of the quote was wrong. Usually we only quote the value of a flag but not the whole flag, because we know the flag name and it won't contain spaces. When someone quotes the whole flag (including the flag name), it suggests the intention of passing the whole flag as a string through a wrapper. Here it doesn't really make a difference but I was nitpicking.

I modified the PR in place for you.

PR #359, opened 2018-07-13
Extract creation of AppEngine testing server

Hexcles:
There are multiple benefits for doing this:

1. The same code that creates a new aetest instance and then a new
   context with a dummy request and two error checks is repeated many
   times throughout the codebase only to get a strongly consistent
   Datastore emulation. This is error-prone and sometimes confusing
   (because of the fake URL).
2. With a central reusable routine, we can easily control all the aetest
   instances. For example, in this PR SuppressDevAppServerLog is added
   to suppress the lengthy and unhelpful messages from dev_appserver to
   make the output of go_test much cleaner.

The utiltiy function is placed inside shared/sharedtest package,
following the Go convention as seen in http/httptest.

### DIFF
--- a/shared/sharedtest/util.go
+++ b/shared/sharedtest/util.go
@@ -1,0 +1,40 @@
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package sharedtest
+
+import (
+	"golang.org/x/net/context"
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/aetest"
+)
+
+// NewAEInstance creates a new aetest instance backed by dev_appserver whose
+// logs are suppressed. It takes a boolean argument for whether the Datastore
+// emulation should be strongly consistent.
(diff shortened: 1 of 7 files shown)


shared/sharedtest/util.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"golang.org/x/net/context"
+	"google.golang.org/appengine"
+	"google.golang.org/appengine/aetest"
+)
+
+// NewAEInstance creates a new aetest instance backed by dev_appserver whose
+// logs are suppressed. It takes a boolean argument for whether the Datastore
+// emulation should be strongly consistent.
+func NewAEInstance(stronglyConsistentDatastore bool) (aetest.Instance, error) {
```
lukebjerring: Any reason not to have the datastore always be strongly-consistent?

Hexcles: I don't know and based on my very limited experience so far, whenever we need Datastore in tests we always want it to be strongly consistent. What about your experience?

lukebjerring: IIUC, it's a _little_ slower.

Hexcles: Then let's land it as is. With the helper function, we can easily change the default behaviour later on.

PR #363, opened 2018-07-19
Fix test-runs

### LINKED ISSUE #360: /test-runs view is broken (empty)
https://github.com/web-platform-tests/wpt.fyi/blob/69100d4611e3cd308c356fbfaed436b193e9fe99/webapp/components/wpt-runs.html#L139

Looks like `this.testRuns` is undefined here.

lukebjerring:
## Description
Adds an integration test that would have caught the breakage, then fixes the breakage.

Fixes #360

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,8 @@ go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webse
 			--chromedriver_path=$$CHROMEDRIVER_PATH \
 			--frame_buffer=$(USE_FRAME_BUFFER) \
 			--staging=$(STAGING) \
-			--browser=$(BROWSER)
+			--browser=$(BROWSER) \
+			$(FLAGS)
 	if [[ "$(USE_FRAME_BUFFER)" == "true" ]]; then $(STOP_XVFB); fi
 
 web_components_test: xvfb firefox chrome node-web-component-tester webserver_deps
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -202,7 +203,7 @@ eslint: node-babel-eslint node-eslint node-eslint-plugin-html
 
 dev_data:
 	cd $(WPTD_GO_PATH)/util; go get -t ./...
-	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS) -static_dir="$(WPTD_GO_PATH)/webapp/static"
+	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS)
```
Hexcles: Why's this change?

Makefile
```diff
@@ -96,7 +96,8 @@ go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webse
 			--chromedriver_path=$$CHROMEDRIVER_PATH \
 			--frame_buffer=$(USE_FRAME_BUFFER) \
 			--staging=$(STAGING) \
-			--browser=$(BROWSER)
+			--browser=$(BROWSER) \
+			$(FLAGS)
```
Hexcles: What's the usage of this? (What are the new flags you are passing to `go_webdriver_test`?)

Makefile
lukebjerring: That flag doesn't exist anymore

Makefile
lukebjerring: -run for filtering to a specific test

Makefile
Hexcles: Oops. It was added in #358 , probably mistakenly. Back in #358 I thought that was the new flag Mark added. Thanks for spotting this.

PR #369, opened 2018-07-20
Support test names with query strings

### LINKED ISSUE #362: Unable to see test results of tests with query parameters
1. Open https://wpt.fyi/results/dom
2. Click ```interfaces.html?exclude=Node```

Expected:
Showing test results of ```interfaces.html?exclude=Node```

Actual:
The new page URL is https://wpt.fyi/results/dom/interfaces.html, and the page doesn't contain any test results.

Hexcles:
## Description

Fixes #362 .

## Review Information

Navigate to `/dom/idlharness` and click around. Everything should work.

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -11,23 +11,23 @@
             value: '/',
             observer: 'pathUpdated',
           },
+          encodedPath: {
+            type: String,
+            computed: 'computeEncodedPath(path)'
+          },
           onpathupdated: Function,
+          encodeTestPath: Function,
+          decodeTestPath: Function,
         };
       }
 
       ready() {
         super.ready();
         this.path = this.urlToPath(window.location);
-      }
-
-      async connectedCallback() {
-        await super.connectedCallback();
         window.onpopstate = () => {
           this.path = this.urlToPath(window.location);
         };
       }
-
-      // encodeTestPath and decodeTestPath are to be implemented by subclasses.
 
       urlToPath(location) {
         let path = location.pathname;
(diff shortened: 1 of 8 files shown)


webapp/components/self-navigator.html
```diff
@@ -11,17 +11,19 @@
             value: '/',
             observer: 'pathUpdated',
           },
+          encodedPath: {
+            type: String,
+            computed: 'computeEncodedPath(path)'
+          },
           onpathupdated: Function,
+          encodeTestPath: Function,
+          decodeTestPath: Function,
```
lukebjerring: I don't think these are needed; just always encode/decode the value as separated properties (`encodedPath` and `path`).

webapp/components/test-runs.html
```diff
@@ -150,6 +150,25 @@
...
+        parts.push(encodeURIComponent(parts.pop()));
+        return '/' + parts.join('/');
+      }
+
+      decodeTestPath(path) {
+        console.assert(path.startsWith('/'));
+        let parts = path.split('/').slice(1);
+        parts.push(decodeURIComponent(parts.pop()));
+        return '/' + parts.join('/');
+      }
```
lukebjerring: These functions can just be done on the base class; I can't imagine we'll ever decode things differently.

webapp/components/self-navigator.html
Hexcles: Done.

webapp/components/test-runs.html
Hexcles: Done. Also, consequently, `test-file-results{,-time-series}.html` no longer call `encodeTestPath`, but take an encoded path as a property instead.

Hexcles:
@lukebjerring can you take another look?

webapp/components/self-navigator.html
```diff
@@ -11,17 +11,17 @@
             value: '/',
             observer: 'pathUpdated',
           },
+          encodedPath: {
+            type: String,
+            computed: 'computeEncodedPath(path)'
```
lukebjerring: Just use `encodeTestPath`, since that's the whole body of `computeEncodedPath`

PR #375, opened 2018-07-23
Use different names for harness and file statuses

### LINKED ISSUE #371: "Harness status" not appropriate for reftests
https://github.com/web-platform-tests/wpt.fyi/pull/290 changed "Test file" to "Harness status" in ([example](https://wpt.fyi/results/css/css-scrollbars/viewport-scrollbar-body.html)) which made me happy because "Test file" didn't quite make sense for testharness.js tests.

However, now "Harness status" doesn't quite make sense for reftests. wptreport.json files calls it "status", so we could just use "Status".

Hexcles:
## Description

Show "Harness status" as the row header only for testharness tests; for
other tests, use "File status" instead. Fixes #371.

## Review Information

Compare:

* /results/infrastructure/server/order-of-metas.any.html
* /results/infrastructure/reftest-wait.html

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -226,7 +226,7 @@ found in the LICENSE file.
         }
         for (let tr of this.testRuns) {
           // "OK"/"ERROR" are testharness-specific statuses.
-          if (tr.subtests['STATUS'].status === 'OK' || tr.subtests['STATUS'].status === 'ERROR') {
+          if (['OK', 'ERROR'].includes(tr.subtests['STATUS'].status)) {
             return 'Harness status';
           }
         }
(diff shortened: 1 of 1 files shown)


webapp/components/test-file-results.html
```diff
@@ -220,6 +220,19 @@
         return `${resultsBase}${testFile}`;
       }
 
+      subtestNameForDisplay(subtestName) {
+        if (subtestName !== 'STATUS') {
+          return subtestName;
+        }
+        for (let tr of this.testRuns) {
+          // "OK"/"ERROR" are testharness-specific statuses.
+          if (tr.subtests['STATUS'].status === 'OK' || tr.subtests['STATUS'].status === 'ERROR') {
```
lukebjerring: nit:
    if (['OK', 'ERROR'].includes(tr.subtests['STATUS'].status)) {

PR #376, opened 2018-07-23
Put history behind a button

lukebjerring:
## Description
Moves the history graph and grids below the test results, and behind a button.

## Review Information
See auto-deployed environment (hard to test locally).


Hexcles:
In the staging instance, clicking on the "show history" button seems always show the grid, even when I'm viewing a directory (when a line chart is supposed to be shown).

webapp/components/self-navigator.html
```diff
@@ -15,6 +15,10 @@
             type: String,
             computed: 'encodeTestPath(path)'
           },
+          isSubfolder: {
+            type: Boolean,
+            computed: '_computeIsSubfolder(path)',
```
Hexcles: Nit: I don't think the leading underscore is used anywhere else in this project.

webapp/components/test-file-results.html
```diff
@@ -196,17 +188,17 @@
...
-              this.subtestNames.push(subtestResult.name);
+            if (!subtestNames.includes(subtestResult.name)) {
+              subtestNames.push(subtestResult.name);
             }
           }
         });
+
+        this.isTestHarness =
+          resultPerTestRun.some(r => ['OK', 'ERROR'].includes(r.status));
+        this.subtestNames = subtestNames;
```
Hexcles: Awesome! Much clearer now!

webapp/components/test-file-results.html
```diff
@@ -221,24 +213,18 @@
 
       resultsURL(testRun) {
         if (testRun.revision === 'diff') {
-          return `${testRun.results_url}&path=${encodeURIComponent(this.testFile)}`;
+          return `${testRun.results_url}&path=${this.encodedPath}`;
```
Hexcles: Actually, I think this should still be `encodedURIComponent(this.path)`, because we are passing the whole path as the value of a query parameter which means we also need to escape slashes. And `encodedPath` only escapes the last component of the path, leaving all the slashes in intact.

webapp/components/wpt-results.html
```diff
@@ -537,6 +587,12 @@
 
         console.log('JSON version:', JSON.stringify(totals));
       }
+
+      showHistoryClicked() {
+        return () => {
+          this.showHistory = true;
+        };
```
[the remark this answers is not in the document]
lukebjerring: Not a dumb question; the lambda doesn't bind `this`, while calling a function does (bind `this` to the element that was clicked). To be able to refer back to the custom element, rather than the button, we return an arrow function.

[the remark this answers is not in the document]
Hexcles: Ahh, I see, same as `bindNavigation` I saw somewhere else. Thx!

PR #383, opened 2018-07-26
Change the interop component to fetch from the client (via the api)

lukebjerring:
## Description
Currently we 404, server-side, resulting in a useless text-only view.

This defers the 404 to the component, meaning that the content of the page except that component will still all load.

## Review Information


Hexcles:
Abandoning this PR in favour of #385.

PR #384, opened 2018-07-26
Propagate the query params to the interop component

lukebjerring:
## Description
Adds an integration test that the label param is piped between `interop` / `results` (which failed).
Then, propagates the (relevant) params to the component.

## Review Information

Note: Probably clashes with #383 , so we should land that first.

### DIFF
--- a/webdriver/webdriver.go
+++ b/webdriver/webdriver.go
@@ -30,15 +30,41 @@ func GetWebDriver() (*selenium.Service, selenium.WebDriver, error) {
 	panic("Invalid --browser value specified")
 }
 
-func FindShadowElements(d selenium.WebDriver, e selenium.WebElement, selector string) ([]selenium.WebElement, error) {
-	i := []interface{}{e}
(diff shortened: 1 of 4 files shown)


webdriver/webdriver.go
```diff
@@ -30,15 +30,41 @@ func GetWebDriver() (*selenium.Service, selenium.WebDriver, error) {
 	panic("Invalid --browser value specified")
 }
 
-func FindShadowElements(d selenium.WebDriver, e selenium.WebElement, selector string) ([]selenium.WebElement, error) {
-	i := []interface{}{e}
-	result, err := d.ExecuteScriptRaw(fmt.Sprintf("return arguments[0].shadowRoot.querySelectorAll('%s')", selector), i)
-	if err != nil {
-		panic(err.Error())
+func FindShadowElements(
```
lukebjerring: Note: the changes to this function allow you to repeatedly resolve shadow dom for multiple elements. 

If `foo` has shadow `bar` which has shadow `baz`, we can now call

    FindShadowElements(wd, foo, "bar", "baz")

instead of having to handle that ourselves at the call site.

Hexcles:
Abandoning this PR in favour of #385.

PR #385, opened 2018-07-26
Propagate params to interop, fetch from the client

lukebjerring:
# Description
Merge of #384 and #383 (with a small amount of clean-up).

# Testing
See auto-deployed enviroment.

### DIFF
--- a/webapp/interop_handler.go
+++ b/webapp/interop_handler.go
@@ -5,73 +5,30 @@
 package webapp
 
 import (
-	"encoding/json"
 	"net/http"
+	"net/url"
 
-	"github.com/web-platform-tests/results-analysis/metrics"
 	"github.com/web-platform-tests/wpt.fyi/shared"
-	"google.golang.org/appengine"
-	"google.golang.org/appengine/datastore"
 )
 
 // interopHandler handles the view of test results broken down by the
 // number of browsers for which the test passes.
 func interopHandler(w http.ResponseWriter, r *http.Request) {
-	ctx := appengine.NewContext(r)
-	passRateType := metrics.GetDatastoreKindName(metrics.PassRateMetadata{})
-	query := datastore.NewQuery(passRateType).Order("-StartTime").Limit(1)
-
-	filters, err := shared.ParseTestRunFilterParams(r)
+	sourceURL, _ := url.Parse("/api/interop")
+	f, err := shared.ParseTestRunFilterParams(r)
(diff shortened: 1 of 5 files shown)


Hexcles:
Is this PR just for testing the combined effect in the staging instance? In other words, shall I review this one or the two individual PRs?

lukebjerring:
Wellll... that's up to you. The two alone are more clear as to the moving pieces, but this is more complete, and is the end-state we'll wind up in either way. So if this is understandable, I'd go with this one.

webapp/interop_handler.go
```diff
@@ -5,73 +5,23 @@
...
 
 // interopHandler handles the view of test results broken down by the
 // number of browsers for which the test passes.
 func interopHandler(w http.ResponseWriter, r *http.Request) {
-	ctx := appengine.NewContext(r)
-	passRateType := metrics.GetDatastoreKindName(metrics.PassRateMetadata{})
-	query := datastore.NewQuery(passRateType).Order("-StartTime").Limit(1)
-
-	filters, err := shared.ParseTestRunFilterParams(r)
+	uiFilters, err := parseTestRunUIFilter(r)
```
Hexcles: This is great! I refactored this handler a while ago to use `ParseTestRunFilterParams` and was planning for a follow-up refactor to use `LoadTestRuns`. It didn't occur to me we could and should move the fetching to the client. This is much more concise and consistent with the results handler!

webdriver/webdriver.go
```diff
@@ -30,15 +30,41 @@ func GetWebDriver() (*selenium.Service, selenium.WebDriver, error) {
 	panic("Invalid --browser value specified")
 }
 
-func FindShadowElements(d selenium.WebDriver, e selenium.WebElement, selector string) ([]selenium.WebElement, error) {
-	i := []interface{}{e}
-	result, err := d.ExecuteScriptRaw(fmt.Sprintf("return arguments[0].shadowRoot.querySelectorAll('%s')", selector), i)
-	if err != nil {
-		panic(err.Error())
+func FindShadowElements(
```
Hexcles: I saw [your comment in the other PR](https://github.com/web-platform-tests/wpt.fyi/pull/384/files#r205461237). This is quite useful, but can we add a short comment to explain the arguments (especially whether the behaviour is similar to CSS selector `a b c` or `a > b > c`; if I read the code correctly, it's the former)?

Issue #389, opened 2018-07-27
Deploy d698063 to prod (2018-08-13)


Hexcles:
Last release was #334 at 4626a5d.

Change list: https://github.com/web-platform-tests/wpt.fyi/compare/4626a5d...2a99cf1

Notable changes:

* Legacy results upload API is deprecated. The migration process is:
    1. Gradually turn on the new version of webapp first.
    2. When the new version of webapp reaches 100%, enable the new processor.
    3. In the interim between 1 & 2, some results upload might fail because the old processor tries to talk to the legacy API that's removed in the new webapp, but it's fine because the task will be retried later.
* `TimeStart` is the new sort field instead of `CreatedAt` #349  (`TimeStart` and `TimeEnd` were added to Go code earlier and included in the previous release), which means:
    * We need to backfill `TimeStart` by running [the migration script](https://github.com/web-platform-tests/data-migration/tree/master/add_time_start) and pushing the new `index.yaml` before turning on the new version. (This won't break the current prod version because it already has the two fields in the struct.)
* We now also have a `dispatch.yaml`, which needs to be deployed manually.

Hexcles:
`TimeStart` backfilled. Indexes updated. New version is available: https://2a99cf1-dot-wptdashboard.appspot.com/results/

Hexcles:
Bumping to f46ed06 to include a crucial fix for the webapp.

Deployed at: https://f46ed06-dot-wptdashboard.appspot.com

Due to this last-minute bug, I won't increase the traffic split until Monday.

mdittmer:
Is deployment urgent? We could just fix the UI issues and then deploy.

Hexcles:
It's not urgent per se, but I'd like to make a release soon-ish to fix many UI and query issues; and while I'd prefer not to add fancy new features before the release, it's better not to block new features for too long and hurt the development velocity (and I'm not in favour of going down the path of creating release branches in the current stage).

In short, I'm fine with waiting for a week to address some top usability issues (especially the styling).

foolip:
Right, if there isn't any other change that's urgent to deploy, then let's just wait. Suggested a fix in https://github.com/web-platform-tests/wpt.fyi/issues/393#issuecomment-409135995, but would love suggestions for something that isn't a UI toggle.

Hexcles:
[...]

PR #390, opened 2018-07-27
Fix test-file-results.html

Hexcles:
## Description

Commit 26b7470752c6ecb4343dafdae5cf369ed6ce987c removed the redundant
`testFile` and `encodedTestFile` properties from test-file-results.html
and switched to `path` & `encodedPath` from the base class instead. Yet
there was one `encodedTestFile` that was left unrenamed, which broke
test-file-results.

## Review Information

Go to any individual test file and results should render properly. (staging.wpt.fyi is showing empty results right now.)


Hexcles:
Blocking prod release #389 .

Hexcles:
No worries. And yeah, let me take the chance to learn a thing or two about WebDriver tests.

PR #397, opened 2018-07-30
update displayed tests if there is only 1

kereliuk:
## Description
When viewing directories with only one test, the history chart will not show up.

## Review Information
see any individual test file in the auto-deploy environment, eg. FileAPI/FileReader/Progress_event_bubbles_cancelable.html

cc/ @lukebjerring

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -214,10 +214,10 @@ found in the LICENSE file.
       }
 
       resultsURL(testRun, path) {
-        path = this.encodeTestPath(path);
         if (testRun.revision === 'diff') {
           return `${testRun.results_url}&path=${encodeURIComponent(path)}`;
         }
+        path = this.encodeTestPath(path);
         // This is relying on the assumption that result files end with '-summary.json.gz'.
         const resultsBase = testRun.results_url.slice(0, testRun.results_url.lastIndexOf('-summary.json.gz'));
         return `${resultsBase}${path}`;
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -205,7 +205,7 @@ found in the LICENSE file.
       }
 
       async loadResultFile(testRun) {
-        const url = this.resultsURL(testRun);
+        const url = this.resultsURL(testRun, this.path);
         const response = await window.fetch(url);
         if (!response.ok) {
           return null;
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-results.html
```diff
@@ -431,6 +431,11 @@ <h3>
         const nextSubdirIndex = this.path === '/' ? 0 : currentPathParts.length;
 
         let displayedTests = [];
+        if (this.computePathIsATestFile(this.path)) {
+          this.displayedTests.push(this.path);
```
Hexcles: This seems wrong. It should be `this.displayedTests = [this.path]`.

A quick debugging confirms my guess. `this.displayedTests` contains two duplicate test files when we're browsing single-test results (the function is called twice), although nothing the UI doesn't show any issue.

lukebjerring:
@Hexcles - `displayedTests` is only used by the `wpt-results` component, and is forwarded to the subcomponents (chart + grid). (Note: probably shouldn't be an _array_ for the grid, which should only take a single item... but hey.) I don't think this changes any assumptions; I just corrects the (single) test that the child elements think is shown, for the case that there's just the one.

lukebjerring:
@kereliuk I clobbered your PR with a helper function :)

lukebjerring:
@Hexcles - fixed the chart, amongst other things.

webapp/components/test-file-results.html
```diff
@@ -205,21 +205,22 @@
...
+        const url = this.resultsURL(testRun, this.path);
         const response = await window.fetch(url);
         if (!response.ok) {
           return null;
         }
         return response.json();
       }
 
-      resultsURL(testRun) {
+      resultsURL(testRun, path) {
```
Hexcles: Ahh, I didn't realize the parameters for `resultsURL` were different between `test-file-results` and `test-results-chart` in my #398 . Thanks!

webapp/components/test-file-results.html
```diff
@@ -205,21 +205,22 @@
...
+        path = this.encodeTestPath(path);
```
Hexcles: This should be done after the if statement for the `diff` case. Otherwise, the path will be double-encoded.

Hexcles: FYI I directly fixed it myself.

Issue #193, opened 2018-05-29
Navigation by middle-click not working in Safari and Firefox


Zirro:
The ability to open results in a new tab by middle-clicking their links seems to be broken in Safari and Firefox. In Safari they open in the same tab, while no navigation at all occurs in Firefox.

I suspect that this comes down to differences in which buttons cause the `click` event to be fired, which I find a bit amusing given the purpose of the site.

Zirro:
Firefox 62.0a1 (latest Nightly). Having tested with Firefox 60 as well, I can confirm that the issue is not present there though.

In Safari I see the issue as described in both stable 11.1.1 and 11.2 (latest Technology Preview).

lukebjerring:
Unsure what'll be changing in FF62, but can definitely repro the issue. Seems to be some combination of how Polymer binds the event and Firefox's creation of the event. A standard `<a>` does not repro the issue.

Aside: Middle click handling is indeed an interoperability problem.

bzbarsky:
@lukebjerring The Safari issue is presumably because the page doesn't check for what sort of click is happening and just navigates.

The  Firefox issue is probably https://bugzilla.mozilla.org/show_bug.cgi?id=1480130

Hexcles:
It might be a Firefox bug. @bzbarsky filed https://bugzilla.mozilla.org/show_bug.cgi?id=1480130 (thx!)

bzbarsky:
Well, the Safari thing is not a Firefox bug.  ;)

foolip:
It also doesn't work in Chrome for me. In https://staging.wpt.fyi/results/?product=chrome[experimental]&product=chrome[stable]&diff I wanted to middle click "resource-timing/" to dig into that separately, but that just opens https://staging.wpt.fyi/results/resource-timing.

I believe we should be rewriting the URLs in the DOM in addition to intercepting navigation to avoid reloading, so that hovering shows the right URL, and things like middle click and context menu works.

Hexcles:
@foolip that's probably an oversight that we forgot to serialize the query string into href which we can easily fix, but it does "work" in the sense that a new tab is opened.

I still don't know why middle clicking breaks in the way it does in Safari and Firefox. Let alone how to fix it...

bzbarsky:
Middle mouse in Firefox was fixed in https://bugzilla.mozilla.org/show_bug.cgi?id=1480130

foolip:
[...]

PR #402, opened 2018-07-31
Don't intercept meta+click

### LINKED ISSUE #400: Command-click does not work correctly in test directory lists
STEPS TO  REPRODUCE:

1)  Use a Mac.
2)  Load https://wpt.fyi/results/
3)  Hold down the "Command" key and click on the "2dcontext/" text.

EXPECTED RESULTS: Opens the "2dcontext" results  in a new tab.

ACTUAL RESULTS: Opens them in the same tab, overwriting the list I am still trying to go through.  This is happening in Chrome, Safari,  and Firefox, so this is not a browser bug; it's clearly a bug in the web page.

If these blue things are normal links, this should Just Work.  Presumably the page is binding (broken!) click event handlers to them instead of just making them normal links that would behave like users expect links to behave...

I expect https://github.com/web-platform-tests/wpt.fyi/issues/193 is somewhat related, but again  this is broken in every single browser I have access to.

I should note that the link to the revision id (right now d892efb842) in the column headers works correctly.

Hexcles:
#401 missed the Meta key, which is the default modifier to open new tabs on macOS (Command).

Fixes #400 again.

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -90,8 +90,8 @@
       }
 
       navigate(event) {
-        // Don't intercept ctrl+click.
-        if (event.ctrlKey) {
+        // Don't intercept Ctrl+click or Meta(Win/Command)+click (open new tabs).
+        if (event.ctrlKey || event.metaKey) {
           return;
         }
         event.preventDefault();


lukebjerring:
@bzbarsky, would you like to verify that the problem is solved for you with this fix (by opening the staging environment deployed above)?

lukebjerring:
Verified on my Mac.

bzbarsky:
So in the staging environment Cmd+click now works in Chrome, but does not work in Firefox.  In Firefox it just does nothing instead of opening a new tab.

@lukebjerring the fix is not deployed to the main site yet, right?  I am still seeing the old behavior there.

lukebjerring:
It should be deployed to https://staging.wpt.fyi/ (which is the continuously deployed `master` branch).

bzbarsky:
Yeah, that has the same "Cmd+click does nothing in Firefox" problem.  I'll dig into that a bit, I guess...

bzbarsky:
I'm on Sierra 10.12.6.  I see it working in Firefox 61 but not nightly.  If you nightly is 61.0.a1, that's a _really_ old nightly, predating Firefox 61...

Anyway, I will dig into this, since clearly something changed on the Firefox end here since Firefox 61.

Hexcles:
@bzbarsky interesting. The other bug (middle clicking) #193 is in the similar situation: can only be reproduced on Firefox Nightly 62 but not on Firefox 61. We did a little bit of digging but wasn't able to find anything conclusive. Perhaps some click handling was changed between 61 and 62?

lukebjerring:
@bzbarsky lol yeah it updated to 63 now and I'm seeing what you're describing :)
or, maybe :( is more appropriate. Let us know if we can help (or fix something we've done wrong).

bzbarsky:
Looks like when we enabled shadow DOM in nightly the behavior changed in Firefox. Presumably some sort of  shadow DOM bug on our end (e.g. around what the browser UI thinks the event target is).  I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1480130 to get that sorted out.

bzbarsky:
> #193 is in the similar situation

Oh, and yeah, that's probably the same exact issue.

PR #404, opened 2018-08-01
[processor] Add support for chunked reports

### LINKED ISSUE #148: Support chunked wptreports (from Taskcluster)
Taskcluster doesn't have a results merger as of now. In order to receive results from it, we'd need the following pieces:

## An adapter reacting to GitHub webhooks

When all Taskcluster jobs for a commit on master finish, the build status on GitHub of that commit will be updated. Listening to this build status event will tell us the list of Taskcluster jobs, from which we can infer the URLs of artifacts, and then we can simply send the list of URLs to the receiver.

## Results merger in processor

The receiver in #142 already supports downloading and passing multiple files to the processor (cuz it's easy to implement). The MVP for the processor #146 , on the other hand, won't have the results merger in the first iteration, so this issue also covers that task.

### LINKED ISSUE #228: [processor] Handle invalid reports
Currently, the results processor doesn't handle invalid reports (e.g. missing metadata, or outright invalid JSON) well. It will just throw an exception and respond with 500. However, TaskQueue will retry the task upon receiving 500, so the task will just sit in TaskQueue retrying forever.

We should have some more prominent logging in such case (even perhaps send an email alert), and remove the task from TaskQueue. As for the invalid report itself, we could move it to somewhere else for later inspection, or we can just leave it in the buffer bucket, which has a 60-day auto GC.

Hexcles:
# Description

The core change is to add chunk-merging support to `WPTReport` so that the processor can handle multi-URL payloads. This fixes #148.

Meanwhile, exception handling is vastly improved, not only to make the code more readable, but also to catch invalid reports in a central place and prevent TaskQueue from continuously retrying the task. This fixes #228.

# Review

The PR is a little big. I carefully rebased the history to make each commit (and its message) meaningful, so it might be easier to review each individual commit.

# Testing

In addition to lots of unit tests, I've promoted the staging instance to the default processor on *staging.wpt.fyi* and manually tested it from end to end. Besides, it also means the branch is now processing incoming uploads from Buildbot, so at least we can be sure there's no regression.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -6,6 +6,8 @@ import re
 import shutil
 import tempfile
 import time
+import traceback
+from http import HTTPStatus
 
 import filelock
 import flask
@@ -58,8 +60,8 @@ def _serial_task(func):
                 return func(*args, **kwargs)
         except filelock.Timeout:
             app.logger.info('%s unable to acquire lock.', func.__name__)
-            # 503 Service Unavailable
-            return 'A result is currently being processed.', 503
+            return ('A result is currently being processed.',
+                    HTTPStatus.SERVICE_UNAVAILABLE)
 
     return decorated_func
 
@@ -70,7 +72,8 @@ def _internal_only(func):
         if not app.debug:
             remote_ip = flask.request.access_route[0]
             if remote_ip != APPENGINE_INTERNAL_IP:
-                return 'External requests not allowed', 403
+                return ('External requests not allowed',
+                        HTTPStatus.FORBIDDEN)
         return func(*args, **kwargs)
 
     return decorated_func
@@ -90,7 +93,8 @@ def liveness_check():
         # Respectively: file not found, invalid content, old timestamp.
         except (IOError, ValueError, AssertionError):
             app.logger.warn('Liveness check failed.')
-            return 'The current task has taken too long.', 500
+            return ('The current task has taken too long.',
+                    HTTPStatus.INTERNAL_SERVER_ERROR)
     return 'Service alive'
 
 
@@ -101,8 +105,30 @@ def readiness_check():
         lock.acquire(timeout=0.1)
         lock.release()
     except filelock.Timeout:
-        return 'A result is currently being processed.', 503
+        return ('A result is currently being processed.',
+                HTTPStatus.SERVICE_UNAVAILABLE)
     return 'Service alive'
+
+
+def _process_chunk(report, gcs_path):
+    match = re.match(r'/([^/]+)/(.*)', gcs_path)
+    assert match
+    bucket_name, blob_path = match.groups()
+
+    gcs = storage.Client()
+    bucket = gcs.get_bucket(bucket_name)
+    blob = bucket.blob(blob_path)
+
+    with tempfile.NamedTemporaryFile(suffix='.json') as temp:
+        blob.download_to_file(temp)
+        temp.seek(0)
+        try:
+            report.load_json(temp)
+        except wptreport.WPTReportError as e:
+            traceback.print_exc()
+            return False
+
+    return True
 
 
 # Check request origins before acquiring the lock.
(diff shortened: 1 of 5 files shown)


Hexcles:
Requesting review from @foolip as Luke is going on vaca. (I imagine there'll be some back-and-forth in this review.) PTAL.

results-processor/wptreport.py
```diff
@@ -59,40 +73,72 @@ def hash_file(self, fileobj):
...
+            'run_info': {},
+        }
         self._summary = dict()
-        # The hexadecimal sha1sum of the (decompressed) report.
-        self.hashsum = ''
+
+    def _add_chunk(self, chunk):
+        self._report['results'].extend(chunk['results'])
+
+        def update_property(key, source, target, conflict_func=None):
```
foolip: This was neat!

results-processor/wptreport.py
```diff
@@ -154,13 +196,17 @@ def write_gzip_json(filepath, payload):
 
     @property
     def results(self):
-        """The 'results' field of the report, or [] if it doesn't exists."""
-        return self._report.get('results', [])
+        """The 'results' field of the report."""
+        return self._report['results']
```
foolip: Looks like the `.get('results', [])` was unnecessary before as well, or was it not guaranteed then that the `InsufficientDataError` bailout point had been crossed?

results-processor/wptreport.py
```diff
@@ -182,8 +228,8 @@ def summarize(self):
         for result in self.results:
             test_file = result['test']
 
-            assert test_file not in self._summary, (
-                'Found duplicate entries for %s' % test_file)
+            if test_file in self._summary:
+                raise ConflictingDataError(test_file)
```
foolip: These are good checks against oddly chunked runs :+1:

results-processor/wptreport.py
```diff
@@ -244,8 +290,9 @@ def product_id(self, separator='-', sanitize=False):
         # os_version isn't required.
         if self.run_info.get('os_version'):
             name += separator + self.run_info['os_version']
-        assert len(self.hashsum) > 0, 'Missing hashsum of the report'
-        name += separator + self.hashsum[:10]
+        hashsum = self.hashsum()
+        assert len(hashsum) > 0, 'Missing hashsum of the report'
```
foolip: Since this is an assert, I guess there's no input which can cause this assert to fail? (Seems like that's the distinction being made elsewhere, which I agree with.)

results-processor/wptreport_test.py
```diff
@@ -29,17 +36,19 @@ def test_write_json(self):
...
         }]}
         tmp_path = os.path.join(self.tmp_dir, 'foo', 'bar.json.gz')
         WPTReport.write_gzip_json(tmp_path, obj)
-        r = WPTReport()
         with open(tmp_path, 'rb') as f:
-            r.load_gzip_json(f)
-        self.assertDictEqual(obj, r._report)
+            with gzip.GzipFile(fileobj=f, mode='rb') as gf:
+                with io.TextIOWrapper(gf, encoding='utf-8') as tf:
+                    round_trip = json.load(tf)
```
foolip: Sweet :)

results-processor/main.py
```diff
@@ -58,8 +60,8 @@ def decorated_func(*args, **kwargs):
                 return func(*args, **kwargs)
         except filelock.Timeout:
             app.logger.info('%s unable to acquire lock.', func.__name__)
-            # 503 Service Unavailable
-            return 'A result is currently being processed.', 503
+            return ('A result is currently being processed.',
```
foolip: Will this cause app engine to spin up another instance if persistent?

results-processor/main.py
```diff
@@ -115,27 +141,21 @@ def task_handler():
...
-
-    match = re.match(r'/([^/]+)/(.*)', gcs_path)
-    assert match
-    bucket_name, blob_path = match.groups()
-
-    gcs = storage.Client()
-    bucket = gcs.get_bucket(bucket_name)
-    blob = bucket.blob(blob_path)
+    assert (
+        (result_type == 'single' and len(gcs_paths) == 1) or
```
foolip: Nit: indentation is a bit odd. do we use autopep8 or something for formatting?

results-processor/main.py
```diff
@@ -115,27 +141,21 @@ def task_handler():
...
 
-    with tempfile.NamedTemporaryFile(suffix='.json') as temp:
-        blob.download_to_file(temp)
-        temp.seek(0)
-        report = wptreport.WPTReport()
-        report.load_json(temp)
+    report = wptreport.WPTReport()
+    for gcs_path in gcs_paths:
+        if not _process_chunk(report, gcs_path):
+            # The chunk is invalid; return 2XX to prevent retry.
```
foolip: Will we have a log of this incident later, or how will we notice if the processor starts rejecting some input?

results-processor/wptreport.py
Hexcles: Yes! It could only happen if I misused the hash library or something went wrong inside.

results-processor/wptreport.py
Hexcles: No, there was no guarantee that we had have checked that and `results` might not exist. In fact, there is still no programmatic guarantee with this PR (there's only the convention that `finalize` should be called before using the data), but in that case `results` will now be an empty array by default.

results-processor/main.py
Hexcles: Hmm, I use https://github.com/Vimjas/vim-python-pep8-indent , which thinks this is the right indentation. What do you think it should look like?

results-processor/main.py
Hexcles: No, I haven't enabled auto scaling for the processor, but it could. See #187 for more details.

Hexcles:
The logging/notification question raised by @foolip is important enough for a standalone comment:

I'm basically making use of the fancy logging support of AppEngine (i.e. built-in Stackdriver integration): AppEngine sends stdout and stderr to Stackdriver by default, which can recognize Python exceptions (and Go panics) from the log and create "errors" from them, which we can subscribe to in the GCP dashboard. I've been using this feature for a while and it's great. This is in fact part of the reason I refactored the code to create a base exception class so that I can control the formatting and add information like paths to raw incoming results in a central place. Here's an example error in the dashboard:

image

results-processor/main.py
foolip: It's just the outermost parens that make it look a bit funny, might have expected a pretty-printer to put `assert ((result_type == 'single' and len(gcs_paths) == 1) or` on the first line and indent the second. Doesn't matter, mostly curious if any style is enforced by checks in the repo itself.

foolip:
Thanks @Hexcles, that looks like a great AppEngine feature, glad we don't need to make our own monitoring infra :)

PR #405, opened 2018-08-02
Add toggle for two file results layouts

mdittmer:
- Default layout: terse (mostly as-is without this change)
- Alternate layout: verbose (previous layout before message-matching and ellipsis were added)
- Common table layout patterns in abstract mixin
- Mixin applied over different base classes for history grid view and two concrete table classes
- Toggle between layouts with styled paper-toggle-button

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -125,7 +125,7 @@ found in the LICENSE file.
       }
 
       async loadResultFile(testRun) {
-        const url = this.resultsURL(testRun);
+        const url = this.resultsURL(testRun, this.path);
         const response = await window.fetch(url);
         if (!response.ok) {
           return null;
@@ -133,13 +133,14 @@ found in the LICENSE file.
         return response.json();
       }
 
-      resultsURL(testRun) {
+      resultsURL(testRun, path) {
         if (testRun.revision === 'diff') {
-          return `${testRun.results_url}&path=${encodeURIComponent(this.path)}`;
+          return `${testRun.results_url}&path=${encodeURIComponent(path)}`;
         }
+        path = this.encodeTestPath(path);
         // This is relying on the assumption that result files end with '-summary.json.gz'.
         const resultsBase = testRun.results_url.slice(0, testRun.results_url.lastIndexOf('-summary.json.gz'));
-        return `${resultsBase}${this.encodedPath}`;
+        return `${resultsBase}${path}`;
       }
     }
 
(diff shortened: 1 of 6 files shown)


webapp/components/test-file-results.html
```diff
@@ -205,84 +125,21 @@
...
+        const url = this.resultsURL(testRun);
         const response = await window.fetch(url);
         if (!response.ok) {
           return null;
         }
         return response.json();
       }
 
-      resultsURL(testRun, path) {
+      resultsURL(testRun) {
```
Hexcles: This looks like a bad merge. Part of e055778117cebff2910b93ce7c92601c73201c10 is reverted, which will break the line chart.

webapp/components/test-results-history-grid.html
```diff
@@ -96,6 +97,13 @@
         return params;
       }
 
+      subtestNameForDisplay(subtestName, isTestHarness) {
```
Hexcles: This is redundant now that `AbstractTestFileResultsTable` is mixed in.

webapp/components/test-file-results.html
mdittmer: Addressed, I think. Plz double-check new diff.

webapp/components/test-results-history-grid.html
mdittmer: Done.

Hexcles:
@mdittmer `paper-toggle-button.html` cannot be found on the staging instance and the button is hence broken. Did you forget to add `paper-toggle-button` to `bowers.json`?

mdittmer:
That is entirely possible. I ran bower install and it did nothing. I
assumed that meant we already depend on it.

On Aug. 7, 2018 2:08 p.m., "Robert Ma" <notifications@github.com> wrote:

@mdittmer <https://github.com/mdittmer> paper-toggle-button.html cannot be
found on the staging instance and the button is hence broken. Did you
forget to add paper-toggle-button to bowers.json?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/web-platform-tests/wpt.fyi/pull/405#issuecomment-411149813>,
or mute the thread

Hexcles:
@mdittmer I think you need to run `bower install --save PolymerElements/paper-toggle-button`, which will add a dependency to `bower.json`.

You can find bower at `webapp/node_moduels/bower/bin`.

Hexcles:
LGTM again. I'm admin-merging this PR because of #413 (and I'm pretty sure this PR won't affect any Go tests).

Issue #413, opened 2018-08-08
Go tests broken on master


Hexcles:
Looks like there's a new version of Google Cloud Go SDK (211) that broke our tests...

Example: https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/411422158

(This very job passed previously with SDK v209 before my retry.)

This is blocking all PRs.

Hexcles:
Darn I cannot reproduce this locally... Let me try to create a PR to increase logging verbosity.

Hexcles:
Seems to be still flaky... Sigh

Hexcles:
Apparently, SDK v212 was just released and the issues *seems* to have gone away since.

I'll keep an eye on it and may attempt to revert #425 if I don't see more errors.

Though I still don't know what's going on...

PR #414, opened 2018-08-08
Test and implement search API

mdittmer:
This corresponds to the search (not autocomplete) part of the MVP [described here](https://docs.google.com/document/d/1qZYldbW8OhtYqg8iTnYG3uIohAQbCJn3bkxys4BgFWQ/edit#heading=h.tjyi0zyz5907).

E.g.,

```
GET /api/search?q=2dcontext
```

```
{
  "runs": <four latest runs, per unqualified GET /api/runs>,
  "results": [<result for every test path containing "2dcontext", each containing "passes" and "total" for each run in "runs">]
}
```

Unqualified `GET /api/search` returns data equivalent to loading summaries for each run returned from unqualified `GET /api/runs`. `run_ids` query param can be used to select runs other than the default.


Hexcles:
The tests will likely fail because of #413 . Don't be surprised... I'm on it.

mdittmer:
Thanks for the heads up, @Hexcles. LMK if you need a second pair of eyes on the GCP SDK issue.

api/search.go
```diff
@@ -0,0 +1,374 @@
...
+func (r byName) Less(i, j int) bool { return r[i].Name < r[j].Name }
+
+type readable interface {
+	Get(string) ([]byte, error)
+}
+
+type readWritable interface {
+	readable
+	Put(string, []byte) error
+}
```
Hexcles: Is there a reason why we can't just use `io.Reader` & `io.ReadWriter`?

api/search.go
```diff
@@ -0,0 +1,374 @@
...
+func (mc memcacheReadWritable) Put(key string, value []byte) error {
+	log.Printf("Writing %d-bytes to memcache object: %s", len(value), key)
+	return memcache.Set(mc.ctx, &memcache.Item{
+		Key:        key,
+		Value:      value,
+		Expiration: 48 * time.Hour,
+	})
+}
+
+type sharedImpl interface {
```
Hexcles: Based on offline discussion, this is mostly for easier testing, so I'm fine with it for now although ideally we should really change the `shared` package to provide an interface and an impl.

However, I still can't come to terms with an interface named impl... Perhaps `sharedInterface`, `abstractShared`?

api/search.go
```diff
@@ -0,0 +1,374 @@
...
+	for _, r := range resMap {
+		resp.Results = append(resp.Results, r)
+	}
+	sort.Sort(byName(resp.Results))
+
+	return resp
+}
+
+func getMemcacheKey(testRun shared.TestRun) string {
+	return "RESULTS_SUMMARY-" + getResultsURL(testRun, "")
```
Hexcles: Are we sure the length of keys won't exceed 250 bytes? I didn't have the length of URL in mind when designing the results receiver. And as the URL is constructed from some user (runner) inputs, they might get too long. I suggest using testRun ID or some other metadata that we have control over as the key.

api/search.go
mdittmer: I'll can revisit that. My initial attempt to use it seemed more complex because this module then needed to manage `RealData <--> Reader/Writer <--> RealData <--> Cache/Store`. That said, I created these interfaces before I bolted `gzip` support on top, so maybe that code will become simpler with "real" `Reader` `ReadWriter` support.

mdittmer: I don't think my first statement was very clear. What I meant to say was some of the APIs we interact with (e.g., memcache) doesn't have a `Reader`/`Writer`-oriented interface, so the wins we should have gotten from "simply decorating" their interface doesn't apply.

api/search.go
mdittmer: I had a similar thought this morning. I'll switch to the run's ID.

api/search.go
Hexcles: Not sure about your initial attempt/motivation, but I think what you have right now is quite clear. I like the abstraction of "readable"/"writable" and the interaction between HTTP, gzip & memcache. It's just that these interfaces are almost the same as the standard ones in `io` (if you rename the methods to `Read` & `Write`). And I think it'd be even clearer to use the standard interfaces and this would become a good example of Go's Reader/Writer paradigm.

mdittmer:
@Hexcles PTAL at new reader/writer-oriented code. It's longer, and puts more onus on code that consumes the APIs (which I don't particularly like). LMK what you think of the early/late binding too (i.e., some things perform potentially failing network ops on read/write, others on close).

api/query/search_test.go
```diff
@@ -16,6 +17,32 @@ import (
...
+	mrw.EXPECT().NewReader(id).Return(r, nil)
+	r.EXPECT().Read(gomock.Any()).DoAndReturn(func(p []byte) (int, error) {
+		assert.True(t, len(p) >= len(v))
+		for i := range v {
+			p[i] = v[i]
+		}
+		return len(v), nil
+	})
+	r.EXPECT().Read(gomock.Any()).Return(0, io.EOF)
+}
```
Hexcles: Can we simply make `store` (or other `Mockreadable`s) return a *fake* reader constructed by `bytes.NewReader` instead of mocking the reader from scratch? I don't think the assertions of buffer sizes in these mocks are related to the test cases. In fact, the documentation of `io.Reader` requires implementations to never return more than `len(p)` bytes, so these assertions are technically incorrect.

api/query/search_test.go
```diff
@@ -43,9 +70,12 @@ func TestLoadSummary_cacheMiss(t *testing.T) {
 
 	// Use channel to synchronize with expected async cache.Put().
 	c := make(chan bool)
-	cache.EXPECT().Get(key).Return(nil, memcache.ErrCacheMiss)
-	store.EXPECT().Get(url).Return(smry, nil)
-	cache.EXPECT().Put(key, smry).DoAndReturn(func(key string, smry []byte) error {
+	w := NewMockWriteCloser(mockCtrl)
```
Hexcles: Is this the only reason why we need to mock the whole `io` package?

I believe this, instead of `MockSuccessfulRead`, warrants to be mocked manually as we'd like to make sure `Close` is called properly. And we can delegate to `bytes.Buffer` internally. E.g.

```go
type MockWriteCloser struct {
	buffer bytes.Buffer
	closed bool
	t *testing.T
}

func (mcw *MockWriteCloser) Write(...) {
	assert.False(t, closed)
	buffer.Write(...)
}

func (mcw *MockWriteCloser) Close(...) {
	mcw.closed = true
}

func NewMockWriteCloser(t *testing.T) {
	return MockWriteCloser{
		buffer: bytes.Buffer.NewBuffer(nil)
		closed: false,
		t: t,
	}
}

assert.Equal(t, "expected", mcw.buffer.String())
```

foolip:
@mdittmer, can you see to updating [the MVP](https://docs.google.com/document/d/1qZYldbW8OhtYqg8iTnYG3uIohAQbCJn3bkxys4BgFWQ/edit#heading=h.tjyi0zyz5907) to match what's being implemented here? I'll leave some more comments there for the specific things I'm thinking of.

api/query/search.go
```diff
@@ -65,140 +67,130 @@ func (r byName) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
...
+var errMemcacheReaderReadAfterClose = errors.New("memcacheReader: Read() after Close()")
 
-func (mc memcacheReadWritable) Get(key string) ([]byte, error) {
+func (mc memcacheReadWritable) NewReader(key string) (io.Reader, error) {
 	item, err := memcache.Get(mc.ctx, key)
-	if item == nil {
+	if err != nil {
 		return nil, err
 	}
+	return bytes.NewReader(item.Value), nil
```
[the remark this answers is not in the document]
Hexcles: Oh, that's just in case if you also want to make memcache a `ReadCloser`. `bytes.Reader` doesn't have `Close()`; that wrapper will give you a no-op one.

api/query/search.go
```diff
@@ -65,140 +67,130 @@ func (r byName) Swap(i, j int)      { r[i], r[j] = r[j], r[i] }
 func (r byName) Less(i, j int) bool { return r[i].Name < r[j].Name }
 
 type readable interface {
-	Get(string) ([]byte, error)
+	NewReader(string) (io.Reader, error)
```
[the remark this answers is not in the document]
Hexcles: @mdittmer did you miss this comment? We need to close `Response.Body` (i.e. this should be a `ReadCloser`).

mdittmer:
Ping @Hexcles did you want to take another pass at this?

Hexcles:
So now that we have `ReadCloser`, where do we `Close()` it?

mdittmer:
Haha, right. In my mind `Close()` was an externality, but in this case, it's not. 🤦‍♂️

mdittmer:
Downside of dropping `io.*` mocks is that we don't have a test for "correctly closed all `io.ReadCloser`s".

foolip:
Weee! https://staging.wpt.fyi/api/search?run_ids=6311104602963968,5132783244541952 is live!

https://staging.wpt.fyi/api/search?run_ids=6311104602963968,5132783244541952&q=xyz works too!

Some post-merge nits here:
* https://staging.wpt.fyi/api/search fails quite cryptically, can it instead say that the requires `run_ids` parameter is missing?
* Documentation in https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md would be great (I had to read code to find `run_ids`)
* `status":[{"passes":1,"total":1}` should be something like "summary_status" or "legacy_summary", as discussed. ("status" should be reserved for test-level status a la wptreport.json)

foolip:
Doing some numbers on size here...

https://staging.wpt.fyi/api/runs?complete=true would currently ask us to fetch there resources:
* https://storage.googleapis.com/wptd-staging/c0a1bc68d7f463a0f7b8a318d3cdc2b56e2cc044/chrome-68.0.3440.106-linux-16.04-9a1850b82a-summary.json.gz
* https://storage.googleapis.com/wptd-staging/c0a1bc68d7f463a0f7b8a318d3cdc2b56e2cc044/edge-17-windows-10-0d0ebb6a1d-summary.json.gz
* https://storage.googleapis.com/wptd-staging/c0a1bc68d7f463a0f7b8a318d3cdc2b56e2cc044/firefox-61.0.2-linux-16.04-3179afc335-summary.json.gz
* https://storage.googleapis.com/wptd-staging/c0a1bc68d7f463a0f7b8a318d3cdc2b56e2cc044/safari-11.1-macos-10.13-35b72cd1b1-summary.json.gz

The new equivalent is https://staging.wpt.fyi/api/search?run_ids=5131930592870400,6291743024611328,4829257402417152,5087753633005568

Previous total uncompressed size was 8573836 bytes, new is cal.

<hr>

Oops. Tinkering stops here. That's <10% of the size, too good to be true. Looks like the search API isn't returning all of the data. @mdittmer, can you investigate?

PR #415, opened 2018-08-08
DO NOT MERGE increase logging verbosity

Hexcles:
Debugging #413


Hexcles:
@mdittmer I swear: 1. there was indeed a SQLite OperationalError (database is locked) that could be reliably reproduced on Travis; 2. I didn't change anything; 3. after several retries the issue suddenly disappeared and hasn't come back since... I wish I could get my 1 hour back.

So this fix now seems to work...

mdittmer:
Sorry this was such a time sink! ¯\_(ツ)_/¯

PR #418, opened 2018-08-09
Rename the "Verbose" toggle to "Expand"

foolip:
As in expand/collapse details.

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -36,7 +36,7 @@ found in the LICENSE file.
     </style>
 
     <div class="right">
-      <label class="pad">Verbose</label>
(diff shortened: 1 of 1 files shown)


foolip:
Alternatively, we could call it "Compact" and change the default state. @mdittmer @Hexcles preferences?

The reason "verbose" doesn't seem right to me is that apart from when you get to the testharness.js failure messages, the compact view doesn't omit any information, it's just more compact. If we do https://github.com/web-platform-tests/wpt.fyi/issues/422 (please?) I think "Compact" will make more sense in those views too.

mdittmer:
I prefer the compact view as a default but I'm not really a member of the target audience.

PR #426, opened 2018-08-10
Force Go medium tests to run in serial

Hexcles:
Another speculative fix for #413.


Hexcles:
Doesn't seem to work. Same error. Abandoning this PR...

PR #434, opened 2018-08-14
Document announcer (revisions) API

mdittmer:


foolip:
Result now live at https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md, thanks @mdittmer! I agree that the JSON is quite a lot more verbose than the other APIs documented, and would suggest abbreviating by only including "daily" and "hourly" in the examples. (Or two other epochs of your choosing.)

jugglinmike:
@mdittmer If the only concern about the JSON examples is their effect on overall document length, then maybe we can just wrap them in a `<details>` element

PR #438, opened 2018-08-15
Set all-permissive CORS for all read-only APIs

### LINKED ISSUE #433: Add CORS headers to /api/
Trying to implement https://github.com/foolip/wpt-rotation/issues/2 as a client-side check doesn't work because "No 'Access-Control-Allow-Origin' header is present on the requested resource." when trying to fetch https://wpt.fyi/api/runs?label=experimental.

Hexcles:
This commit adds Access-Control-Allow-Origin: * to the headers of all
read-only APIs. Fixes #433.

### DIFF
--- a/webapp/routes_test.go
+++ b/webapp/routes_test.go
@@ -79,11 +79,13 @@ func TestApiResultsBoundCORS(t *testing.T) {
 func TestApiResultsUploadBoundHSTS(t *testing.T) {
 	assertHandlerIs(t, "/api/results/upload", "api-results-upload")
 	assertHSTS(t, "/api/results/upload")
+	assertNoCORS(t, "/api/results/upload")
 }
 
 func TestApiResultsCreateBoundHSTS(t *testing.T) {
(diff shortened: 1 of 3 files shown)


webapp/routes_test.go
```diff
@@ -40,33 +40,53 @@ func TestInteropAnomaliesBound(t *testing.T) {
...
 func TestRunsBoundHSTS(t *testing.T) {
+	assertHandlerIs(t, "/test-runs", "test-runs")
 	assertHSTS(t, "/test-runs")
 }
 
-func TestApiDiffBound(t *testing.T) {
-	assertBound(t, "/api/diff")
+func TestApiDiffBoundCORS(t *testing.T) {
+	assertHandlerIs(t, "/api/diff", "api-diff")
+	assertCORS(t, "/api/diff")
```
foolip: Are the APIs that mutate that we should similarly `assertNoCORS` for?

PR #439, opened 2018-08-15
Fix the setup of integration tests and skip failing tests

Hexcles:
(This is a painful fix...)

I tried my best to split commits logically and write plenty of explanation in the commit messages. Please review commit by commit and pay attention to the messages (and merge by rebasing).

I also tried to fix the actual tests that have been broken for a supposedly long time, but unfortunately I couldn't hone up my modern JavaScript & Web Component skills fast enough and there are simply too many failures -- many tests are quite outdated and need to be completely rewritten. Therefore, after spending one whole day of trying, I decided to skip the failing tests for now and fix the setup first to stop further bleeding.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,21 +89,23 @@ go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webse
 	export GECKODRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)geckodriver/ -type f -name '*geckodriver')"; \
 	export CHROMEDRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)chromedriver/ -type f -name '*chromedriver')"; \
 	$(START_XVFB); \
-	cd $(WPTD_PATH)webdriver; go test -v -tags=large \
-			--selenium_path=$$SELENIUM_SERVER_PATH \
-			--firefox_path=$(FIREFOX_PATH) \
-			--geckodriver_path=$$GECKODRIVER_PATH \
-			--chrome_path=$(CHROME_PATH) \
-			--chromedriver_path=$$CHROMEDRIVER_PATH \
-			--frame_buffer=$(USE_FRAME_BUFFER) \
-			--staging=$(STAGING) \
-			--browser=$(BROWSER) \
-			$(FLAGS); \
+	cd $(WPTD_PATH)webdriver; \
+	go test -v -tags=large \
+		--selenium_path=$$SELENIUM_SERVER_PATH \
+		--firefox_path=$(FIREFOX_PATH) \
+		--geckodriver_path=$$GECKODRIVER_PATH \
+		--chrome_path=$(CHROME_PATH) \
+		--chromedriver_path=$$CHROMEDRIVER_PATH \
+		--frame_buffer=$(USE_FRAME_BUFFER) \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -120,22 +120,25 @@ webapp_deps: go_deps bower_components
...
-	  $(WPTD_PATH)webdriver/install.sh $$HOME/browsers; \
-		sudo ln -s $(FIREFOX_PATH) /usr/bin/firefox; \
+firefox:
+	if [[ -z "$$(which firefox)" ]]; then \
+		make firefox_install; \
 	fi
 
-browser_deps: wget java
-	sudo apt-get install -qy --no-install-suggests $$(apt-cache depends firefox-esr chromedriver | grep Depends | sed "s/.*ends:\ //" | tr '\n' ' ')
+firefox_install: firefox_deps bzip2 wget java
```
foolip: This bzip2 dependency will never be met, so this target will always be built. It depends on the `if [[ -z "$$(which firefox)" ]]` in the firefox rule to not always repeat the work. Given that, wouldn't it be just as clear to always inline these bits into the firefox target?

foolip: OK, you get to pick the style :)

webdriver/chrome.go
```diff
@@ -39,10 +39,10 @@ func ChromeWebDriver() (*selenium.Service, selenium.WebDriver, error) {
 		options = append(options, selenium.StartFrameBuffer())
 	}
 	// Specify the path to ChromeDriver in order to use Chrome.
+	// FIXME: chromeDriverPath seems to be ignored?
 	// Output debug information to STDERR.
 	options = append(options, selenium.Output(os.Stderr))
 
-	selenium.SetDebug(true)
```
mdittmer: Could we plumb a flag into here in case we want to debug webdriver tests? I'd be fine with commenting out the code an adding a `TODO`, since this PR's already so large.

Makefile
```diff
@@ -54,8 +54,8 @@ go_lint: go_deps golint_deps go_test_tag_lint
 	cd $(WPTD_GO_PATH); golint -set_exit_status shared/
 	cd $(WPTD_GO_PATH); golint -set_exit_status util/
 	cd $(WPTD_GO_PATH); golint -set_exit_status webapp/
-	# Printing files with differences between current/gofmt'd output, asserting empty...
-	cd $(WPTD_GO_PATH); ! gofmt -d $(GO_FILES) 2>&1 | read || ! echo $$(gofmt -l $(GO_FILES))
+	# Print out gofmt diff and fail if `gofmt -l` gives non-empty output.
+	cd $(WPTD_GO_PATH); ! (gofmt -l ./ | read && echo "Found gofmt issues:" && gofmt -d ./)
```
mdittmer: Review limitations warning: This looks right to me. Have you tried deliberately adding poorly formatted code to verify that this works? If so, this commit LGTM.

webdriver/chrome.go
Hexcles: Added TODO (and in `firefox.go` as well).

Makefile
Hexcles: Yes I verified locally. However, we probably need to get rid of this anyway because of #459 , but let's deal with that later.

Hexcles:
@mdittmer PTAL again

Issue #62, opened 2018-04-17
Pass/total numbers inflated by "Test file" for testharness.js tests (harness status)


foolip:
Moved from https://github.com/web-platform-tests/results-collection/issues/427. I wrote:

> https://wpt.fyi/gyroscope is a directory where currently all tests except https://wpt.fyi/gyroscope/Gyroscope_insecure_context.html are actually failing, but the overall pass rates listed in https://wpt.fyi/ are 10/50.
>
> This is because "Test file" counts as passing if there aren't any harness errors, for testharness.js tests. For test suites that are split into many files, this makes it harder to spot the "not supported at all" cases, and it would also inflate the 4/4 counts for https://github.com/w3c/wptdashboard/issues/83. Example:
https://metrics5-dot-wptdashboard.appspot.com/metrics/gyroscope/Gyroscope-disabled-by-feature-policy.https.html
>
> @mdittmer @lukebjerring FYI ^
>
> Since "Test file" is the main and only result for reftests the solution would not be to just omit this line. And knowing about harness errors is useful, so omitting it for just testharness.js tests is also not a complete solution.

And:

> I think this might work:
> * Call it "Harness error" instead of "Test file" for testharness.js tests
> * Don't count that in any summary numbers
> * Include harness errors in the upcoming "anomalies" view.

foolip:
Additional thoughts since then. If we make an effort to eliminate harness errors wpt-wide, and make harness errors blocking for PRs (which seems within reach) then we could include harness errors in an "anomalies" view and exclude it from the counts.

foolip:
I just noticed this is a particular problem in the interop view.

In https://wpt.fyi/interop/infrastructure/expected-fail?label=stable&aligned=true, failing-test.html is supposed to fail consistently, and does, but the harness status makes it look like there might be a problem, with "1 / 2" in the 4/4 column.

More generally, the harness status will inflate the "4 / 4" column and make things seem better than they are, numerically at least.

foolip:
@lukebjerring this is easy-ish to fix now with the search cache, right? Or do the summary files ruin it still?

lukebjerring:
It's possible - The issue is diverging from the summary files (used when not executing a search).

foolip:
So I guess the options are making /api/search serve all requests or recomputing all summary files? The former is appealing, but I guess it's not known how much traffic we can handle?

fred-wang:
Hi. I wonder if there is any update on this issue?

[...]

PR #443, opened 2018-08-15
Fix "File/Harness status"

### LINKED ISSUE #419: "File status" still shown for testharness.js tests
A follow-up to https://github.com/web-platform-tests/wpt.fyi/issues/371.

"File status" is still shown for these test:
https://staging.wpt.fyi/results/compat/webkit-text-fill-color-currentColor.html
https://staging.wpt.fyi/results/css/css-flexbox/flexbox_first-letter.html
https://staging.wpt.fyi/results/dom/historical.html

Probably all tests are affected.

Hexcles:
In #405, we turned `test-file-results` into a thin layer was only responsible for fetching data and delegated the UI logic to new embedded components `test-file-results-table-{terse,verbose}` (which were subclasses of `AbstractTestFileResultsTable(TestRunsBase)`).

Therefore, information needs to be explicitly passed into the new components. Yet we forgot to pass `isTestHarness`, which caused the regression. This PR turns `isTestHarness` into `hasSubtests` (which is really a more accurate name as we don't know the type of a test for sure without consulting the manifest), and passes the property down to the embedded components.

This PR also removes the `is` getter from `abstract-test-file-results-table`; the abstract class is not a DOM template and doesn't need the `is` getter.

Fixes #419 .

## Review Information

Check out a testharness test and a reftest in the staging instance. They should show "Harness status" and "File status" respectively.

### DIFF
--- a/webapp/components/abstract-test-file-results-table.html
+++ b/webapp/components/abstract-test-file-results-table.html
@@ -24,7 +24,7 @@ found in the LICENSE file.
         if (subtestName !== 'STATUS') {
           return subtestName;
         }
-        return hasSubtests ? 'Harness status' : 'File status';
+        return hasSubtests ? 'Harness status' : 'Test status';
       }
 
       subtestResultForTestRun(testRun, subtestName) {
--- a/webapp/components/abstract-test-file-results-table.html
+++ b/webapp/components/abstract-test-file-results-table.html
@@ -14,29 +14,17 @@ found in the LICENSE file.
             type: Array,
             value: [],
           },
-          isTestHarness: {
+          hasSubtests: {
             type: Boolean,
-            computed: 'computeIsTestHarness(subtestNames)',
           },
         };
       }
 
-      computeIsTestHarness(subtestNames) {
-        // subtestNames is always updated after testRuns[].subtests are updated,
-        // so as long as 'STATUS' exists in subtestNames, we can safely access
-        // testRuns[].subtests['STATUS'].
-        if (subtestNames.includes('STATUS')) {
-          return false;
-        }
-        const statuses = this.testRuns.map(tr => tr.subtests['STATUS'].status);
-        return statuses.includes('OK') || statuses.includes('ERROR');
-      }
-
-      subtestNameForDisplay(subtestName, isTestHarness) {
+      subtestNameForDisplay(subtestName, hasSubtests) {
         if (subtestName !== 'STATUS') {
           return subtestName;
         }
-        return isTestHarness ? 'Harness status' : 'File status';
+        return hasSubtests ? 'Harness status' : 'File status';
       }
 
       subtestResultForTestRun(testRun, subtestName) {
--- a/webapp/components/abstract-test-file-results-table.html
+++ b/webapp/components/abstract-test-file-results-table.html
@@ -8,17 +8,72 @@ found in the LICENSE file.
   <script>
     // eslint-disable-next-line no-unused-vars
     const AbstractTestFileResultsTable = superClass => class extends superClass {
-      static get is() {
-        return 'abstract-test-file-results-table';
-      }
-
       static get properties() {
         return {
           subtestNames: {
             type: Array,
             value: [],
           },
+          isTestHarness: {
+            type: Boolean,
+            value: false,
+          },
         };
+      }
+
+      static get observers() {
+        return ['fetchTestFile(path, testRuns)'];
+      }
+
+      async fetchTestFile(path, testRuns) {
+        if (!path || !testRuns) {
+          this.subtestNames = [];
+          return;
+        }
+
+        const resultPerTestRun = await Promise.all(
+          testRuns.map(tr => this.loadResultFile(tr)));
+
+        let isTestHarness = false;
+        const subtestNames = ['STATUS'];
+        resultPerTestRun.forEach((resultData, i) => {
+          if (!resultData) {
+            testRuns[i].subtests = {};
+            testRuns[i].subtests['STATUS'] = { status: '(results not found)' };
+            return;
+          }
+          testRuns[i].subtests = {};
+          testRuns[i].subtests['STATUS'] = { status: resultData.status };
+          if (['OK', 'ERROR'].includes(resultData.status)) {
+            isTestHarness = true;
+          }
+
+          for (let subtestResult of resultData.subtests) {
+            testRuns[i].subtests[subtestResult.name] = subtestResult;
+            if (!subtestNames.includes(subtestResult.name)) {
+              subtestNames.push(subtestResult.name);
+            }
+          }
+        });
+
+        this.isTestHarness = isTestHarness;
+        this.subtestNames = subtestNames;
+      }
+
+      async loadResultFile(testRun) {
+        const url = this.resultsURL(testRun, this.path);
+        const response = await window.fetch(url);
+        if (!response.ok) {
+          return null;
+        }
+        return response.json();
+      }
+
+      resultsURL(testRun, path) {
+        path = this.encodeTestPath(path);
+        // This is relying on the assumption that result files end with '-summary.json.gz'.
+        const resultsBase = testRun.results_url.slice(0, testRun.results_url.lastIndexOf('-summary.json.gz'));
+        return `${resultsBase}${path}`;
       }
 
       subtestNameForDisplay(subtestName, isTestHarness) {
--- a/webapp/components/abstract-test-file-results-table.html
+++ b/webapp/components/abstract-test-file-results-table.html
@@ -8,72 +8,17 @@ found in the LICENSE file.
   <script>
     // eslint-disable-next-line no-unused-vars
     const AbstractTestFileResultsTable = superClass => class extends superClass {
+      static get is() {
+        return 'abstract-test-file-results-table';
+      }
+
       static get properties() {
         return {
           subtestNames: {
             type: Array,
             value: [],
           },
-          isTestHarness: {
-            type: Boolean,
-            value: false,
-          },
         };
-      }
-
-      static get observers() {
-        return ['fetchTestFile(path, testRuns)'];
-      }
-
-      async fetchTestFile(path, testRuns) {
-        if (!path || !testRuns) {
-          this.subtestNames = [];
-          return;
-        }
-
-        const resultPerTestRun = await Promise.all(
-          testRuns.map(tr => this.loadResultFile(tr)));
-
-        let isTestHarness = false;
-        const subtestNames = ['STATUS'];
-        resultPerTestRun.forEach((resultData, i) => {
-          if (!resultData) {
-            testRuns[i].subtests = {};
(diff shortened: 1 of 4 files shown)


webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,19 +8,73 @@
...
+          },
         };
       }
 
+      static get observers() {
+        return ['fetchTestFile(path, testRuns)'];
+      }
+
+      async fetchTestFile(path, testRuns) {
+        if (!path || !testRuns) {
```
foolip: `this.subtestNames = []; // Clear any existing rows` before this wasn't copied over. Is it no longer possible that `this.subtestNames` should become empty after being non-empty? I haven't read too carefully to discover the invariants, sorry...

webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,19 +8,73 @@
...
+        const response = await window.fetch(url);
+        if (!response.ok) {
+          return null;
+        }
+        return response.json();
+      }
+
+      resultsURL(testRun, path) {
+        path = this.encodeTestPath(path);
+        // This is relying on the assumption that result files end with '-summary.json.gz'.
```
foolip: Aside: I don't think anyone will be too sad if @mdittmer makes these lines go *poof* :)

webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,19 +8,73 @@
...
+        resultPerTestRun.forEach((resultData, i) => {
+          if (!resultData) {
+            testRuns[i].subtests = {};
+            testRuns[i].subtests['STATUS'] = { status: '(results not found)' };
+            return;
+          }
+          testRuns[i].subtests = {};
+          testRuns[i].subtests['STATUS'] = { status: resultData.status };
+          if (['OK', 'ERROR'].includes(resultData.status)) {
+            isTestHarness = true;
```
foolip: I can't quite tell if the fix is around here and hard to spot because of the moved code, or if the refactoring itself fixes this as a side effect. Either works, but if there are hard-to-spot changes here, two commits would be appreciated for a nervous reviewer :)

foolip:
Fix confirmed with https://fix-status-dot-wptdashboard-staging.appspot.com/results/compat/webkit-text-fill-color-currentColor.html

webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,19 +8,74 @@
   <script>
```
mdittmer: Broad point of clarification: Aren't these tables usually embedded in the context of other components that perform fetch? Are we sure that this addition isn't double-fetching for some view configurations?

webapp/components/abstract-test-file-results-table.html
Hexcles: I think this is no longer needed, but in case I missed something, I added it back, but to a different place: inside the `if` block right before `return`. If the condition doesn't hold, `this.subtestNames` will always be updated at the end of this function.

webapp/components/abstract-test-file-results-table.html
Hexcles: No, the real fix is the two new lines in `test-file-results.html`. I've split the changes into two commits.

Hexcles:
@mdittmer just pointed out that my first commit (the refactor) causes the frontend to fetch individual test results a second time when the user clicks the toggle...

webapp/components/abstract-test-file-results-table.html
Hexcles: Thanks for spotting this! Although my refactor might be a little bit cleaner logically, we don't want double fetching. I just reverted my refactor. Let's fix the bug first.

Hexcles:
Ping @mdittmer ; this is now much shorter (just a fix without refactoring). Does it look good to you?

Hexcles:
Sorry please disregard... This apparently doesn't fix this issue. My original patch worked in a way I didn't fully understand...

Hexcles:
Sorry for the back and forth. I misunderstood the code I was trying to fix. This should now be a cleaner fix that actually works. I've also updated the PR description. PTAL, @foolip & @mdittmer . Thanks!

webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,19 +8,30 @@
...
 
+      computeIsTestHarness(subtestNames) {
+        // subtestNames is always updated after testRuns[].subtests are updated,
+        // so as long as 'STATUS' exists in subtestNames, we can safely access
+        // testRuns[].subtests['STATUS'].
+        if (subtestNames.includes('STATUS')) {
+          return false;
+        }
+        const statuses = this.testRuns.map(tr => tr.subtests['STATUS'].status);
+        return statuses.includes('OK') || statuses.includes('ERROR');
```
foolip: This all seems roundabout to me. Why not just check if there are any subtests? And if there aren't, it might be a single harness error, but I think we'd need the manifest to know for sure what kind of test it was? "TIMEOUT" might be either kind, for example.

Hexcles: There's always at least a subtest as we currently spoof `testRuns[].subtests` to always include a fake `STATUS` subtest for the file. We could check if there's other non-'STATUS' subtest, but that would fail if the test is a testharness test and all browsers produce harness errors. Eventually we should rely on manifest. The search API may give us the ability to do more useful things on the server side.

mdittmer: Where do these data come from? From our sharded individual test files? If so, I believe they contain separate `"status": ...` and `"subtest": [ ... status ... ]`. Perhaps the better long term solution (this may have been in your previous refactor, I'm not sure) would be to disentangle the test status and subtest statuses in whatever component processes data from GCS.

mdittmer: ... to be clear, I would be fine with filing an issue to handle that and landing this code.

Hexcles:
The file-level status is prepended to subtests as a special subtest called "STATUS" here:
https://github.com/web-platform-tests/wpt.fyi/blob/33fcde2fcbf0d21ad698c5b774060fd086952ef0/webapp/components/test-file-results.html#L110

And I agree this isn't pretty.

foolip:
Long comment so hoisting out of review thread.

When we load https://staging.wpt.fyi/results/compat/webkit-text-fill-color-currentColor.html, the data for Chrome is from https://storage.googleapis.com/wptd-staging/2dda7b8c10c7566fa6167a32b09c85d51baf2a85/chrome-68.0.3440.106-linux-16.04-edf200244e/compat/webkit-text-fill-color-currentColor.html. Verbatim response body: `{"test": "/compat/webkit-text-fill-color-currentColor.html", "status": "OK", "message": null, "duration": 118, "subtests": [{"status": "PASS", "message": null, "name": "Test that currentColor is not affected by -webkit-text-fill-color"}]}`

A reftest, https://storage.googleapis.com/wptd-staging/2dda7b8c10c7566fa6167a32b09c85d51baf2a85/chrome-68.0.3440.106-linux-16.04-edf200244e/compat/webkit-text-fill-color-property-001a.html, however, has this response body: `{"test": "/compat/webkit-text-fill-color-property-001a.html", "status": "PASS", "message": null, "duration": 536, "subtests": []}`

And here's a webdriver response body from https://staging.wpt.fyi/results/webdriver/tests/accept_alert/accept.py: `{"test": "/webdriver/tests/accept_alert/accept.py", "status": "OK", "message": null, "duration": 3047, "subtests": [{"status": "PASS", "message": null, "name": "test_null_response_value"}, {"status": "PASS", "message": null, "name": "test_no_browsing_context"}, {"status": "PASS", "message": null, "name": "test_no_user_prompt"}, {"status": "PASS", "message": null, "name": "test_accept_alert"}, {"status": "PASS", "message": null, "name": "test_accept_confirm"}, {"status": "PASS", "message": null, "name": "test_accept_prompt"}]}`

In other words, we can't tell testharness.js from wdspec, but we *can* tell if there are subtests, and when there are we call the the test-level status a "harness status".

@Hexcles, if the data has been mangled by the time it reaches the web component, can file a follow up issue to fix this with a more straightforward test based on the above data formats?

foolip:
> The file-level status is prepended to subtests as a special subtest called "STATUS" here

Yep, that's all part of https://github.com/web-platform-tests/wpt.fyi/issues/62, to stop treating harness status as a test in itself.

For now, perhaps we can simply compute if there are subtests *before* we add the fake one, and use that bool later?

Hexcles:
@foolip good point. See my latest commit.

webapp/components/test-file-results.html
```diff
@@ -108,19 +110,19 @@
...
+              testRuns[i].subtests[subtestResult.name] = subtestResult;
+              if (!subtestNames.includes(subtestResult.name)) {
+                subtestNames.push(subtestResult.name);
+              }
             }
           }
         });
-
-        this.isTestHarness = isTestHarness;
+        this.hasSubtests = true;
```
foolip: Should be `= hasSubtests`?

webapp/components/test-file-results.html
```diff
@@ -108,19 +110,19 @@
...
-          if (['OK', 'ERROR'].includes(resultData.status)) {
-            isTestHarness = true;
-          }
-
-          for (let subtestResult of resultData.subtests) {
-            testRuns[i].subtests[subtestResult.name] = subtestResult;
-            if (!subtestNames.includes(subtestResult.name)) {
-              subtestNames.push(subtestResult.name);
+          if (resultData.subtests && resultData.subtests.length > 0) {
+            // Since we spoof the file-level status also as a "subtest", we
```
foolip: test-level. And maybe comment here that this is still imperfect, since a testharness.js test that throws an exception before registering the first test will still say "Test status" instead of "Harness status".

foolip:
Doesn't work as expected at https://fix-status-dot-wptdashboard-staging.appspot.com/results/battery-status/battery-iframe.https.html

What kinds of tests should we have for this? WCT?

Hexcles:
@foolip fixed in the latest commit ("fixup"). Yeah I'll add a web component test for this when I fix the existing tests for this component. The test setup is broken now.

webapp/components/abstract-test-file-results-table.html
```diff
@@ -8,24 +8,23 @@
...
         };
       }
 
-      subtestNameForDisplay(subtestName, isTestHarness) {
+      subtestNameForDisplay(subtestName, hasSubtests) {
         if (subtestName !== 'STATUS') {
           return subtestName;
         }
-        return isTestHarness ? 'Harness status' : 'File status';
+        return hasSubtests ? 'Harness status' : 'File status';
```
foolip: Wasn't this fixed to "Test status" in this PR already?

Hexcles: Sorry... I have too many local branches and confused myself when merging. Thanks for double checking on this!

PR #445, opened 2018-08-15
Let docs/api.md point to api/README.md

foolip:


Hexcles:
Where's `docs/api.md` linked from? Or do you think this is a more natural place people might look for docs?

foolip:
Yes, I just went looking for the API docs and couldn't find them :)

docs/api.md
```diff
@@ -0,0 +1 @@
+See [api/README.md](api/README.md).
```
Hexcles: The hyperlink doesn't work. Perhaps try `/api/README.md`?

foolip: Ha, I had the slash, but didn't like how it looked, and removed both :)

PR #448, opened 2018-08-17
Refer to test-level status as "Test status", not "File status"

foolip:
Most tests with no subtests are reftests and the paths really are files,
but let's still consistently refer to the manifest entries as "tests".


foolip:
It would be best to merge this *after* https://github.com/web-platform-tests/wpt.fyi/pull/443.

foolip:
It works! https://file-test-status-dot-wptdashboard-staging.appspot.com/results/css/CSS2/abspos/abspos-containing-block-initial-001.xht

Issue #441, opened 2018-08-15
Inconsistent runs on the frontpage


foolip:
https://wpt.fyi/results/ now shows this:
![image](https://user-images.githubusercontent.com/498917/44139369-2966123e-a077-11e8-8b36-c510dc4d7a34.png)

The Chrome results are for 86e157b387 and the others are eea0b54014.

I see a request to https://wpt.fyi/api/runs in devtools. That should be https://wpt.fyi/api/runs?complete=true and I'm pretty sure I've seen that used in devtools before, since that's how I learned of the `complete=true` param.

foolip:
The original issue was https://github.com/web-platform-tests/wpt.fyi/issues/27 and this was then resolved by https://github.com/web-platform-tests/wpt.fyi/pull/72. `git log -S complete=true -- shared/fetch_runs.go webapp/test_results_handler.go` suggests this might have regressed in #318 or #343.

foolip:
With https://github.com/web-platform-tests/wpt.fyi/issues/432 fixed this is arguably more serious, since there is a recent enough run which is complete:
https://wpt.fyi/results/?sha=c0a1bc68d7

@Hexcles @mdittmer, you reviewed the two suspected PRs that might have caused this regression, can you take a look?

PR #450, opened 2018-08-17
Use ?complete=true as default query for /api/runs

foolip:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/441.

### DIFF
--- a/webapp/components/test-runs.html
+++ b/webapp/components/test-runs.html
@@ -153,7 +153,7 @@ found in the LICENSE file.
 
         // Insist on complete runs iff neither sha nor products are specified.
         if (!sha && (!products || products.length === 0)) {
-          params['complete'] = true;
+          params.complete = true;
         }
 
         return params;
--- a/webapp/components/test-runs.html
+++ b/webapp/components/test-runs.html
@@ -95,7 +95,7 @@ found in the LICENSE file.
         if ((this.products && this.products.length)
           || (!preloaded && !this.testRunResources)) {
           const byProduct =
-            fetch(`/api/runs${this.query}`).then(r => r.ok && r.json());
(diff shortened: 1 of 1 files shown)


foolip:
The fix works at least: https://api-runs-complete-dot-wptdashboard-staging.appspot.com/results/

mdittmer:
@foolip and @Hexcles what do you think about the commits I've added? I think we should answer "when do we _not_ want to insist on complete runs?" and encode that logic where the query parameters are computed.

One thing that I haven't included, but we should consider, is whether or not we should add `complete=true` when `sha=latest` (in addition to when `sha` is not specified).

foolip:
This is a much better approach, thanks Mark! Some things that need tuning IMHO.

https://api-runs-complete-dot-wptdashboard-staging.appspot.com/results/?complete=false still shows complete results, i.e. I don't think there's any way to show a mix of runs now.

https://api-runs-complete-dot-wptdashboard-staging.appspot.com/results/?label=stable looks great, that's what I think should be on our front page now.

https://api-runs-complete-dot-wptdashboard-staging.appspot.com/results/?label=experimental also looks good, although I don't understand why. Does /api/runs have a limit on how far back it will look?

If ?complete=false makes that view look like https://wpt.fyi/results/?label=experimental does now (including an Edge run) then I think this will be all good!

foolip:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/468 about the behavior I'm not sure about.

Hexcles:
Hmm, looks like we need to add `complete` as a property of `TestRunsBase`, similar to other params like `sha`. And pass the value of `complete` from Go to the web components.

Issue #24, opened 2018-04-10
change warning to say doesnt test current versions of browsers


lukebjerring:
_From @dstorey on October 22, 2017 6:48_

Currently says doesn't test pre-release versions of Edge and Safari

Actually doesn't even test current versions of Edge and Safari.

_Copied from original issue: w3c/wptdashboard#170_

lukebjerring:
_From @jeffcarp on October 22, 2017 15:49_

We should be testing at least the current public versions of browsers, otherwise that's a bug.

lukebjerring:
_From @jgraham on October 22, 2017 18:27_

I think we aren't necessarily because Sauce is a little behind. There is talk about moving to BrowserStack which allegedly might be more  up to date, at least with Edge.

We would of course do much better here if MS and Apple weren't tying browser version to OS versions, and then placing restrictive license requirements on those OSes that makes it impractical to provide them through a service like TravisCI.

lukebjerring:
_From @jgraham on October 22, 2017 18:27_

(and I think there has been discussion about getting Microsoft and Apple to provide their own results, but so far I haven't seen anything come of that).

foolip:
Current warning is "Each browser listed here is currently tested at its latest stable release." That's correct for the default view, but I'll still send a PR to remove it.

PR #452, opened 2018-08-17
Simplify the caveat

foolip:
Hyphenated "work-in-progress" acts like an adjective and not correct here:
https://www.grammarly.com/blog/hyphen-with-compound-modifiers/

And we now have both stable and experimental browsers.

Related: https://github.com/web-platform-tests/wpt.fyi/issues/24

### DIFF
--- a/webapp/templates/caveat.html
+++ b/webapp/templates/caveat.html
@@ -1,3 +1,3 @@
 <section class="caveat">
-  The web-platform-tests dashboard is a work-in-progress. The reported results do not necessarily reflect the true capabilities of each web browser, so they should not be used evaluate or compare feature support. Each browser listed here is currently tested at its latest stable release.
+  wpt.fyi is a work in progress. The reported results do not necessarily reflect the true capabilities of each web browser, so they should not be used evaluate or compare feature support.
 </section>


webapp/templates/caveat.html
```diff
@@ -1,3 +1,3 @@
 <section class="caveat">
-  The web-platform-tests dashboard is a work-in-progress. The reported results do not necessarily reflect the true capabilities of each web browser, so they should not be used evaluate or compare feature support. Each browser listed here is currently tested at its latest stable release.
+  wpt.fyi is a work in progress. The reported results do not necessarily reflect the true capabilities of each web browser, so they should not be used evaluate or compare feature support.
```
Hexcles: Or remove `a`, i.e. `wpt.fyi is work-in-progress`, which avoids the technicality of countable `work`.

No just kidding. This LGTM.

Hexcles: In "wpt.fyi is work-in-progress", the hyphenated phrase is a predicative adjective instead of attributive, unless such hyphenated form isn't usually used predicatively.

foolip: Ah, without "a". Yes, that would be grammatically correct I think. Didn't get that's what you meant initially.

Enough grammar for today :)

PR #436, opened 2018-08-14
Test and implement autocomplete API

mdittmer:
@Hexcles this is now ready for review.

PR #453, opened 2018-08-17
Query params: handle no "run_ids"; test parsing "run_ids" and "q"

mdittmer:
Not ready for review until #436 lands (and this is synced to it). It seemed easier to write this in a post-QueryAPIsRefactor world.


shared/params.go
```diff
@@ -553,7 +553,11 @@ func ParseBooleanParam(r *http.Request, name string) (result *bool, err error) {
 }
 
 func ParseRunIDsParam(r *http.Request) (ids []int64, err error) {
-	idStrs := strings.Split(r.URL.Query().Get("run_ids"), ",")
+	str := r.URL.Query().Get("run_ids")
+	if str == "" {
+		return nil, nil
```
mdittmer: @Hexcles should we use `nil` slice or the already established `ErrMissing` here?

Hexcles: If `run_ids` is not required, then let's return nil and no error if run_ids is empty. I think ErrMissing should be reserved for missing mandatory fields.

mdittmer:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/469

PR #454, opened 2018-08-17
Document /api/search

mdittmer:


api/README.md
```diff
@@ -613,3 +614,89 @@ __Example JSON__
...
+`q` value as a substring of the test name will be returned. Defaults to the
+empty string, which will yield all test results for the selected runs.
+
+#### Examples
+
+- https://staging.wpt.fyi/api/search?run_ids=6311104602963968,5132783244541952&q=xyz
+
+__Example JSON__
+
+<details>
```
mdittmer: @jugglinmike what am I doing wrong? ( see https://github.com/web-platform-tests/wpt.fyi/blob/748ee3c1e134d69565abc8f0bda661a66feccedc/api/README.md#apisearch )

jugglinmike: May be a HTML<->GFM parsing thing. Try separating the tags and the backticks with newlines. Also, don't forget a `<summary>`.

mdittmer: Thanks! More newlines seems to fix it.

jugglinmike: Newlines are the developer's club soda

mdittmer:
Done. Only put `<details>` around "Examples" for /api/search. Do you want it around all "Examples" sections? In this PR?

mdittmer:
So, to be clear, my most recent change put a very short list of `Examples` into a `<details>` element, *in addition to* the long `Example JSON`. Other sections have `Example JSON` and *not* `Examples` inside a `<details>` element. Should I revert the extra `<details>` around the one example link to make it consistent with others?

Hexcles:
I see. No I don't think it matters unless it bothers you :)

PR #465, opened 2018-08-21
Fix and re-enable WebDriver integration tests

Hexcles:
This PR is based on #439 . I plan to merge that one first, and rebase this PR onto master.

This PR improves the setup of our WebDriver tests and fixes the broken tests disabled in #439.

Again, please review commit by commit.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,7 @@ go_lint: go_deps golint_deps go_test_tag_lint
 	cd $(WPTD_GO_PATH); golint -set_exit_status shared/
 	cd $(WPTD_GO_PATH); golint -set_exit_status util/
 	cd $(WPTD_GO_PATH); golint -set_exit_status webapp/
+	cd $(WPTD_GO_PATH); golint -set_exit_status webdriver/
 	# Print out gofmt diff and fail if `gofmt -l` gives non-empty output.
 	cd $(WPTD_GO_PATH); ! (gofmt -l ./ | read && echo "Found gofmt issues:" && gofmt -d ./)
 
--- a/Makefile
+++ b/Makefile
@@ -85,22 +85,24 @@ go_chrome_test: chrome | go_webdriver_test
 
 go_webdriver_test: STAGING := false
 go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webserver_deps
-	export SELENIUM_SERVER_PATH="$(shell find $(NODE_SELENIUM_PATH)selenium-server/ -type f -name '*server.jar')"; \
-	export GECKODRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)geckodriver/ -type f -name '*geckodriver')"; \
-	export CHROMEDRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)chromedriver/ -type f -name '*chromedriver')"; \
-	$(START_XVFB); \
+	# This Go test manages Xvfb itself, so we don't start/stop Xvfb for it.
+	# The following variables are defined here because we don't know the
+	# paths before installing node-web-component-tester as the paths
+	# include version strings.
+	SELENIUM_SERVER_PATH="$(shell find $(NODE_SELENIUM_PATH)selenium-server/ -type f -name '*server.jar')"; \
+	GECKODRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)geckodriver/ -type f -name '*geckodriver')"; \
+	CHROMEDRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)chromedriver/ -type f -name '*chromedriver')"; \
 	cd $(WPTD_PATH)webdriver; \
-	go test -v -tags=large \
-		--selenium_path=$$SELENIUM_SERVER_PATH \
-		--firefox_path=$(FIREFOX_PATH) \
-		--geckodriver_path=$$GECKODRIVER_PATH \
-		--chrome_path=$(CHROME_PATH) \
-		--chromedriver_path=$$CHROMEDRIVER_PATH \
-		--frame_buffer=$(USE_FRAME_BUFFER) \
-		--staging=$(STAGING) \
-		--browser=$(BROWSER) \
-		$(FLAGS) || (($(STOP_XVFB)) && exit 1); \
-	$(STOP_XVFB)
+	go test -v -tags=large -args \
+		-selenium_path=$$SELENIUM_SERVER_PATH \
+		-firefox_path=$(FIREFOX_PATH) \
+		-geckodriver_path=$$GECKODRIVER_PATH \
+		-chrome_path=$(CHROME_PATH) \
+		-chromedriver_path=$$CHROMEDRIVER_PATH \
+		-frame_buffer=$(USE_FRAME_BUFFER) \
+		-staging=$(STAGING) \
+		-browser=$(BROWSER) \
+		$(FLAGS)
 
 web_components_test: xvfb firefox chrome node-web-component-tester webserver_deps
 	$(START_XVFB); \
--- a/webdriver/search_test.go
+++ b/webdriver/search_test.go
@@ -11,8 +11,6 @@ import (
 )
 
 func TestSearch(t *testing.T) {
-	// TODO(Hexcles): Re-enable this test.
-	t.SkipNow()
 	app, err := NewWebserver()
 	if err != nil {
 		panic(err)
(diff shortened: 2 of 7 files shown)


webdriver/search_test.go
```diff
@@ -67,9 +66,6 @@ func testSearch(t *testing.T, wd selenium.WebDriver, app AppServer, path string,
 	assert.Equal(t, "2dcontext/", text)
 }
 
-// NOTE(lukebjerring): Firefox, annoyingly, throws a TypeError querying shadowRoot because of a
```
mdittmer: IMHO this shouldn't be moved. It belongs here. Perhaps adding something about how it refers to both `case "firefox":` blocks (i.e., the blocks in both functions below) would be appropriate to clarify why it's here.

webdriver/webapp_server.go
```diff
@@ -132,17 +132,17 @@ func NewDevAppServer() (s DevAppServerInstance, err error) {
...
+		port:    pickUnusedPort(),
+		apiPort: pickUnusedPort(),
 	}
 
 	i.cmd = exec.Command(
 		"dev_appserver.py",
 		fmt.Sprintf("--port=%d", i.port),
-		fmt.Sprintf("--admin_port=%d", i.adminPort),
 		fmt.Sprintf("--api_port=%d", i.apiPort),
+		// Let dev_appserver find a free port itself.
```
mdittmer: Was this change to reduce the number of things we have to specify in go (using `pickUnusedPort()`) or is there a technical reason that we _need_ to let dev_appserver pick the port? If it's the latter, that reason should probably be added to this comment. Otherwise, ignore me.

webdriver/webdriver.go
```diff
@@ -13,28 +15,71 @@ var (
 	startFrameBuffer = flag.Bool("frame_buffer", frameBufferDefault(), "Whether to use a frame buffer")
 	seleniumPath     = flag.String("selenium_path", "", "Path to the selenium standalone binary.")
 	seleniumHost     = flag.String("selenium_host", "localhost", "Host to run selenium on")
-	seleniumPort     = flag.Int("selenium_port", 8888, "Port to run selenium on")
+	seleniumPort     = flag.Int("selenium_port", 0, "Port to run selenium on")
```
mdittmer: Will this yield picking a random port, or will it cause a failure when `selenium_port` is not specified?

webdriver/webdriver.go
```diff
@@ -13,28 +15,71 @@ var (
...
 )
 
 func frameBufferDefault() bool {
 	return runtime.GOOS != "darwin"
 }
 
+func pickUnusedPort() int {
```
mdittmer: Do I understand correctly that this is kind of "best effort"? There's nothing that prevents a race condition where another unrelated process acquires the port after `l.Close()` and before the intended subprocess has a chance to acquire it.

webdriver/webdriver.go
mdittmer: Reading further I see it's the former. Maybe add that to the doc string.

webdriver/webdriver.go
```diff
@@ -13,28 +15,71 @@ var (
...
+
+// GetWebDriver starts a WebDriver service (server) and creates a remote
+// (client).
+// Make sure to close both the service and the WebDriver instances, e.g.
+//
+// server, driver, err := GetWebDriver()
+// if err != nil {
+//   panic(err)
+// }
+// defer server.Stop()
```
mdittmer: Does order matter here? If so, we should mention it. It's easy to forget the order in which defers stack.

webdriver/webdriver.go
```diff
@@ -13,28 +15,71 @@ var (
...
 // FindShadowElements finds the shadow DOM children via the given query
-// selectors, recursively.
-// e.g. FindShadowElements(wd, foo, "bar", "baz") would be similar to
-// A "foo bar baz" CSS selector, except it crosses the shadow boundaries for
-// each separate selector.
+// selectors, recursively. The function takes a variable number of selectors;
+// the selectors are combined together similar to CSS descendant combinators.
+// However, all but the the last selector are expected to match to hosts of
+// shadow DOM, and the shadow DOM boundaries will be crossed.
+// e.g. FindShadowElements(wd, node, "bar", "baz blah"). All matches of "bar"
```
mdittmer: nit: Add nothing-but-`//` line before e.g. to break large comment into logical chunks.

webdriver/webapp_server.go
```diff
@@ -41,7 +40,7 @@ type remoteAppServer struct {
 }
 
 func (i *remoteAppServer) GetWebappURL(path string) string {
-	return fmt.Sprintf("http://%s%s", i.host, path)
+	return fmt.Sprintf("https://%s%s", i.host, path)
```
mdittmer: I didn't see anything that changed encryption settings for `webapp` in this PR. Will this work if `i.host` is `localhost`? Usually dev_appserver runs unencrypted.

Hexcles: This is for `remoteAppServer` only, which in foreseeable future will only be `staging.wpt.fyi`, which is encrypted.

There's another `GetWebappURL` for local `DevAppServerInstance` that uses `http`.

webdriver/webapp_server.go
Hexcles: It was mainly to reduce the likelihood of race conditions, as dev_appserver will bind to the free port given by the kernel directly. I'll make this clearer in the comment.

webdriver/webdriver.go
Hexcles: Yes, it's best-effort. The details are a bit nuanced but it generally works:

Closing a TCP socket puts it into TIME_WAIT, which has a short timeout (~2min) during which the kernel won't reuse the port. However, if a program specifically asks to bind that port, the kernel will allow that to happen as there is no active listening socket on the port. In addition, kernel assigns ephemeral ports pseudo-randomly to mitigate some types of attacks. However, a user/program could accidentally/maliciously race to bind to the same port explicitly.

In our case, the likelihood is low and acceptable. Besides, I try to use this trick as little as possible (only use it when we need to know the port number or the subprocess doesn't support binding to 0).

Makefile
```diff
@@ -85,22 +86,23 @@ go_chrome_test: chrome | go_webdriver_test
 
 go_webdriver_test: STAGING := false
 go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webserver_deps
-	export SELENIUM_SERVER_PATH="$(shell find $(NODE_SELENIUM_PATH)selenium-server/ -type f -name '*server.jar')"; \
-	export GECKODRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)geckodriver/ -type f -name '*geckodriver')"; \
-	export CHROMEDRIVER_PATH="$(shell find $(NODE_SELENIUM_PATH)chromedriver/ -type f -name '*chromedriver')"; \
-	$(START_XVFB); \
+	# This Go test manages Xvfb itself, so we don't start/stop Xvfb for it.
+	# The following variables are defined here because we don't know the
```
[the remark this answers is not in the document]
mdittmer: Yeah. With that info incorporated into the comment, this LGTM.

webdriver/search_test.go
```diff
@@ -49,11 +47,12 @@ func testSearch(t *testing.T, wd selenium.WebDriver, app AppServer, path string,
 	if err != nil {
 		panic(err)
 	}
-
 	const query = "2dcontext"
 	if err := searchBox.SendKeys(query); err != nil {
 		panic(err)
 	}
+	// Wait for debouncing (500ms).
```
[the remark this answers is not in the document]
mdittmer: Polling with a timeout SG. However, if you just make the comment explain what you just said, I'm fine with us waiting for flakiness to _actually_ be a problem before we invest in a poll/timeout helper. (Unless there's an out-of-the-box helper that we can invoke; then we should probably just update the comment _and_ switch to polling+timeout.)

webdriver/webdriver.go
Hexcles: Yeah, I found a good article that can serve as explanation and will link to it.

webdriver/webdriver.go
Hexcles: After some experiment I believe we need to quit the driver session (remote) first, which is also logical (if we kill the whole server process first, we can't really ask the server to close a session). I tweaked the comment.

webdriver/webdriver.go
```diff
@@ -13,28 +15,80 @@ var (
...
+	// processes can still explicitly bind to this port immediately.
+	if err := l.Close(); err != nil {
+		panic(err)
+	}
+	return port
+}
+
+// GetWebDriver starts a WebDriver service (server) and creates a remote
+// (client).
+// Note: Make sure to close the service first and the remote later, e.g.
```
mdittmer: Other way around! Curse you, `defer`, and your stacking behaviour!!!

foolip:
When the individual commits have good commit messages and each would work in order, then merging without squashing is nice, I think.

Hexcles:
@mdittmer sorry for the inconvenience. I too usually do squash-merge, but in rare cases like this I try to keep a clean, logical commit history and rebase-merge.

In the future, if I choose to rebase-merge again, I'd use the fixup pattern to make code review easier: i.e. if something needs to fix in commit "foo", I'll push the fix in a new commit "fixup! foo". And when the code review is finished and before the PR is merged, I'll interactively rebase the branch to squash all "fixup..." commits to their respective original commits, force-push to GitHub, and click the rebase-merge button.

PR #472, opened 2018-08-23
Fix go_chrome_test and make sure we actually test all browsers on Travis

Hexcles:
See individual commits.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,19 +9,24 @@
 # the correct version of tools are installed and environment variables are
 # set appropriately.
 
+# Prefer simply expanded variables (:=) to avoid confusion caused by recursion.
+# All variables can be overridden in command line by `make TARGET FOO=BAR`.
+
 SHELL := /bin/bash
-
 GOPATH := $(shell go env GOPATH)
-
 # WPTD_PATH will have a trailing slash, e.g. /home/user/wpt.fyi/
 WPTD_PATH := $(dir $(abspath $(lastword $(MAKEFILE_LIST))))
 WPT_PATH := $(dir $(WPTD_PATH)/../)
-WPT_GO_PATH ?= $(GOPATH)/src/github.com/web-platform-tests
-WPTD_GO_PATH ?= $(WPT_GO_PATH)/wpt.fyi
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -9,19 +9,24 @@
 # the correct version of tools are installed and environment variables are
 # set appropriately.
 
-SHELL := /bin/bash
+# Prefer simply expanded variables (:=) to avoid confusion caused by recursion.
+# All variables can be overridden in command line by `make TARGET FOO=BAR`.
```
mdittmer: nit: lowercase `target` or even `make_target`. Possibly lowercase `bar` too. Conveys naming conventions for different fragments.

Makefile
```diff
@@ -79,13 +81,19 @@ integration_test: go_all_browsers_test web_components_test
...
 go_chrome_test: BROWSER := chrome
-go_chrome_test: chrome | go_webdriver_test
-
-go_webdriver_test: STAGING := false
-go_webdriver_test: var-BROWSER java go_deps xvfb node-web-component-tester webserver_deps
+go_chrome_test: chrome chromedriver | _go_webdriver_test
+
+# _go_webdriver_test is not intended to be used directly; use go_firefox_test or
+# go_chrome_test instead.
+# This target has to be explicitly marked as .PHONY; otherwise, make won't
```
mdittmer: Nit: Extra nothing-but-`#` line between paragraphs.

Makefile
```diff
@@ -130,7 +138,13 @@ chrome:
 		if [[ -z "$$(which chromium)" ]]; then \
 			make apt-get-chromium; \
 		fi; \
-		sudo ln -s "$$(which chromium)" /usr/bin/google-chrome; \
+		sudo ln -s "$$(which chromium)" $(CHROME_PATH); \
```
mdittmer: This never runs on docker images destined for prod, does it? using `which chromium` to decide for us how to find a binary we need makes the security-worryer in me uneasy.

Makefile
Hexcles: Done

Makefile
Hexcles: Removed `.PHONY` altogether

Makefile
Hexcles: Well, I agree this is ugly.

PR #474, opened 2018-08-23
Integrate autocomplete into web components

mdittmer:
- `test-search` recieves `autocompleteURL`

  - ... contains logic for parsing result

  - ... manages `<datalist>` containing results

  - `wpt-results` controls bound `autocompleteURL`

- `test-search` emits event when user commits to query with <Enter>

  - `interop` and `wpt-results` respond accordingly

- `interop`'s only component using old `executeQuery` from `test-search`

  - ... move that logic to `interop`


Hexcles:
Also diff view is broken: https://search-and-autocomplete-dot-wptdashboard-staging.appspot.com/results/?product=chrome&product=edge&diff

Hexcles:
Sorry, the diffing issue is probably not caused by this PR. It's already broken on staging: #477

mdittmer:
I've responded to UX comments inline. @Hexcles I'm interested in how many of these you think should get added to this PR (if any).
> This PR introduces a behavioural change: users now need to press enter for the search to take effect, which might not be obvious, especially to existing/power users. What's the rationale behind it? If there's a technical reason, can we perhaps prompt users? Something as simple as appending "and press enter" to the placeholder in the search box would still be better. Or perhaps losing focus should also trigger the search?

- The technical reason is we don't want to spam `/api/search` with half-formed queries (this was covered in the related design doc);
- I think a tooltip (and update to placeholder) would be good; maybe appears on debounce along with autocomplete fetch;
- Re:losing focus, I think it would be good to have three categories: (a) press <Enter> to "commit" the query; (b) press <Esc> to "clear" the query; (c) loss of focus for some other reason to "abort" the query (the UI will switch back to whatever query appeared when first focused). Other thoughts on the interaction here are welcome.

> The requirement to press enter is especially confusing when one clicks an autocomplete suggestion.
Loading search results is a bit slow (which I think is acceptable), but it'd be great to show a loading indicator. Otherwise, the slowness coupled with the requirement of pressing enter might further confuse users how/when search is triggered.

- We should add a spinner (or similar) for the duration of in-flight query/queries.

> Just to double check: not including search keywords in URL is an intentional decision, right?

- Yes and no. I am aware of the omission. In the future, I think the committed query should be bound to `q=...` in the URL (using history API). IMHO, this one is definitely future work wrt this PR.

Hexcles:
IMHO, a spinner & some kind of prompts for pressing enter would solve most, if not all, of my confusion interacting with this UI. All the rest are optional / good-to-have.

lukebjerring:
FWIW, I'd expect to submit a search when an autocomplete result is selected (so both populate the search box with the selected completion, and run a search for it).

mdittmer:
@lukebjerring, yes. That's a bug. I need to figure out what event fires there and add it to this PR.

mdittmer:
@Hexcles fixed up about as many features as should probably go in this one PR. Will still need to de-conflict if #479 lands first, but the nature of the changes under review will remain the same.

webapp/components/wpt-results.html
```diff
@@ -4,13 +4,13 @@
...
-<link rel="import" href="../bower_components/polymer/lib/elements/dom-if.html">
-<link rel="import" href="../bower_components/polymer/lib/elements/dom-repeat.html">
 <link rel="import" href="../bower_components/paper-button/paper-button.html">
+<link rel="import" href="../bower_components/paper-spinner/paper-spinner-lite.html">
 <link rel="import" href="../bower_components/paper-styles/color.html">
 <link rel="import" href="../bower_components/paper-toast/paper-toast.html">
-<link rel="import" href="../bower_components/paper-button/paper-button.html">
+<link rel="import" href="../bower_components/polymer/lib/elements/dom-if.html">
+<link rel="import" href="../bower_components/polymer/lib/elements/dom-repeat.html">
+<link rel="import" href="../bower_components/polymer/polymer-element.html">
```
Hexcles: Just curious: does the import order matter?

webapp/components/wpt-results.html
```diff
@@ -370,9 +392,57 @@ <h3>
...
+      disconnectedCallback() {
+        this.shadowRoot.querySelector('test-search')
+          .removeEventListener('commit', this.onSearchCommit);
+        super.disconnectedCallback();
+      }
+
       ready() {
         super.ready();
         this.loadRuns().then(async runs => {
+          this.fetchResults();
```
Hexcles: Was this intended?

webapp/components/wpt-results.html
mdittmer: No. It only matters when we use `PolymerElement` as a base class.

webapp/components/wpt-results.html
mdittmer: Yes. We don't get an initial fetch without this.

Issue #468, opened 2018-08-23
Sort out the "complete" API parameter and its interaction with the frontend


foolip:
https://wpt.fyi/api/runs?complete=true does something different from https://wpt.fyi/api/runs but it is not documented in https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md

lukebjerring:
Not sure if there's an open issue for it, but the definition of complete actually needs addressing / changing as well as documentation.

`complete` should only be used when `sha` is `latest`, and should mean "the most recent run that contains all of the products in the query". Currently though, it is implemented as "the most recent run that contains all of the default products". These two definitions have the same meaning when the query does not contain specific products (and uses the default product set).

Hexcles:
I just discussed offline with @mdittmer . The problem is multifaceted:

* The "complete" parameter of `/api/runs` is a bit confusing as said above. We need to document its behaviour or even rework the parameter to better suit use cases (speaking of that, we should revisit the use case and intention).
* The frontend currently doesn't know about the "complete" parameter, which prevents it from applying more advanced logic related to the parameter.
     * Background: The parsing and deserialization of query parameters happen on the server side. Go code then fills the param struct back into Go templates to pass the parameters as properties to web components. Yet currently the web component doesn't have a "complete" parameter.
     * A demonstration of this problem is that adding `?complete=true` or `?complete=false` to front-end URLs doesn't do anything.

Mark and I think we should use a simple hack to provide a somewhat natural user experience so that we can make a prod release next week. After that, we can think about how to fix the issue properly.

lukebjerring:
> rework the parameter to better suit use cases

Addressed by #491 

> The frontend currently doesn't know about the "complete" parameter, 

Addressed by #498

Additional problems:
- `complete` has no effect in /interop (addressed by #488)
- `products` is always "chrome","edge","firefox","safari" (addressed by #491)
- `GetCompleteSHAs` implementation ignores labels (addressed by #491)

We're in a healthier state after those PRs land, by which stage we'll confirm expectations/use-cases, and document things

lukebjerring:
#532 renamed `complete` to `aligned`,  better reflecting it's meaning.

foolip:
Thanks Luke, that makes a lot more sense!

PR #476, opened 2018-08-23
Reland "Use ?complete=true as default query for /api/runs"

### LINKED ISSUE #441: Inconsistent runs on the frontpage

Hexcles:
Reverts web-platform-tests/wpt.fyi#475, i.e. relands https://github.com/web-platform-tests/wpt.fyi/pull/450

Fixes #441 . Also see #468 .

### DIFF
--- a/webapp/components/test-runs.html
+++ b/webapp/components/test-runs.html
@@ -151,8 +151,9 @@ found in the LICENSE file.
           params['max-count'] = maxCount;
         }
 
-        // Insist on complete runs iff neither sha nor products are specified.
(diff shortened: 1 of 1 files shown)


Hexcles:
@mdittmer PTAL again

PR #479, opened 2018-08-24
Fix diff view in wpt-results

### LINKED ISSUE #477: Diffing is broken on staging
e.g. https://staging.wpt.fyi/results/2dcontext?product=chrome&product=edge&diff=true

mdittmer:
Fixes #477

- Misspelled CSS class names
- Incorrect parameters to helper functions
- Broken/needless fetch of diff
- Fix application logic for 'none' CSS class:
  - Early return needed `!isDiff &&`
  - Do not use 'delta none'; other num (passes or total) may be non-zero
- nit: `getDiffDeltaStr` always returns string

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -200,9 +200,9 @@ found in the LICENSE file.
               </template>
               <template is="dom-if" if="[[diffShown]]">
                 <td class$="numbers [[ testResultClass(node, index, diffRun, 'passes') ]]">
-                  <span class$="passing [[ testResultClass(node, index, diffRun, 'passes') ]]">{{ getNodeResultDataByPropertyName(node, diffRun, 'passing') }}</span>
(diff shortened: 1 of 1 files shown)


webapp/components/wpt-results.html
```diff
@@ -507,11 +490,10 @@ <h3>
         }
 
         const result = node.results[index];
-        if (typeof result === 'undefined') {
+        const isDiff = this.isDiff(testRun);
+        if (!isDiff && typeof result === 'undefined') {
           return 'none';
```
mdittmer: Should this also check that `prop === 'total'`? It's not really none unless there are no tests (no passes doesn't matter).

Hexcles: Let's discuss tomorrow. I'm not quite sure about this.

mdittmer:
I think this reworking of the logic is much simpler to follow (also, comments are good). Basically, `none` is only used when the total number of tests is 0 (not by diff, by number of tests). Other than that, I just tried to group cases together logically.

@Hexcles LMK what you think of this formulation.

Hexcles:
LGTM again

PR #480, opened 2018-08-24
Lint subdirectories recursively

### LINKED ISSUE #460: Lint subdirectories recursively
Our `go_lint` presubmit check only lints the top-level directories. We need to run `golint foo/...` in order to recursively lint foo.

I tried locally, but there were too many previously undiscovered errors in `revisions/`, so it'd take some effort. cc @mdittmer

Hexcles:
Fixes #460 .


Hexcles:
Looks like most (if not all) of the errors are just missing comments. @mdittmer , could you take over this branch and add the missing comments when you have time? This is by no means urgent at all.

PR #484, opened 2018-08-26
Document impact of static_runs flag

### LINKED ISSUE #478: interop view is unusable locally, making debugging/development hard
We should also fetch some interop data from prod in `populate_dev_data.go`.

mdittmer:
Documents impact of `-static_runs=true` for `populate_dev_data.go`.

Fixes #478.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,14 @@ execute the script which leverages `util/populate_dev_data.go` by running:
 ./util/docker-dev/dev_data.sh
 ```
 
+*NOTE*: Some web UI views, such as `/interop`, only work in the presence of so-called "static runs". If some views
+do not work while debugging locally, try clearing your local datastore and memcache, then run:
+
+```sh
+source ./util/commands.sh
+wptd_exec go run ./util/populate_dev_data.go -static_runs=true
+```
+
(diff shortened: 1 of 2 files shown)


mdittmer:
Other options to fix this include:

- Make `static_runs` default value `true`
- Make `util/docker-dev/dev_data.sh` specify `-static_runs=true`

README.md
```diff
@@ -34,6 +34,14 @@ execute the script which leverages `util/populate_dev_data.go` by running:
 ./util/docker-dev/dev_data.sh
 ```
 
+*NOTE*: Some web UI views, such as `/interop`, only work in the presence of so-called "static runs". If some views
+do not work while debugging locally, try clearing your local datastore and memcache, then run:
```
Hexcles: @lukebjerring this is no longer true with your #487 , right?

lukebjerring: Yes, the intention is that `populate_dev_data` works locally with or without the static_runs. I've got to do another PR, to enumerate `complete` and not, first before I'll make the claim that we're in that state.

mdittmer: Found this old PR lying around. @lukebjerring you've been working on the related code. Is any of this PR still relevant or should I abandon it?

mdittmer:
*ping* @lukebjerring (see inline comment above)

PR #486, opened 2018-08-27
Add terse/verbose toggle to results and interop (dir-level) views

### LINKED ISSUE #422: Use the "verbose" toggle at the directory level too
https://staging.wpt.fyi/results/css/css-flexbox has some very long test names, and in general navigating across directories causes the layout to change a lot. Using a fixed-size layout for all views would fix this.

mdittmer:
Fixes #422.

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -108,12 +108,12 @@ found in the LICENSE file.
       padding: 0;
     }
 
-    td.path-part {
+    .terse td.path-part {
       min-width: 200px;
       background-color: white;
     }
 
-    td path-part {
+    .terse td path-part {
       box-sizing: border-box;
       height: 100%;
       left: 0;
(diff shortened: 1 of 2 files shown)


lukebjerring:
So, without diving into the code just yet:
I'm not sure we should have the toggle on directories. While some paths are (too) long, the ellipsis should rarely cause ambiguity at the folder level, and hovering resolves that in any case.

Hexcles:
I noticed a small UI quirk: expand the results in https://dir-verbose-dot-wptdashboard-staging.appspot.com/results/credential-management , and you'll see the first lines of the long test names have a small indent that doesn't exist on staging.wpt.fyi.

Hexcles:
A more general request: if we are changing the table layout, we should take the chance to improve the readability of the UI, namely to separate the rows more clearly in the expanded view.

mdittmer:
> I'm not sure we should have the toggle on directories. While some paths are (too) long, the ellipsis should rarely cause ambiguity at the folder level, and hovering resolves that in any case.

@foolip I think you requested this feature. WDYT?

foolip:
> @foolip I think you requested this feature. WDYT?

https://dir-verbose-dot-wptdashboard-staging.appspot.com/results/fullscreen/api?complete=true is a case where this feature helps a bit, but now I see that at the top level the feature is a bit confusing. It just slightly changes the layout and it might appear to just not do anything useful.

So, I'm going to apologize (sorry!) for not quite thinking this through, and agree with @lukebjerring that probably we don't need this.

PR #488, opened 2018-08-29
Handle ?complete=true in /api/interop

lukebjerring:
## Description
Handles ?complete=true (in the `/api/interop` endpoint), by consolidating the handling of params when querying for any non-default `interop` request.

## Review Information
- See [auto-deployed enviroment](https://complete-interop-dot-wptdashboard-staging.appspot.com/)
- Visit various param combos, e.g.
  - https://complete-interop-dot-wptdashboard-staging.appspot.com/api/interop/?complete=true
  - https://complete-interop-dot-wptdashboard-staging.appspot.com/api/interop/?complete=true&label=experimental
  - https://complete-interop-dot-wptdashboard-staging.appspot.com/api/interop/?complete=false&label=stable

### DIFF
--- a/api/interop.go
+++ b/api/interop.go
@@ -31,8 +31,7 @@ func apiInteropHandler(w http.ResponseWriter, r *http.Request) {
 	if !filters.IsDefaultQuery() {
 		// Load default browser runs for SHA.
 		// Ignore any max-count; makes no sense for a interop run.
-		one := 1
-		filters.MaxCount = &one
+		filters.MaxCount = nil
 		runs, err := LoadTestRunsForFilters(ctx, filters)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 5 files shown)


api/interop.go
```diff
@@ -32,8 +32,8 @@ func apiInteropHandler(w http.ResponseWriter, r *http.Request) {
 		// Load default browser runs for SHA.
 		// Ignore any max-count; makes no sense for a interop run.
 		one := 1
-		runs, err := shared.LoadTestRuns(
-			ctx, filters.GetProductsOrDefault(), filters.Labels, []string{filters.SHA}, filters.From, filters.To, &one)
+		filters.MaxCount = &one
```
Hexcles: Is this equivalent to just assigning `nil` to `filters.MaxCount`?

lukebjerring: Yeah; changed to use `nil`.

lukebjerring:
You're right that the UI won't work; will make this PR about handling ?complete=true correctly, and then follow up with a PR to fix the UI behaviour.

lukebjerring:
FYI - integration tests are failing in a way that is the `complete` feature WAI.
The static runs we set up for the test are not `complete`, so it causes the `interop` view to 404 (and potentially has other ramifications).

Hexcles:
LGTM

PR #489, opened 2018-08-29
Refactor the way 'complete' is implemented

lukebjerring:
## Description
Migrates the `getCompleteSHAs` code to the shared codebase, and re-jigs it a bit, such that we grab the default products set, and find (any) product that matches the run. Since the default product specs are versionless, this is a no-op, intended to be followed up with actual changes in another PR.

## Review Information
In particular, review the code for `ProductSpec.Matches(TestRun)`

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -146,3 +146,61 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
 		Filter(fieldName+" >=", fmt.Sprintf("%s.", versionPrefix)).
 		Filter(fieldName+" <=", fmt.Sprintf("%s.%c", versionPrefix, '9'+1))
 }
+
+// GetCompleteRunSHAs returns an array of the SHA[0:10] for runs that
+// exists for all initially-loaded browser names (see GetDefaultBrowserNames),
+// ordered by most-recent.
+func GetCompleteRunSHAs(ctx context.Context, from, to *time.Time, limit *int) (shas []string, err error) {
+	query := datastore.
+		NewQuery("TestRun").
+		Order("-TimeStart").
+		Project("Revision", "BrowserName")
+
+	// TODO(lukebjerring): Pass in products.
+	products := GetDefaultProducts()
+
+	if from != nil {
+		query = query.Filter("TimeStart >=", *from)
+	}
+	if to != nil {
+		query = query.Filter("TimeStart <", *to)
+	}
+
+	bySHA := make(map[string]mapset.Set)
+	done := mapset.NewSet()
+	it := query.Run(ctx)
+	for {
+		var testRun TestRun
+		var matchingProduct *ProductSpec
+		_, err := it.Next(&testRun)
+		if err == datastore.Done {
+			break
+		} else if err != nil {
+			return nil, err
+		} else {
+			for _, product := range products {
+				if product.Matches(testRun) {
+					matchingProduct = &product
+					break
+				}
+			}
+		}
+		if matchingProduct == nil {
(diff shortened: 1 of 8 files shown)


shared/datastore.go
```diff
@@ -146,3 +146,60 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
...
 		Filter(fieldName+" <=", fmt.Sprintf("%s.%c", versionPrefix, '9'+1))
 }
+
+// GetCompleteRunSHAs returns an array of the SHA[0:10] for runs that
+// exists for all initially-loaded browser names (see GetDefaultBrowserNames),
+// ordered by most-recent.
+func GetCompleteRunSHAs(ctx context.Context, from, to *time.Time, limit *int) (shas []string, err error) {
+	query := datastore.
+		NewQuery("TestRun").
+		Order("-TimeStart")
```
lukebjerring:
```diff
-		Project("Revision", "BrowserName")
```

shared/datastore.go
```diff
@@ -146,3 +146,60 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
...
+
+	// TODO(lukebjerring): Pass in products.
+	products := GetDefaultProducts()
```
lukebjerring:
```diff
-	browserNames := shared.GetDefaultBrowserNames()
+	products := GetDefaultProducts()
```

shared/params.go
```diff
@@ -73,6 +73,29 @@ type ProductSpec struct {
...
+func (p ProductSpec) Matches(run TestRun) bool {
+	if run.BrowserName != p.BrowserName {
+		return false
+	}
+	if !IsLatest(p.Revision) && p.Revision != run.Revision {
+		return false
+	}
+	if p.Labels != nil && p.Labels.Cardinality() > 0 {
+		runLabels := run.LabelsSet()
+		if runLabels.Intersect(p.Labels).Cardinality() < p.Labels.Cardinality() {
```
Hexcles: `p.IsSubset(runLabels)`

shared/params.go
```diff
@@ -73,6 +73,29 @@ type ProductSpec struct {
...
+			return false
+		}
+	}
+	if p.BrowserVersion != "" {
+		// Make "6" not match "60.123" by adding trailing dots to both.
+		if strings.Index(run.BrowserVersion+".", p.BrowserVersion+".") != 0 {
```
Hexcles: I think `HasPrefix` would be slightly faster and clearer.

```go
if runVersion := run.BrowserVersion+"."; !runVersion.HasPrefix(p.BrowserVersion+".") {
```

shared/datastore.go
```diff
@@ -146,3 +146,60 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
...
+					break
+				}
+			}
+		}
+		if matchingProduct == nil {
+			continue
+		}
+		set, ok := bySHA[testRun.Revision]
+		if !ok {
+			bySHA[testRun.Revision] = mapset.NewSetWith(*matchingProduct)
```
Hexcles: I think we can simply store pointers in the mapset. Not sure if it'd be cheaper.

shared/params.go

shared/params.go
lukebjerring: I think it's `strings.HasPrefix(s, prefix)`, but, done nonetheless.

Hexcles: Oops. Apparently I couldn't wrap my head around Python and Go.

PR #491, opened 2018-08-29
Allow custom products when computing `complete`

lukebjerring:
## Description
Builds on #489 to actually "fix" the behaviour of combining `labels` + `complete` when getting SHAs.

We're currently just ignoring labels when we get a list of complete SHAs, which may mean that if we limit the number to 1, we get a SHA that isn't complete for those labels.

e.g.
- https://staging.wpt.fyi/api/shas?label=stable&complete&max-count=1 (currently says `["6a9c5cbeae"]`)
- https://staging.wpt.fyi/test-runs?labels=stable (shows the most recent complete is `d2fa1bdc90`)

This fixes the issue, e.g. 
https://complete-from-products-dot-wptdashboard-staging.appspot.com/api/shas?label=stable&complete&max-count=1 now shows `d2fa1bdc90`

Also supports the corrected meaning of `complete` to be product combos, e.g.
https://complete-from-products-dot-wptdashboard-staging.appspot.com/api/shas?label=experimental&products=chrome,firefox&complete

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -150,14 +150,22 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
 // GetCompleteRunSHAs returns an array of the SHA[0:10] for runs that
 // exists for all initially-loaded browser names (see GetDefaultBrowserNames),
 // ordered by most-recent.
-func GetCompleteRunSHAs(ctx context.Context, from, to *time.Time, limit *int) (shas []string, err error) {
+func GetCompleteRunSHAs(
+	ctx context.Context,
+	products ProductSpecs,
+	labels mapset.Set,
+	from,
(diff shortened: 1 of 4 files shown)


shared/datastore.go
```diff
@@ -150,14 +150,22 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
 // GetCompleteRunSHAs returns an array of the SHA[0:10] for runs that
 // exists for all initially-loaded browser names (see GetDefaultBrowserNames),
```
Hexcles: Please update the comment accordingly.

lukebjerring:
Will wait for a redeploy with master merged in, as there's some behavioural changes around `complete` and `label=experimental` to be confirmed.

PR #493, opened 2018-08-29
Add a complete and non-complete run in populate_dev_data

lukebjerring:
Ensures that both `?complete=true` and not will have some results locally.


lukebjerring:
@Hexcles this is ready for review once #489 has landed, as there's a bunch of common changes. The only real changes are in `populate_dev_data`.

PR #496, opened 2018-08-30
File-level view data mirrors view structure

### LINKED ISSUE #494: Inconsistent harness status between terse & verbose views after navigation
Steps to reproduce:

In an **incognito window**:

1. Go to https://staging.wpt.fyi/results/dom/historical.html
2. Click on the "expand" toggle to switch to verbose view
3. Navigate to `/dom/interface-objects.html` by *following hyperlinks*.
4. Click the toggle. You'll see inconsistent harness status between the two views (namely, for Edge 17).

The theory is that only the harness status in the active view gets updated after navigation.

mdittmer:
Fixes #494.

Polymer was only updating table rows when their name column happened to change. This was a consequence of depending on `testRuns` and `subtestName`. In the case of the "Harness status" row, the `subtestName` didn't change and Polymer doesn't know that `testRuns[i].subtests.STATUS.*` changed, so so we observed #494.

This change replaces most of the `test-file-results*.html` properties with a `resultsTable` that contains an array of`{name, results: [{status, message}]` where the inner array is parallel to the `testRuns` array. This mirrors the structure of the views themselves and simplifies the binding logic, while also fixing #494.

### DIFF
--- a/webapp/components/test-file-results.html
+++ b/webapp/components/test-file-results.html
@@ -87,54 +87,49 @@ found in the LICENSE file.
       }
 
       async fetchTestFile(path, testRuns) {
-        this.subtestNames = []; // Clear any existing rows.
+        this.resultsTable = []; // Clear any existing rows.
         if (!path || !testRuns) {
           return;
         }
         const resultsPerTestRun = await Promise.all(
           testRuns.map(tr => this.loadResultFile(tr)));
 
-        // resultsTable[0].name set after discovering subtests. Use "" to
-        // remain first in resultsTable.sort() later.
+        // resultsTable[0].name set after discovering subtests.
         let resultsTable = [{
-          name: '',
           results: resultsPerTestRun.map(data => {
             return {status: data && data.status};
           }),
         }];
 
-        let subtestNamesSet = new Set();
-        resultsPerTestRun
-          .forEach(data => data && data.subtests && data.subtests
-            .forEach(sub => subtestNamesSet.add(sub.name)));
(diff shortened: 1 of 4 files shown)


Hexcles:
So it indeed crashes on this example I sent you earlier: https://test-file-results-dot-wptdashboard-staging.appspot.com/results/webdriver/tests/actions/bounds.py

webapp/components/test-file-results.html
```diff
@@ -97,33 +91,43 @@
...
-        this.hasSubtests = hasSubtests;
-        this.subtestNames = subtestNames;
+        // resultsTable[0].name set after discovering subtests.
+        let resultsTable = [{
+          results: resultsPerTestRun.map(data => {
+            return {status: data && data.status};
+          }),
+        }];
+
+        let subtestNamesSet = new Set();
```
Hexcles: Is Javascript `Set` ordered? If not, this changes the behaviour, i.e. the order of subtests is no longer preserved. Although the current behaviour (sorting based on the first test run) isn't perfect either, I do think it's better than not sorting subtests at all. For tests like idlharness.js tests, we do want to try our best to preserve the ordering.

I guess this can be changed to a simple array, which will increase the complexity of the insert-if-not-exist operation below to O(N^2), but I think that's fine.

Hexcles:
https://test-file-results-dot-wptdashboard-staging.appspot.com/results/webdriver/tests/actions/bounds.py?label=stable

PR #499, opened 2018-08-30
Include URL params in href (especially path-part)

### LINKED ISSUE #193: Navigation by middle-click not working in Safari and Firefox

Hexcles:
So that "open in new tab" (or equivalant) can preserve URL parameters
correctly. Towards #193 resolution.

## Review Information

Go to `/results/?complete=true&label=stable` on the staging instance. Hovering on various links (especially the top path just below the tab switcher, and the list of subdirectories/files), and you should see the URL parameters included in the linked paths.

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -121,9 +121,9 @@ found in the LICENSE file.
   <section class="search">
     <div class="path">
       <a href="/interop/" on-click="navigate">wpt</a>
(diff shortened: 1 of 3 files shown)


lukebjerring:
Note: The leading `wpt` the split paths isn't including the query

PR #506, opened 2018-08-31
Fail nicely when an attempted run fetch fails.

### LINKED ISSUE #478: interop view is unusable locally, making debugging/development hard

lukebjerring:
(Last step that) fixes #478


lukebjerring:
Note that we don't actually have any `complete` runs for `label=experimental`, which is now a query that actually behaves as expected (and 404s).

PR #507, opened 2018-08-31
Handle no test runs better.

lukebjerring:
## Description
Pulls the `.info` styles out into a reusable component, and surfaces another one for the case that we're finished loading but there's no test runs.

@mdittmer - this also re-jigs the code around the spinner (needed `isLoading` elsewhere) - let me know if the `newValue` / `oldValue` stuff was important in some way that I'm missing.

## Review Information
See autodeployed environment, load something with no results, e.g. /results?sha=1234567890


webapp/components/info-banner.html
```diff
@@ -0,0 +1,56 @@
...
+        float: right;
+      }
+      .banner.error {
+        background-color: var(--paper-red-100);
+        border-left: solid 4px var(--paper-red-300);
+      }
+    </style>
+
+    <section class$="banner [[type]]">
+      <slot></slot>
```
mdittmer: What's this for? Does it denote where content prior to content inside `<... slot="small">...</...>` will go?

webapp/components/wpt-results.html
```diff
@@ -173,56 +164,64 @@
...
-              <th><test-run test-run="[[testRun]]"></test-run></th>
-            </template>
-            <template is="dom-if" if="[[diffShown]]">
-              <th><test-run test-run="[[diffRun]]"></test-run></th>
-            </template>
-          </tr>
-        </thead>
-
-        <tbody>
+    <template is="dom-if" if="[[hasRuns]]">
```
mdittmer: This diff seems needlessly hard to follow. Did you just wrap existing markup in `<template is="dom-if" if="[[hasRuns]]">...</template>`, or is there more here?

webapp/components/wpt-results.html
```diff
@@ -406,20 +412,12 @@ <h3>
...
-      observeLoadingCount(newValue, oldValue) {
-        if (newValue === oldValue) {
-          return;
-        }
-        const elem = this.shadowRoot.querySelector('paper-spinner-lite');
-        if (!elem) {
-          return;
-        }
+      computeIsLoading(loadingCount) {
+        return loadingCount;
```
mdittmer: nit: Explicit coercing to Boolean a bit clearer: `!!loadingCount`.

webapp/components/info-banner.html
lukebjerring: Correct - nameless is the default `slot`, which takes the (unslotted) child elements.
The other slot is named, so takes the element with `slot="small"`

https://www.polymer-project.org/2.0/docs/devguide/shadow-dom#shadow-dom-and-composition

webapp/components/wpt-results.html
lukebjerring: Yep.

webapp/components/wpt-results.html

api/query/autocomplete.go
```diff
@@ -123,7 +123,7 @@ func prepareAutocompleteResponse(limit int, filters *shared.QueryFilter, testRun
 		}
 	}
 
-	files := make([]AutocompleteResult, 0, fileSet.Cardinality()/len(testRuns))
+	var files []AutocompleteResult
```
Hexcles: Hmm, why is this changed?

lukebjerring: Drive-by fix for 500 when no test runs are found for an autocomplete query.

Hexcles: Hmm, sorry could you explain what was wrong with the original statement?

lukebjerring: `len(test runs)` is zero when no test runs get loaded. Division by zero is bad.

Hexcles: Ahh I see. Thanks!

PR #509, opened 2018-08-31
Make search and autocomplete case-insensitive

### LINKED ISSUE #508: Search and autocomplete are case-sensitive
They should be case-insensitive.

mdittmer:
This creates a helper for "canonicalizing" strings in the context of the `api/query` package. This is used sparingly in the search and autocomplete implementations to perform case-insensitive comparisons.

Fixes #508.

### DIFF
--- a/api/query/search_test.go
+++ b/api/query/search_test.go
@@ -7,13 +7,15 @@
 package query
 
 import (
+	"sort"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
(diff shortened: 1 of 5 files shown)


api/query/search_test.go
```diff
@@ -27,7 +27,7 @@ func TestPrepareSearchResponse(t *testing.T) {
 	}
 	filters := shared.QueryFilter{
 		RunIDs: runIDs,
-		Q:      "/b/",
+		Q:      "/B/",
```
Hexcles: I'd like to see one more test where the paths are capital but the query string is lowercase.

PR #512, opened 2018-09-03
Copy immutable `legacy_status` data by value into mutable `results` entries

### LINKED ISSUE #510: Summary numbers incorrect when navigating into subdirectories
Repro steps:

Start at https://wpt.fyi/results/webaudio/the-audio-api?complete=true. You should see this:
screen shot 2018-08-31 at 10 25 22 pm

Then, click "the-analysernode-interface/" to reach https://wpt.fyi/results/webaudio/the-audio-api/the-analysernode-interface?complete=true. You should see this:
screen shot 2018-08-31 at 10 26 22 pm

Note how the summary numbers are the same for the first row! Reload the second page to instead see this:
screen shot 2018-08-31 at 10 27 10 pm

mdittmer:
Fixes #510.

`wptResultsInstance.searchResults[i].legacy_status` contains the pass rate data for a single file. However, it was being copied by reference into `wptResultsInstance.searchResults.displayedNodes[j].results`, which models (sometimes) the pass rate data for an entire directory. Hence, when the `wptResultsInstance.searchResults.displayedNodes[j].results` was mutated to accumulate all data for its directory, "immutable-after-fetched" `searchResults` nodes were being mutated.

This fixes the problem by copying the values inside `legacy_status` properties into new `results` properties.

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -528,7 +528,9 @@ found in the LICENSE file.
                 acc.push({
                   path: `${prefix}${name}`,
                   isDir,
-                  results: Array.from(r.legacy_status),
+                  // Copy legacy_status node contents.
+                  results: r.legacy_status
+                    .map(stats => Object.assign({}, stats)),
                 });
               } else {
                 const rs = r.legacy_status;


foolip:
> too bad we can't easily write a test for this issue right now

Why is that the case? Would the test have to build on some existing test that is disabled? Adding tests that work is something we can currently do, right?

PR #518, opened 2018-09-04
Abstract GAE-compatible logging interface

mdittmer:
Logging in GAE Standard Environment goes nowhere when the golang `"log"` package is used. This change introduces an abstract `Logger` interface that is `".../appengine/log"`-compatible. Other implementations are available for GAE Flex (`stdLogger`) and test (`nilLogger`) environments.

Might be easier to review commit-by-commit.

### DIFF
--- a/shared/datastore_medium_test.go
+++ b/shared/datastore_medium_test.go
@@ -1,6 +1,6 @@
 // +build medium
 
-package shared
+package shared_test
 
 import (
 	"strconv"
@@ -9,6 +9,7 @@ import (
 	"time"
 
 	mapset "github.com/deckarep/golang-set"
+	"github.com/web-platform-tests/wpt.fyi/shared"
 	"github.com/web-platform-tests/wpt.fyi/shared/sharedtest"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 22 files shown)


shared/util.go
```diff
@@ -40,3 +47,116 @@ func ToStringSlice(set mapset.Set) []string {
...
+)
+
+// NewGAELogger returns a Google App Engine Standard Environment logger bound to
+// the given context.
+func NewGAELogger(ctx context.Context) Logger {
+	return gaeLogger{ctx}
+}
+
+// NewSTDLogger returns a new standard logger.
+func NewSTDLogger() Logger {
```
Hexcles: I think the convention is `Std`.

shared/datastore_medium_test.go
```diff
@@ -1,6 +1,6 @@
 // +build medium
 
-package shared
+package shared_test
```
Hexcles: Nit: `sharedtest` to match the name of the directory.

shared/sharedtest/util.go
```diff
@@ -5,8 +5,8 @@
 package sharedtest
 
 import (
+	"github.com/web-platform-tests/wpt.fyi/shared"
```
Hexcles: Nit: move this import to the end. And while we are at it, we should use `context` instead of `golang.org/x/net/context`.

shared/datastore_medium_test.go
mdittmer: This is actually a deliberate convention. `foo/*_test.go` files may be `package foo` or `package foo_test`. This is `shared_test` to a void a dependency cycle between `shared` and `shared_test`.

Hexcles: Hah I see, so it's also different from `shared/sharedtest`.

Hexcles:
@mdittmer Mark, could you try accessing the affected APIs on the staging instance and verify that logs show up on GCP dashboard?

PR #521, opened 2018-09-05
Fix two query API bugs

mdittmer:
1. `gzip` reader/writers do not close their delegates
    1. Wrap in new `compositeReadWriteCloser`
1. `*memcacheWriteCloser` ptr type (not `memcacheWriteCloser` value type) should be `io.WriteCloser`

### DIFF
--- a/api/query/query.go
+++ b/api/query/query.go
@@ -64,7 +64,17 @@ func (gz gzipReadWritable) NewReadCloser(id string) (io.ReadCloser, error) {
 	if err != nil {
 		return nil, err
 	}
-	return gzip.NewReader(r)
+
+	gzr, err := gzip.NewReader(r)
+	if err != nil {
+		return nil, err
+	}
+
+	return compositeReadWriteCloser{
+		reader:   gzr,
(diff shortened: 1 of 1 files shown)


Hexcles:
https://golang.org/doc/faq#methods_on_values_or_pointers

api/query/query.go
```diff
@@ -97,17 +135,17 @@ func (mc memcacheReadWritable) NewReadCloser(key string) (io.ReadCloser, error)
 }
 
 func (mc memcacheReadWritable) NewWriteCloser(key string) (io.WriteCloser, error) {
```
Hexcles: @mdittmer I know you've merged the PR but still want to share a thought on this:

The fact that changing the return value doesn't change the function signature reminds me of a Go convention that we didn't follow here: take *interfaces* as arguments, and return *concrete* types. In this example, we should return `*memcacheWriteCloser`. Explicitness in return values benefits the callers (and callers don't have to care - they can always use `:=` or even assign the return value to an interface type), while the polymorphism in arguments makes the function more generic.

Issue #451, opened 2018-08-17
Interop tab is broken when starting from /results/?label=stable


foolip:
Starting at https://wpt.fyi/results/?label=stable, the interop tab link is https://wpt.fyi/interop/?label=stable, but that view fails to load because https://wpt.fyi/api/interop?label=stable is 404.

We should probably keep only the parameters we know are supported for the interop view, or alternatively keep them, strip them internally in the interop view, and show a message that the showed results don't match what was requested.

foolip:
This is probably a regression from https://github.com/web-platform-tests/wpt.fyi/pull/385.

Hexcles:
Travis is now failing on master because of this regression: https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/419475391#L3901

Hexcles:
@mdittmer , I was wrong about this being "easy to fix".

The query parameters never work reliably for the interop view.

Since we don't embed test runs inside interop result records anymore (because Datastore doesn't support nested arrays), we use a hack to query interop results: for complicated queries, we find the four *test runs* that match the query first, and then search for the interop run that is calculated from the four test runs.

However, we only calculate interop results once a day, which means there is a significant gap in each day where no interop results are available for the latest stable/experimental runs.

I thought we could fix it by attempting multiple times: i.e. get several *sets* of test runs and see if any set is included in an interop run, but this is tricky as the stable and experimental runs aren't necessarily complete (SHA-aligned) runs (see the [related discussion](https://github.com/web-platform-tests/results-analysis/issues/45)), which means we don't naturally have "sets" of test runs, intead we are looking at a combinotorial problem...

The other way around is to fetch the recent N interop runs, and see if any of them has the test runs that can satisfy the query parameters. The downside of this approach is that it'd be hard to query the interop result at an old SHA.

I wish we could join the tables.

Hexcles:
@foolip as a stopgap, we can drop support for all query parameters except "label" in the interop view. I think this is the most important use case, more so than finding the interop results at a particular historical revision. WDYT?

Issue #514, opened 2018-09-03
Interop view empty?


gsnedders:
Currently, going to https://wpt.fyi and clicking on "Interoperability" shows an empty page, with no browsers and no paths loaded.

```
[Error] Failed to load resource: the server responded with a status of 404 () (https://wpt.fyi/api/interop?complete=true, line 0)
[Error] Unhandled Promise Rejection: Failed to fetch interop data
	(anonymous function) (data:text/javascript;charset=utf-8,%0A…%0A:70)
	asyncFunctionResume
	(anonymous function)
	promiseReactionJob
```

lukebjerring:
The reason for this issue is that the CRON job that computes the interop data is running daily, but occassionally (most of the time) we have complete runs land sometime throughout the day, without the interop being re-computed.

Ideally, we need to:
- change to a push-based model
- run the interop more frequently / on update
- have the UI fall back to the most recent instead of 404'ing

In the meantime, whenever the complete run is ahead of interop, it can be quick-fixed by (manually) re-running the Travis CRON job for `labels=""` @ https://travis-ci.org/web-platform-tests/results-analysis

@Hexcles and I have discussed that push-events can be done from the results-receiver (when a new run is created). We'll need an "subscriber" endpoint for executing a check as to interop scores being up-to-date or not.

mdittmer:
This occurs because interop computations are only run daily, but the client code assumes that the latest runs that match query params must have interop data (which is often not the case). @lukebjerring and I have discussed a few long- and short-term fixes to this.

The most immediate fix is to move those cron jobs to run hourly so that the data are correct more often than they are now.

PR #523, opened 2018-09-06
Fall back to most-recent interop run when SHA is latest

lukebjerring:
## Description
Works towards preventing https://github.com/web-platform-tests/wpt.fyi/issues/514

## Review Information
To test, you'll need to manually ensure that the latest `complete` run does not have interop data, by deleting it if it does.

- Visit the [interop](https://interop-fallback-dot-wptdashboard-staging.appspot.com/interop) tab
  - Observe that you can see some interop scores for a complete run, and what the SHA is
- Visit the corresponding [results](https://interop-fallback-dot-wptdashboard-staging.appspot.com/results) (Test Results) tab
  - Observe that the SHA is different and the runs more recent

## Changes
To achieve this behaviour, `GetCompleteSHAs` has been rejigged to return not only the SHAs in chronological order, but a map of SHA to TestRun keys.

`LoadTestRuns` is now a wrapper for calls to both `LoadTestRunKeys` and `LoadTestRunsForKeys`. This ensures that we can run key-only queries in different contexts efficiently.

Interop has also changed from "find the (one run only) test runs matching the query, then find the single interop data that covers those test runs" to "find up to 128 test runs that match the query, then iterate through the interop runs until we find one where all 4 `TestRunIDs` are present in the set of keys that matched the query.


shared/datastore.go
```diff
@@ -21,23 +21,20 @@ func LoadTestRun(ctx context.Context, id int64) (*TestRun, error) {
 	return &testRun, nil
 }
 
-// LoadTestRuns loads the TestRun entities for the given parameters.
+// LoadTestRunKeys loads the keys for the TestRun entities for the given parameters.
```
mdittmer: Can we get tests for the two new functions in this file?

mdittmer: Related note: GitHub's crappy diff is making it hard to do a thorough review of this file. I'm sure once I review your tests for the functions I'll be confident that the code is correct :smiley:

shared/util.go
```diff
@@ -16,6 +16,9 @@ import (
 // ExperimentalLabel is the implicit label present for runs marked 'experimental'.
 const ExperimentalLabel = "experimental"
 
+// LatestSHA is a helper for the 'latest' keyword/special-case.
```
mdittmer: nit: `special case`.

util/populate_dev_data.go
```diff
@@ -229,19 +229,22 @@ func copyProdRuns(ctx context.Context, filters shared.TestRunFilter) {
...
-		var localRunCopies []shared.TestRun
-		localRunCopies, err = shared.LoadTestRuns(ctx, shared.GetDefaultProducts(), filters.Labels, shas, nil, nil, &one)
-		if len(localRunCopies) != len(prodPassRateMetadata.TestRunIDs) {
-			sha := "latest"
+			var shas []string
+			var keys map[string][]*datastore.Key
+			shas, keys, _ = shared.GetCompleteRunSHAs(ctx, shared.GetDefaultProducts(), filters.Labels, nil, nil, &one)
 			if len(shas) > 0 {
 				sha = shas[0]
+				localRunCopies, _ = shared.LoadTestRunsByKeys(ctx, keys[sha])
```
mdittmer: Check for and log error?

util/populate_dev_data.go
```diff
@@ -229,19 +229,22 @@ func copyProdRuns(ctx context.Context, filters shared.TestRunFilter) {
...
 		if complete {
-			shas, _ = shared.GetCompleteRunSHAs(ctx, shared.GetDefaultProducts(), filters.Labels, nil, nil, &one)
-		}
-		var localRunCopies []shared.TestRun
-		localRunCopies, err = shared.LoadTestRuns(ctx, shared.GetDefaultProducts(), filters.Labels, shas, nil, nil, &one)
-		if len(localRunCopies) != len(prodPassRateMetadata.TestRunIDs) {
-			sha := "latest"
+			var shas []string
+			var keys map[string][]*datastore.Key
+			shas, keys, _ = shared.GetCompleteRunSHAs(ctx, shared.GetDefaultProducts(), filters.Labels, nil, nil, &one)
```
mdittmer: Check for and log error?

shared/datastore.go
```diff
@@ -164,7 +164,8 @@ func VersionPrefix(query *datastore.Query, fieldName, versionPrefix string, desc
 }
 
 // GetCompleteRunSHAs returns an array of the SHA[0:10] for runs that
-// exists for all the given products, ordered by most-recent.
+// exists for all the given products, ordered by most-recent, as well as a map
+// of those SHAs to the TestRun keys for the complete run.
```
mdittmer: nit: Discourage this sense of the word "complete".

lukebjerring:
Note that this also fixes #451

webdriver/label_test.go
```diff
@@ -31,11 +31,8 @@ func TestLabelParam(t *testing.T) {
 	} else {
 		// Local static data only have 2 experimental browsers.
 		testLabel(t, wd, app, "/", "experimental", "wpt-results", 2)
-		// TODO(Hexcles): "/interop?label=" is broken on staging (#451),
-		// so we only run this test locally. Once the regression is
-		// fixed, we should run it against staging, too.
-		testLabel(t, wd, app, "/interop", "stable", "wpt-interop", 4)
 	}
+	testLabel(t, wd, app, "/interop", "stable", "wpt-interop", 4)
```
lukebjerring: Note that this was tested locally against the deployment of this PR, by executing:

    make go_firefox_test FIREFOX_PATH=~/browsers/firefox/firefox FLAGS="-remote_host=interop-fallback-dot-wptdashboard-staging.appspot.com -staging=true -test.run=Label"

api/interop.go
```diff
@@ -18,53 +20,123 @@ import (
...
+				http.Error(w, "No metrics runs found", http.StatusNotFound)
+				return
+			}
+			keysChecker = checkKeysMatchQuery(keys)
+		}
+	}
+
+	// Iterate until we find interop data where its TestRunIDs match the query.
+	var interop metrics.PassRateMetadata
+	it := query.Run(ctx)
```
Hexcles: I'm not seeing any explicit limit on the query. In the worst case, how many iterations will we do?

lukebjerring: I'll cap to 1000. Though, interestingly, this points out a problem with loading really, really old interop data (e.g. for a very specific SHA).

lukebjerring:
I've deployed this as a staging branch in the production environment to better demonstrate its effect.
https://interop-fallback-dot-wptdashboard.appspot.com/interop/?label=stable&aligned=true

api/interop.go
```diff
@@ -18,53 +20,123 @@ import (
...
+	keysFilter := mapset.NewSet()
+	for _, key := range keys {
+		keysFilter.Add(key.IntID())
+	}
+	return func(ids shared.TestRunIDs) bool {
+		all := true
+		for _, id := range ids {
+			if !keysFilter.Contains(id) {
+				all = false
+				break
```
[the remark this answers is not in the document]
lukebjerring: Uh, not quite. We're in a nested loop here, so need to consider each of the other sets of `aligned` keys (returning false would return on the first check, ignoring all other aligned sets).

api/interop.go
```diff
@@ -18,53 +20,123 @@ import (
...
+				return
+			} else if len(shaKeys) < 1 {
+				// Bail out early - can't find any complete runs.
+				http.Error(w, "No metrics runs found", http.StatusNotFound)
+				return
+			}
+			keysChecker = checkKeysAreAligned(shaKeys)
+		} else {
+			// Load default browser runs for SHA.
+			// Force any max-count to one; more than one of each product makes no sense for a interop run.
```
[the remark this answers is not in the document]
lukebjerring: Revised.

api/interop.go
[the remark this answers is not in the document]
Hexcles: Hmm, I don't see a nested loop here; there's one in `checkKeysAreAligned`. Anyway, it doesn't matter much.

api/interop.go
Hexcles: Let's make this 100.

Hexcles:
`sha` by itself works: https://interop-fallback-dot-wptdashboard.appspot.com/interop/?sha=9614def367

lukebjerring:
So, the problem was a lack of support for the legacy format of the data. I've included the fix, but also I've deleted runs with the nested entities in datastore (both prod and staging), which tees us up to remove the legacy struct in a follow-up PR.

PR #526, opened 2018-09-10
Default to complete (aligned), stable runs

### LINKED ISSUE #468: Sort out the "complete" API parameter and its interaction with the frontend

lukebjerring:
## Description
Further steps toward #468

## Review Information
- See [homepage](https://stable-default-dot-wptdashboard-staging.appspot.com/).
  - Observe aligned, stable run.
- See [`?complete=false`](https://stable-default-dot-wptdashboard-staging.appspot.com/?complete=false)
  - Observe non-aligned, not-just-stable, runs

## Changes
This migrates the "fall back to a default query of `{complete, stable}`" behaviour from the client to the server, such than there is never any ambiguity of the `?complete=false` case (which was previously ignored, since "not setting complete" and "complete is false" were indistinguishable in the component.


webapp/test_results_handler.go
```diff
@@ -12,6 +12,8 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/deckarep/golang-set"
```
Hexcles: Nit: can we explicitly alias this to `mapset` like we always do? I think it's a good practice because otherwise it's not obvious what the package name is.

webapp/test_results_handler.go
```diff
@@ -99,6 +101,13 @@ func parseTestRunUIFilter(r *http.Request) (filter testRunUIFilter, err error) {
 		return filter, err
 	}
 
+	// Default to complete (aligned), stable runs.
+	if testRunFilter.IsDefaultQuery() {
```
Hexcles: IMHO overriding `testRunFilter` here is a bit confusing, especially because we have already a similar behaviour in the `ToQuery` method with a subtle difference that `ToQuery(true)` only sets `complete`, but does not set `label=stable`. 

I know `ToQuery` has a different purpose and isn't suitable here. What about hoisting the logic of setting "default" params into another method, e.g. `SetDefaultParamsIfEmpty`, and calling that member both here and in `ToQuery`?

lukebjerring: I, uh, took the general idea but called it `OrDefault()`... looks kinda fluent:

    filters.OrDefault().ToQuery().Encode()

PR #529, opened 2018-09-10
[processor] Save the merged report and infer labels from browser_channel

Hexcles:
This change makes the results processor to save the merged report
instead of just the last chunk of the report if the original report is
chunked when uploaded.

Besides, it also sets the channel label (stable/experimental) based on a
new property in the report, `run_info.browser_channel`, introduced in
https://github.com/web-platform-tests/wpt/pull/12679 .

## Review

The integration, including the modified upload process for full raw reports, is tested by deploying to staging.wpt.fyi. The heuristics for adding labels are tested in unit tests.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -168,11 +168,19 @@ def task_handler():
         # to tell TaskQueue to drop the task.
         return ('', HTTPStatus.NO_CONTENT)
 
-    resp = "{} results loaded from {}\n".format(len(report.results), gcs_path)
+    resp = "{} results loaded from {}\n".format(
+        len(report.results), str(gcs_paths))
 
     raw_results_gcs_path = '/{}/{}/report.json'.format(
         config.raw_results_bucket(), report.sha_product_path)
-    gsutil.copy('gs:/' + gcs_path, 'gs:/' + raw_results_gcs_path)
+    if result_type == 'single':
+        # If the original report isn't chunked, we store it directly without
+        # the roundtrip to serialize it back.
+        gsutil.copy('gs:/' + gcs_paths[0], 'gs:/' + raw_results_gcs_path)
+    else:
+        with tempfile.NamedTemporaryFile(suffix='.json.gz') as temp:
+            report.serialize_gzip(temp.name)
+            gsutil.copy(temp.name, 'gs:/' + raw_results_gcs_path, gzipped=True)
 
     tempdir = tempfile.mkdtemp()
     try:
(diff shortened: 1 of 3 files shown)


results-processor/main.py
```diff
@@ -168,11 +168,19 @@ def task_handler():
         # to tell TaskQueue to drop the task.
         return ('', HTTPStatus.NO_CONTENT)
 
-    resp = "{} results loaded from {}\n".format(len(report.results), gcs_path)
+    resp = "{} results loaded from {}\n".format(
+        len(report.results), str(gcs_paths))
```
foolip: For other reviewers, what changed here is `str(gcs_paths)`.

results-processor/wptreport.py
```diff
@@ -414,6 +414,16 @@ def finalize(self):
         self.sha_product_path
         self.test_run_metadata
 
+    def serialize_gzip(self, filepath):
+        """Serializes and gzips the in-memory report to a file.
+
+        finalize() should be called first.
```
foolip: What happens otherwise? Can anything be asserted to make sure it has been called?

results-processor/wptreport.py
```diff
@@ -438,9 +448,12 @@ def prepare_labels(report, labels_str, uploader):
     # Empty labels may be generated here, but they will be removed later.
     for label in labels_str.split(','):
         labels.add(label.strip())
-    # Set the default channel to stable.
     if ('stable' not in labels) and ('experimental' not in labels):
-        labels.add('stable')
+        # Extract browser_channel into a label; default to "stable".
+        if report.run_info.get('browser_channel') in {'nightly', 'dev'}:
+            labels.add('experimental')
```
foolip: Can you split out this change? Letting the experimental label be implied seems risky if we're going to let it mean "dev with extra features enabled".

Hexcles:
@foolip PTAL again.

results-processor/wptreport.py
foolip: Thanks! I guess you usually merge for PRs where you've made the history nice?

WDYT about "Letting the experimental label be implied seems risky"?

Hexcles:
@foolip ah you're actually the only reviewer for this PR :) I sent the Go part to Luke (#536 ).

Regarding the risks of inferring labels, I think it's pretty low: we only infer it when the uploader doesn't set stable or experimental, and when `browser_channel` is present in the report. And `browser_channel` is pretty reliable.

results-processor/wptreport.py
Hexcles: Ahh those are beta results. Let me add the "beta" label here.

Hexcles: @foolip just updated the second commit in this PR to add the "beta" label.

foolip: Cool! Is there an issue for adding beta icons as well?

PR #531, opened 2018-09-11
Add TestRunIDs.LoadTestRuns test

lukebjerring:
## Description
Adds a test for the `TestRunIDs.LoadTestRuns` helper method.


shared/models_test.go
```diff
@@ -0,0 +1,46 @@
+// +build medium
+
+package shared_test
```
mdittmer: Vain curiosity: Is there are reason this is using `shared_test` (i.e., black-box testing)? Most of our code just uses `foo` instead of `foo_test`.

shared/models_test.go
```diff
@@ -0,0 +1,46 @@
...
+	for range testRuns {
+		keys = append(keys, datastore.NewIncompleteKey(ctx, "TestRun", nil))
+	}
+	keys, err = datastore.PutMulti(ctx, keys, testRuns)
+	assert.Nil(t, err)
+	for i, key := range keys {
+		testRuns[i].ID = key.IntID()
+	}
+
+	assert.Nil(t, err)
```
mdittmer: nit: Duplicate code?

shared/models_test.go
lukebjerring: There was a circular dep; perhaps I should change back and solve it properly.

shared/models_test.go
lukebjerring: Removed.

shared/models_test.go
lukebjerring: Ok so to document the offline discussion:
As you mention, `shared_test` is a special-case is Golang that lets you test a package as though you're external to it.
We _were_ depending on `shared` from `shared/sharedtest`, which was circular if running tests from `shared`. I've removed that somewhat unnecessary dependency.

PR #532, opened 2018-09-11
Rename complete to aligned

lukebjerring:
## Description
Treats `complete` as a legacy fallback, and uses `aligned` instead.

`aligned` is much more descriptive of the effect of the param; for the requested products, return (the first / most recent) runs which share the same SHA.

`complete` was chosen under the assumption that all four browsers had run for the SHA, and does not take into account a custom set of products (added in #491)

### DIFF
--- a/docs/ui.md
+++ b/docs/ui.md
@@ -4,7 +4,7 @@ The UI consists of base HTML templates served by the Go App Engine app in `webap
 
 ## UI Principles
 
-- The dashboard should not surface any overall metrics that compare complete runs of different browsers against each other.
+- The dashboard should not surface any overall metrics that compare aligned runs of different browsers against each other.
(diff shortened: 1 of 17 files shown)


docs/ui.md
```diff
@@ -4,7 +4,7 @@ The UI consists of base HTML templates served by the Go App Engine app in `webap
 
 ## UI Principles
 
-- The dashboard should not surface any overall metrics that compare complete runs of different browsers against each other.
+- The dashboard should not surface any overall metrics that compare aligned runs of different browsers against each other.
```
foolip: This instance of "complete" was actually not about the same sha alignment concept, but rather about summarizing a full run into a single metric to compare browsers by.

PR #533, opened 2018-09-11
Implement LoggerMux

mdittmer:
This will enable use cases such as, in metrics computation:

- Create `logrus` logger -> stdout with log level `Warning`
- Create `logrus` logger -> file with log level `Debug`
- Compose two loggers in `context.Context` key for logging during metrics computation

### DIFF
--- a/shared/util.go
+++ b/shared/util.go
@@ -130,7 +130,7 @@ var (
 // NewLoggerMux creates a multiplexing Logger that writes all log operations to
 // all delegates.
 func NewLoggerMux(delegates []Logger) Logger {
-	if delegates == nil || len(delegates) == 0 {
+	if len(delegates) == 0 {
 		return NewNilLogger()
 	}
 	return loggerMux{delegates}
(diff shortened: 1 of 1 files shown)


mdittmer:
Could do. Maybe extensible is better, or maybe [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)? I can see it both ways, but so far we only have a need for forward-to-two-loggers.

shared/util.go
```diff
@@ -94,6 +127,15 @@ var (
 	lck = LoggerCtxKey{}
 )
 
+// NewLoggerMux creates a multiplexing Logger that writes all log operations to
+// all delegates.
+func NewLoggerMux(delegates []Logger) Logger {
+	if delegates == nil || len(delegates) == 0 {
```
Hexcles: Nit: `len(delegates) == 0` would suffice, since `delegates` has a concrete slice type (you can't do `len(nil)`, but you can `len([]int(nil))`)

mdittmer: Done.

PR #536, opened 2018-09-11
Implement a GitHub webhook for Taskcluster

Hexcles:
The webhook listens to GitHub "status" events (i.e. all status updates
for all commits from all CI integration), but only reacts to the
"success" event from Taskcluster on master. When it receives such
events, it will query Taskcluster API for metadata, from which it
extracts task info, constructs a list of artifact URLs
(wptreport.json.gz), and sends the URLs to the results receiver.

## Review Information

In addition to unit tests, the code has already been deployed in my personal GCP and correctly sending results to the staging receiver: https://staging.wpt.fyi/test-runs?labels=taskcluster

### DIFF
--- a/api/webhook/taskcluster.go
+++ b/api/webhook/taskcluster.go
@@ -6,9 +6,11 @@ package webhook
 
 import (
 	"context"
+	"crypto/hmac"
+	"crypto/sha1"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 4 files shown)


Hexcles:
@lukebjerring HMAC has been added and tested in my staging deployment. PTAL.

api/webhook/taskcluster.go
```diff
@@ -0,0 +1,282 @@
...
+func getAuth(ctx context.Context) (username string, password string, err error) {
+	var u shared.Uploader
+	key := datastore.NewKey(ctx, "Uploader", "taskcluster", 0, nil)
+	err = datastore.Get(ctx, key, &u)
+	return u.Username, u.Password, err
+}
+
+func getSecret(ctx context.Context) (token string, err error) {
+	var t shared.Token
+	key := datastore.NewKey(ctx, "Token", "github-webhook-secret", 0, nil)
```
lukebjerring: Is this a unique enough secret name?

PR #540, opened 2018-09-12
Pull CachedStore out to shared/

mdittmer:
This is in anticipation of putting `TestRun` entities into memcache and fetching them from there when the IDs of the runs in question are known.


shared/iotest/io.go
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package iotest
```
Hexcles: I'd suggest putting this into `shared/sharedtest`. There are two reasons: 1. `iotest` collides with `testing/iotest`, which is a commonly used package in unit tests; 2. this file also similarly provides shared utility functions for unit tests as `sharedtest/util.go`.

Hexcles:
LGTM again.

Note: I think @lukebjerring made a similar change to resolve the circular dependency in another PR. I'll let you two fight for solving merge conflicts :)

PR #541, opened 2018-09-12
Show the build source in /test-runs

lukebjerring:
## Description
Shows an icon of the run's source (if labelled) against the icon shown for the run.
Should help distinguish `buildbot` and `taskcluster` runs (in case useful).

See https://icons-dot-wptdashboard-staging.appspot.com/test-runs

cc @jugglinmike


lukebjerring:
@thejohnjansen might have an opinion on that. The windows icon would be cool IMHO.

PR #542, opened 2018-09-12
Correct description of results format

jugglinmike:
The `time_start` and `time_end` properties are cited as being inserted
by the WPT CLI. That tool does indeed define those properties [1], but
the values are in units of milliseconds, not microseconds. This is
documented by the MozLog project [2] which the WPT CLI uses to provide
structured logging:

> time - The timestamp of the message in ms since the epoch (int).

[1] https://github.com/web-platform-tests/wpt/blob/632a3f59238036b6e24b28d47218ba9986ff4c62/tools/wptrunner/wptrunner/formatters.py#L13-L25
[2] https://firefox-source-docs.mozilla.org/mozbase/mozlog.html#data-format

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -342,8 +342,8 @@ The JSON file roughly looks like this:
 ```json
 {
   "results": [...],
-  "time_start": MICROSECONDS_SINCE_EPOCH,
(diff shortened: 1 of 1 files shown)


jugglinmike:
Certainly!

PR #544, opened 2018-09-13
Add a loading spinner + failure message for interop

lukebjerring:
## Description
Re-uses the `loadingCount` and `isLoading` properties in a `loading-state` behaviour component.

## Review Information
- Visit the auto-deployed env's interop tab. See a spinner.
- add `?label=foo`. See a failure message.

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -7,7 +7,10 @@ found in the LICENSE file.
 <link rel="import" href="../bower_components/polymer/polymer-element.html">
 <link rel="import" href="../bower_components/polymer/lib/elements/dom-if.html">
 <link rel="import" href="../bower_components/polymer/lib/elements/dom-repeat.html">
+<link rel="import" href="../bower_components/paper-spinner/paper-spinner-lite.html">
 <link rel="import" href="../bower_components/paper-styles/color.html">
+<link rel="import" href="info-banner.html">
+<link rel="import" href="loading-state.html">
 <link rel="import" href="path-part.html">
 <link rel="import" href="test-runs.html">
 <link rel="import" href="test-file-results.html">
@@ -33,10 +36,15 @@ found in the LICENSE file.
       border-bottom: solid 1px #ccc;
       padding-bottom: 1em;
       margin-bottom: 1em;
+      position: relative;
     }
-
     section.search .path {
       margin-top: 1em;
+    }
+    section.search paper-spinner-lite {
+      position: absolute;
+      top: 0;
+      right: 0;
     }
 
     /* Direct access to test-search from local shadowRoot prevents using
@@ -115,6 +123,8 @@ found in the LICENSE file.
       </template>
     </div>
 
+    <paper-spinner-lite active="[[isLoading]]" class="blue"></paper-spinner-lite>
+
     <test-search class$="search-[[pathIsATestFile]]" query="[[search]]">
     </test-search>
 
@@ -130,6 +140,12 @@ found in the LICENSE file.
       </div>
     </template>
   </section>
+
+  <template is="dom-if" if="[[interopLoadFailed]]">
+    <info-banner type="error">
+      Failed to fetch interop data.
+    </info-banner>
+  </template>
 
   <template is="dom-if" if="[[!pathIsATestFile]]">
     <section class="runs">
@@ -198,8 +214,8 @@ found in the LICENSE file.
     [118, 255,   3], // --paper-light-green-a200
   ];
 
-  /* global TestRunsBase, SelfNavigation */
-  class WPTInterop extends SelfNavigation(TestRunsBase) {
+  /* global TestRunsBase, SelfNavigation, LoadingState */
+  class WPTInterop extends SelfNavigation(LoadingState(TestRunsBase)) {
     static get is() {
       return 'wpt-interop';
     }
@@ -235,12 +251,16 @@ found in the LICENSE file.
         },
         searchResults: Array,
         onSearchCommit: Function,
+        interopLoadFailed: Boolean,
       };
     }
 
     constructor() {
       super();
       this.onSearchCommit = this.handleSearchCommit.bind(this);
+      this.onLoadingComplete = () => {
+        this.interopLoadFailed = !(this.passRateMetadata && this.allMetrics);
+      }
     }
 
     connectedCallback() {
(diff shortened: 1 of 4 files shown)


webapp/components/wpt-interop.html
```diff
@@ -269,8 +289,11 @@
 
     async ready() {
       await super.ready();
+      this.loadingCount++;
+      const that = this;
```
mdittmer: If you don't need the `arguments` object in `function(r) {...}` below, then prefer using `r => {...}` and dropping `that = this`.

webapp/components/wpt-interop.html
```diff
@@ -308,6 +331,10 @@
       return metadata && metadata.test_runs;
     }
 
+    computeInteropNotFound(isLoading, interop) {
+      return !isLoading && !interop;
```
mdittmer: nit: Don't rename `passRateMetadata` to `interop.

Won't this yield `interopNotFound = true` until `this.ready()` executes? Perhaps we should observe `isLoading` and check `this.passRateMetadata` exactly when `prevIsLoading = true && isLoading = false`.

webapp/components/wpt-interop.html
```diff
@@ -7,7 +7,10 @@
 <link rel="import" href="../bower_components/polymer/polymer-element.html">
```
mdittmer: Vain curiosity: Why bother with the rename?

webapp/components/wpt-interop.html
lukebjerring: Done.

webapp/components/wpt-interop.html
mdittmer: :+1:

webapp/components/wpt-interop.html
lukebjerring: Yeah, there was a small jank of 'not found' briefly. Done.

webapp/components/loading-state.html
```diff
@@ -21,13 +21,28 @@
           isLoading: {
             type: Boolean,
             computed: 'computeIsLoading(loadingCount)',
+            observer: 'isLoadingChanged',
           },
+          /**
+           * wasLoading is true when something was actually loading, which
+           * distinguishes !isLoading for the not-yet-started case.
+           */
+          wasLoading: {
```
mdittmer: I think this is more than we need (see comment below).

webapp/components/wpt-interop.html
```diff
@@ -330,8 +333,8 @@
       return metadata && metadata.test_runs;
     }
 
-    computeInteropNotFound(isLoading, interop) {
-      return !isLoading && !interop;
+    computeInteropNotFound(wasLoading, isLoading, passRateMetadata) {
```
mdittmer: Here and below. If we observe nothing but `isLoading`, then the observer function can see previous value and new value. Iff old value was `true` and new value is `false`, we can then manually check `this.passRateMetadata` this obviates the need for `wasLoading`. This also applies to `computeRunsNotFound` in `wpt-results.html`.

webapp/components/wpt-results.html
```diff
@@ -324,7 +324,7 @@ <h3>
           },
           runsNotFound: {
             type: Boolean,
-            computed: 'computeRunsNotFound(isLoading, testRuns)',
+            computed: 'computeRunsNotFound(wasLoading, isLoading, testRuns)',
```
mdittmer: Ditto.

webapp/components/interop.html
```diff
@@ -235,12 +251,16 @@
...
         onSearchCommit: Function,
+        interopLoadFailed: Boolean,
       };
     }
 
     constructor() {
       super();
       this.onSearchCommit = this.handleSearchCommit.bind(this);
+      this.onLoadingComplete = () => {
+        this.interopLoadFailed = !(this.passRateMetadata && this.allMetrics);
```
mdittmer: It's not immediately obvious why `&& this.allMetrics` is included. Maybe add a comment about this predicate?

webapp/components/wpt-results.html
```diff
@@ -321,17 +321,8 @@ <h3>
...
-          },
-          isLoading: {
-            type: Boolean,
-            computed: 'computeIsLoading(loadingCount)',
-          },
-          runsNotFound: {
-            type: Boolean,
-            computed: 'computeRunsNotFound(isLoading, testRuns)',
+          resultsLoadFailed: {
+            type:Boolean, value: false,
```
mdittmer: nits: Newline for each item; space after `type:`.

webapp/components/wpt-results.html
```diff
@@ -412,17 +403,14 @@ <h3>
...
-      computeRunsNotFound(isLoading, testRuns) {
-        return !isLoading && !(testRuns && testRuns.length);
-      }
-
       constructor() {
         super();
         this.onSearchCommit = this.handleSearchCommit.bind(this);
+        this.onLoadingComplete = () => {
+          this.resultsLoadFailed =
+            !(this.testRuns && this.testRuns.length
```
mdittmer: Is this susceptible to transient effects? E.g., might `testRuns` start out as `undefined` or `[]` before being populated, and thus `(this && part)` is false, making `!(...)` true, and causing a transient error message to be shown?

webapp/components/wpt-results.html
```diff
@@ -437,37 +425,45 @@ <h3>
...
       async fetchResults() {
         this.loadingCount++;
         try {
-          const response = await window.fetch(this.searchURL);
-          if (response.status !== 200) {
-            throw response;
-          }
-          const json = await response.json();
+          const json = await window.fetch(this.searchURL)
+            .then(async r => {
```
mdittmer: What change in behaviour are we trying to get by switching to `.then` inside an already async function? It seems like this change could just be a change in what gets thrown  when `response.status !== 200`.

webapp/components/interop.html
lukebjerring: The metadata is the runs and where to find the interop JSON blob, the allMetrics is the fetched JSON blob. If either fails, then the view is borked.

webapp/components/wpt-results.html
lukebjerring: no; `testRuns` only gets set as the parsed JSON for the response, once.

webapp/components/wpt-results.html
lukebjerring: Consistency with the exact same chunk of code in `interop`, which will make for a cleaner PR when the 2 get consolidated.

webapp/components/wpt-results.html
mdittmer: Right. I thought this was a persistent binding, but I see now that it's inside `onLoadingComplete`, triggered when data is loaded. Ignore me.

webapp/components/wpt-results.html
```diff
@@ -322,7 +322,8 @@ <h3>
             computed: 'computeAutocompleteURL(testRuns, search)',
           },
           resultsLoadFailed: {
-            type:Boolean, value: false,
+            type:Boolean,
```
mdittmer: nit: Still missing space after colon.

lukebjerring: sure; In future, just tweak the PR :)

PR #545, opened 2018-09-13
Fix test-runs WCT tests

lukebjerring:
## Description
Cleans up some redundant `Promise.resolve(true)` code, and removes the `ready()` test (we don't call ready anymore).


webapp/test/test-runs.html
```diff
@@ -35,15 +35,15 @@
...
+              resolve();
             }
           }, 0);
         });
       }
 
       setup(() => {
         sandbox = sinon.sandbox.create();
-        sandbox.stub(window, 'fetch', () => Promise.resolve(new Response(fetches[0])));
+        sandbox.stub(window, 'fetch', () => Promise.resolve(new Response(fetches)));
```
Hexcles: Why did you remove `[0]`?

lukebjerring: It's fetching `/api/runs` which returns an array, so the result should be the array. Not sure why it was a single item originally.

lukebjerring:
Will merge after #546 for simplicity.

PR #552, opened 2018-09-14
Fix the wpt-interop WCT tests

lukebjerring:
## Description
Clean up and re-enable the (very basic) tests for wpt-interop.
Also pulls the re-used helper methods into their own shared component.


webapp/components/interop.html
```diff
@@ -255,28 +255,24 @@
...
     }
 
-    connectedCallback() {
-      super.connectedCallback();
-      this.shadowRoot.querySelector('test-search')
-        .addEventListener('commit', this.onSearchCommit);
-    }
-
-    disconnectedCallback() {
+    detached() {
```
lukebjerring: Note that `this.shadowRoot` is only available once `super.ready()` has been called (and child elements have been created). The tests were failing preceding this change.

webapp/components/test/interop.html
```diff
@@ -86,116 +48,29 @@
         });
       });
 
-      // TODO(Hexcles): Re-enable this test.
-      suite.skip('static get properties()', () => {
-        test('passRateMetadata', () => {
```
lukebjerring: Note that I removed these tests because "test that I do have a property name foo with a type bar" is a meaningless test; it is, of course, going to be true until something changes (such as the type or existence), and you're getting no assurances out of it (except that Polymer is doing what it says it does).

webapp/components/test-search.html
```diff
@@ -81,27 +81,23 @@
         this.onBlur = this.handleBlur.bind(this);
       }
 
-      connectedCallback() {
-        super.connectedCallback();
+      ready() {
```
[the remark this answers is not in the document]
lukebjerring: The reason I moved this is because it doesn't work as-is. `shadowRoot` is null until `ready` is called.
This setup isn't correct anyway; the input element belongs to the custom element (with a 1:1 pairing), so we don't need to remove the event listener at all.

lukebjerring:
Fixed the `ready` vs `connectedCallback` issue by using HTML attribute bindings instead. PTAL.

lukebjerring:
Webdriver breakages are not related to this PR, so I'll force-merge.

PR #553, opened 2018-09-14
Ignore tests when deciding whether to deploy

lukebjerring:
## Description
Do an inverted `egrep` on `_test.go$` (and, webapp/components/test/ for webapp)

### DIFF
--- a/util/travis-deploy-staging.sh
+++ b/util/travis-deploy-staging.sh
@@ -25,8 +25,7 @@ if [[ "${APP_PATH}" == "webapp" ]]; then APP_DEPS="${APP_DEPS}|api|shared"; fi
 if [[ "${APP_PATH}" == "revisions/service" ]]; then APP_DEPS="${APP_DEPS}|revisions|shared"; fi
 APP_DEPS_REGEX="^(${APP_DEPS})/"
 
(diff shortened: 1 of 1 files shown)


util/travis-deploy-staging.sh
```diff
@@ -25,6 +25,9 @@ if [[ "${APP_PATH}" == "webapp" ]]; then APP_DEPS="${APP_DEPS}|api|shared"; fi
 if [[ "${APP_PATH}" == "revisions/service" ]]; then APP_DEPS="${APP_DEPS}|revisions|shared"; fi
 APP_DEPS_REGEX="^(${APP_DEPS})/"
 
+EXCLUSIONS="_test.go$$"
+if [[ "${APP_PATH}" == "webapp" ]]; then EXCLUSIONS="${EXCLUSIONS}|webapp/components/test/"
```
Hexcles: Do we need the `if`? Can we just always ignore these files?

lukebjerring: Fair point. Done.

Hexcles: BTW, you're also missing a `fi` here, but I guess you're removing the whole `if` anyway.

PR #555, opened 2018-09-14
Add (the beginnings of) a test run query builder component

lukebjerring:
## Description
Adds an (unused) `test-runs-query-builder` component, and some WCT tests.

This involved some refactoring of a `TestRunQuery` behaviour class, and a `DisplayLogo` component, both of which belong to the `test-runs` and `wpt-results` originally.

I recommend iterating through the commits, which I tried to separate out for easier review.

See http://localhost:8080/components/test/test-runs-query-builder.html for example :)


Hexcles:
Is there a demo that I can see? Or would you like me to just glance the code for obvious mistakes?

PR #556, opened 2018-09-14
Migrate web-platform-tests/results-analysis/metrics GetLogger() & fix announcer deployment

mdittmer:
This will allow us to drop https://github.com/web-platform-tests/results-analysis/blob/683c4b09af849e543a0473c4a0b7fe36fc8f1e33/metrics/logging.go and use the `shared` implementation instead. (Also, subsequent PRs will use `GetLogger()` in wpt.fyi.)

Also, copy local version of `wpt.fyi` code into announcer deployment Docker image before attempting to download its dependencies (and then build the announcer service).

### DIFF
--- a/revisions/service/Dockerfile
+++ b/revisions/service/Dockerfile
@@ -8,9 +8,8 @@ RUN apt-get install -qy --no-install-suggests git
 WORKDIR /go/src/app
 COPY *.go .
 
-RUN /usr/local/go/bin/go get -d .
-RUN rm -rf /root/go/src/github.com/web-platform-tests/wpt.fyi/*
 COPY wpt.fyi /root/go/src/github.com/web-platform-tests/wpt.fyi/
+RUN /usr/local/go/bin/go get -d .
 RUN /usr/local/go/bin/go build -o app .
 
 # Application image.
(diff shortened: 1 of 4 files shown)


mdittmer:
Blocked on landing https://github.com/sirupsen/logrus/pull/823 or something like it to fix the missing symbol in logrus AppEngine builds.

revisions/service/Dockerfile
```diff
@@ -8,9 +8,8 @@ RUN apt-get install -qy --no-install-suggests git
 WORKDIR /go/src/app
 COPY *.go .
 
-RUN /usr/local/go/bin/go get -d .
-RUN rm -rf /root/go/src/github.com/web-platform-tests/wpt.fyi/*
 COPY wpt.fyi /root/go/src/github.com/web-platform-tests/wpt.fyi/
+RUN /usr/local/go/bin/go get -d .
```
Hexcles: Could you briefly explain this change? Preferably in the PR description so that it'll be in the commit message.

shared/util.go
```diff
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	mapset "github.com/deckarep/golang-set"
+	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
```
Hexcles: While you're at it, could you change this to just `context`?

Hexcles:
LGTM and merged.

PR #557, opened 2018-09-14
Implement CachedStore strategy: memcache+datastore

mdittmer:
This is in anticipation of fetching `TestRun` entities using a memcache-fallback-to-datastore strategy.

### DIFF
--- a/shared/storage.go
+++ b/shared/storage.go
@@ -7,6 +7,7 @@ package shared
 import (
 	"bytes"
 	"compress/gzip"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -15,6 +16,7 @@ import (
 	"time"
 
 	"golang.org/x/net/context"
+	"google.golang.org/appengine/datastore"
 	"google.golang.org/appengine/memcache"
 	"google.golang.org/appengine/urlfetch"
 )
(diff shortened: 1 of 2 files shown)


shared/storage.go
```diff
@@ -284,3 +286,125 @@ func (cs byteCachedStore) Get(cacheID, storeID, iValue interface{}) error {
...
+
+func (oc jsonObjectCache) Get(id, value interface{}) error {
+	r, err := oc.delegate.NewReadCloser(id)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		err := r.Close()
+		if err != nil {
+			logger := oc.ctx.Value(DefaultLoggerCtxKey()).(Logger)
```
lukebjerring: This is another instance where we should use a fallback/helper (when getting the logger).

shared/storage.go
```diff
@@ -284,3 +286,125 @@ func (cs byteCachedStore) Get(cacheID, storeID, iValue interface{}) error {
...
+	return jsonObjectCache{ctx, delegate}
+}
+
+type datastoreObjectStore struct {
+	ctx  context.Context
+	kind string
+}
+
+func (s datastoreObjectStore) Get(iID, value interface{}) error {
+	id, ok := iID.(int64)
```
lukebjerring: So, this is fine, but I question the need to abstract to non-int64-IDs. Do we ever identify something by a different ID type?

mdittmer: Yes. Memcache uses strings.

shared/storage.go
mdittmer: Will do once #556 lands.

PR #558, opened 2018-09-17
Install go1.11 gofmt as part of go_deps

mdittmer:
Unfortunately, there is no go1.11 image under the `gcr.io/gcp-runtimes/go1-builder` docker dossier. All we really need for tooling consistency is go1.11 `gofmt`. This manually installs it from source to a location earlier in `PATH` than the default go toolchain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,17 +212,14 @@ node: curl gpg
 gofmt:
 	if [[ "$$(which gofmt)" != "$(GOPATH)/bin/gofmt" ]]; then \
 		TMP_DIR=$$(mktemp -d); \
-		pushd $$TMP_DIR > /dev/null 2>&1; \
-		git clone "https://github.com/golang/go.git"; \
+		cd $$TMP_DIR; \
+		git clone --depth 1 "https://github.com/golang/go.git" -b "release-branch.go1.11"; \
 		cd go; \
-		git checkout "origin/release-branch.go1.11"; \
 		mv "src/cmd" "$(GOPATH)/src/cmd"; \
-		popd > /dev/null 2>&1; \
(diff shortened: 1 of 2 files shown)


mdittmer:
*ping* @Hexcles

Makefile
```diff
@@ -209,6 +209,22 @@ node: curl gpg
...
 	fi
 
+gofmt:
+	if [[ "$$(which gofmt)" != "$(GOPATH)/bin/gofmt" ]]; then \
+		TMP_DIR=$$(mktemp -d); \
+		pushd $$TMP_DIR > /dev/null 2>&1; \
+		git clone "https://github.com/golang/go.git"; \
+		cd go; \
+		git checkout "origin/release-branch.go1.11"; \
+		mv "src/cmd" "$(GOPATH)/src/cmd"; \
```
Hexcles: Do we have to move `src/cmd` to `GOPATH`? Can we build it inside the temp directory?

Makefile
```diff
@@ -209,6 +209,22 @@ node: curl gpg
...
+		popd > /dev/null 2>&1; \
+		rm -rf "$$TMP_DIR"; \
+		pushd "$(GOPATH)/src" > /dev/null 2>&1; \
```
Hexcles: We can simply `cd` here. The paired `popd` happens at the very end, at which point we don't care about CWD any more (recipes run in a subshell so we don't need to restore CWD at the end of a recipe).

Hexcles:
Sorry, I think I lost a comment somewhere...

I was asking a high-level question: how much overhead (in terms of time) would this introduce? I'm a bit concerned as this checks out the whole Go repo.

Issue #561, opened 2018-09-17
Deploy 58dd5a0 to prod (2018-09-17)


Hexcles:
Changelist: https://github.com/web-platform-tests/wpt.fyi/compare/9923185...58dd5a0

Notable changes:

* @mdittmer 's improvements of the new query/autocomplete APIs:
    * Case-insensitive search (#509)
    * Bug fixes (#521) and performance optimiaztion (#528)
* Scale up to 2 results processors (#530) **needs manual deployment of `queue.yaml`**
* New epochs in revision announcer (#522 , already manually deployed to prod by @mdittmer )
* @lukebjerring 's various UI improvements/fixes: beta icons, runner icons (overlay), better error handling, limit the max number of test runs returned, etc.
* Overhaul of the `complete` parameter (now `aligned`), with many significant observable changes:
    * Default view is changed to aligned, stable runs.
    * The interop view should always work for stable/experimental runs (without specific SHA). When there is no interop result for the most recent runs, the view should fall back to an earlier one (#523 ).
* A webhook is added to listen to GitHub status updates to extract artifact URLs from Taskcluster runs (#536). **GitHub integration needs to be set up manually. A secure token also needs to be generated.**
* Results processor is now able to merge chunked reports and infer channel labels from the reports (#529).

Hexcles:
Webapp is up (currently serving 0 traffic): https://58dd5a0-dot-wptdashboard.appspot.com/

Anyone interested to run the checklist? (And I think part of the checklist needs to be updated.)

lukebjerring:
@Hexcles checklist was updated to reflect `aligned` and I strikethroughed one that should be omitted.
Looking good.

Hexcles:
New versions of processor & announcer are also up.

I'll turn up the traffic tomorrow.

Hexcles:
* Webapp running at 50%
* Processor & announcer running at 100%
* `queue.yaml` has been manually deployed (to run 2 tasks concurrently)

Hexcles:
No errors detected or reported by users. Turning the new version of webapp to 100%. Closing the issue.

PR #562, opened 2018-09-18
Check aligned in TestLabelParam webdriver test

lukebjerring:
## Description
Our integration tests should cover the full [launch checklist](https://github.com/web-platform-tests/wpt.fyi/issues/561#issuecomment-422177245).

This PR adds coverage for:
- `/results?label=stable&aligned`
- `/interop?label=stable&aligned`

Checks that the SHAs of the loaded test runs match.

## Review Information
    make go_firefox_test USE_FRAME_BUFFER=false FIREFOX_PATH=~/browsers/firefox/firefox FLAGS='-test.run=^TestLabel'

and 

    make go_firefox_test STAGING=true USE_FRAME_BUFFER=false FIREFOX_PATH=~/browsers/firefox/firefox FLAGS='-test.run=^TestLabel'

### DIFF
--- a/webdriver/install.sh
+++ b/webdriver/install.sh
@@ -44,21 +44,21 @@ then
 fi
 cd ${INSTALL_DIR}
 
-# Firefox 60
+# Firefox 62
 FIREFOX="firefox"
 case "${UNAME_OUT}" in
     Darwin*)
         FIREFOX_OS="mac"
-        FIREFOX_DMG="Firefox 60.0.dmg"
+        FIREFOX_DMG="Firefox 62.0.dmg"
         FIREFOX_SRC="${FIREFOX_DMG}"
         ;;
     Linux*|*)
         FIREFOX_OS="linux-x86_64"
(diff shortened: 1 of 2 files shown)


Hexcles:
Tests are failing:

e.g. Chrome: `panic: no such element: no such element: Unable to locate element: {"method":"css selector","selector":"wpt-results"}`

Firefox also fails a seemingly unrelated test: `panic: no such element: Unable to locate element: input.query`.

I think Firefox has started to (or will soon) ship native shadow DOM, so we'll need to drop the special case for Firefox when finding shadow elements, but I'm not sure if this is causing the failures.

lukebjerring:
At a guess, the `selenium` node package might have bumped their driver versions. FF62 seems to at least pass locally on my machine (both in docker + out of docker). Chrome passes fine outside docker, but not in docker. Guessing that's because my version of the driver on my box is older (ChromeDriver 2.40.565383).

lukebjerring:
FF seems happy - which means it's flaky. Which isn't newly caused by this PR. I'll merge as-is.

PR #564, opened 2018-09-18
Use memcache+datastore to fetch TestRun entities

mdittmer:
Depends on #557


mdittmer:
:trophy: unconditional LGTM :trophy:

First try

**First try!**

Issue #535, opened 2018-09-11
Add a graphical test run query builder


lukebjerring:
Sub-task of https://github.com/web-platform-tests/wpt.fyi/issues/379

Currently, we propagate and support a range of params that help query for specific results, but there is no UI around building the query.

lukebjerring:
Mock of what this might (eventually) look like:
https://docs.google.com/presentation/d/1f5MsezY4TRz4W-JTnqPuW99WFtMp24WNYaPZ6OSdmTM/edit#slide=id.g34863e07cf_0_58

lukebjerring:
This is now launched.

PR #571, opened 2018-09-20
Expand on the test-runs-query-builder component

lukebjerring:
## Description
See https://github.com/web-platform-tests/wpt.fyi/issues/535
(and, see #555 for the original component's definition.) 

This extends the builder's functionality to include adding + removing, and selecting a channel.

## Review Information
See http://localhost:8080/components/test/test-runs-query-builder.html for a playground.


webapp/components/product-info.html
```diff
@@ -0,0 +1,126 @@
+<dom-module id="browser-info">
```
lukebjerring: Browser info is just migrated code from `test-run` and `test-run-query` below, for re-use.

webapp/components/test-runs-query-builder.html
```diff
@@ -7,38 +7,297 @@
...
+      </template>
+    </template>
     <div>
-      <template is="dom-repeat" items="[[products]]" as="product">
-        <display-logo product="[[product]]"></display-logo>
-        [[product.browser_name]]
-        <template is="dom-if" if="[[product.browser_version]]">
-          [[product.browser_version]]
-        </template>
+      <template is="dom-repeat" items="[[products]]" as="p" index-as="i">
```
lukebjerring: Note that nested fields on an object, which is inside an array, aren't observable from the parent, which is why they fields are separated in `product-builder`.

webapp/.eslintrc.json
```diff
@@ -12,6 +12,7 @@
   "globals": {
     "assert": true,
     "fixture": true,
+    "flush": true,
```
Hexcles: Looks like we should also add `expect` here and remove `expect` from:
https://github.com/web-platform-tests/wpt.fyi/blob/06b80f2cf093ba41becf318bdd661890d8f82b88/webapp/components/test/test-runs-query.html#L13

Hexcles:
Looks like I have to checkout this branch locally to play with the demo, since we don't deploy `test`.

webapp/.eslintrc.json
lukebjerring: Done.

lukebjerring:
Yeah, unfortunately you'll need to run this locally for now, since the feature is not yet used.

webapp/components/test/test-runs-query-builder.html
```diff
@@ -23,77 +23,79 @@
   </test-fixture>
 
   <script>
-    // Set up the playground.
-    document.querySelector('test-runs-query-builder').addProduct();
-    document.querySelector('test-runs-query-builder').addProduct();
+    document.addEventListener('WebComponentsReady', () => {
+      // Set up the playground.
+      document.querySelector('test-runs-query-builder').addProduct();
+      document.querySelector('test-runs-query-builder').addProduct();
```
Hexcles: I understand these setups need to happen after the component is ready, but do the tests also need to? We don't wait for components ready in other tests, so it seems a bit inconsistent.

webapp/components/test-runs-query.html
```diff
@@ -12,17 +13,19 @@
      * Behaviour class for re-use of test-runs fetching behaviour.
      */
     // eslint-disable-next-line no-unused-vars
-    const TestRunsQuery = (superClass) => class extends QueryBuilder(superClass) {
+    const TestRunsQuery =
+      (superClass) => class extends QueryBuilder(ProductInfo(superClass)) {
       static get properties() {
         return {
           /* Parsed product objects, computed from the spec strings */
```
Hexcles: Is this comment still true? IIUC `products` is now the parameter set by parents while `productSpecs` is the computed value.

lukebjerring: Actually, both; if you set the product-specs value initially, it'll tee up the `products` array from those specs.

lukebjerring:
Will address both of those comments in #580, which introduces the refresh button.

webapp/components/test/test-runs-query-builder.html
lukebjerring: Eh, I just waited for it either way, but, reverted.

Issue #566, opened 2018-09-18
WebDriver tests become very flaky (again)


Hexcles:
Even when we exclude the unrelated transient network issues (during `npm install` and alike), there are still lots of flaky failures in our `go_{chrome,firefox}_test`.

Symptoms I've seen so far:

* Elements not found (page empty? `dev_appserver` dead or unable to start?)
* Timeout
* Expected X runs/files but only found 0

PR #573, opened 2018-09-20
Tweak Travis retry

Hexcles:
Retry more network-related operations (namely building the Docker image,
installing system dependencies and pulling Go packages). Stop retrying
the test itself.

This is yet another effort towards #566 .

## Review Information

Travis should pass.

## Changes

The core change is in `.travis.yml`. However, to make it possible to retry `util/docker-dev/run.sh` (which includes installing system dependencies), `util/docker-dev/run.sh` needs to be changed so that `-q` implies removing existing instances.

Also removed `-u $(id -u $USER):$(id -g $USER)` from `.travis.yml` as it's not necessary anymore (it's the default user).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,12 +40,8 @@ before_install: |
   # Decrypt client-secret.json for Appengine.
   openssl aes-256-cbc -K $encrypted_c8659b25fe66_key -iv $encrypted_c8659b25fe66_iv -in client-secret.json.enc -out client-secret.json -d
 
-  docker build -t "${DOCKER_IMAGE}" .
-  bash ./util/docker-dev/run.sh -d -q
-
-install:
-  - | # Set the user as home dir owner in the docker instance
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -65,5 +61,6 @@ script:
     fi
   - | # Run tests
     if [ "${MAKE_TEST_TARGET}" != "" ]; then
-      travis_retry docker exec -t -u $(id -u $USER):$(id -g $USER) "${DOCKER_INSTANCE}" make "${MAKE_TEST_TARGET}" ${MAKE_TEST_FLAGS}
+      travis_retry docker exec -t "${DOCKER_INSTANCE}" make go_packages;
```
lukebjerring: Do we use Golang for everything?

Hexcles: No, not really, but this is one area I know that tends to be flaky (because `go get` needs to clone some GitHub repos), and it's relatively small so we aren't wasting too much time on jobs that don't need `go_packages`.

Hexcles: On a second thought, I moved this line to `install` and added a comment.

Issue #440, opened 2018-08-15
Discussion: put breaking UI changes and new UI features behind flags by default


Hexcles:
Today during my discussion with @foolip , I pointed out that the recent deployment (#389) was delayed by more than two weeks partly because of some usability issues in the new UI features or some other breaking UI changes.

The release was much needed to fix some important bugs reported by users, show the new logo for Edge Insider, and make some new parameters that had been advertised in various discussions & presentations finally available to users. And when we realized the code wasn't ready for release, we kinda had to put off landing new features and focus on fixing / working around bugs to make the release possible.

It's important to be able to release quickly when needed (avoiding the feature-fix-release cycle), while it's equally important to maintain developer velocity. Given the size of our team, we probably don't want to maintain different branches for bug fixes and new features (or release channels). The other option that I can think of is to put all breaking UI changes and new UI features behind flags *by default*. And by flags, I mean anything including URL params (or hash) and buttons/toggles. These ad-hoc "flags" should be good for now, and the actual implementation will be case-by-case, decided by authors and reviewers. We might consider some more elaborate flag system in the future.

@mdittmer @lukebjerring please feel free to give your opinions.

foolip:
A cookie and `/flags` SGTM!

lukebjerring:
`/flags` and `admin/flags` are now tightly integrated into the common workflow.

PR #579, opened 2018-09-21
Add component for configuring features

lukebjerring:
## Description
As discussed in https://github.com/web-platform-tests/wpt.fyi/issues/440, we should be putting new features behind flags by default.

This PR adds an `wpt-flags` component which uses `window.localStorage` to store feature state.
Also configurable with `wpt-flags-editor` component, which uses checkboxes.

## Review Information
Visit [/flags](https://flags-dot-wptdashboard-staging.appspot.com/flags) (might need a query parameter if your browser's stored a redirect from `/flags` -> `/results/flags`.

### DIFF
--- a/webapp/components/wpt-flags.html
+++ b/webapp/components/wpt-flags.html
@@ -12,51 +12,30 @@ found in the LICENSE file.
 `<wpt-flags>` is a component for checking wpt.fyi feature flags.
 -->
 <dom-module id="wpt-flags">
-  <template>
-    <style>
-    </style>
-
-  </template>
-
   <script>
     const WPT_FYI_FEATURES = ['queryBuilder'];
 
-    class WPTFlags extends window.Polymer.Element {
+    const _wptFlags = (superClass, opt_writable) => class extends superClass {
       static get is() {
         return 'wpt-flags';
       }
 
       static get properties() {
-        const props = {
-        };
-
+        const props = {};
         for (const feature of WPT_FYI_FEATURES) {
           const stored = localStorage.getItem(`features.${feature}`);
           const value = stored && JSON.parse(stored);
           props[feature] = {
             type: Boolean,
+            readOnly: !opt_writable,
             notify: true,
             value: value,
           };
         }
         return props;
       }
-
-      ready() {
-        super.ready();
-        for (const feature of WPT_FYI_FEATURES) {
-          this._createMethodObserver(`valueChanged(${feature}, '${feature}')`);
-        }
-      }
-
-      valueChanged(value, feature) {
-        return localStorage.setItem(
-          `features.${feature}`,
-          JSON.stringify(value));
-      }
     }
-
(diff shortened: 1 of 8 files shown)


webapp/flags_handler.go
```diff
@@ -0,0 +1,15 @@
+// Copyright 2017 The WPT Dashboard Project. All rights reserved.
```
Hexcles: Nit: 2018

lukebjerring: Just edit the PR in cases like this :)

lukebjerring:
I was seeing that behaviour (redirecting) locally.

We'd simply need a `<wpt-flags></wpt-flags>` component somewhere, e.g.
https://github.com/web-platform-tests/wpt.fyi/pull/580/files#diff-ff3585b170ed88df8ef0c2fa8bcc61ecR476

Hexcles:
Yeah that generally looks fine. The only minor concern I have is about race conditions. It'd be good to clearly separate "editor" and read-only "reader".

lukebjerring:
Done; I've separated the readonly into a mixin, and the writable into an actual component (since the readonly behaviour is intended as just checking the flags). PTAL.

webapp/components/wpt-flags.html
```diff
@@ -0,0 +1,71 @@
...
+<link rel="import" href="../bower_components/paper-checkbox/paper-checkbox.html">
+
+<!--
+`<wpt-flags>` is a component for checking wpt.fyi feature flags.
+-->
+<dom-module id="wpt-flags">
+  <script>
+    const WPT_FYI_FEATURES = ['queryBuilder'];
+
+    const _wptFlags = (superClass, opt_writable) => class extends superClass {
```
Hexcles: The name `opt_writable` is a bit unclear to me. What is `opt`? Optional?

Also, what about using `readOnly` directly as the argument to avoid the extra negation?

webapp/components/wpt-flags.html
```diff
@@ -0,0 +1,71 @@
...
+  </template>
+
+  <script>
+    class WPTFlagsEditor extends _wptFlags(window.Polymer.Element, /*writable*/ true) {
+      static get is() { return 'wpt-flags-editor'; }
+
+      ready() {
+        super.ready();
+        for (const feature of WPT_FYI_FEATURES) {
+          this._createMethodObserver(`valueChanged(${feature}, '${feature}')`);
```
Hexcles: The leading underscore made me a bit worried at first, but looks like it's documented: https://www.polymer-project.org/2.0/docs/devguide/observers

lukebjerring: I had the same reaction.

webapp/components/wpt-flags.html
lukebjerring: Done.

lukebjerring:
We'll use this to hide the component I'm building for https://github.com/web-platform-tests/wpt.fyi/issues/535

Issue #408, opened 2018-08-08
Preserve order of products as given in URL


foolip:
These two views both show experimental first and then stable:
https://staging.wpt.fyi/results/?product=chrome[experimental]&product=chrome[stable]&diff
https://staging.wpt.fyi/results/?product=chrome[stable]&product=chrome[experimental]&diff

When comparing two runs, I'll normally want to list the "baseline" first, so that red means possible regressions. Changing the default order would fix it in this case, but just preserving the order given in the URL would be better.

Somewhat related to discussion with @mdittmer and @Hexcles yesterday about /api/search behavior and order.

foolip:
@jugglinmike mentioned that this would be useful for comparing Safari as well. In particular, this URL:
https://wpt.fyi/results/?product=safari[stable]@0398567bac&product=safari[experimental]@0398567bac&diff

mdittmer:
I think there's an argument to be made to show runs in a predefined order
instead. That's how Confluence works: browsers are always in release date
order (then browser + os name/version).

Is there a compelling use case or UX reason that users will benefit from
being able to specify order? If so, IMHO the right thing to do is provide
UI controls for order.

jugglinmike:
A visual comparison of the results between browser releases can be very
useful in discussion, but the pre-defined order may not support the
point being made. In my case, I wanted to show how much better Safari
Technology Preview performs than the "stable" release of Safari.  The
link @foolip shared has that information, but because the diff is in
terms of Safari Technology Preview, the visualization reads more like
"how much worse Safari performs compared to Safari Technology Preview."

To be clear: this is an issue of connotation, not correctness, so the
feature should probably be prioritized below more pressing work.

lukebjerring:
A more blunt / concise version of Mike's comment:
When diffing, "before / after" order is a useful thing.

Hexcles:
Woohoo! This is already coming in handy when I compare Taskcluster vs BuildBot. Thanks, Luke!

foolip:
Thanks Luke indeed! My original links work now :)

PR #582, opened 2018-09-24
Preserve product order

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/408

Uses `[]string`, instead of `mapset.Set` which has [deliberately non-deterministic ordering (in Golang)](https://blog.golang.org/go-maps-in-action#TOC_7.)

Also fixes support for `before` and `after` params that are spec strings (not base64 TestRun entities).

## Review Information
- Visit /results?product=safari&product=chrome&diff
  - Observe safari on the left
- Visit /results?before=edge&after=chrome
  - Observe edge on the left

### DIFF
--- a/api/manifest.go
+++ b/api/manifest.go
@@ -15,7 +15,6 @@ import (
 	"regexp"
 	"time"
 
-	"github.com/deckarep/golang-set"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine/datastore"
 	"google.golang.org/appengine/memcache"
@@ -211,12 +210,12 @@ func getGitHubReleaseAssetForSHA(client gitHubClient, sha string) (fetchedSHA st
 
 // filterManifest filters items in the the given manifest JSON, omitting anything that isn't an
 // item which has a URL beginning with one of the given paths.
-func filterManifest(manifestBytes []byte, paths mapset.Set) (result []byte, err error) {
+func filterManifest(manifestBytes []byte, paths []string) (result []byte, err error) {
 	var manifest shared.Manifest
 	if err = json.Unmarshal(manifestBytes, &manifest); err != nil {
 		return nil, err
 	}
-	if manifest, err = manifest.FilterByPath(paths); err != nil {
+	if manifest, err = manifest.FilterByPath(paths...); err != nil {
 		return nil, err
 	}
 	return json.Marshal(manifest)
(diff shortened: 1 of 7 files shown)


api/manifest.go
```diff
@@ -211,12 +210,12 @@ func getGitHubReleaseAssetForSHA(client gitHubClient, sha string) (fetchedSHA st
 
 // filterManifest filters items in the the given manifest JSON, omitting anything that isn't an
 // item which has a URL beginning with one of the given paths.
-func filterManifest(manifestBytes []byte, paths mapset.Set) (result []byte, err error) {
+func filterManifest(manifestBytes []byte, paths []string) (result []byte, err error) {
```
Hexcles: Please make it clear in the description that you also preserved the input order (of paths) in the manifest API, not just the `product` param.

shared/models.go
```diff
@@ -162,18 +162,18 @@ type Manifest struct {
 }
 
 // FilterByPath filters all the manifest items by path.
-func (m Manifest) FilterByPath(paths mapset.Set) (result Manifest, err error) {
+func (m Manifest) FilterByPath(paths ...string) (result Manifest, err error) {
```
Hexcles: Is there a caller of this function that actually needs to pass a variable length argument list? It seems to me the caller unpacks a slice using trailing `...` and the function packs it up again using leading `...`.

shared/models.go
```diff
@@ -192,7 +192,7 @@ type ManifestItem map[string][][]*json.RawMessage
 
 // FilterByPath culls out entries in the ManifestItem that don't have any items with
 // a URL that starts with the given path.
-func (m ManifestItem) FilterByPath(paths mapset.Set) (item ManifestItem, err error) {
+func (m ManifestItem) FilterByPath(paths ...string) (item ManifestItem, err error) {
```
Hexcles: Ditto.

api/manifest.go
lukebjerring: I don't think order has any effect (it's just multiple things to filter by; any of which is acceptible).

shared/models.go
lukebjerring: No, but I'd doubt that this leads to any performance hit

Hexcles: I wasn't concerned about performance, but the ergonomics of the API.

PR #583, opened 2018-09-25
Test order in the product spec integration test

lukebjerring:
## Description
Changes the test to fetch 2 products, and ensures their order is correct.

### DIFF
--- a/webdriver/product_test.go
+++ b/webdriver/product_test.go
@@ -27,21 +27,19 @@ func TestProductParam(t *testing.T) {
 	defer service.Stop()
 	defer wd.Quit()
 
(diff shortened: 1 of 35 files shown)


webdriver/product_test.go
```diff
@@ -65,7 +63,7 @@ func testProduct(
 	if err != nil {
 		panic(fmt.Sprintf("Failed to get test runs: %s", err.Error()))
 	}
-	assert.Lenf(t, testRuns, 1, "Expected exactly 1 TestRun.")
+	assert.Lenf(t, testRuns, len(products), "Expected %v TestRun(s).", len(products))
```
Hexcles: Nit: perhaps move this into `assertProducts`.

Issue #563, opened 2018-09-18
Search query is dropped when switching tabs (interop/results)


lukebjerring:
Should be preserved because you'll be looking at roughly the same content.
In essence, `interop` is really just a different format that search results take.

PR #585, opened 2018-09-25
Propagate the query param when navigating

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/563

The query param is not part of the test-runs query, but a (separate) filtering query. Additionally, you can't actually override the `computed` field (on a property) value, so we can't branch what constitutes the query.

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -109,10 +109,10 @@ found in the LICENSE file.
     }
   </style>
 
-  <results-navigation tab="interop"
-                      path="[[encodedPath]]"
-                      query="[[query]]">
-  </results-navigation>
+  <results-tabs tab="interop"
+                path="[[encodedPath]]"
+                query="[[query]]">
+  </results-tabs>
 
   <section class="search">
     <div class="path">
@@ -125,7 +125,7 @@ found in the LICENSE file.
 
     <paper-spinner-lite active="[[isLoading]]" class="blue"></paper-spinner-lite>
 
-    <test-search class$="search-[[pathIsATestFile]]" query="[[search]]">
+    <test-search class$="search-[[pathIsATestFile]]" query="{{search}}">
     </test-search>
 
     <template is="dom-if" if="{{ pathIsATestFile }}">
@@ -214,11 +214,12 @@ found in the LICENSE file.
     [118, 255,   3], // --paper-light-green-a200
   ];
 
-  /* global TestRunsBase, SelfNavigation, LoadingState */
-  class WPTInterop extends SelfNavigation(LoadingState(TestRunsBase)) {
-    static get is() {
-      return 'wpt-interop';
-    }
+  /* global TestRunsBase, SelfNavigation, LoadingState, ResultsNavigation */
+  class WPTInterop extends ResultsNavigation(
+    SelfNavigation(LoadingState(TestRunsBase)),
+    'interopQueryParams(sha, labels, productSpecs, maxCount, search)') {
+    static get is() { return 'wpt-interop'; }
+
     static get properties() {
       return {
         passRateMetadata: Object,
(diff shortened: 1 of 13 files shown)


webapp/components/test-search.html
```diff
@@ -49,8 +49,6 @@
             type: String,
             value: 'Search test files, like \'cors/allow-headers.htm\', then press <Enter>',
           },
-          items: Array,
-          itemKey: String,
```
lukebjerring: These fields are unused, and mislead me in the binding setup.

webapp/components/wpt-results.html
```diff
@@ -378,28 +375,29 @@ <h3>
           .filter(name => name.startsWith(path));
       }
 
-      computeSearchURL(testRuns, search) {
-        let url = '/api/search?';
+      computeSearchURL(testRuns, query) {
+        let url = new URL('/api/search', window.location);
```
lukebjerring: This is a drive-by cleanup.

webapp/components/interop.html
```diff
@@ -214,11 +214,12 @@
...
-  /* global TestRunsBase, SelfNavigation, LoadingState */
-  class WPTInterop extends SelfNavigation(LoadingState(TestRunsBase)) {
-    static get is() {
-      return 'wpt-interop';
-    }
+  /* global TestRunsBase, SelfNavigation, LoadingState, ResultsNavigation */
+  class WPTInterop extends ResultsNavigation(
+    SelfNavigation(LoadingState(TestRunsBase)),
+    'interopQueryParams(sha, labels, productSpecs, maxCount, search)') {
+    static get is() { return 'wpt-interop'; }
```
mdittmer: Does our linter like this one-liner?

webapp/components/interop.html
```diff
@@ -214,11 +214,12 @@
...
 
-  /* global TestRunsBase, SelfNavigation, LoadingState */
-  class WPTInterop extends SelfNavigation(LoadingState(TestRunsBase)) {
-    static get is() {
-      return 'wpt-interop';
-    }
+  /* global TestRunsBase, SelfNavigation, LoadingState, ResultsNavigation */
+  class WPTInterop extends ResultsNavigation(
+    SelfNavigation(LoadingState(TestRunsBase)),
+    'interopQueryParams(sha, labels, productSpecs, maxCount, search)') {
```
mdittmer: This is starting to feel like C++ template coding horror :scream:

webapp/components/interop.html
```diff
@@ -262,6 +263,9 @@
         // passRateMetadata contains the url for the JSON blob of allMetrics;
         // both fetches need to succeed + parse.
         this.interopLoadFailed = !(this.passRateMetadata && this.allMetrics);
+        if (!this.interopLoadFailed && this.search) {
+          this.shadowRoot.querySelector('test-search').commitQuery();
```
mdittmer: This feels wrong. The desired effect is that _this_ component runs application logic associated with having both interop data and a query. 

I would prefer to write a helper for `handleCommitQuery(e)`, like `onQuery(q)`; `handleCommitQuery` can delegate to `onQuery` after extracting `q`, and this line of code becomes `this.onQuery(this.search)`.

webdriver/search_test.go
```diff
@@ -24,58 +32,62 @@ func TestSearch(t *testing.T) {
...
+	var pathParts []selenium.WebElement
+	var err error
 	filteredPathPartsCondition := func(wd selenium.WebDriver) (bool, error) {
-		pathParts, err := getPathPartElements(wd, elementName)
+		pathParts, err = getPathPartElements(wd, elementName)
 		if err != nil {
 			return false, err
 		}
-		return len(pathParts) > 0 && len(pathParts) < numInitialPathParts, nil
+		return len(pathParts) == 1, nil
```
mdittmer: Does this test use static input data? If not, the introduction of something in another folder named `2dcontext` could cause an unexpected breakage.

webapp/components/interop.html
lukebjerring: re-jigged `handleSearchCommit` to take just the query. PTAL.

webdriver/search_test.go
lukebjerring: It does locally, but does not in staging. Been this way a long time.. I'll leave it for you to have the chance to say "told you so" later ;)

PR #586, opened 2018-09-25
Request-level caching for query APIs

mdittmer:
## Description

This PR introduces request-level caching and integrates it into the search and autocomplete APIs.

## Review Information

- Load any "results" (not "runs" or "interop") page with the Network tab open in DevTools

  - Observe at most 1 `/api/search` and `/api/autocomplete` request

  - Reload

    - Observe `/api/search` and `/api/autocomplete` requests take 300-800ms

    - Observe two server-side logs: `Serving cached data from cache key: URL-/api/[...]`

  - Type and commit a search query; type and commit a different search query; type and commit first search query again

    - Observe second load of same query'ified `/api/search` and`/api/autocomplete` faster (200-500ms, depending on query)

    - Observe two server-side logs: `Serving cached data from cache key: URL-/api/[...]` with query

## Changes

- Ensure `wpt-results` component waits for run IDs to commit search+autocomplete queries

- Implement `cachingHandler`, a `http.Handler` implementation, and `cachingResponseWriter`, an `http.ResponseWriter` implementation

- Decorate search and autocomplete HTTP handlers with `cachingHandler` instances

  - Reuse memcache `ReadWritable` from entity-level caching for request-level caching

### DIFF
--- a/api/query/query.go
+++ b/api/query/query.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"sort"
 	"strconv"
 	"sync"
 	"time"
@@ -150,4 +151,28 @@ func (qh queryHandler) loadSummary(testRun shared.TestRun) ([]byte, error) {
 
 func getMemcacheKey(testRun shared.TestRun) string {
 	return "RESULTS_SUMMARY-" + strconv.FormatInt(testRun.ID, 10)
+}
+
+func isRequestCacheable(r *http.Request) bool {
+	ids, err := shared.ParseRunIDsParam(r)
+	return err == nil && len(ids) > 0
+}
+
+func getRequestCacheKey(r *http.Request) interface{} {
+	q := r.URL.Query()
+	qKeys := make([]string, 0, len(q))
+	for key := range q {
+		qKeys = append(qKeys, key)
+	}
+	sort.Strings(qKeys)
+
+	cacheKey := "URL-" + r.URL.Path + "?"
+	for _, key := range qKeys {
+		values := q[key]
+		for _, value := range values {
+			cacheKey += key + "=" + value + "&"
+		}
+	}
+
+	return cacheKey
 }
--- a/shared/request_caching.go
+++ b/shared/request_caching.go
@@ -1,0 +1,159 @@
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import (
+	"bytes"
+	"fmt"
+	io "io"
+	"io/ioutil"
+	"net/http"
+)
+
+// CachingResponseWriter is an http.ResponseWriter that can produce a new
+// io.Reader instances that can replay the response.
+type CachingResponseWriter interface {
+	http.ResponseWriter
+
+	WriteTo(io.Writer) (int64, error)
+}
+
+type cachingResponseWriter struct {
+	delegate   http.ResponseWriter
+	b          *bytes.Buffer
+	statusCode int
+}
+
+func (w *cachingResponseWriter) Header() http.Header {
+	return w.delegate.Header()
+}
+
+func (w *cachingResponseWriter) Write(data []byte) (int, error) {
+	if w.statusCode == 0 {
+		w.statusCode = http.StatusOK
+	}
+
+	var err error
+	i := 0
+	for err == nil && i < len(data) {
+		var n int
+		n, err = w.b.Write(data[i:])
+		i += n
+	}
+	if err != nil {
+		w.b.Reset()
+		return 0, err
+	}
+
+	i = 0
+	for err == nil && i < len(data) {
+		var n int
+		n, err = w.delegate.Write(data[i:])
+		i += n
+	}
+	if err != nil {
+		w.b.Reset()
+		return i, err
+	}
+
+	return i, nil
+}
+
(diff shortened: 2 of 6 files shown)


shared/request_caching.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+func (w *cachingResponseWriter) Write(data []byte) (int, error) {
+	if w.statusCode == 0 {
+		w.statusCode = http.StatusOK
+	}
+
+	var err error
+	i := 0
+	for err == nil && i < len(data) {
```
lukebjerring: nit: this `i` is never re-used, can collapse into the `for`

api/query/query.go
```diff
@@ -151,3 +152,27 @@ func (qh queryHandler) loadSummary(testRun shared.TestRun) ([]byte, error) {
 func getMemcacheKey(testRun shared.TestRun) string {
 	return "RESULTS_SUMMARY-" + strconv.FormatInt(testRun.ID, 10)
 }
+
+func isRequestCacheable(r *http.Request) bool {
+	ids, err := shared.ParseRunIDsParam(r)
+	return err == nil && len(ids) > 0
+}
+
+func getRequestCacheKey(r *http.Request) interface{} {
```
lukebjerring: As discussed offline, the burden of this code probably doesn't benefit us much; client-code controls the incoming URLs, and manually typed URLs are unlikely to hit cache anyway.

Let's just use the (full) URL as the cache key for the HTTP cache.

Hexcles:
I realized we were essentially implementing a transparent reverse HTTP proxy with cache. I was going to suggest setting ETags properly and using things like [Cloud CDN](https://cloud.google.com/cdn/). However, it seems that Cloud CDN doesn't have out-of-box integration with AppEngine.

If we were to use an external cache service, the basic structure of the code would still be the same. Instead of writing to memcache, we would compute some hashes and set ETags in the decorator (or explicitly disallow cache if the request is not cacheable).

shared/request_caching.go
```diff
@@ -0,0 +1,159 @@
...
+	getCacheKey func(*http.Request) interface{}
+}
+
+func (h cachingHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := NewAppEngineContext(r)
+	logger := GetLogger(ctx)
+
+	// Case 1: Not cacheable.
+	if !h.isCacheable(r) {
+		logger.Infof("Not cacheable: %s", r.URL.String())
```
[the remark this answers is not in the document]
mdittmer: With the current things that are cached, we actually hit this case relatively rarely. The client usually looks up the run IDs that will mark its request as cacheable.

api/query/query.go
```diff
@@ -151,3 +151,12 @@ func (qh queryHandler) loadSummary(testRun shared.TestRun) ([]byte, error) {
...
+	return r.URL.String()
```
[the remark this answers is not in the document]
mdittmer: The key size limit is 250 bytes. We could use `[pathname][query-params]` parts only. WDYT?

//CC @lukebjerring; we were discussing this yesterday.

shared/request_caching.go
```diff
@@ -0,0 +1,158 @@
...
+	return w.delegate.Header()
+}
+
+func (w *cachingResponseWriter) Write(data []byte) (int, error) {
+	if w.statusCode == 0 {
+		w.statusCode = http.StatusOK
+	}
+
+	var err error
+	for i := 0; err == nil && i < len(data); {
```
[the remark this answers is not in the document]
mdittmer: I'm happy to simplify the code, assuming all-or-nothing writes. Silly me thinking that `n, err` return value would contain two pieces of meaningful information :stuck_out_tongue_closed_eyes:

mdittmer:
Thanks for the suggestion of Cloud CDN (too bad it won't work here). I didn't know we have such a service that lets you use application logic to determine `isCacheable` on each request.

shared/request_caching.go
```diff
@@ -23,6 +23,7 @@ type CachingResponseWriter interface {
 type cachingResponseWriter struct {
 	delegate   http.ResponseWriter
 	b          *bytes.Buffer
+	err        error
```
Hexcles: It is unclear which writer the error comes from.

shared/request_caching.go
```diff
@@ -35,7 +36,9 @@ func (w *cachingResponseWriter) Write(data []byte) (int, error) {
 		w.statusCode = http.StatusOK
 	}
 
-	return w.b.Write(data)
+	_, w.err = w.b.Write(data)
+
+	return w.delegate.Write(data)
```
Hexcles: What about writing to `delegate` first, and early returning (without writing to buffer) if error occurs?

shared/request_caching.go
```diff
@@ -36,9 +36,14 @@ func (w *cachingResponseWriter) Write(data []byte) (int, error) {
 		w.statusCode = http.StatusOK
 	}
 
-	_, w.err = w.b.Write(data)
+	n, err := w.delegate.Write(data)
+	if err != nil {
+		return n, err
+	}
+
+	_, w.bufErr = w.b.Write(data)
```
Hexcles: If you simply `return w.b.Write(data)` here, you won't need `bufErr` at all, right?

Or do you think error writing to buffer needs to be handled differently (from error writing to response)?

mdittmer: We want the consistency from decorated Write() fails iff delegate write failed. Problems with the decorator are handled externally.

mdittmer:
@lukebjerring please revisit.

PR #590, opened 2018-09-26
Add aligned + labels to query builder

lukebjerring:
## Description
Further work on #535 

Adds a checkbox for aligning the runs, and a textbox for comma-separated labels.


lukebjerring:
Ping.

Hexcles:
The "Labels" input doesn't seem to work.

lukebjerring:
@foolip I merged #608 into this PR to fix the problems @Hexcles was seeing.

The issue with the labels on the products getting dropped was `Set` vs `Array` (why they had to have distinct names for size/length, has/contains, I'll never know).

webapp/components/test-runs-query.html
```diff
@@ -9,59 +9,10 @@
...
 <dom-module id="test-runs-query">
   <script>
-    const DEFAULT_PRODUCT_SPECS = [
-      'chrome',
-      'edge',
-      'firefox',
-      'safari',
-    ];
-    const DEFAULT_PRODUCTS = DEFAULT_PRODUCT_SPECS.map(p => parseProductSpec(p));
-
```
Hexcles: Glad to see getting rid of the duplicated code here.

webapp/components/product-info.html
```diff
@@ -64,44 +113,11 @@
...
-          labelsStr = labelsStr.substr(0, labelsStr.length - 1);
-          labelsStr.split(',').forEach(l => labels.add(l));
-          spec = spec.substr(0, arrayIndex);
-        }
-        // product
-        const product = this.parseProduct(spec);
-        product.revision = revision;
-        product.labels = Array.from(labels);
-        return product;
+        return window.parseProductSpec(spec);
```
Hexcles: I don't know much about the isolation of web components. Would these two methods leak into the global `window`?

Besides, why do we need to put them in the global namespace in the first place?

lukebjerring: I use them above to populate `DEFAULT_PRODUCTS`, and couldn't think how to refer to the function otherwise.

lukebjerring:
The test failures are some weird flakiness with `aetest` / Golang.

foolip:
I'll leave the code review to @Hexcles unless asked to look a specific thing. Feedback as a user:
* Channel "Dev" doesn't work, related to https://github.com/web-platform-tests/wpt.fyi/issues/611
* If that is fixed, something like Edge Dev still doesn't exist. It would be great to limit the exposed channels to ones that have at least 1 run for the selected browser.
* For labels, knowing more labels that exist would be helpful. It's a bit odd that stable is included in the example, will users not wonder whether to use labels or the channel for it?
* Dragging to reorder products would be great. I got them in the wrong order and had to start over.

One final thing is that resulting URLs are less nice than the hand-crafted onces because of percent encoding: https://query-builder-4-dot-wptdashboard-staging.appspot.com/results/?product=firefox%5Bbeta%5D&product=chrome&aligned=true

Is there a another pair of characters other than `[]` that could be used that won't be escaped by this or any software? Looks like `()` isn't escaped, but the I have noticed that URL detectors will tend to think that URLs end at `)` so that would also be annoying.

Hexcles:
> * Channel "Dev" doesn't work, related to #611

There's a misunderstanding here. It's not really related to this PR; "dev" doesn't work because we currently don't have "dev" channels (in our terminologies). See my comment in the issue.

> * Dragging to reorder products would be great. I got them in the wrong order and had to start over.

+1, but I think we can do it in another PR.

Hexcles:
Luke, there are also failures in web component tests that don't seem like infra issues.

webapp/components/test-runs-query.html
```diff
@@ -137,7 +91,7 @@
           if (maxCount) {
             params['max-count'] = maxCount;
           }
-          if (this.aligned) {
+          if (aligned) {
```
foolip: I have a small feature request, the change wouldn't be here, but I also don't know where. For params that have a boolean value, would it not be nicer to just append `&aligned` instead of `&aliged=true`? C.f. `&diff`. Came here because I just found myself writing a URL by hand and was not at all tempted to append `=true`.

lukebjerring: Noted. Will play around in a follow-up PR.

PR #591, opened 2018-09-26
Remove the workaround for golint and fix test tag lint

Hexcles:
This mainly fixes two things:

* `go get -u golang.org/x/lint/golint` now works, so we can remove the manual checkout.
* Fix the lint for test tags which has been broken for a while (again, because of multi-line).

Along with some drive-by cleanups.

## Testing

With the first commit, the missing tag in `api/webhook/taskcluster_test.go` was found (and fixed in the second commit).

### DIFF
--- a/api/webhook/taskcluster_test.go
+++ b/api/webhook/taskcluster_test.go
@@ -1,3 +1,5 @@
+// +build small
+
 // Copyright 2018 The WPT Dashboard Project. All rights reserved.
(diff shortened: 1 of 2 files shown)


api/webhook/taskcluster_test.go
```diff
@@ -1,3 +1,5 @@
+// +build small
```
lukebjerring: unrelated?

Issue #449, opened 2018-08-17
Distinguish harness status from subtests in test results view


foolip:
https://wpt.fyi/results/quirks/classname-query-after-sibling-adoption.html?sha=c0a1bc68d7 looks like this:
![image](https://user-images.githubusercontent.com/498917/44257138-681ef780-a20b-11e8-98df-d3036ab8a143.png)

https://wpt.fyi/results/visual-viewport/viewport-unscaled-scale-iframe.html?sha=c0a1bc68d7, with harness errors:
![image](https://user-images.githubusercontent.com/498917/44256623-ff834b00-a209-11e8-95cc-e27d3fadd000.png)

(The "File status" misnomer is https://github.com/web-platform-tests/wpt.fyi/issues/419 + https://github.com/web-platform-tests/wpt.fyi/pull/448, should be "Harness status" here.)

Harness statuses aren't passes and failures, but rather a failure to correctly run the test. While harness errors aren't always a serious problem, I believe we should treat them as a separate problem and drive them down close to zero across wpt: https://github.com/web-platform-tests/wpt/issues/10877.

Harness statuses also shouldn't be part of summary numbers, see https://github.com/web-platform-tests/wpt.fyi/issues/62.

Taking this together, I believe we should present them as distinct from subtests. I suggest gray for "OK", and purple (as in https://github.com/web-platform-tests/wpt.fyi/issues/421#issuecomment-412017365) for anything else. Mockups of the above examples:
![image](https://user-images.githubusercontent.com/498917/44257718-2c852d00-a20d-11e8-98c4-9c7c7fb446c4.png)
![image](https://user-images.githubusercontent.com/498917/44257653-f9429e00-a20c-11e8-926d-d70ba24f20cb.png)

foolip:
Oh, but https://github.com/web-platform-tests/wpt.fyi/issues/449 might not be the issue that fixed the color the first time, I see.

PR #595, opened 2018-09-27
Separate the first row in test-file-results display

lukebjerring:
## Description
More subtle fix for https://github.com/web-platform-tests/wpt.fyi/issues/449

### DIFF
--- a/webapp/components/abstract-test-file-results-table.html
+++ b/webapp/components/abstract-test-file-results-table.html
@@ -33,7 +33,7 @@ found in the LICENSE file.
         background-color: rgb(242, 90, 90);
       }
       tbody tr:first-child {
-        border: 8px 0 solid white;
+        border-bottom: 8px solid white;
       }
     </style>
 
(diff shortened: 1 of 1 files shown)


foolip:
Looking at https://rowsplit-dot-wptdashboard-staging.appspot.com/results/ambient-light/AmbientLightSensor-disabled-by-feature-policy.https.html?label=stable&aligned=true, I can't tell the difference. Am I looking at the wrong thing?

lukebjerring:
Sorry, child PR didn't seem to work so well. Reset now that #594 is landed. PTAL.

foolip:
Still can't see a difference, tried https://rowsplit-dot-wptdashboard-staging.appspot.com/results/dom/historical.html?label=stable&aligned=true this time. Do you have a URL that works for you?

lukebjerring:
There's an extra 8px of whitespace under the headers.

foolip:
Can you share a screenshot of what you see? Seems like we must be getting different content server or something?

lukebjerring:
So.. it's not different because the change is already present somehow (bad merge?) My bad.

lukebjerring:
.. I will tweak this here.

Issue #328, opened 2018-06-28
Strange error message when requesting diff view


kereliuk:
**How to reproduce**
Visit [https://staging.wpt.fyi/results/?before=chrome@044cb1b9a7&after=edge](https://staging.wpt.fyi/results/?before=chrome@044cb1b9a7&after=edge).
I see `invalid character 'y' looking for beginning of value` with an HTTP status code of 400.

There probably isn't a full run for edge in this case (since there is for chrome I know), but we should handle the error more descriptively.

Hexcles:
I'm not exactly sure why this is happening. The code is supposed to find the last run for the browser if no revision is specified, so it shouldn't matter if Edge doesn't have a run at 044cb1b9a7. The problem seems to only happen with Edge. I tried Safari and some other combinations of `before` & `after` and couldn't reproduce the issue.

lukebjerring:
weirdly, it words fine with `edge@latest`:
https://staging.wpt.fyi/results/?before=edge@latest&after=chrome

Will dig into this (and at least verify that `edge` is a valid product spec...)

lukebjerring:
lol. so... `edge` actually successfully decodes to `y�`

https://play.golang.org/p/tFZjxM7E8eS

PR #599, opened 2018-09-27
Fall back to base64 instead of opening with it for before/after

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/328

### DIFF
--- a/webapp/test_results_handler.go
+++ b/webapp/test_results_handler.go
@@ -131,36 +131,33 @@ func parseTestRunUIFilter(r *http.Request) (filter testRunUIFilter, err error) {
 
 		const singleRunURL = `/api/run?sha=%s&product=%s`
 		var specs []string
-		if beforeDecoded, err := base64.URLEncoding.DecodeString(before); err == nil {
-			var run shared.TestRun
-			if err = json.Unmarshal([]byte(beforeDecoded), &run); err != nil {
-				return filter, err
+		beforeSpec, err := shared.ParseProductSpec(before)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


webapp/test_results_handler.go
```diff
@@ -131,36 +131,33 @@ func parseTestRunUIFilter(r *http.Request) (filter testRunUIFilter, err error) {
 
 		const singleRunURL = `/api/run?sha=%s&product=%s`
 		var specs []string
-		if beforeDecoded, err := base64.URLEncoding.DecodeString(before); err == nil {
-			var run shared.TestRun
-			if err = json.Unmarshal([]byte(beforeDecoded), &run); err != nil {
-				return filter, err
+		beforeSpec, err := shared.ParseProductSpec(before)
```
mdittmer: Why do we treat `before`/`after` two different ways? IIUC, we either:

1. Find a `before`/`after` spec in plaintext and add it to `specs`, OR
1. Find a `before`/`after` spec base64-encoded and copy it to `filter.{Before|After}TestRun`.

In non-error cases we do one of these things (write to `specs` OR write to `filter`), but not both. Why the two different ways of interpreting the same input?

lukebjerring: Specs: `before=chrome@latest`
Base64: `before=[encodedblobofchars]` where decoding the blob is a JSON object `{browser_name: "foo", ... }`

It's an old use-case for passing in arbitrary runs that aren't in the datastore, while still being able to see them in the UI. This PR doesn't change behaviour, just the order in which we try base64'ing

Issue #225, opened 2018-06-05
Show something when searching doesn't match anything


foolip:
I opened https://wpt.fyi/results/ and started typing "ishistory" in the search box before it was loaded. I was left with a just the header, not sure if it was because the results didn't load correctly, or if it was because the search didn't match anything. "No results" or something similar when the search doesn't have any matches would be great.

lukebjerring:
So, we actually do show something now.. but it's 

> Failed to fetch test runs.

If the request succeeds, and `testRuns` are found, it would be good to show a `No results found` message instead.

PR #602, opened 2018-09-28
Distinguish 'no results' from failed fetch.

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/225

## Review Information
- Visit /results?q=this-wont-match-anything
  - Observe 'No results'
- Visit /results?label=foo
  - Observe 'Failed to fetch results'

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -25,7 +25,9 @@
 
       ready() {
         super.ready();
-        this.path = this.urlToPath(window.location);
+        if (!this.path) {
+          this.path = this.urlToPath(window.location);
+        }
         window.onpopstate = () => {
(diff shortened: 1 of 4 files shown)


foolip:
Confirmed to work:
https://no-results-dot-wptdashboard-staging.appspot.com/results/?q=this-wont-match-anything
https://no-results-dot-wptdashboard-staging.appspot.com/results/?label=foo

lukebjerring:
Hmmm.. Not convinced a test would have much value here; we'd essentially be testing whether `dom-if` is doing its job. Will add one anyway.

foolip:
Don't add tests if they're not valuable :)

Issue #282, opened 2018-06-18
Expose per-test execution time


foolip:
As https://github.com/web-platform-tests/wpt/issues/10026 the wptreport JSON has the execution time at a per-test level. Exposing this in the UI would be great, and even greater would be the ability to search for slow tests.

foolip:
@lukebjerring did you close this accidentally? Test-level execution time is not yet exposed in the UI.

lukebjerring:
Yes, I misinterpreted this as overall execution time. My mistake.

foolip:
This would have helped in https://github.com/web-platform-tests/wpt/pull/19527 where I suspect the test is slow, but determining this required downloading the full report and reading the JSON manually.

Issue #204, opened 2018-05-30
Differing dates of runs become confusing when showing complete runs


foolip:
After https://github.com/web-platform-tests/wpt.fyi/pull/72, https://staging.wpt.fyi/results/ shows the same commit for each run, but different dates.

Using the commit date of the commit shown would probably be most useful at this point. Being able to hover it and see the exact time would also be neat.

lukebjerring:
TBH I disagree about it being confusing. When the SHA landed isn't as useful as when the run occurred.

foolip:
Alright, I would not be offended if the issue is closed as "won't fix in the way you want" :)

lukebjerring:
Closing as won't fix / WAI.

PR #603, opened 2018-09-28
Switch date of runs to time_start and show execution time

lukebjerring:
## Description
Improves https://github.com/web-platform-tests/wpt.fyi/issues/204 and adds run execution-time data for a misinterpretation of https://github.com/web-platform-tests/wpt.fyi/issues/282

Changes the hover content to show how long a run took (and what time it started). This disambiguates the date shown.

### DIFF
--- a/webapp/components/test-run.html
+++ b/webapp/components/test-run.html
@@ -62,7 +62,7 @@ See models.go for more details.
         <paper-tooltip offset=0>
           {{displayName(testRun.browser_name)}} {{testRun.browser_version}}<br>
           Labels: {{displayLabels(testRun.labels)}}<br>
-          Started {{timeFormat(testRun.time_start)}} (took {{timeTaken(testRun)}})<br>
+          Started {{timeFormat(testRun.time_start)}} {{timeTaken(testRun)}}<br>
           {{moreTooltip(testRun)}}
         </paper-tooltip>
       </template>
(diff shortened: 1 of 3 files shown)


foolip:
Cool! The staging instance still shows different dates, but it's probably less likely to happen this way.

foolip:
In other words, I don't think this should close https://github.com/web-platform-tests/wpt.fyi/issues/204, but perhaps reduce the priority.

webapp/components/test-run.html
```diff
@@ -92,7 +93,29 @@
...
+          hour: 'numeric',
+          minute: 'numeric',
+        });
+        return `${date} at ${time}`;
+      }
+
+      timeTaken(testRun) {
+        const hour = 1000*60*60;
+        const elapsed = new Date(testRun.time_end) - new Date(testRun.time_start);
+        return `${Math.round(elapsed / hour * 10) / 10}h`;
```
foolip: Maybe use .toFixed(1) here?

foolip:
Oh, actually this doesn't totally fix https://github.com/web-platform-tests/wpt.fyi/issues/282 either, that was about exposing test-level duration, which we have in the reports now.

Still, this PR is great :)

Hexcles:
Unfortunately, the execution time isn't always reliable.

Some runners do not provide `time_start` and `time_end`, in which case the results receiver will fill the current timestamp into these two fields, essentially making `time_start` the same as `created_at` and causing the execution time to be almost 0.

One solution is to change the results receiver to stop faking `time_end` and having the frontend to hide execution time if `time_end` is null.

lukebjerring:
I'll leave it to you to tweak this PR then :)

PR #608, opened 2018-10-01
Update the URL when submitting a query

lukebjerring:
## Description
One step further for https://github.com/web-platform-tests/wpt.fyi/issues/535

Updates the url when a query is submitted.

## Review Information
- Visit /flags, enable query builder
- Visit /results, build a query, submit
- See that the query params have updated.

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -19,7 +19,7 @@
             type: Boolean,
             computed: 'computeIsSubfolder(path)',
           },
-          onpathupdated: Function,
+          onLocationUpdated: Function,
         };
       }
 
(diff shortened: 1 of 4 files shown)


foolip:
I had some fun playing with the query builder. There's one thing I couldn't get to work, which does work on staging.wpt.fyi: add Chrome stable and Chrome beta with or without "Align runs" checked. The resulting URL is https://query-builder-5-dot-wptdashboard-staging.appspot.com/results/?product=chrome&product=chrome, i.e., the channel isn't part of the URL. The result is that only one column is shown.

lukebjerring:
Merging this into `query-builder-5` to resolve the issues seen there. One PR to rule them all.

Issue #411, opened 2018-08-08
Finding regressions in diff view is too hard


foolip:
https://staging.wpt.fyi/results/?product=chrome%5Bexperimental%5D&product=chrome%5Bstable%5D&diff shows this:
![image](https://user-images.githubusercontent.com/498917/43826603-fd69d76c-9af7-11e8-8cd1-e2d85eb708f4.png)

The numbers in the 3rd column are the differences between the pass / total numbers in columns 1 and 2.

Because the number of *failures* isn't counted, when both the total number of tests and total number of passes has changed, it's difficult to see if the number of failures has changed. The appropriate comparison would be in the number of failures, which is `(totalA - passA) - (totalB - passB)`. That's `(totalA - totalB) - (passA - passB)`, so as it turns out one can compare the numerator and denominator in column 3 to check if there are more failures, but that is not possible to do at a glance.

(https://github.com/web-platform-tests/wpt.fyi/issues/408 is a separate issue about the order, which also makes it tricky to skim the view for interesting things.)

foolip:
Actually, that formula doesn't quite cut it either. There is a regression in https://staging.wpt.fyi/results/css/css-values?product=chrome[experimental]&product=chrome[stable]&diff=true (viewport-units-css2-001.html) but it wasn't possible to see that from the high-level numbers, I just happened to drill down there because the numbers for css were large and I wanted to see why.

jgraham:
A comparison based on totals rather than actual differences seems rather problematic; finding that e.g. nightly is +8 passes vs stable doesn't mean there aren't regressions if it's actually +10 -2 i.e. 10 tests started passing, but two regressed.

Separate to that the view would be much clearer if it filtered the columns to only show directories/tests with changes.

foolip:
Yep, these are both changes @lukebjerring and I have discussed. This view is based on comparing summary numbers. It would be improved by comparing the pass/total numbers on the test level and aggregating the differences from tests that could be regressions.

To make it entirely reliable the comparison has to be done using the full results including subtests, but one can probably emulate that quite well using just the summary numbers in different way.

foolip:
The problem with comparing totals became apparent here:
https://staging.wpt.fyi/results/?product=firefox[experimental,buildbot]@8657903435&product=firefox[experimental,taskcluster]@8657903435&diff=true

[...]

PR #609, opened 2018-10-01
Show diff passes/regression numbers separately

lukebjerring:
## Description
See https://github.com/web-platform-tests/wpt.fyi/issues/411

Separates the `delta` number into 2 numbers (pass increases, pass decreases), such that a new pass and a new fail don't cancel each other out.

## Review Information
Visit a diff view, e.g. /results?label=taskcluster&product=chrome[beta]&product=chrome[stable]&diff

### DIFF
--- a/shared/run_diff.go
+++ b/shared/run_diff.go
@@ -88,8 +88,8 @@ func FetchRunResultsJSON(ctx context.Context, r *http.Request, run TestRun) (res
 	return results, nil
 }
 
-// GetResultsDiff returns a map of test name to an array of [count-different-tests, total-tests], for tests which had
-// different results counts in their map (which is test name to array of [count-passed, total-tests]).
+// GetResultsDiff returns a map of test name to an array of [newly-passing, newly-failing, total-delta], for tests which had
+// different results counts in their map (which is test name to array of [count-passed, total]).
 //
 func GetResultsDiff(before map[string][]int, after map[string][]int, filter DiffFilterParam) map[string][]int {
 	diff := make(map[string][]int)
@@ -104,22 +104,28 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
 				if !filter.Deleted {
 					continue
 				}
-				diff[test] = []int{resultsBefore[1], resultsBefore[1]}
+				diff[test] = []int{resultsBefore[0], resultsBefore[1] - resultsBefore[0], resultsBefore[1]}
 			} else {
 				if !filter.Changed && !filter.Unchanged {
 					continue
 				}
-				passDiff := abs(resultsBefore[0] - resultsAfter[0])
+				improved, regressed, delta := 0, 0, resultsBefore[0]-resultsAfter[0]
+				if delta >= 0 {
+					improved = delta
+				} else {
+					regressed = abs(delta)
+				}
 				countDiff := abs(resultsBefore[1] - resultsAfter[1])
-				changed := passDiff != 0 || countDiff != 0
+				changed := delta != 0 || countDiff != 0
 				if (!changed && !filter.Unchanged) || changed && !filter.Changed {
 					continue
 				}
 				// Changed tests is at most the number of different outcomes,
 				// but newly introduced tests should still be counted (e.g. 0/2 => 0/5)
 				diff[test] = []int{
-					max(passDiff, countDiff),
-					max(resultsBefore[1], resultsAfter[1]),
+					improved,
+					regressed,
+					abs(resultsBefore[1] - resultsAfter[1]),
 				}
 			}
 		}
@@ -132,7 +138,7 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
 
 			if _, ok := before[test]; !ok {
 				// Missing? Then N / N tests are 'different'
-				diff[test] = []int{resultsAfter[1], resultsAfter[1]}
+				diff[test] = []int{resultsAfter[0], resultsAfter[1] - resultsAfter[0], resultsAfter[1]}
 			}
 		}
 	}
--- a/shared/run_diff.go
+++ b/shared/run_diff.go
@@ -104,16 +104,16 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
 				if !filter.Deleted {
 					continue
 				}
-				diff[test] = []int{resultsBefore[0], resultsBefore[1] - resultsBefore[0], resultsBefore[1]}
+				diff[test] = []int{resultsBefore[0], resultsBefore[1] - resultsBefore[0], -resultsBefore[1]}
 			} else {
 				if !filter.Changed && !filter.Unchanged {
 					continue
 				}
 				improved, regressed, delta := 0, 0, resultsBefore[0]-resultsAfter[0]
-				if delta >= 0 {
-					improved = delta
+				if delta < 0 {
+					improved = abs(delta)
 				} else {
-					regressed = abs(delta)
+					regressed = delta
 				}
 				countDiff := abs(resultsBefore[1] - resultsAfter[1])
 				changed := delta != 0 || countDiff != 0
@@ -125,7 +125,7 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
 				diff[test] = []int{
 					improved,
 					regressed,
-					abs(resultsBefore[1] - resultsAfter[1]),
+					resultsAfter[1] - resultsBefore[1],
 				}
 			}
 		}
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -621,10 +621,11 @@ found in the LICENSE file.
       getDiffDelta(node, prop) {
         const [before, after] = node.results;
         switch (prop) {
-        case 'regressions':
+        case 'regressions': {
           const failingBefore = before.total - before.passes;
           const failingAfter = after.total - after.passes;
           return Math.min(failingBefore - failingAfter, 0);
+        }
         case 'total':
           return before.total - after.total;
         case 'passes':
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -90,10 +90,10 @@ found in the LICENSE file.
       .passes-80 { background-color: hsl(0, 85%, 85%); }
       .passes-100 { background-color: var(--paper-light-green-a400); }
 
-      span.delta.negative {
+      span.delta.regressions {
         color: var(--paper-red-700);
       }
-      span.delta.positive {
+      span.delta.passes {
         color: var(--paper-green-700);
       }
       td.none {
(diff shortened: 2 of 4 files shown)


shared/run_diff.go
```diff
@@ -104,22 +104,28 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
...
-				passDiff := abs(resultsBefore[0] - resultsAfter[0])
+				improved, regressed, delta := 0, 0, resultsBefore[0] - resultsAfter[0]
+				if delta < 0 {
+					improved = abs(delta)
+				} else {
+					regressed = delta
+				}
 				countDiff := abs(resultsBefore[1] - resultsAfter[1])
-				changed := passDiff != 0 || countDiff != 0
+				changed := delta != 0 || countDiff != 0
```
foolip: Given that `countDiff` now isn't used for anything else, can't this be simplified to just `changed := resultsAfter != resultsBefore`? This does end up collapsing to just that, but it took me a while to tell.

webapp/components/wpt-results.html
```diff
@@ -237,9 +237,11 @@
                 </template>
                 <template is="dom-if" if="[[diffShown]]">
                   <td class$="numbers [[ testResultClass(node, index, diffRun, 'passes') ]]">
-                    <span class$="passes [[ testResultClass(node, index, diffRun, 'passes') ]]">{{ getNodeResultDataByPropertyName(node, -1, diffRun, 'passes') }}</span>
+                    <span class="delta passes">{{ getNodeResultDataByPropertyName(node, -1, diffRun, 'passes') }}</span>
```
foolip: Maybe class name "improvements" to match the documentation?

webapp/components/wpt-results.html
```diff
@@ -237,9 +237,11 @@
...
                     /
-                    <span class$="total [[ testResultClass(node, index, diffRun, 'total') ]]">{{ getNodeResultDataByPropertyName(node, -1, diffRun, 'total') }}</span>
+                    <span class="delta regressions">{{ getNodeResultDataByPropertyName(node, -1, diffRun, 'regressions') }}</span>
```
foolip: That the zeros here are red makes it quite a bit harder to spot regressions. The ability to filter to only regressions would be great, but failing that maybe the text should be black if the number is zero?

webapp/components/wpt-results.html
foolip: Commenting on a random changed line in this file. I think that only you and I will know what the 3 numbers in the view means. Can you add some legend under the diff icon, maybe where hovering shows extra information like for runs?

foolip:
Here's a great view:
https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/?product=chrome%5Bstable%5D&product=chrome%5Bbeta%5D&aligned=true&diff=true

Compare that to staging:
https://staging.wpt.fyi/results/?product=chrome%5Bstable%5D&product=chrome%5Bbeta%5D&aligned=true&diff=true

The difference in how WebCryptoAPI is presented is interesting. Because those subtests were simply missing in dev, it doesn't show up as a regression. However, **bug alert**, the sign of the difference seems wrong, it says +2 but should be -2. And I thought I reviewed the signs of each number very carefully :/

foolip:
Hmm, something seems to be wrong with how the regression numbers are aggregated too. Check this view:
https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/service-workers/service-worker?sha=ee2e69bfb1&product=safari-12.0&product=safari-12.1&diff=true

I see -3 and -1, but stepping out the aggregated number is -2.

foolip:
https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/css?sha=ee2e69bfb1&product=safari-12.0&product=safari-12.1&diff=true is another case, where there's a -33 for css/vendor-imports/ which doesn't propagate up.

foolip:
I've also found a view where the results are perhaps a bit confusing:
https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/web-animations/animation-model/animation-types?sha=ee2e69bfb1&product=safari-12.0&product=safari-12.1&diff=true

Subtests that were previously missing are now running and failing. We can't fix this with the current data model, but it's something to include in a bug about subtest-level diffing.

shared/run_diff.go
```diff
@@ -104,22 +104,28 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
 				if !filter.Deleted {
 					continue
 				}
-				diff[test] = []int{resultsBefore[1], resultsBefore[1]}
+				diff[test] = []int{resultsBefore[0], resultsBefore[1] - resultsBefore[0], -resultsBefore[1]}
 			} else {
 				if !filter.Changed && !filter.Unchanged {
```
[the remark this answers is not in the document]
lukebjerring: Yeah I can do that in a follow-up.

shared/run_diff.go
```diff
@@ -104,22 +104,28 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
...
 					continue
 				}
-				passDiff := abs(resultsBefore[0] - resultsAfter[0])
+				improved, regressed, delta := 0, 0, resultsBefore[0] - resultsAfter[0]
+				if delta < 0 {
```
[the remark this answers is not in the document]
lukebjerring: Well, actually this issue causes the aggregation issues we're seeing (many of) below. Ultimately, we can't compute these numbers on the client (because by that stage the summary numbers have already been squashed). We'll need to append the `diff` arrays to the `search` api.

webapp/components/wpt-results.html
lukebjerring: I'm re-using the existing styling; the less changes the better.

foolip: Oh, didn't realize it was an existing class.

shared/run_diff.go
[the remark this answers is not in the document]
foolip: Now I'm a bit confused about where this data is coming from. /api/diff doesn't show up in devtools at all, how does this fit together?

In any event, it doesn't sound quite right that the aggregation bugs are due to this if-else branch at bottom. If we have the [newly-passing, newly-failing, total-delta] array of numbers for each test, then the aggregate array for some directory is just the "vector sum" of all of the test-level arrays.

What is the place where numbers are squashed in a way that isn't the "vector sum"?

shared/run_diff.go
lukebjerring: Done.

webapp/components/wpt-results.html
lukebjerring: Maybe in a follow-up; let's improve cosmetics separately.

shared/run_diff.go
```diff
@@ -104,22 +104,28 @@ func GetResultsDiff(before map[string][]int, after map[string][]int, filter Diff
...
 					continue
 				}
 				// Changed tests is at most the number of different outcomes,
 				// but newly introduced tests should still be counted (e.g. 0/2 => 0/5)
 				diff[test] = []int{
-					max(passDiff, countDiff),
-					max(resultsBefore[1], resultsAfter[1]),
+					improved,
+					regressed,
+					resultsAfter[1] - resultsBefore[1],
```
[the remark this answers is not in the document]
lukebjerring: With the latest commit, the negation should be obvious. Let's debate there. "+N failing tests" felt weird to me.

lukebjerring:
I wrote https://github.com/web-platform-tests/wpt.fyi/pull/609/commits/d7d5a0f4a6a2a6a81d7c6d3ac10ab53a1fbcba4c to fetch the API endpoint asynchronously. PTAL when it's deployed.

lukebjerring:
So, I've noticed a bug with the diff endpoint that it doesn't take into account the `aligned` param, which causes the runs being compared to differ from those shown in the client. For this reason, I've thrown the `/api/diff` fetch behind a flag (see `/flags` to enable it).

lukebjerring:
I've added a filter icon in https://github.com/web-platform-tests/wpt.fyi/pull/609/commits/0ad3456a9b5614c5fe87bce8f89b29f6de3b7666 if you visit `/flags` and enable it. I feel like this, if we're happy with it, would cover #411 ?

foolip:
https://diff-3-numbers-dot-announcer-dot-wptdashboard-staging.appspot.com/ is 404 right now, was it garbage collected too soon, or never deployed?

Hexcles:
@foolip you clicked the wrong link. That's the announcer.

foolip:
@Hexcles it's the link in https://github.com/web-platform-tests/wpt.fyi/pull/609#issuecomment-426407511. Is that itself a bug? I've found the right URL, can now review :)

foolip:
Oh, wait, I guess this is staging instances of the different services. That was not easy to notice :)

Hexcles:
@foolip there should be two comments with two links to the webapp and announcer respectively. They are usually next to each other, but somehow it wasn't the case in this PR.

foolip:
Cool, aggregation issues seem to be fixed now.

https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/?sha=ee2e69bfb1&product=safari-11.1&product=safari-12.0&diff is a pretty complex diff (the one sent to [webkit-dev](https://lists.webkit.org/pipermail/webkit-dev/2018-October/030197.html)) which is now more fun to look at. There are a lot of red numbers because more tests are running and failing. I think this is a good default mode, but perhaps some filtering make it only count PASS&rarr;not-PASS later.

*BUG*: with /api/diff enabled, https://diff-3-numbers-dot-wptdashboard-staging.appspot.com/results/css/css-text/i18n/zh?label=taskcluster&product=chrome[stable]&product=chrome[beta]&diff=true has very wrong numbers, and this aggregates into a seeming huge regression in css/css-text/ which actually has no new failures.

webapp/components/test-run.html
```diff
@@ -58,14 +58,18 @@
...
+      <paper-tooltip offset=0>
+        <template is="dom-if" if="{{ !isDiff(testRun.browser_name) }}">
           {{displayName(testRun.browser_name)}} {{testRun.browser_version}}<br>
           Labels: {{displayLabels(testRun.labels)}}<br>
           Started {{timeFormat(testRun.time_start)}} {{timeTaken(testRun)}}<br>
           {{moreTooltip(testRun)}}
-        </paper-tooltip>
-      </template>
+        </template>
+        <template is="dom-if" if="{{ isDiff(testRun.browser_name }}">
```
foolip: This tooltip shows up as part of the browser tooltips too, which I think isn't intended. I don't even see how it's possible given that the conditions are negated. Found https://github.com/Polymer/polymer/issues/2165 while looking for a construct that would make the bug impossible.

webapp/components/test-run.html
```diff
@@ -58,14 +58,18 @@
...
+          diff numbers are for:<br>
+          [Newly passing] / [Newly failing] / [Total count delta]
```
foolip: Mind if I lowercase there? Looks a bit odd, to my eyes.

webapp/components/wpt-results.html
```diff
@@ -210,7 +210,12 @@
                 <th><test-run test-run="[[testRun]]"></test-run></th>
               </template>
               <template is="dom-if" if="[[diffShown]]">
-                <th><test-run test-run="[[diffRun]]"></test-run></th>
+                <th>
+                  <test-run test-run="[[diffRun]]"></test-run>
+                  <template is="dom-if" if="[[diffFilter]]">
+                    <paper-icon-button icon="filter-list" onclick="[[toggleDiffFilter]]"></paper-icon-button>
```
foolip: This is great! One improvement would be to give this some state or button feel, so that one can tell that clicking it does something.

webapp/components/wpt-results.html
```diff
@@ -353,6 +364,7 @@ <h3>
             type: Boolean,
             value: false,
           },
+          onlyShowDifferences: Boolean,
```
foolip: Future feature request: only show regressions (not sure about UI)

webapp/components/wpt-results.html
```diff
@@ -482,26 +511,45 @@ <h3>
                 revision: 'diff',
                 browser_name: 'diff',
               };
+              this.fetchDiff();
             }
           })
         );
       }
 
       async fetchResults() {
```
foolip: Using `async` here still looks like a bug. Even if `this.load` return a promise (not checked) it's not returned, so fetchResults() would look like `Promise.resolve(undefined)` with side-effects happening after the promise is settled.

webapp/components/wpt-results.html
```diff
@@ -482,26 +511,45 @@ <h3>
...
-          this.loadingCount--;
+            })
+            .then(json => {
+              this.searchResults = json.results;
+              this.refreshDisplayedNodes();
+            })
+        );
+      }
+
+      async fetchDiff() {
```
foolip: `async` oddity here too.

webapp/components/test-run.html
lukebjerring: Ah; missing a trailing `)`

Issue #604, opened 2018-10-01
Some Taskcluster runs aren't uploaded to wpt.fyi successfully


foolip:
https://wpt.fyi/test-runs currently shows stable and beta runs from Taskcluster, but not the dev run.

https://api.github.com/repos/web-platform-tests/wpt/commits/ee2e69bfb1d44c4013a8ce94ca6932f86d63aa31/statuses does include "TaskGroup: success" pointing to https://tools.taskcluster.net/groups/ecpaEJHuRfmPunkTiCmsKQ which appears to be in good condition.

This is the weekly run, which is why I noticed. It's possible that runs are being randomly dropped in a way that is less noticeable.

@Hexcles, can you investigate?

Hexcles:
The working theory is that my webhook takes too long to respond, sometimes close to 1min, which is well beyond [GitHub's 10s timeout for webhooks](https://developer.github.com/v3/guides/best-practices-for-integrators/#favor-asynchronous-work-over-synchronous). The client (GitHub server) then closes the connection, which *sometimes* causes AppEngine to terminate the request handling thread, leading to incomplete results.

We should make sure the webhooks respond within 10 seconds, which I think is totally possible given that there are quite a few HTTP requests we can parallelize.

foolip:
@Hexcles would it work to download the results at the other end of the task queue, or is the contract that once the results receiver API has returned, that the client doesn't need to keep storing the results for some unknown number of minutes/hours?

Hexcles:
@foolip yeah that would also work, but I think the download should be well within 10s once parallelized.

Hexcles:
I did a little more digging into the logs. In addition to the potential timeout issue, Taskcluster API itself also sometimes fails. Namely, the endpoint for downloading the test artifacts has failed a few times in the past week. I suspect we are hitting the endpoint too fast and/or too soon after a task finishes (the artifacts might not be available on their cloud storage yet). I'll add retry.

Hexcles:
The latency of the effective webhook requests (the ones that actually upload results, not the no-op ones) has decreased from 85s to 5s. And with the retry mechanism built in, I believe this is largely solved.

Now we just need to wait for a prod release.

Hexcles:
Closing this unless it happens again.

PR #613, opened 2018-10-02
[webhook] Upload test results in parallel

Hexcles:
This PR changes the Taskcluster webhook to upload test results to the
results receiver API in parallel in order to reduce the webhook response
time (in half, as we currently run two browsers in each task group).

The goroutine and error handling are tested with simple HTTP testing
servers.

Part of #604 .

### DIFF
--- a/api/webhook/taskcluster.go
+++ b/api/webhook/taskcluster.go
@@ -16,11 +16,11 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
-	"google.golang.org/appengine/log"
 	"google.golang.org/appengine/urlfetch"
 
 	"github.com/web-platform-tests/wpt.fyi/shared"
@@ -37,19 +37,20 @@ func tcWebhookHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	ctx := appengine.NewContext(r)
+	ctx := shared.NewAppEngineContext(r)
+	log := shared.GetLogger(ctx)
 
 	payload, err := ioutil.ReadAll(r.Body)
 	r.Body.Close()
 	if err != nil {
-		log.Errorf(ctx, "%v", err)
+		log.Errorf("%v", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
 
 	secret, err := getSecret(ctx)
 	if err != nil {
-		log.Errorf(ctx, "%v", err)
+		log.Errorf("%v", err)
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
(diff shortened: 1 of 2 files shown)


mdittmer:
What are the specs of the instances that run this code? (In particular: how many cores?)

Hexcles:
@mdittmer just one, but IIUC, even if there's only 1 goroutine executing at a time (which I think is unlikely), we will still see big performance benefits as `urlfetch` should yield when waiting for IO.

> A Go runtime instance can handle multiple requests concurrently, though only one running goroutine will get CPU time at a time. Other goroutines can perform I/O (read a file, make a system or API call) while that happens.

https://cloud.google.com/appengine/docs/standard/go/runtime

api/webhook/taskcluster.go
```diff
@@ -252,6 +251,31 @@ func verifySignature(message []byte, signature string, secret string) bool {
...
+			if err != nil {
+				log.Errorf("%v", err)
+				errors <- err
+			}
+		}(browser, urls)
+	}
+	wg.Wait()
+	close(errors)
+	if _, ok := <-errors; ok {
+		return fmt.Errorf("error(s) occured when talking to %s", api)
```
mdittmer: As we discussed, one possibility here is to concat strings that constitute the errors. I would say that's worth doing if:

1. This error string may get propagated to somewhere other than `stderr` (e.g., in a HTTP response body), and
1. Someone without access to our logs may find themselves reading (somewhere other than `stderr`) and wondering which runs succeeded and which runs failed.

mdittmer: ... in which case we should probably use `for ... range errors { ... }` to capture the error, and also make sure that the error strings contain information about which run/browser caused the error(s).

api/webhook/taskcluster_test.go
```diff
@@ -94,3 +97,29 @@ func TestVerifySignature(t *testing.T) {
...
+
+	err := createAllRuns(logrus.New(), nil, server.URL, "username", "password",
+		map[string][]string{"chrome": []string{"1"}, "firefox": []string{"1", "2"}},
+	)
+	assert.Nil(t, err)
+	assert.Equal(t, 2, requested)
+}
+
+func TestCreateAllRuns_error(t *testing.T) {
+	server := httptest.NewServer(http.NotFoundHandler())
```
mdittmer: It would be nice to have a test that arranges for a specific request to error-out.

api/webhook/taskcluster.go
```diff
@@ -122,18 +124,15 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
...
-		slowCtx, cancel := context.WithTimeout(ctx, time.Minute)
-		client := urlfetch.Client(slowCtx)
-		err := createRun(client, api, username, password, urls)
-		cancel()
-		if err != nil {
-			return false, err
-		}
+
+	// Set timeout to 1 min (the default is 5s) to give the receiver enough time to download the reports.
+	slowCtx, cancel := context.WithTimeout(ctx, time.Minute)
```
mdittmer: Let's abstract this `time.Minute` into a package-local variable. That way, we can write a test that verifies we get a deadline error without the test having to run for a full minute.

api/webhook/taskcluster.go
Hexcles: Done.

api/webhook/taskcluster_test.go
Hexcles: Done. Now there are two test cases.

Hexcles:
@mdittmer PTAL again

Issue #12, opened 2018-04-10
Use a pluralization library


lukebjerring:
_From @lukebjerring on January 22, 2018 14:28_

For the web components which have plurals in title, switch the ad-hoc code to a 3rd-party library.
e.g. https://github.com/w3c/wptdashboard/pull/349

_Copied from original issue: w3c/wptdashboard#379_

PR #619, opened 2018-10-04
Use 'pluralize' to pluralize browser(s)

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/12


lukebjerring:
@rwaldron a blast from the past here, but was originally your feedback to switch to a library.

Issue #615, opened 2018-10-03
https://wpt.fyi/test-runs shows commits out of order


foolip:
https://wpt.fyi/test-runs current shows this:
![image](https://user-images.githubusercontent.com/498917/46397977-0f8cb680-c6f4-11e8-8def-af7f4cc6e9c7.png)

I found it surprising that there are stable runs for two consecutive commits and suspected a bug, but actually it's just a problem with the order. Probably because the Edge and Safari results for 67152fdecd arrived many hours later than the others, that row has a later date. The other rows with Buildbot results (f15449622a and 63641fb5fe) also appear out of order compared to the commit time order.

https://wpt.fyi/test-runs?label=taskcluster does match the commit order currently, but that's probably not guaranteed because the Taskcluster runs for two consecutive commits will be racing to finish.

foolip:
Even though it wouldn't fix the order in all cases, just sorting by time of earliest arrived results rather than latest arrived results would fix most of this.

jugglinmike:
[We discussed this previously and landed on "sort by `test_start` time" as a rough approximation](https://github.com/web-platform-tests/wpt.fyi/issues/310). I think the issue here isn't about when the results are uploaded, but when the builds were started. On Sauce Labs, we don't start collecting from Safari until we've finished with Edge. Those results have a delayed `test_start`, and that's probably to blame here.

Hexcles:
Thanks for the information, @jugglinmike . We are already doing what @foolip suggested in the first comment, but I wasn't aware of the SauceLabs situation. I guess the reason why we only start Safari after Edge is capacity/quota?

jugglinmike:
Capacity means we can't run all the builds in parallel, but it doesn't mean we can't schedule collection for the browsers to start and finish at roughly the same time. That would be possible if we interleaved the segments for Safari and Edge. Even then, though, the results would be arriving long after the Firefox and Chrome results (from both Buildbot and TaskCluster).

Hexcles:
*Arriving* much later isn't an issue, as long as they *start* as soon as possible, the list recent runs will be ordered more correctly (closer to commit history).

I don't know how much change it needs and how risky it is to start Sauce and Edge at the same time. If it's practical, it'd be nice to have.

foolip:
[...]

PR #621, opened 2018-10-04
Order SHAs by first run time

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/615

## Review Information
Visit /test-runs, observe ordering

### DIFF
--- a/webapp/components/wpt-runs.html
+++ b/webapp/components/wpt-runs.html
@@ -185,8 +185,8 @@ found in the LICENSE file.
           }, {});
 
           // We flatten into an array of objects so Polymer can deal with them.
-          const firstRunTime = runs => {
-            return Array.from(Object.values(runs))
+          const firstRunDate = runs => {
+            return Object.values(runs)
               .reduce((oldest, runs) => {
                 for (const time of runs.map(r => new Date(r.time_start))) {
                   if (time < oldest) {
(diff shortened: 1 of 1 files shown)


webapp/components/wpt-runs.html
```diff
@@ -79,7 +80,7 @@
           <tr>
             <th width="[[computeThWidth(browsers)]]">SHA</th>
             <template is="dom-repeat" items="{{ browsers }}" as="browser">
-              <th width="[[computeThWidth(browsers)]]">[[browser]]</th>
+              <th width="[[computeThWidth(browsers)]]">[[displayName(browser)]]</th>
```
foolip: Ah, capitalized, I had noticed the lower case, this is better!

foolip:
I haven't checked against a commit log, but https://test-run-order-dot-wptdashboard-staging.appspot.com/test-runs looks as expected now, with lots of runs between fec3928f35 and 67152fdecd. \o/

webapp/components/wpt-runs.html
```diff
@@ -184,16 +185,26 @@
           }, {});
 
           // We flatten into an array of objects so Polymer can deal with them.
-          const firstRunTime = x => Object.values(x.runs)[0][0].created_at;
-          const flattened = Object.entries(testRunsBySha)
-            .map(({0: sha, 1: runs}) => ({ sha, runs }))
-            .sort((a, b) => firstRunTime(a) < firstRunTime(b) ? 1 : -1);
+          const firstRunTime = runs => {
+            return Array.from(Object.values(runs))
```
foolip: `Array.from` is redundant because `Object.values(runs)` returns an array.

webapp/components/wpt-runs.html
```diff
@@ -184,16 +185,26 @@
...
+                for (const time of runs.map(r => new Date(r.time_start))) {
+                  if (time < oldest) {
+                    oldest = time;
+                  }
+                }
+                return oldest;
+              }, new Date()); // Existing runs should be historical...
+          };
+          const flattened = Array.from(Object.entries(testRunsBySha))
+            .map(({0: sha, 1: runs}) => ({ sha, runs, firstRunTime: firstRunTime(runs) }))
```
foolip: Unless there's more than meets the eye here, `({0: sha, 1: runs})` can be `([sha,runs])`. (`Object.entries` returns array of arrays, which also means that `Array.from` isn't needed.)

webapp/components/wpt-runs.html
```diff
@@ -184,16 +185,26 @@
           }, {});
 
           // We flatten into an array of objects so Polymer can deal with them.
-          const firstRunTime = x => Object.values(x.runs)[0][0].created_at;
-          const flattened = Object.entries(testRunsBySha)
-            .map(({0: sha, 1: runs}) => ({ sha, runs }))
-            .sort((a, b) => firstRunTime(a) < firstRunTime(b) ? 1 : -1);
+          const firstRunTime = runs => {
```
foolip: Maybe rename to `firstRunDate` to be more accurate, and `firstRunTime.getTime()` wouldn't look as redundant below.

Hexcles:
This is great! I didn't know the recent runs view was sorted on the client side. Thanks, Luke!

PR #623, opened 2018-10-05
Drop redundant `=true` when building a query string

lukebjerring:
## Description
Makes the query strings a bit more readable.

## Review Information
`searchParams.set(key, value)` _requires_ value, and stringifys it (so undefined doesn't work). An empty string results in 'foo='.

Hence, a hacky replace-all it is.

### DIFF
--- a/webapp/components/results-navigation.html
+++ b/webapp/components/results-navigation.html
@@ -21,7 +21,7 @@
             url.searchParams.set(k, params[k]);
           }
         }
-        return url.search;
+        return `${url.search}`.replace(/=true/g, '');
       }
     };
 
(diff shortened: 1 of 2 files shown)


foolip:
Hmm, does it work? I tried it and ended up with https://no-equals-true-dot-wptdashboard-staging.appspot.com/results/?label=stable&aligned=true via the query builder.

lukebjerring:
I'll dig into why queryBuilder is not following the same behaviour.

foolip:
It doesn't work yet, but maybe it hasn't deployed. I'll assume it works and reopen the issue if it doesn't work on staging after this is merged.

PR #629, opened 2018-10-09
Fix single-item diffs

lukebjerring:
## Description
Fixes single-product views, broken by https://github.com/web-platform-tests/wpt.fyi/pull/609

## Review Information
Load `?product=chrome`, see that results load as expected.

PR #630, opened 2018-10-09
Put product builder in collapsible

lukebjerring:
## Description
Puts the builder in a collapsible that closes when you submit.

## Review Information
- Visit /flags, enable `queryBuilder`
- Visit /esults, click the pencil, submit a query
- See it collapse

### DIFF
--- a/webapp/bower.json
+++ b/webapp/bower.json
@@ -20,7 +20,8 @@
     "paper-listbox": "2",
     "paper-item": "2",
     "paper-checkbox": "2",
-    "pluralize": "^7.0.0"
+    "pluralize": "^7.0.0",
+    "iron-collapse": "2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


foolip:
Expanding and collapsing work (neat!) but after I'm done, I end up at https://builder-collapse-dot-wptdashboard-staging.appspot.com/results/?label=stable&product=chrome&aligned with no results loaded. Loading that in a new tab directly is also broken. Ought to be a regression in this PR because staging.wpt.fyi works.

lukebjerring:
That'll be the regression from https://github.com/web-platform-tests/wpt.fyi/pull/629

foolip:
Should the fix go in before this PR? I'll review the code here in any case.

foolip:
Approved, by wait, can this be tested?

lukebjerring:
No, I don't think it can be tested easily.

PR #631, opened 2018-10-09
Test and implement test status enum for use with Cloud Spanner

mdittmer:
## Description

Test and implement a unified enum for (sub)test status as they will be modeled in Cloud Spanner.

### DIFF
--- a/shared/spanner.go
+++ b/shared/spanner.go
@@ -35,35 +35,68 @@ const (
 	// TestStatusCrash indicates that the WPT test runner crashed attempting to run the test.
 	TestStatusCrash int64 = 7
 
+	// TestStatusNameUnknown is the string representation for an uninitialized
+	// TestStatus and should not be used.
+	TestStatusNameUnknown string = "UNKNOWN"
+
+	// TestStatusNamePass is the string representation of a test result where the
+	// test passed.
+	TestStatusNamePass string = "PASS"
+
+	// TestStatusNameOK is the string represnetation of a test result where the
+	// test ran completely but may not have passed (and/or not all of its subtests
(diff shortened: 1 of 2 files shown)


shared/spanner.go
```diff
@@ -0,0 +1,87 @@
...
+	// TestStatusDefault is the default value used when a status string cannot be
+	// interpreted.
+	TestStatusDefault int64 = 0
+
+	// TestStatusNameDefault is the default string used when a status value cannot
+	// be interpreted.
+	TestStatusNameDefault string = "UNKNOWN"
+)
+
+var testStatusValues = map[string]int64{
```
Hexcles: I can't think of a use case for converting from string to int. If we don't have one right now, let's remove this map, the `TestStatusNameDefault` constant and the related function. We can always add them back later.

shared/spanner.go
```diff
@@ -0,0 +1,87 @@
...
+	"OK":      2,
+	"ERROR":   3,
+	"TIMEOUT": 4,
+	"NOT_RUN": 5,
+	"FAIL":    6,
+	"CRASH":   7,
+}
+
+var testStatusNames = map[int64]string{
+	0: "UNKNOWN",
```
Hexcles: The idiomatic way to do this in Go (seen in `net/http/status.go`) is to use the constants declared above as the keys of the map. I think it's a good idea to avoid duplication (and hence the chances of mistakes).

shared/spanner.go
mdittmer: The use case is parsing WPT reports for storing data in Cloud Spanner.

PR #632, opened 2018-10-09
 Receiver create_run: copy datastore key into TestRun.ID in response

mdittmer:
Also contains an ugly custom data type to work around no appengine-context-free mechanism for constructing datastore keys with predefined kind and integral ID.

### DIFF
--- a/api/receiver/appengine.go
+++ b/api/receiver/appengine.go
@@ -19,10 +19,18 @@ import (
 	"google.golang.org/appengine/user"
 )
 
+// DatastoreKey is a context-free constructable form of datastore.Key that
+// contains identifying data for a datastore object that was assigned an
+// integral ID with no parent key.
+type DatastoreKey struct {
+	Kind string
+	ID   int64
+}
+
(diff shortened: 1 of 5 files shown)


api/receiver/appengine.go
```diff
@@ -19,10 +19,18 @@ import (
 	"google.golang.org/appengine/user"
 )
 
+// DatastoreKey is a context-free constructable form of datastore.Key that
+// contains identifying data for a datastore object that was assigned an
+// integral ID with no parent key.
```
Hexcles: Oh right, `datastore.Key` has a context... Sad. I might choose returning only the `ID`, but this is good, too.

Hexcles:
I'll leave it to you to decide whether you'd like to squash or rebase merge (either is fine with me).

api/receiver/appengine.go
mdittmer: I initially returned the ID, but we had tests that verify that the kind we get back is `TestRun`. Believing that the tests are meaningful, I changed the code to preserve both pieces of data from the key that might be relevant.

PR #633, opened 2018-10-09
Results processor: Get test run ID from results receiver API payload

mdittmer:
This PR depends on #632 landing first.

### DIFF
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -204,9 +204,10 @@ def task_handler():
 
     ds = datastore.Client()
     secret = ds.get(ds.key('Uploader', '_processor'))['Password']
-    id = wptreport.create_test_run(report, labels, uploader, secret,
-                                   results_gcs_path, raw_results_gcs_path)
-    assert (isinstance(id, int) and id != 0)
+    test_run_id = wptreport.create_test_run(report, labels, uploader, secret,
+                                            results_gcs_path,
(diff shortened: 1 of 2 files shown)


results-processor/main.py
```diff
@@ -204,8 +204,9 @@ def task_handler():
 
     ds = datastore.Client()
     secret = ds.get(ds.key('Uploader', '_processor'))['Password']
-    wptreport.create_test_run(report, labels, uploader, secret,
-                              results_gcs_path, raw_results_gcs_path)
+    id = wptreport.create_test_run(report, labels, uploader, secret,
```
Hexcles: `id` is a keyword in Python; we should avoid overwriting a keyword.

results-processor/main.py
```diff
@@ -204,8 +204,9 @@ def task_handler():
...
+                                   results_gcs_path, raw_results_gcs_path)
+    assert (isinstance(id, int) and id != 0)
```
Hexcles: This will fail... There's a high chance that the test run ID is of type `long`, not `int`, because of its range. And `long` isn't a subclass of `int`. In fact, they don't even have a common base class. (Well, they have an *abstract* common base class, `numbers.Integral`.)

I'd say just `assert id` (which makes sure it's not `None` or 0) and don't worry about the type, the Pythonic way :)

PR #634, opened 2018-10-10
Abstract basic auth in AppEngine APIs out of receiver package

mdittmer:
This is so that the same authentication strategy can be used for requesting that runs be written out to Cloud Spanner.

### DIFF
--- a/api/auth/appengine.go
+++ b/api/auth/appengine.go
@@ -15,17 +15,14 @@ import (
 // only be accessed by services in this AppEngine project via Datastore.
 const InternalUsername = "_processor"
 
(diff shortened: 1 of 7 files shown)


Hexcles:
To briefly capture the offline discussion:

I prefer duplicating (references) of the context in each API struct because conceptually we are creating a low-level, modular AppEngineAPI structs that can be composited with other modules into a high-level struct (via type embedding) to be used in a specific service. Imagine in addition to `auth`, we might have `taskqueue` factored out in a similar fashion in the future. With this type hierarchy in mind, there are two logical placements of the context: either the high-level/embedding struct holds it (and the low-level/embedded structs would need references to the embedder) or we duplicate the context everywhere.

PR #635, opened 2018-10-10
Initial implementation of spanner API-as-service

mdittmer:
## Description

This establishes the configuration and code for a handler that accepts authenticated requests to push runs in Datastore and GCS to Cloud Spanner. The request handler returns immediately after constructing a (composite) identifier for the request that can be used to track push logs. A `TODO` is included to implement Datastore/GCS pull and Cloud Spanner push.


api/spanner/auth.go
```diff
@@ -0,0 +1,76 @@
...
+func (a *nopAuthenticator) Authenticate(ctx context.Context, r *http.Request) bool {
+	return true
+}
+
+var nopa = &nopAuthenticator{}
+
+// NewNopAuthenticator returns an authenticator that always
+// authenticates successfully.
+func NewNopAuthenticator() Authenticator {
+	return nopa
```
Hexcles: Nit: we can simply create a new `nopAuthenticator` here. It's an empty struct so we don't need to worry about memory. That way, we avoid a global (though unexported) variable.

api/spanner/service/app.yaml
```diff
@@ -0,0 +1,24 @@
...
+env: flex
+service: spannerwrite
+
+# Should match something at https://cloud.google.com/compute/docs/machine-types.
+resources:
+  # g1-small
+  cpu: 0.5
+  memory_gb: 1.70
+
+# Setting values under automatic_scaling yields error:
```
Hexcles: What values did you try to set?

mdittmer: [These two](https://github.com/web-platform-tests/wpt.fyi/pull/635/commits/3f92a83de8f035deaf47fbdb2bd20394b22c1432)

PR #637, opened 2018-10-11
Fix bug in /api/diff counts

lukebjerring:
## Description
Deleted tests were being counted as passes + regressions, when we only count them as a change in the total test count.


webapp/components/wpt-results.html
```diff
@@ -493,7 +493,7 @@ <h3>
           if (state) {
             const builder = this.shadowRoot.querySelector('test-runs-query-builder');
             if (builder) {
-              builder.submitQueryParams(state);
+              builder.updateQueryParams(state);
```
lukebjerring: Drive-by fix for a breakage from #609

PR #646, opened 2018-10-12
Use -moz-max-content so that cells show up on hover on Firefox.

emilio:
We're working on unprefixing these keywords in:

  https://bugzilla.mozilla.org/show_bug.cgi?id=1322780

But it'd be nice if it worked before that's fixed :)


lukebjerring:
Thanks!

emilio:
Thank you both for the quick turnaround!

PR #648, opened 2018-10-12
[receiver] Concurrently fetch results and retry failures

Hexcles:
## Description

This is a two-fold approach to address #604 : 

1. Retry fetching results (for URL payloads). External network requests are not reliable. And in reality, we have seen intermittent timeout & 404s when downloading test results from Taskcluster.
2. Execute the requests in parallel (in goroutines). This will drastically reduce the response time and avoid timeout.

## Review Information

Extensive testing is added. Tests should pass.

## Changes

In addition to the two fixes, this PR also un-exports a few methods in `appengine.go`. There are no users of these methods outside of the receiver. And the `AppEngineAPI` interface was created for ease of testing (mocking), not code reuse. The APIs aren't very generic, and should only be exported when needed.


api/receiver/receive_results.go
```diff
@@ -98,38 +107,65 @@ func handleURLPayload(a AppEngineAPI, uploader string, urls []string, extraParam
 
 	var payloadType string
 	gcs := make([]string, 0, len(urls))
+	errors := make(chan error, len(urls))
```
mdittmer: I would advocate for one of two strategies here:

1. Move `errors`, `wg` into the `if len(urls) > 1` block, or
2. Drop the `if len(urls) > 1` block and wait on a single goroutine.

I think I prefer 2. One complication with 1 is that `saveFileToGCS` may need to support a `nil` `sync.WaitGroup`.

api/receiver/receive_results.go
```diff
@@ -98,38 +107,65 @@ func handleURLPayload(a AppEngineAPI, uploader string, urls []string, extraParam
...
+	}
+	defer f.Close()
+	// TODO: Detect whether the fetched blob is gzipped.
+	if err := a.uploadToGCS(gcsPath, f, true); err != nil {
+		e <- err
+	}
+}
+
+func fetchFile(a AppEngineAPI, url string) (io.ReadCloser, error) {
+	// It is safe to reuse the request as long as we are not modifying it.
```
mdittmer: There is no `*http.Request` here. What is this comment meant to convey?

api/receiver/receive_results.go
```diff
@@ -98,38 +107,65 @@ func handleURLPayload(a AppEngineAPI, uploader string, urls []string, extraParam
...
+	log := shared.GetLogger(a.Context())
+	sleep := time.Millisecond * 500
+	for retry := 0; retry < NumRetries; retry++ {
```
mdittmer: I like https://github.com/avast/retry-go for this sort of thing, but I don't feel strongly about it. If you think using it would make the code clearer then you can make the change.

api/receiver/receive_results_test.go
```diff
@@ -7,18 +7,41 @@
...
 	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine/taskqueue"
 )
 
+type regexMatcher struct {
```
mdittmer: Can we get a comment on the rationale for implementing a custom matcher? I'm sure reading the code eventually reveals it, but it would be helpful to have it up front.

api/receiver/receive_results.go
Hexcles: The purpose of the `if` block here is to use different schemas for `gcsPath`. When we have multiple reports for a run, the path is one level deeper for easier debugging. I rejigged the code a bit; ptal if it's better now.

api/receiver/receive_results.go
Hexcles: Sorry forgot to remove the comment.

api/receiver/receive_results_test.go
```diff
@@ -22,6 +22,8 @@ import (
 	"google.golang.org/appengine/taskqueue"
 )
 
+// regexMatcher is a gomock.Matcher that verifies whether a string argument
```
mdittmer: nit: Rather than emphasizing _what_ regexMatcher is, let's have this comment focus on _why_ it's here. Give an example or pattern in this test file where a matcher that matches a regexp is the right verification tool for these tests.

PR #650, opened 2018-10-12
Add default flags in datastore

lukebjerring:
## Description
Supports us toggling a flag without the need for re-deploying the project, and allows us to toggle a flag for a whole environment instead of per-user.

## Review Information
Note that I've added a `Flag` entity for enabling `queryBuilder` in staging, so you should be able to see the query builder on the landing page, and ticked in `/flags`, in the auto-deployed env.

### DIFF
--- a/webapp/dynamic-components/wpt-env-flags.html
+++ b/webapp/dynamic-components/wpt-env-flags.html
@@ -1,0 +1,30 @@
+<!--
+Copyright 2018 The WPT Dashboard Project. All rights reserved.
+Use of this source code is governed by a BSD-style license that can be
+found in the LICENSE file.
+-->
+
+<!--
+`<wpt-env-flags>` is a component for checking default enviroment wpt.fyi
+  feature flags.
+-->
+<dom-module id="wpt-env-flags">
+  <script>
+    (() => {
+      // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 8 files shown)


webapp/dynamic-components/wpt-env-flags.html
```diff
@@ -0,0 +1,30 @@
...
+
+<!--
+`<wpt-env-flags>` is a component for checking default enviroment wpt.fyi
+  feature flags.
+-->
+<dom-module id="wpt-env-flags">
+  <script>
+    (() => {
+      // eslint-disable-next-line no-unused-vars
+      const DEFAULTS = JSON.parse({{ .Flags }});
```
Hexcles: Is `DEFAULTS` used anywhere else? If not, can we `range` over the `Flags` slice in Go template instead of doing it in JS, which seems a bit convoluted to me?

And even if we need to do it in JS, `html/template` supports directly embedding a struct or slice in a `<script>` context, and the templating engine will automatically JSON serialize the object:

```
Non-string values can be used in JavaScript contexts. If {{.}} is

struct{A,B string}{ "foo", "bar" }
in the escaped template

<script>var pair = {{.}};</script>
then the template output is

<script>var pair = {"A": "foo", "B": "bar"};</script>
```

lukebjerring: Done.

Issue #653, opened 2018-10-13
Navigation is broken


foolip:
Start at https://staging.wpt.fyi/results/?sha=d647a1bc74&product=safari%5Bstable%5D&product=safari%5Bexperimental%5D&diff with no flags enabled (discovered with query builder and diff filter enabled, but repros without)

Then, click 2dcontext. The URL changes to https://staging.wpt.fyi/results/2dcontext?sha=d647a1bc74&product=safari%5Bstable%5D&product=safari%5Bexperimental%5D&diff but the view does not, the top-level view is still shown!

Now, click 2dcontext again. Now you will end at the right view. From here, click the "wpt" above the table to navigate back to the top level. Here something very bizarre happens. The URL is https://staging.wpt.fyi/results/?sha=d647a1bc74&product=safari%5Bstable%5D&product=safari%5Bexperimental%5D&diff but the view is the contents of 2dcontext with "2dcontext/" pretended to each path.

Overall, it seems like the navigation is delayed by one step, plus some confusion of state.

foolip:
Screenshot of the final bizarre state here. I discovered this in Chrome for Android, not verified in a desktop browser.

![screenshot_20181013-124018](https://user-images.githubusercontent.com/498917/46904528-8a6c8300-cee5-11e8-8cdc-1326798a2524.png)

CYBAI:
I can reproduce this issue on Firefox Nightly 64.0a1.

CYBAI:
After doing more investigation, I found we'll trigger `click` event and do `submit` so the correct `this.path` is overridden when doing the `submit` (the `window.location` is still `https://staging.wpt.fyi/results/?sha=` instead of `https://staging.wpt.fyi/results/2dcontext?sha=`)

1. Update `path` correctly from `click` event via the `event.target`

https://github.com/web-platform-tests/wpt.fyi/blob/3dea9f3be299f2461151a6b60ab03a398bf5f659/webapp/components/self-navigator.html#L96-L103

2. The observer will trigger `onLocationUpdated` so it will trigger the `submit`

https://github.com/web-platform-tests/wpt.fyi/blob/3dea9f3be299f2461151a6b60ab03a398bf5f659/webapp/components/self-navigator.html#L67-L72

3. the `submit` is triggered 3 times and `path` is `/` in last 2 times

[...]

PR #657, opened 2018-10-15
Fix navigation

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/653

When handling a change in navigation state, we are force-submitting on _all_ changes, to ensure the builder always reflects the right state, and thus state is "submitted" as the current query. This is overkill for the case that only the path changes (no reload needed), so this PR checks that the _query_ is actually different, before triggering a data load (which was clobbering the path in the process).


lukebjerring:
Yes; I'll add an integration test in the near future.

PR #658, opened 2018-10-15
Spanner push run: Compare result counts

mdittmer:
## Description

This implements the initial check for pushing a run to Cloud Spanner: Does the report contain more distinct test results than Cloud Spanner? If `reportResultCount > spannerRowCount` for the given run ID, then (in a follow-up PR) the service will write all results to Cloud Spanner.

## Review Information

If you wish to run locally, make use of the flags defined in `api/spanner/service/main.go` to setup credentials and a port number. If you wish to use a staging instance, then lookup the HTTP basic auth password for the `_spanner` user in Datastore.

`PUT /api/spanner_push_run?run_id=[integral-datastore-id-for-run]`

Logging should show the result of your request.

### DIFF
--- a/api/spanner/api.go
+++ b/api/spanner/api.go
@@ -4,10 +4,60 @@
 
 package spanner
 
-type API struct {
+import (
+	"context"
+	"fmt"
+
+	"cloud.google.com/go/datastore"
+	"cloud.google.com/go/spanner"
+	"google.golang.org/api/option"
+)
+
+type API interface {
 	Authenticator
 
-	ProjectID          string
-	Database           string
-	GCPCredentialsFile *string
+	WithCredentialsFile(string) API
+	DatastoreConnect(context.Context) (*datastore.Client, error)
+	SpannerConnect(context.Context) (*spanner.Client, error)
+}
+
+type apiImpl struct {
+	Authenticator
+
+	projectID          string
+	instance           string
+	database           string
+	gcpCredentialsFile *string
+}
+
+func (a *apiImpl) DatastoreConnect(ctx context.Context) (*datastore.Client, error) {
+	if a.gcpCredentialsFile != nil {
+		return datastore.NewClient(ctx, a.projectID, option.WithCredentialsFile(*a.gcpCredentialsFile))
+	}
+
+	return datastore.NewClient(ctx, a.projectID)
+}
+
+func (a *apiImpl) SpannerConnect(ctx context.Context) (*spanner.Client, error) {
+	db := fmt.Sprintf("projects/%s/instances/%s/databases/%s", a.projectID, a.instance, a.database)
+	if a.gcpCredentialsFile != nil {
+		return spanner.NewClient(ctx, db, option.WithCredentialsFile(*a.gcpCredentialsFile))
+	}
+
+	return spanner.NewClient(ctx, db)
+}
+
+func (a *apiImpl) WithCredentialsFile(gcpCredentialsFile string) API {
+	a2 := *a
+	a2.gcpCredentialsFile = &gcpCredentialsFile
+	return &a2
+}
+
+func NewAPI(a Authenticator, projectID, instance, database string) API {
+	return &apiImpl{
+		Authenticator: a,
+		projectID:     projectID,
+		instance:      instance,
+		database:      database,
+	}
 }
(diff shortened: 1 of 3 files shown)


api/spanner/api.go
```diff
@@ -0,0 +1,69 @@
...
+func (a *apiImpl) SpannerConnect(ctx context.Context) (*spanner.Client, error) {
+	db := fmt.Sprintf("projects/%s/instances/%s/databases/%s", a.projectID, a.instance, a.database)
+	if a.gcpCredentialsFile != nil {
+		return spanner.NewClient(ctx, db, option.WithCredentialsFile(*a.gcpCredentialsFile))
+	}
+
+	return spanner.NewClient(ctx, db)
+}
+
+func (a *apiImpl) WithCredentialsFile(gcpCredentialsFile string) API {
```
lukebjerring: If you just pass by value, you won't need to make a copy.

api/spanner/push_run.go
```diff
@@ -49,3 +67,179 @@ func HandlePushRun(ctx context.Context, auth Authenticator, w http.ResponseWrite
...
+
+	shared.GetLogger(ctx).Infof("NOT IMPLEMENTED: Spanner push run would now push run if number of missing rows=%d > 0", n)
+}
+
+// loadRun loads shared.TestRun data from datastore, given an integral ID
+// (Datastore key).
+func loadRun(ctx context.Context, client *datastore.Client, id int64) (*shared.TestRun, error) {
+	logger := shared.GetLogger(ctx)
+
+	logger.Infof("Loading TestRun entity with integra key %d", id)
```
lukebjerring: integra?

api/spanner/push_run.go
```diff
@@ -49,3 +67,179 @@ func HandlePushRun(ctx context.Context, auth Authenticator, w http.ResponseWrite
...
+
+	logger.Infof("Read report for run ID=%d", run.ID)
+
+	return &report, nil
+}
+
+// countReportResults counts the number of meaningfully distinct test results
+// detailed in report.
+func countReportResults(ctx context.Context, report *metrics.TestResultsReport) int64 {
+	count := int64(0)
```
lukebjerring: Are we likely to exceed 2M entries?

api/spanner/push_run.go
```diff
@@ -49,3 +67,179 @@ func HandlePushRun(ctx context.Context, auth Authenticator, w http.ResponseWrite
...
+}
+
+// countSpannerResults counts the number of test results bound to the given
+// runID in Cloud Spanner.
+func countSpannerResults(ctx context.Context, client *spanner.Client, runID int64) (int64, error) {
+	params := map[string]interface{}{
+		"run_id": runID,
+	}
+	s := spanner.Statement{
+		SQL:    countStmt,
```
lukebjerring: I'd probably prefer `countStmt` declared locally in this method, to make the code more readable.

api/spanner/push_run.go
```diff
@@ -49,3 +67,179 @@ func HandlePushRun(ctx context.Context, auth Authenticator, w http.ResponseWrite
...
+		Params: params,
+	}
+
+	shared.GetLogger(ctx).Infof("Spanner query: \"%s\" with %v", countStmt, params)
+
+	itr := client.Single().WithTimestampBound(spanner.MaxStaleness(1*time.Minute)).Query(ctx, s)
+	defer itr.Stop()
```
lukebjerring: I'm surprised this is needed, given that we exhaust the iterator?

api/spanner/push_run.go
mdittmer: No, but Spanner will give us an `int64` against which to compare, so this keeps the types the same.

api/spanner/push_run.go
mdittmer: My thinking was that keeping SQL statements together at the top of the file may ease any future schema updates without missing a SQL string by mistake. WDYT of this rationale?

api/spanner/push_run.go
mdittmer: [Docs](https://godoc.org/cloud.google.com/go/spanner#RowIterator.Stop) seems to imply that this should be invoked unconditionally.

mdittmer:
@lukebjerring PTAL at comment responses. Not all comments addressed/resolved in last push.

api/spanner/push_run.go
lukebjerring: Plausible; sure.

PR #660, opened 2018-10-16
Cleanup staging on master

lukebjerring:
## Description
Might not work just yet; prompting @Hexcles for collaboratively getting this set up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -216,6 +216,9 @@ deploy_staging: gcloud webapp_deps package_service var-BRANCH_NAME var-APP_PATH 
 	rm -rf $(WPTD_PATH)revisions/service/wpt.fyi
 	rm -rf $(WPTD_PATH)api/spanner/service/wpt.fyi
 
+cleanup_staging_versions: gcloud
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -216,6 +216,9 @@ deploy_staging: gcloud webapp_deps package_service var-BRANCH_NAME var-APP_PATH
 	rm -rf $(WPTD_PATH)revisions/service/wpt.fyi
 	rm -rf $(WPTD_PATH)api/spanner/service/wpt.fyi
 
+cleanup_staging_versions: gcloud
+	$(WPTD_GO_PATH)/util/cleanup-versions.sh
```
lukebjerring: This script executes `gcloud versions delete` - does that need a `-q` flag to avoid the requirement of user interaction?

Hexcles: Indeed, I just pushed a change in `cleanup-versions.sh` to always delete stale versions without asking (by adding `--quiet`), as I think it should be pretty safe to do.

PR #661, opened 2018-10-16
Results processor: Invoke push-to-spanner API

mdittmer:


results-processor/main.py
```diff
@@ -202,13 +202,23 @@ def task_handler():
...
     test_run_id = wptreport.create_test_run(report, labels, uploader, secret,
                                             results_gcs_path,
                                             raw_results_gcs_path)
     assert test_run_id
 
+    # Authenticate as "_spanner" for push-to-spanner API.
+    secret = ds.get(ds.key('Uploader', '_spanner'))['Password']
+    response = requests.post(
+        config.project_baseurl() + '/api/spanner_push_run',
+        auth=('_processor', secret),
```
Hexcles: Username should be `_spanner`.

results-processor/main.py
```diff
@@ -202,13 +202,23 @@ def task_handler():
...
+        data=json.dumps({'run_id': test_run_id}})
```
Hexcles: Use `json={'run_id': test_run_id}`. It's shorter, and sets the `Content-Type` correctly.

results-processor/main.py
```diff
@@ -202,13 +202,23 @@ def task_handler():
...
+    response.raise_for_status()
+
+
```
Hexcles: One empty line is enough.

results-processor/main.py
```diff
@@ -202,13 +202,23 @@ def task_handler():
...
                                             raw_results_gcs_path)
     assert test_run_id
 
+    # Authenticate as "_spanner" for push-to-spanner API.
+    secret = ds.get(ds.key('Uploader', '_spanner'))['Password']
+    response = requests.post(
+        config.project_baseurl() + '/api/spanner_push_run',
+        auth=('_processor', secret),
+        data=json.dumps({'run_id': test_run_id}})
+    response.raise_for_status()
```
Hexcles: We don't want to raise here (at least for now). We want to return 201 even if this fails; otherwise TaskQueue will retry the task and end up creating duplicate runs in Datastore.

`app.logger.info` any error here.

results-processor/main.py
```diff
@@ -202,13 +203,22 @@ def task_handler():
...
                                             raw_results_gcs_path)
     assert test_run_id
 
+    # Authenticate as "_spanner" for push-to-spanner API.
+    secret = ds.get(ds.key('Uploader', '_spanner'))['Password']
+    response = requests.put(
+        '%s/api/spanner_push_run?run_id=%d' % (config.project_baseurl(), test_run_id)
+        auth=('_spanner', secret))
+    if not response.ok:
+        app.logger.info('ERROR: Bad status code from push-to-spanner API: %d' % response.status_code)
```
Hexcles: Nit: `app.logger.error`, and then you don't need "ERROR: " in the message.

PR #664, opened 2018-10-16
Implement spanner push run

mdittmer:
- Don't bother if numResults(report) <= numResults(spanner)
- Use 5 retries around each major step
- Push up to 1000 batches of 1000 mutations concurrently
- Push tests, checking for (sub)test name hash collisions first
  - Then push interleaved data in three stages: `Runs` row, `RunResults` rows, `RunResultTests` rows
- Log warning on retryable failure
- Log error on push-breaking failure

## Review Information

If you wish to run locally, make use of the flags defined in `api/spanner/service/main.go` to setup credentials and a port number. If you wish to use a staging instance, then lookup the HTTP basic auth password for the `_spanner` user in Datastore.

`PUT /api/spanner_push_run?run_id=[integral-datastore-id-for-run]`

Logging on the server should show the result of your request.

### Runs already pushed to wptdashboard-staging (can be used to test already-in-spanner case)

5428053320663040


api/spanner/push_run.go
```diff
@@ -239,7 +559,41 @@ func numRowsToUpload(ctx context.Context, client *spanner.Client, runID int64, r
...
-	shared.GetLogger(ctx).Infof("Run %d contains %d rows (according to GCS); %d found in Spanner", runID, totalRows, existingRows)
+	shared.GetLogger(ctx).Infof("Run %d contains %d results (according to GCS); %d found in Spanner", runID, totalRows, existingRows)
 
 	return totalRows - existingRows, nil
 }
+
+// computeTestKey computes a stable int64 ID for a test+(optional)subtest pair.
+func computeTestKey(test string, subtest *string) TestKey {
+	key := TestKey{
+		TestID: int64(farm.Fingerprint64([]byte(test))),
```
Hexcles: When reading the schema doc, I didn't realize this was how the IDs would be generated. I thought it was some semi-random unique ID automatically generated by Spanner.

My question is: what do we about hash collision? I know 2^64 is a large enough space that hash collision is highly unlikely, but IIUC, if this ever happens, the current implementation can't really handle it gracefully.

I'm in particular worried about subtests. Given the nature of WPT, this surface is wide open for exploitation. People can just create a new test file which has two subtests with the same farm64 hash.

api/spanner/push_run.go
```diff
@@ -73,38 +278,153 @@ func pushRun(ctx context.Context, api API, id int64) {
...
+		if end != len(rows) {
+			wg.Add(1)
+			s.Acquire(ctx, 1)
+			log.Infof("Writing small batch for %d-row run %d: [%d,%d)", len(rows), run.ID, end-batchSize, len(rows))
+			go writeBatch(s, &wg, rows[0:], end-batchSize, len(rows))
+			log.Infof("Wrote small batch for %d-row run %d: [%d,%d)", len(rows), run.ID, end-batchSize, len(rows))
+		}
+		return &wg
+	}
+
```
Hexcles: Shall we use transactions here? Or is it too much data?

api/spanner/push_run.go
```diff
@@ -6,25 +6,37 @@ package spanner
...
 const (
-	countStmt = "SELECT COUNT(*) FROM results WHERE run_id = @run_id"
+	numConcurrentBatches    = 1000
+	batchSize               = 1000
+	testsTableName          = "Tests"
+	resultsTableName        = "Results"
+	runsTableName           = "Runs"
+	runResultsTableName     = "RunResults"
+	runResultTestsTableName = "RunResultTests"
+	countStmt               = "SELECT COUNT(*) FROM RunResultTests WHERE RunID = @RunID"
```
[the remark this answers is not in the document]
mdittmer: Using a string like this is the appropriate way to issue _queries_ via the golang Cloud Spanner API. All other interactions in this change are actually _writes_ and use different API patterns.

api/spanner/push_run.go
mdittmer: A currently have no solution for hash collisions. I have verified (though this is the opposite of a future guarantee) that we _currently_ have no hash collisions in the most recent 1k runs. We could implement read-then-write on all `Tests` table rows, but that seems prohibitively expensive: there are O(1million) per test run.

W.R.T. duplicate test+subtest strings, that's what [this line](https://github.com/web-platform-tests/wpt.fyi/blob/dfc6e954faa6cc4067cffb73687c19a18daf9f6b/api/spanner/push_run.go#L534) is for. All it does is log, but there is not much else we can do because subtests do not complete in a deterministic order between runs. I think I mentioned this to @foolip and he said that we should monitor logging the inconsistency and work to fix tests that come up.

@Hexcles do you think there are additional measures we should take in this PR? One potential compromise to find genuine hash collisions is to perform a check when writing them into the upcoming in-memory cache. That will be faster, but fail to detect collisions involving really old runs that may never get cached.

api/spanner/push_run.go
mdittmer: IIRC this is way too much data for a transaction.

api/spanner/service/main.go
```diff
@@ -21,7 +21,7 @@ const (
 	// maintained for same-name databases across projects. If schema changes are
 	// needed, usually a new database is created, then code containing assumptions
 	// about schema can be updated alongside this constant in a single change.
-	spannerDatabase = "results-apep"
+	spannerDatabase = "results-caspilly"
```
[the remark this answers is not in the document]
mdittmer: Done.

api/spanner/push_run.go
Hexcles: I'm less concerned about actually duplicate strings, but more about different strings with the same hash (collision). I don't think the log you pointed to will catch collisions?

To continue the discussion regarding the hash algorithm, IIUC, FarmHash is optimized for the hash-table use case, where the tradeoff is slightly tilted towards performance. If we want to use the hash as ID, perhaps we should use something more cryptographically strong? SHA256 is probably better, and no external dependencies required, but I honestly don't know how well it does when we only take the first 64 bits...

mdittmer:
The other option would be to use a byte array type in both go and spanner
and keep all 256 bits. The larger size may limit how many runs we can keep
cached and hurt performance on both systems.

IMHO at least detecting collisions that are likely to matter with checks in
the cache is a reasonable tradeoff.

@Hexcles do you want a more detailed design of the process to be used in
the cache to detect collisions? What collisions would and wouldn't be
caught, what would actually end up in the database in those cases, etc.?

foolip:
If collisions are unlikely but would result in corrupt (confused) data if it occurs, can we detect when it's about to happen and fail?

Or, give each thing a unique ID and use hashes as the way to narrow down a search for an existing thing, more like a hash map?

mdittmer:
Regarding the latter suggestion, incrementing IDs is strongly discouraged in Cloud Spanner, but we might be safe if we make the ID prefix be the hash value to distribute the data.

If we want to detect collisions before committing them, we need to either:

1. Read-then-write every `Tests` row, or, more likely,
1. Read all test rows into memory and check new rows' strings before writing them (and fail the whole push if a collision is detected).

@foolip and @Hexcles do you prefer the abundance of caution `{hash, counter-that-is-virtually-always-0}`, or deferring the problem by failing if we ever find a collision?

foolip:
> deferring the problem by failing if we ever find a collision

Is it possible for us to detect that a collision has happened even after the fact unless we take some precautions? If we can notice and it won't lead to data loss/corruption, then just monitoring for it seems fine.

mdittmer:
Not entirely sure what that would look like. We could keep an auxiliary table of collisions to know about them, but if we're going to consult the table on every query ("just in case" the query involves tests with collisions) I'd say we should just implement the cautious approach.

It depends on what's meant by loss/corruption. The consequence of a collision is that we begin treating two different tests as the same test in terms of counting pass rates and displaying tests. Essentially one of the the two tests would "disappear" from the dashboard, and some results for the disappeared test may be reported as the result of a different test. I guess the extent of the problem depends on the use case. For aggregate data, a collision amounts to a rounding error. For an API owner that doesn't understand why one of the tests that's clearly in WPT never appears in the folder where it belongs, the issue is more serious.

foolip:
Each individual mis-aliasing, if we can call it that, is a pretty minor issue once understood. My concern is whether we would discover that it's happened and if we'll be able to recover with anything short of a full rebuild of the Spanner data. The person looking at the bug when it happens might not have read this issue :)

Hexcles:
Talked with Mark offline, mostly regarding two approaches (roughly in the same vein):

1. Insert (not update) into the table, and catch the error when there is duplicate primary key. Then check the existing row for collision. No transaction is needed because we never modify rows (only insert). It is expected that in the long run most insertions will fail because of duplication, so we almost always need 2 queries per row.
2. Use some native Spanner read-then-write (CAS) or transaction primitive.

I proposed the former because of my hunch that atomic transactions would have larger overhead.

Regardless, I think either is fine. We will sacrifice the performance quite a bit, but as long as it is on the same magnitude then I think it's acceptable.

mdittmer:
Updated to feature similar batching strategy for insert(-then-check) strategy for `Tests` table before writing any other data from run.

api/spanner/push_run.go
```diff
@@ -73,38 +304,294 @@ func pushRun(ctx context.Context, api API, id int64) {
...
+					if err != nil {
+						return err
+					}
+					var subtestName spanner.NullString
+					err = row.ColumnByName(subtestNameColumnName, &subtestName)
+					if err != nil {
+						return err
+					}
+					if t.TestName != testName || t.SubtestName != subtestName {
+						return fmt.Errorf(`Hash collision: Test identifier <%d, %v> mapped to different test+subtest names: "%s".%v != "%s".%v`, t.TestID, t.SubtestID, t.TestName, t.SubtestName, testName, subtestName)
```
mdittmer: @lukebjerring do you think we should accumulate these collisions, remove them from the result set, then push the nearly-complete run, rather than giving up entirely?

Hexcles: I also concur with Luke here. It's fine to lose a few lines, and we still want the otherwise nearly complete run. We just need to log the error here. Basically, I agree with you this is unlikely to happen, so we shouldn't worry about it prematurely, but we do want to know when/if it happens.

mdittmer:
Code now removes tests that constitute hash collisions and continues with the push. @Hexcles PTAL.

Hexcles:
Tests would be great, but I think it'd be hard and tedious to write tests
for this (first we need to test the generation of mutations from a report,
then we might need to generate a mock of Spanner client to test the
scheduling of the mutations), and I'm not sure if it's worth it for the
reason I said earlier.

mdittmer:
I am comfortable closing (and possibly later reopening) this PR. The
current thinking in this space is that we should focus on efficient qeuries
against fresh runs using different infrastructure. The Spanner piece, as
currently proposed would be launched after.

Issue #639, opened 2018-10-11
Order of products isn't always preserved when using comma-separated ?products=a,b,c


foolip:
https://wpt.fyi/results/css/CSS2/selectors/first-letter-punctuation-280.xht?products=chrome%5Bstable%5D,edge%5Bstable%5D,firefox%5Bstable%5D,firefox%5Bexperimental%5D,safari%5Bstable%5D&sha=b0d2d2da21

In the URL, I've put Firefox stable before experimental, but it shows up  in the other order:
<img width="808" alt="screen shot 2018-10-11 at 10 10 14 pm" src="https://user-images.githubusercontent.com/498917/46831003-74a27500-cda2-11e8-8e17-0acf7b7425c3.png">

Note also that https://staging.wpt.fyi/results/css/CSS2/selectors/first-letter-punctuation-280.xht?products=chrome%5Bstable%5D,edge%5Bstable%5D,firefox%5Bstable%5D,firefox%5Bexperimental%5D,safari%5Bstable%5D&sha=b0d2d2da21 is missing Firefox experimental entirely, so that's a regression.

lukebjerring:
Well, the cause for the latter is actually that we don't support `[label]` syntax for the `products` param (but should). Order is preserved since #408.

foolip:
Oh, I didn't get that, thanks Luke!

lukebjerring:
The provided link no longer exhibits a problem - I believe this is fixed.

PR #669, opened 2018-10-17
Handle multiple labels for products

lukebjerring:
## Description
Step towards resolving https://github.com/web-platform-tests/wpt.fyi/issues/639

Hackily avoids having to properly parse multiple `products` values by replacing the nested commas temporarily.

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -383,8 +383,8 @@ func ParseProductsParam(r *http.Request) (ProductSpecs, error) {
 	repeatedParam := r.URL.Query()["product"]
 	pluralParam := r.URL.Query().Get("products")
 	// Replace nested ',' in the label part with a placeholder
-	nestedCommas := regexp.MustCompile("(\\[[^\\]]*),")
-	const comma = "%COMMA%"
+	nestedCommas := regexp.MustCompile(`(\[[^\]]*),`)
+	const comma = `%COMMA%`
 	for nestedCommas.MatchString(pluralParam) {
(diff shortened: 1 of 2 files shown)


shared/params.go
```diff
@@ -380,10 +380,22 @@ func ParseProductParam(r *http.Request) (product *Product, err error) {
 // It parses the 'products' parameter, split on commas, and also checks for the (repeatable)
 // 'product' params.
 func ParseProductsParam(r *http.Request) (ProductSpecs, error) {
-	productParams := ParseRepeatedParam(r, "product", "products")
+	repeatedParam := r.URL.Query()["product"]
+	pluralParam := r.URL.Query().Get("products")
+	// Replace nested ',' in the label part with a placeholder
+	nestedCommas := regexp.MustCompile("(\\[[^\\]]*),")
```
Hexcles: Use

    `pattern`

so that you don't need to escape all the backslashes.

shared/params.go
```diff
@@ -380,10 +380,22 @@ func ParseProductParam(r *http.Request) (product *Product, err error) {
...
+	const comma = "%COMMA%"
```
Hexcles: GitHub isn't happy with this string. It's trying to highlight it as a format string. I don't know whether it's just a code highlighting bug in GitHub. We could just use backticks instead of quotes.

shared/params.go

shared/params.go

Issue #671, opened 2018-10-18
https://wpt.fyi/test-runs?product=chrome-68 is 404


foolip:
https://wpt.fyi/api/runs?product=chrome-68, however, does return something.

foolip:
What I see in the console is that https://wpt.fyi/api/runs?from=2018-09-18T00%3A00%3A00Z&product=chrome-68 failed to load with 404. 

https://wpt.fyi/api/runs?product=chrome-68&max-count=500 shows that the most recent run is from 2018-09-04, so the data parameter is probably the problem.

foolip:
https://staging.wpt.fyi/test-runs?product=chrome-68 still doesn't work for me.

lukebjerring:
https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/443267938 failed to deploy is all :)

lukebjerring:
Oh; just realised this is actually a separate bug. #673 only fixed `/results?product=chrome-68`

lukebjerring:
In the same way that typing products manually on /results becomes non-aligned, /runs falls back to a default of max-count 1 when you use any non default query.

i.e. add max-count=50

foolip:
What's the reason for that fallback, would something weird happen if we use the same max-count unconditionally?

lukebjerring:
No; I'll send another PR.

Issue #587, opened 2018-09-26
https://wpt.fyi/api/runs?max-count=1000&label=stable returns 500 Server Error


foolip:
It says "API error 1 (datastore_v3: BAD_REQUEST): cannot get more than 1000 keys in a single call".

This URL is used in [ecosystem-infra-stats](https://github.com/foolip/ecosystem-infra-stats/blob/bce9e3cd153eb2a7427abec1e05bc5878f093499/wpt-dashboard-stats.py#L10) with reference to https://github.com/web-platform-tests/wpt.fyi/issues/3.

The cutoff point between working and not working seems to be 302/303:
https://wpt.fyi/api/runs?max-count=302&label=stable (works)
https://wpt.fyi/api/runs?max-count=303&label=stable (500)

That's a surprising cutoff point, not a divisor of 1000, so perhaps it won't be stable over time.

What I really want is an API to enumerate *all runs ever*, or possibly all runs after a specified date.

foolip:
Note that with Taskcluster runs in prod, this problem will get worse. 300 runs will be roughly 3 weeks of activity in wpt, not enough to compute stats for the past quarter as we'll want to do.

lukebjerring:
This is essentially a duplicate of #569

foolip:
I'll open a feature request instead.

Issue #570, opened 2018-09-19
API: Results ordering limits utility of `from` parameter


jugglinmike:
When the data returned by `/api/runs` is truncated by `max-count`, it prioritizes more recent data. Because the server sets hard limits on the number of results it will return, this behavior means older results can't be retrieved:

    $ curl 'https://wpt.fyi/api/runs?from=2018-07-01T00:00:00-04:00&max-count=5' | python -c 'import json, sys; print "\n".join([x["created_at"] for x in json.loads(sys.stdin.read())])'
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
    100 13300  100 13300    0     0  25829      0 --:--:-- --:--:-- --:--:-- 25875
    2018-09-19T19:15:43.418629Z
    2018-09-19T18:37:49.814379Z
    2018-09-19T09:14:49.484807Z
    2018-09-19T05:19:12.087341Z
    2018-09-19T06:47:34.644255Z
    2018-09-19T06:08:01.782948Z
    2018-09-18T06:24:24.505601Z
    2018-09-17T03:15:25.279373Z
    2018-09-16T03:49:12.577577Z
    2018-09-15T06:49:07.095779Z
    2018-09-19T17:34:12.912039Z
    2018-09-19T03:21:12.631356Z
    2018-09-18T22:13:45.732254Z
    2018-09-18T14:34:40.634335Z
    2018-09-18T12:41:28.159062Z
    2018-09-19T09:34:15.331304Z
    2018-09-18T09:53:21.336091Z
    2018-09-17T04:56:38.038864Z
    2018-09-16T21:49:29.386713Z
    2018-09-16T09:32:36.593057Z

It does look like the parameter is correctly filtering results. That's not evident in the above example, but it can be observed by setting the parameter to a recent date:

    curl 'https://wpt.fyi/api/runs?from=2018-09-18T00:00:00-04:00&max-count=5' | python -c 'import json, sys; print "\n".join([x["created_at"] for x in json.loads(sys.stdin.read())])'                                                                                                                                                                                                            
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed
    100  8103  100  8103    0     0  15764      0 --:--:-- --:--:-- --:--:-- 15764
    2018-09-19T19:15:43.418629Z
    2018-09-19T18:37:49.814379Z
[...]

PR #673, opened 2018-10-18
Iterate keys instead of fetching

### LINKED ISSUE #569: API: Default value of `max-count` is too large
[There's no documentation for a default value of the `max-count` parameter](https://github.com/web-platform-tests/wpt.fyi/tree/03411bc097a6759fe0d44235ffc4a809295b44d8/api#apiruns), but it looks like the current implementation selects a value which is too large:

    $ curl 'https://wpt.fyi/api/runs?from=2018-07-01T00:00:00-04:00'
    API error 1 (datastore_v3: BAD_REQUEST): cannot get more than 1000 keys in a single call
    $ curl 'https://wpt.fyi/api/runs?from=2018-07-01T00:00:00-04:00&max-count=1'
    [{"id":5640432071475200,  (...etc.)

lukebjerring:
## Description
Re-fixes #https://github.com/web-platform-tests/wpt.fyi/issues/587 without fixed-limiting the key count.
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/671

Fixes #569 and might help #570 too...

## Review Information
/api/runs?product=chrome-68

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -79,16 +79,12 @@ func LoadTestRunKeys(
 				break
 			} else if err != nil {
 				return result, err
+			} else if (limit != nil && *limit <= len(keys)) || len(keys) >= MaxCountMaxValue {
+				break
+			} else if prefiltered != nil && !(*prefiltered).Contains(key.String()) {
+				continue
 			}
-			if (limit == nil || *limit > len(keys)) && len(keys) < MaxCountMaxValue {
-				if prefiltered == nil || (*prefiltered).Contains(key.String()) {
-					keys = append(keys, key)
-				}
-			}
-		}
-
-		if limit != nil && len(keys) > *limit {
-			keys = keys[:*limit]
+			keys = append(keys, key)
 		}
 		result = append(result, keys...)
 	}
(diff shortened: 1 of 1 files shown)


shared/datastore.go
```diff
@@ -73,13 +71,16 @@ func LoadTestRunKeys(
...
-			if limit == nil || *limit > len(keys) && len(keys) < MaxCountMaxValue {
+		iter := query.KeysOnly().Run(ctx)
+		for {
+			key, err := iter.Next(nil)
+			if err == datastore.Done {
+				break
+			} else if err != nil {
+				return result, err
+			}
+			if (limit == nil || *limit > len(keys)) && len(keys) < MaxCountMaxValue {
```
Hexcles: We should add an `else` branch for this `if` and break the loop there to prevent unnecessary full table scan.

lukebjerring: Done (and other simplification).

shared/datastore.go
```diff
@@ -79,16 +79,12 @@ func LoadTestRunKeys(
 				break
 			} else if err != nil {
 				return result, err
+			} else if (limit != nil && *limit <= len(keys)) || len(keys) >= MaxCountMaxValue {
```
Hexcles: Nit: can we make the two inequalities of the same direction?

Issue #670, opened 2018-10-17
Bikeshedding: rename/redirect https://wpt.fyi/test-runs to https://wpt.fyi/runs


foolip:
https://wpt.fyi/runs would be shorter and more consistent with https://wpt.fyi/results/ which does not have a "test-" prefix.

Real reason: I type out this URL by hand sometimes.

lukebjerring:
I approve this sentiment, feel free to send the PR.

PR #677, opened 2018-10-18
Only set from if the query is the default

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/671 (and other bugs with the same cause).

Drive by fix for https://github.com/web-platform-tests/wpt.fyi/issues/670 too.

## Review Information
Visit /test-runs?product=chrome-68
Observe a result

### DIFF
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -40,7 +40,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/interop/{path:.*}", "interop", interopHandler)
 
 	// List of all test runs, by SHA[0:10]
-	shared.AddRoute("/test-runs", "test-runs", testRunsHandler)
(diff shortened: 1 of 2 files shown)


webapp/routes.go
```diff
@@ -40,7 +40,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/interop/{path:.*}", "interop", interopHandler)
 
 	// List of all test runs, by SHA[0:10]
-	shared.AddRoute("/test-runs", "test-runs", testRunsHandler)
+	shared.AddRoute("/runs", "test-runs", testRunsHandler)
```
foolip: Thanks :) Can you point to https://github.com/web-platform-tests/wpt.fyi/issues/670 in the commit message?

lukebjerring: Done.

Issue #672, opened 2018-10-18
The /results/?product=browser@sha URL doesn't accept full SHAs


foolip:
This works:
https://wpt.fyi/results/?product=chrome@a0f75fcad8

This does not:
https://wpt.fyi/results/?product=chrome@a0f75fcad81bee1dbe35d30c94aed9b7630d9dda

Discovered as I was trying to create a URL to compare Chrome 68 and 69.

foolip:
For the record, here is that comparison:
https://staging.wpt.fyi/results/?product=chrome-68@a0f75fcad8&product=chrome-69@7c50c21608&diff

And 67&rarr;68: https://staging.wpt.fyi/results/?product=chrome-67@68318aca8a&product=chrome-68@0589266b84&diff

PR #684, opened 2018-10-22
Handle full SHAs in product specs

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/672 by re-using Full SHA parsing on the product SHA part.

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -202,7 +202,7 @@ func ParseSHAParam(r *http.Request) (runSHA string, err error) {
 }
 
 // ParseSHA validates the given 'sha' value, cropping it to 10 chars.
-// It returns "latest" by default. (and in error cases).
+// It returns "latest" by default (and in error cases).
 func ParseSHA(sha string) (runSHA string, err error) {
 	sha, err = ParseSHAFull(sha)
 	if err != nil || !SHARegex.MatchString(sha) {
--- a/shared/params.go
+++ b/shared/params.go
@@ -215,7 +215,6 @@ func ParseSHA(sha string) (runSHA string, err error) {
 // It returns "latest" by default (and in error cases).
 func ParseSHAParamFull(r *http.Request) (runSHA string, err error) {
 	// Get the SHA for the run being loaded (the first part of the path.)
-	runSHA = "latest"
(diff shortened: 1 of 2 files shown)


shared/params.go
```diff
@@ -201,17 +201,29 @@ func ParseSHAParam(r *http.Request) (runSHA string, err error) {
...
+		return sha, err
+	}
+	return sha[:10], nil
+}
+
 // ParseSHAParamFull parses and validates the 'sha' param for the request.
 // It returns "latest" by default (and in error cases).
 func ParseSHAParamFull(r *http.Request) (runSHA string, err error) {
 	// Get the SHA for the run being loaded (the first part of the path.)
 	runSHA = "latest"
```
CYBAI: This variable seems unused now in `ParseSHAParamFull`?

shared/params.go
```diff
@@ -201,17 +201,28 @@ func ParseSHAParam(r *http.Request) (runSHA string, err error) {
 	return sha[:10], nil
 }
 
+// ParseSHA validates the given 'sha' value, cropping it to 10 chars.
+// It returns "latest" by default. (and in error cases).
```
foolip:
```suggestion
// It returns "latest" by default (and in error cases).
```

shared/params_test.go
```diff
@@ -344,6 +337,18 @@ func TestParseProductSpec(t *testing.T) {
...
 }
 
+func TestParseProductSpec_FullSHA(t *testing.T) {
+	sha := "0123456789aaaaabbbbbcccccdddddeeeeefffff"
+	r := httptest.NewRequest("GET", "http://wpt.fyi/?product=chrome@"+sha, nil)
+	filters, err := ParseTestRunFilterParams(r)
+	assert.Nil(t, err)
+	products := filters.GetProductsOrDefault()
+	assert.Len(t, products, 1)
+	if len(products) > 0 {
```
foolip: Isn't this condition redundant given the above assertion? Or does the test keep running on failed asserts?

lukebjerring: Unfortunately no, Golang keeps going on a failed assert.

foolip: Oh, weird :)

PR #683, opened 2018-10-22
Add diff to query builder

lukebjerring:
## Description
Adds a tickbox for showing a diff to the builder.

This required an unfortunate amount of refactoring :(

## Review Information
Select 2 products in the query builder, tick 'diff', submit.

PR #685, opened 2018-10-23
Add source dropdown to the query builder

lukebjerring:
## Description
Continues from #683, more work on https://github.com/web-platform-tests/wpt.fyi/issues/535, 
Adding a drop-down for 'Source' (taskcluster, buildbot, msedge).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,10 @@ matrix:
         - DEPLOY_STAGING=true
         - MAKE_TEST_TARGET=go_large_test
         - MAKE_TEST_FLAGS="STAGING=true"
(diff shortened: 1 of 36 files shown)


foolip:
@lukebjerring can you rebase to resolve conflicts and only leave the relevant commits? I'll review just the "Add source dropdown" commit, but am not clean if "Fix inheritance order" is new in this PR or not.

foolip:
Trying it out manually I can confirm it work.

Request for follow-up: for both channels and sources, gray out the ones for each product that we know cannot match anything. As it is people have to try all sorts of combinations to discover what data we have and not.

Issue #687, opened 2018-10-23
Query builder should allow selecting sha(s)


jgraham:
For https://github.com/web-platform-tests/wpt/commit/fbc56f388ac5d3a9cfcdfc161917ba0cfa2280f5 we wanted to compare the results against the pirior revision of the wpt repo, to ensure that updating the OS didn't unexpectedly change the test results. Therefore it would be good if the query builder would allow selecting specific sha1s for each selected browser.

lukebjerring:
#689 covered a SHA for the whole run (now in staging.wpt.fyi), but we'll need to follow up with having a single SHA that's per-product instead, which is collapsed + expanded the same way shared channel labels are.

foolip:
@Hexcles @stephenmcgruer this came up in discussion on #testing today again. If you construct the URL https://wpt.fyi/results/?product=chrome%404a0bac098f&product=chrome%405d6a919353 by hand, there's no way to edit it in the UI.

Aside: do we think it's weird that all the stuff you can use to get to a specific run is in a param called `product`?

PR #689, opened 2018-10-24
Add field for a SHA, with autocomplete

lukebjerring:
## Description
Starting towards #687 (for the all-the-same SHA case)

Adds a WPT flag for `queryBuilderSHA` which supports fetching the `/api/shas` endpoint for the current query, and populating the autocomplete of the SHA field using possibilities.

## Review Information
Visit `/flags`, enable both `queryBuilder` and `queryBuilderSHA`, type stuff.

### DIFF
--- a/webapp/components/test-runs-query-builder.html
+++ b/webapp/components/test-runs-query-builder.html
@@ -176,20 +176,23 @@ found in the LICENSE file.
         this.set(`products.${i}`, product);
       }
 
+      // Respond to query changes by computing a new shas URL.
       computeSHAsURL(query) {
         const url = new URL(`/api/shas${query || ''}`, window.location);
         url.searchParams.delete('sha');
         return url;
       }
 
+      // Respond to shas URL changing by fetching the shas
       shasURLUpdated(url) {
         fetch(url).then(r => r.json()).then(s => {
           this.shas = s;
         });
       }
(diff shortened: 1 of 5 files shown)


webapp/components/test-runs-query-builder.html
```diff
@@ -148,6 +175,52 @@ <h3>
...
+          shas = shas.filter(s => s.startsWith(sha));
+        }
+        shas = shas.slice(0, 10);
+        // Check actually different from current.
+        const curr = this.shasAutocomplete || [];
+        if (curr.length === shas.length) {
+          let same = true;
+          for (let i = 0; i < shas.length && same; i++) {
+            if (i >= curr.length || shas[i] !== curr[i]) {
+              same = false;
```
Hexcles: A few suggestions about this loop:

1. Why do we need to check `i >= curr.length`? The whole loop only runs if `curr` and `shas` have the same length.
2. We can simply break here, instead of checking `same` every time.
3. Once 1 & 2 are done, then we can simplify the 3-clause for into a for...of (which, unlike for...in, is guaranteed to iterate in order).

lukebjerring: Updated to match the versions autocomplete logic I simplified earlier today.

PR #691, opened 2018-10-25
Add /api/versions endpoint

lukebjerring:
## Description
Lists the versions that we have for a given product. Fairly sure `Distinct` means the sheer quantity of runs (>1000) won't be a problem.

Relates to being consistent with the autocomplete work in #689 

## Review Information
Visit `/api/versions?product=chrome-70`

### DIFF
--- a/api/versions.go
+++ b/api/versions.go
@@ -27,15 +27,14 @@ func apiVersionsHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := shared.NewAppEngineContext(r)
 	query := datastore.NewQuery("TestRun").Filter("BrowserName =", product.BrowserName)
+	distinctQuery := query.Project("BrowserVersion").Distinct()
 	var queries []*datastore.Query
 	if product.BrowserVersion == "" {
-		queries = []*datastore.Query{query}
+		queries = []*datastore.Query{distinctQuery}
 	} else {
 		queries = []*datastore.Query{
 			query.Filter("BrowserVersion =", product.BrowserVersion).Limit(1),
-			shared.VersionPrefix(query, "BrowserVersion", product.BrowserVersion, false /*desc*/).
-				Project("BrowserVersion").
-				Distinct(),
+			shared.VersionPrefix(distinctQuery, "BrowserVersion", product.BrowserVersion, false /*desc*/),
 		}
 	}
 
--- a/api/versions.go
+++ b/api/versions.go
@@ -1,4 +1,4 @@
-// Copyright 2017 The WPT Dashboard Project. All rights reserved.
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
@@ -27,13 +27,16 @@ func apiVersionsHandler(w http.ResponseWriter, r *http.Request) {
 
 	ctx := shared.NewAppEngineContext(r)
 	query := datastore.NewQuery("TestRun").Filter("BrowserName =", product.BrowserName)
-	queries := []*datastore.Query{query}
-	if product.BrowserVersion != "" {
-		queries = make([]*datastore.Query, 2)
-		queries[0] = query.Filter("BrowserVersion =", product.BrowserVersion).Limit(1)
-		queries[1] = shared.VersionPrefix(query, "BrowserVersion", product.BrowserVersion, false).
-			Project("BrowserVersion").
-			Distinct()
+	var queries []*datastore.Query
+	if product.BrowserVersion == "" {
+		queries = []*datastore.Query{query}
+	} else {
+		queries = []*datastore.Query{
+			query.Filter("BrowserVersion =", product.BrowserVersion).Limit(1),
+			shared.VersionPrefix(query, "BrowserVersion", product.BrowserVersion, false /*desc*/).
+				Project("BrowserVersion").
+				Distinct(),
+		}
 	}
 
 	var runs shared.TestRuns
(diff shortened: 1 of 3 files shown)


api/versions.go
```diff
@@ -0,0 +1,67 @@
...
+		return
+	}
+
+	ctx := shared.NewAppEngineContext(r)
+	query := datastore.NewQuery("TestRun").Filter("BrowserName =", product.BrowserName)
+	queries := []*datastore.Query{query}
+	if product.BrowserVersion != "" {
+		queries = make([]*datastore.Query, 2)
+		queries[0] = query.Filter("BrowserVersion =", product.BrowserVersion).Limit(1)
+		queries[1] = shared.VersionPrefix(query, "BrowserVersion", product.BrowserVersion, false).
```
mdittmer: IIUC, `len(queries) == 1` before this line runs. Should this be `queries = append(queries, [...])`?

mdittmer: nit: Comment reminding meaning of `false` here would make this more readable.

mdittmer: I think `VersionPrefix` is inclusive (`VersionPrefix(foo)` over `["foo", "foobar", "bar"]` yields `["foo", "foobar"]`). Doesn't the inclusion of `"foo"` make `queries[0]` redundant?

api/versions.go
```diff
@@ -0,0 +1,67 @@
...
+			Project("BrowserVersion").
+			Distinct()
+	}
+
+	var runs shared.TestRuns
```
mdittmer: This will initialize to `nil`. Is that safe for the first `append()`, or should we have `runs := make(shared.TestRuns, 0)`?

api/versions.go
```diff
@@ -0,0 +1,67 @@
...
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte("[]"))
+		return
+	}
+
+	versions := make([]string, len(runs))
+	for i := range runs {
+		versions[i] = runs[i].BrowserVersion
+	}
+	sort.Sort(sort.Reverse(sort.StringSlice(versions)))
```
mdittmer: nit: This will put `10.0` before `2.0`, but consistently modeling versions is out of scope here. Maybe add a `TODO`?

api/versions.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if len(runs) < 1 {
+		w.WriteHeader(http.StatusNotFound)
+		w.Write([]byte("[]"))
+		return
+	}
+
+	versions := make([]string, len(runs))
+	for i := range runs {
+		versions[i] = runs[i].BrowserVersion
```
mdittmer: Follow-up from above: This will probably repeat the one result from `queries[0]` (if there is one) because it will also appear in the result set from `queries[1]`.

api/versions.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2017 The WPT Dashboard Project. All rights reserved.
```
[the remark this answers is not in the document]
lukebjerring: GET WITH THE TIMES

api/versions.go
lukebjerring: You're right and wrong. I don't want the original solo query in the case that a version prefix is needed; I want "exactly that version" and "stuff starting with that version" (which I can't query for both of at once, because, Golang + datastore reasons).

Otherwise the query as built so far is the only one. I'll, uh, make it an else branch for readability.

lukebjerring: Done

lukebjerring: No; because of an edge-case where chrome-7 would match chrome-70, we only find prefixes (and not exact matches).

api/versions.go
lukebjerring: `nil` arrays and empty arrays are equal (`len(foo) == 0`, etc)

api/versions.go
lukebjerring: Done.

api/versions.go
lukebjerring: The tests cover exact matching (and pass) with `chrome-1` :)

https://github.com/web-platform-tests/wpt.fyi/pull/691/files/0086c84e37c341dbed0bedd1e1cf31275285b3d1#diff-8f642c273c44c9c74435d70606cb4148R65

api/versions.go
mdittmer: I guess what I don't understand is why isn't it true that "foo" starts with "foo" (because, it does, right?) If it's just something weird about Datastore queries though, then I'm happy to let the tests prove that this is necessary and sufficient.

api/versions.go
mdittmer: You're right that append nil-of-type-TestRuns will do what you expect (which is what matters here).

Aside: TIL, though, that an uninitialized TestRuns and make(TestRuns, 0) have the same length, but only the former is equal to nil.

api/versions_medium_test.go
```diff
@@ -33,8 +33,8 @@ func TestApiVersionsHandler(t *testing.T) {
 	json.Unmarshal(bytes, &versions)
 	assert.Equal(t, []string{}, versions)
 
-	// Add test runs
-	someVersions := []string{"2", "1.1.1", "1.1", "1.0", "1"}
+	// Add test runs (duplicating 1.1 is deliberate)
```
mdittmer: nit: Comment to end with period.

api/versions.go
lukebjerring: Because chrome-7 shouldn't match chrome-70, we cheat by using a prefix of `7.`
This leaves exactly one edge-case; a full string match.

api/versions_medium_test.go
lukebjerring:
```suggestion
	// Add test runs (duplicating 1.1 is deliberate).
```

Issue #690, opened 2018-10-24
We should (quite temporarily) cache /api/shas


lukebjerring:
With a regular fetch to the endpoint, we should optimize (see https://github.com/web-platform-tests/wpt.fyi/pull/689 and #687)

We can either have a reasonable timeout (e.g. 5min), or we could respond to new revisions and invalidate?

PR #692, opened 2018-10-25
Request-level caching of /api/shas + bug fix

mdittmer:
## Description

Add request-level caching to `/api/shas`. To keep tests happy, also apply a bug fix detailed in b4392d9.

## Review Information

Observe logging about cache missing and serving from cache when visiting any `/api/shas?[params]` URL.

### DIFF
--- a/shared/request_caching.go
+++ b/shared/request_caching.go
@@ -157,15 +157,6 @@ func (h cachingHandler) delegateAndCache(w http.ResponseWriter, r *http.Request,
 // NewCachingHandler produces a caching handler with an underlying delegate
 // handler, cache, cacheability decision function, and cache key producer.
 func NewCachingHandler(ctx context.Context, delegate http.Handler, cache ReadWritable, isCacheable func(*http.Request) bool, getCacheKey func(*http.Request) interface{}, shouldCache func(int, []byte) bool) http.Handler {
-	if isCacheable == nil {
-		isCacheable = AlwaysCachable
-	}
-	if getCacheKey == nil {
-		getCacheKey = URLAsCacheKey
-	}
-	if shouldCache == nil {
-		shouldCache = CacheStatusOK
-	}
 	return cachingHandler{ctx, delegate, cache, isCacheable, getCacheKey, shouldCache}
 }
 
--- a/shared/request_caching.go
+++ b/shared/request_caching.go
@@ -18,6 +18,7 @@ type CachingResponseWriter interface {
 	http.ResponseWriter
 
 	WriteTo(io.Writer) (int64, error)
+	StatusCode() int
 }
 
 type cachingResponseWriter struct {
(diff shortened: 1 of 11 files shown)


mdittmer:
This should fix #690.

shared/request_caching.go
```diff
@@ -75,8 +77,13 @@ type cachingHandler struct {
 	cache       ReadWritable
 	isCacheable func(*http.Request) bool
 	getCacheKey func(*http.Request) interface{}
+	shouldCache func(int) bool
 }
 
+func defaultIsCacheable(*http.Request) bool          { return true }
```
lukebjerring: So, in another in-flight PR I've made this public and actually called this 'AlwaysCachable':
https://github.com/web-platform-tests/wpt.fyi/pull/674/files#diff-6c9911785ad4f162692fbb1bcc25984aR149

shared/request_caching.go
```diff
@@ -75,8 +77,13 @@ type cachingHandler struct {
...
+func defaultGetCacheKey(r *http.Request) interface{} { return r.URL.String() }
```
lukebjerring: And this `URLAsCacheKey`:
https://github.com/web-platform-tests/wpt.fyi/pull/674/files#diff-6c9911785ad4f162692fbb1bcc25984aR158

shared/request_caching.go
```diff
@@ -75,8 +77,13 @@ type cachingHandler struct {
...
+func defaultShouldCache(status int) bool             { return status == http.StatusOK }
```
lukebjerring: Guess I didn't have status-based/output 'cachability' before, but same pattern applies. If you like the idea that the name implies the behaviour (instead of saying "I'm the default" and reader must work it out an extra reading step away), then consolidate whatever names you think are good - I didn't scrutinize my naming, just the implications of the naming.

shared/request_caching.go
mdittmer: Patched that in and added `CacheStatusOK` for the third default function.

shared/request_caching.go
```diff
@@ -154,15 +156,34 @@ func (h cachingHandler) delegateAndCache(w http.ResponseWriter, r *http.Request,
 
 // NewCachingHandler produces a caching handler with an underlying delegate
 // handler, cache, cacheability decision function, and cache key producer.
-func NewCachingHandler(delegate http.Handler, cache ReadWritable, isCacheable func(*http.Request) bool, getCacheKey func(*http.Request) interface{}, shouldCache func(int) bool) http.Handler {
+func NewCachingHandler(ctx context.Context, delegate http.Handler, cache ReadWritable, isCacheable func(*http.Request) bool, getCacheKey func(*http.Request) interface{}, shouldCache func(int, []byte) bool) http.Handler {
```
lukebjerring: So, I don't think we should ever pass in nil (so that it's instead quite obvious what the value is at the call-site).

lukebjerring:
Related, can I please have a separate + similar PR for `/api/versions` (following on #691)?

mdittmer:
@lukebjerring sure. We should discuss the client use cases for these APIs and make sure that cache skew between endpoints can't break pages.

PR #695, opened 2018-10-26
Add env-flags for different default queries

lukebjerring:
## Description
Adds 2 flags `experimentalByDefault` and `experimentalAlignedExceptEdge` which allow for some different defaults as we've discussed offline.

## Review Information
A little hard to test without actually enabling them in staging, but locally they're added by default when you run dev_data (against this PR) and you can test that it works to revert back to normal when you delete them from the datastore using the admin console @ 0.0.0.0:8000


lukebjerring:
@foolip - note that I have followed up the merge by enabling both the features in https://staging.wpt.fyi

foolip:
Oh, nice! Is a prod release planned soonish? If people notice staging is much better, they'll just spread those links around, and then staging will effectively be as important to keep stable as prod.

PR #696, opened 2018-10-26
Add autocomplete to the Version query builder field

lukebjerring:
## Description
Builds on https://github.com/web-platform-tests/wpt.fyi/pull/689, using the same technique as SHA.

## Review Information
Open the query builder, see autocomplete in the `Version` field for each product.

### DIFF
--- a/webapp/components/test-runs-query-builder.html
+++ b/webapp/components/test-runs-query-builder.html
@@ -248,10 +248,15 @@ found in the LICENSE file.
         </paper-dropdown-menu>
 
         <br>
-        <paper-input always-float-label
-                      label="Version"
-                      placeholder="(Any version)"
-                      value="{{ browserVersion::input }}"></paper-input>
+        <paper-input-container always-float-label>
+          <label slot="label">Version</label>
+          <input slot="input"
+                  placeholder="(Any version)"
+                  list="versions-datalist"
+                  value="{{ browserVersion::input }}">
+          <datalist id="versions-datalist"></datalist>
+        </paper-input-container>
+      </template>
       </div>
     </paper-card>
   </template>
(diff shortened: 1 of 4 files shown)


webapp/components/test-runs-query-builder.html
```diff
@@ -383,6 +401,52 @@ <h3>
...
+          return;
+        }
+        if (version) {
+          versions = versions.filter(s => s.startsWith(version));
+        }
+        versions = versions.slice(0, 10);
+        // Check actually different from current.
+        const curr = this.versionsAutocomplete || [];
+        if (curr.length === versions.length) {
+          let same = true;
```
mdittmer: This looks correct. Would the code be simpler if we added versions to a set and check its length?

webapp/components/test-runs-query-builder.html
```diff
@@ -383,6 +401,52 @@ <h3>
           labels: (product.labels || []).filter(l => !CHANNELS.has(l)).concat(channel)
         });
       }
+
```
mdittmer: It's hard to keep the flow straight my method name alone in this block of methods. Could we add short header comments like `// Respond to [foo] changing by [action]`, where `[foo]` and `[action]` are more than just a property or URL name. E.g., "Respond to list of versions to be suggested in data list changing by ..."

webapp/components/test-runs-query-builder.html
lukebjerring: Simplified.

webapp/components/test-runs-query-builder.html
lukebjerring: Done.

lukebjerring:
Error is just announcer deployment, quota issue. Force-merging.

Issue #679, opened 2018-10-19
Compute autocomplete client-side


lukebjerring:
Currently we do a server-side fetch for all of the paths across a set of runs. This is actually redundant, since we can compute this information client-side once the run summaries arrive, which will reduce the latency on the client.

PR #697, opened 2018-10-28
Compute autocomplete client-side + navigate on selection

### LINKED ISSUE #643: Navigate directly to suggested test from search dropdown
Go to https://wpt.fyi/results/ and type "historical" into the search box. A dropdown with suggestions appears:
search-dropdown

Picking a suggestion, however, merely searched for the string:
image

One should end up at https://wpt.fyi/results/url/historical.any.worker.html. I'm hesitant to call this an enhancement because it feels fundamental to the UI, so more like necessary polish.

lukebjerring:
## Description
Fixes #643 and #679

Compute the autocomplete options from the set of test results paths available on the client, reducing latency.
Add an event hook for changing the input value to the exact value of an autocomplete item, and leverage that to navigate to the file. 

## Review Information
@mdittmer, I've picked you for this because you know the effect of executing a search, and can deem whether this is problematic (autocompleting based on a subset of paths when doing a search).

I think this fixes a current bug with `history` push/pop too (clicking 'back' was causing a push to the stack).

### DIFF
--- a/webapp/components/interop.html
+++ b/webapp/components/interop.html
@@ -127,7 +127,8 @@ found in the LICENSE file.
 
     <test-search class$="search-[[pathIsATestFile]]"
                  query="{{search}}"
-                 test-runs="[[testRuns]]">
+                 test-runs="[[testRuns]]"
+                 test-paths="[[testPaths]]">
     </test-search>
 
     <template is="dom-if" if="{{ pathIsATestFile }}">
@@ -257,6 +258,10 @@ found in the LICENSE file.
         searchResults: Array,
         onSearchCommit: Function,
         interopLoadFailed: Boolean,
+        testPaths: {
+          type: Set,
+          computed: 'computeTestPaths(fileMetrics)',
+        },
       };
     }
 
@@ -264,6 +269,9 @@ found in the LICENSE file.
       super();
       this.onSearchCommit = (e) => {
         this.handleSearchCommit(e.detail.query);
+      };
+      this.onSearchAutocomplete = (e) => {
+        this.handleSearchAutocomplete(e.detail.path);
       };
       this.onLoadingComplete = () => {
         // passRateMetadata contains the url for the JSON blob of allMetrics;
@@ -279,6 +287,8 @@ found in the LICENSE file.
       super.connectedCallback();
       this.shadowRoot.querySelector('test-search')
         .addEventListener('commit', this.onSearchCommit);
+      this.shadowRoot.querySelector('test-search')
+        .addEventListener('autocomplete', this.onSearchAutocomplete);
     }
 
     disconnectedCallback() {
@@ -333,6 +343,13 @@ found in the LICENSE file.
       return metadata && metadata.test_runs;
     }
 
+    computeTestPaths(fileMetrics) {
+      if (!fileMetrics) {
+        return [];
+      }
+      return new Set(fileMetrics.map(m => m.dir));
+    }
+
     computeFileMetrics(allMetrics) {
       let fileMetrics = [];
       for (const metric of allMetrics.data) {
@@ -369,6 +386,11 @@ found in the LICENSE file.
 
       this.searchResults = this.fileMetrics
         .filter(t => t.dir.toLowerCase().includes(q));
+    }
+
+    handleSearchAutocomplete(path) {
+      this.shadowRoot.querySelector('test-search').clear();
+      this.navigateToPath(path);
     }
 
     computeDisplayedNodes(path, displayedTests) {
--- a/webapp/components/test-search.html
+++ b/webapp/components/test-search.html
@@ -19,10 +19,11 @@ found in the LICENSE file.
     </style>
 
     <div>
-      <input value="{{ q::input }}"
+      <input value="{{ queryInput::input }}"
              class="query"
              list="query-list"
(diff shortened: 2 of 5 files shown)


webapp/components/test-search.html
```diff
@@ -86,50 +86,36 @@
...
-        if (q) {
-          url.searchParams.set('q', q);
-        }
-        return url;
-      }
-
       ready() {
         super.ready();
-        this.q = this.query;
+        this._createMethodObserver('pathsUpdated(query, testPaths)');
```
mdittmer: This sounds like a Polymer implementation detail. Is this the recommended way of creating an observer with no explicit output property?

webapp/components/test-search.html
```diff
@@ -86,50 +86,36 @@
...
-        const json = await response.json();
-
         const datalist = this.shadowRoot.querySelector('datalist');
         datalist.innerHTML = '';
-        if (!json.results) {
-          return;
+        let matches = Array.from(paths);
+        if (query) {
+          matches = matches
+            .filter(p => p.indexOf(query) >= 0)
```
mdittmer: Nit: Prefer [`p.includes(query)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes).

webapp/components/test-search.html
```diff
@@ -86,50 +86,36 @@
...
-          return;
-        }
+      queryUpdated(query) {
+        this.queryInput = query;
+      }
 
-        const response = await window.fetch(newValue);
-        if (response.status !== 200) {
-          throw response;
+      pathsUpdated(query, paths) {
```
mdittmer: nit: This depends on query too, and what it _really_ does is update the datalist. Maybe name it `updateDatalist` or similar?

webapp/components/test-search.html
```diff
@@ -141,11 +127,11 @@
       }
 
       latchQuery() {
-        this.query = this.q.toLowerCase();
+        this.query = (this.queryInput || '').toLowerCase();
```
mdittmer: Are paths `toLowerCase()`d?

webapp/components/test-search.html
```diff
@@ -170,7 +156,10 @@
         const value = e.target.value.toLowerCase();
 
         if (opts.includes(value)) {
-          this.commitQuery();
+          this.dispatchEvent(new CustomEvent('autocomplete', {
```
mdittmer: As discussed offline, we may want to drop this and let the `'commit'` listener detect when a query matches a path.

webapp/components/wpt-results.html
```diff
@@ -412,6 +417,16 @@ <h3>
         return true;
       }
 
+      computeTestPaths(searchResults) {
+        if (!searchResults) {
+          return [];
```
mdittmer: ditto

webapp/components/interop.html
```diff
@@ -333,6 +343,13 @@
       return metadata && metadata.test_runs;
     }
 
+    computeTestPaths(fileMetrics) {
+      if (!fileMetrics) {
+        return [];
+      }
+      return new Set(fileMetrics.map(m => m.dir));
```
mdittmer: nit: We use `reduce()` instead in the other `computeTestPaths` which is more lines of code, but has the advantage of not building an array for nearly immediate garbage collection. We should pick one strategy for both implementations.

webapp/components/interop.html
```diff
@@ -265,6 +270,9 @@
       this.onSearchCommit = (e) => {
         this.handleSearchCommit(e.detail.query);
       };
+      this.onSearchAutocomplete = (e) => {
```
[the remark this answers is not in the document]
lukebjerring: Done.

webapp/components/interop.html
```diff
@@ -333,6 +343,13 @@
       return metadata && metadata.test_runs;
     }
 
+    computeTestPaths(fileMetrics) {
+      if (!fileMetrics) {
+        return [];
```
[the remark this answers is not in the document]
lukebjerring: Fixed.

webapp/components/interop.html
lukebjerring: Consolidated to use `Set` ctor

webapp/components/test-search.html
lukebjerring: Yes, it's essentially an observer of multiple properties (with no output property).

webapp/components/test-search.html
lukebjerring: Done.

webapp/components/test-search.html
lukebjerring: Done.

webapp/components/test-search.html
lukebjerring: Uh, no.. added `normalizedTestPaths`.

webapp/components/test-search.html
lukebjerring: I think we still want to expose the `autocomplete` event. I doubt a user will type a full path (though they may paste one), and we don't want to need a double-enter for autocompletion.

webapp/components/wpt-results.html

foolip:
Oh, nice, I love it! Thanks @lukebjerring!

PR #698, opened 2018-10-29
Add 'full' param to TestRun query params

lukebjerring:
## Description
This param will be used by the front page to omit non-full runs (such as PRs)

## Review Information
Not yet utilized in the UI, and loading `/api/runs?full` will be empty results right now - we need to change the infra to flag the full runs as such.

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -9,6 +9,8 @@ import (
 	"encoding/json"
 	"net/http"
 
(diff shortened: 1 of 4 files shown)


lukebjerring:
As discussed offline, we may move to a model of executing the full suite for PRs, so this does not achieve the desired result.

PR #700, opened 2018-10-29
Consolidate color scheme

### LINKED ISSUE #423: Use the same color scheme in results and interop view

lukebjerring:
## Description
Resolves #423

### DIFF
--- a/webapp/components/wpt-colors.html
+++ b/webapp/components/wpt-colors.html
@@ -15,6 +15,7 @@
 
   <script>
     const wpt = window.wpt || {};
+    // RGB values from https://material.io/design/color/
     wpt.colors = [
       { class: 'passes-none', colorVar: '--paper-red-400', rgb: [239, 83, 80] },
       { class: 'passes-hardly', colorVar: '--paper-orange-400', rgb: [255, 167, 38] },
--- a/webapp/components/wpt-colors.html
+++ b/webapp/components/wpt-colors.html
@@ -1,0 +1,58 @@
+<link rel="import" href="../bower_components/paper-styles/color.html">
+
+<dom-module id="wpt-colors">
+  <template>
+    <style>
+      .passes-none { background-color: var(--paper-red-400); }
+      .passes-hardly { background-color: var(--paper-orange-400); }
+      .passes-a-few { background-color: var(--paper-amber-400); }
+      .passes-half { background-color: var(--paper-yellow-400); }
+      .passes-lots { background-color: var(--paper-lime-400); }
+      .passes-most { background-color: var(--paper-light-green-400); }
+      .passes-all { background-color: var(--paper-green-400); }
+    </style>
+  </template>
+
+  <script>
+    const wpt = window.wpt || {};
+    wpt.colors = [
+      { class: 'passes-none', colorVar: '--paper-red-400', rgb: [239, 83, 80] },
+      { class: 'passes-hardly', colorVar: '--paper-orange-400', rgb: [255, 167, 38] },
+      { class: 'passes-a-few', colorVar: '--paper-amber-400', rgb: [255, 202, 40] },
+      { class: 'passes-half', colorVar: '--paper-yellow-400', rgb: [255, 238, 88] },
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-colors.html
```diff
@@ -0,0 +1,58 @@
...
+      .passes-lots { background-color: var(--paper-lime-400); }
+      .passes-most { background-color: var(--paper-light-green-400); }
+      .passes-all { background-color: var(--paper-green-400); }
+    </style>
+  </template>
+
+  <script>
+    const wpt = window.wpt || {};
+    wpt.colors = [
+      { class: 'passes-none', colorVar: '--paper-red-400', rgb: [239, 83, 80] },
```
foolip: Can you link to the canonical definition of these that these numbers should match?

webapp/components/wpt-colors.html
```diff
@@ -0,0 +1,58 @@
...
+      { class: 'passes-hardly', colorVar: '--paper-orange-400', rgb: [255, 167, 38] },
+      { class: 'passes-a-few', colorVar: '--paper-amber-400', rgb: [255, 202, 40] },
+      { class: 'passes-half', colorVar: '--paper-yellow-400', rgb: [255, 238, 88] },
+      { class: 'passes-lots', colorVar: '--paper-lime-400', rgb: [212,225,87] },
+      { class: 'passes-most', colorVar: '--paper-light-green-400', rgb: [156, 204, 101] },
+      { class: 'passes-all', colorVar: '--paper-green-400', rgb: [102, 187, 106] },
+    ];
+    wpt.passRateClass = (passes, total) => {
+      return wpt.getColor(passes, total)['class'];
```
foolip:
```suggestion
      return wpt.getColor(passes, total).class;
```

webapp/components/wpt-colors.html
```diff
@@ -0,0 +1,58 @@
...
+    };
+    wpt.passRateColorVar = (passes, total) => {
+      return wpt.getColor(passes, total)['colorVar'];
```
foolip:
```suggestion
      return wpt.getColor(passes, total).colorVar;
```

webapp/components/wpt-colors.html
```diff
@@ -0,0 +1,58 @@
...
+    };
+    wpt.passRateColorRGBA = (passes, total, alpha) => {
+      const color = wpt.getColor(passes, total);
+      return `rgba(${color['rgb'].join(', ')}, ${alpha})`;
```
foolip:
```suggestion
      return `rgba(${color.rgb.join(', ')}, ${alpha})`;
```

webapp/components/wpt-colors.html
foolip: (Or use `className` if necessary to avoid a keyword, like `element.className`.)

webapp/components/wpt-colors.html
foolip: Linked above.

foolip:
To answer my own question about whether 400 is the sweet spot, here's with 300, 400 and 500, in that order:
screen shot 2018-10-29 at 10 59 52 pm
screen shot 2018-10-29 at 11 01 11 pm
screen shot 2018-10-29 at 11 00 42 pm

I think 400 is nice, 500 too dark, and if we get feedback about legibility perhaps we can move to 300. (We could also try a different font, bold text, etc.)

Issue #699, opened 2018-10-29
Label runs with their WPT branch


lukebjerring:
Runs for our current status-quo should be labelled with the branch they came from, so that the default query can distinguish `master` on the homepage of wpt.fyi

PR #703, opened 2018-10-30
Add the github branch as a label

lukebjerring:
## Description
This will allow us to distinguish `master` runs.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/699

### DIFF
--- a/api/webhook/taskcluster.go
+++ b/api/webhook/taskcluster.go
@@ -81,23 +81,11 @@ func tcWebhookHandler(w http.ResponseWriter, r *http.Request) {
 
 // https://developer.github.com/v3/activity/events/types/#statusevent
 type statusEventPayload struct {
-	Sha       string       `json:"sha"`
-	State     string       `json:"state"`
-	Context   string       `json:"context"`
-	TargetURL string       `json:"target_url"`
-	Branches  []branchInfo `json:"branches"`
-}
-
-// GetBranch returns the first branch that has the revision of the commit
-// as its HEAD state.
-func (s statusEventPayload) GetBranch() string {
-	// Only process the master branch.
-	for _, branch := range s.Branches {
-		if branch.Commit.Sha == s.Sha {
-			return branch.Name
-		}
-	}
-	return ""
+	SHA       string      `json:"sha"`
+	State     string      `json:"state"`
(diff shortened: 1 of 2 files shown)


api/webhook/taskcluster.go
```diff
@@ -88,8 +88,34 @@ type statusEventPayload struct {
...
+
 type branchInfo struct {
-	Name string `json:"name"`
+	Name   string     `json:"name"`
+	Commit commitInfo `json:"commit"`
+}
+
+type commitInfo struct {
+	Sha string `json:"sha"`
+	URL string `json:"url"`
```
Hexcles: Do we need the URL of a commit? If not, let's remove this field. The API payload from both GitHub and Taskcluster are big and complicated, so I like to include only the necessary fields in various structs, which simplifies the struct definition and avoids distraction.

api/webhook/taskcluster.go
```diff
@@ -88,8 +88,34 @@ type statusEventPayload struct {
 	Branches  []branchInfo `json:"branches"`
 }
 
+// GetBranch returns the first branch that has the revision of the commit
+// as its HEAD state.
+func (s statusEventPayload) GetBranch() string {
```
Hexcles: Using this to decide whether a status update comes from the master branch might be problematic.

IIUC, the `branches` field contains a list of branches that contain this commit, each one with its *current* head. When a Taskcluster run finishes for a commit and posts the "success" status, that commit is very likely no longer the head of master.

I think we need two methods, like `IsOnMaster` and `HeadingBranches`, and only the latter looks at the SHA and will be used in the PR case.

api/webhook/taskcluster.go
```diff
@@ -88,8 +88,34 @@ type statusEventPayload struct {
...
+}
+
 type branchInfo struct {
-	Name string `json:"name"`
+	Name   string     `json:"name"`
+	Commit commitInfo `json:"commit"`
+}
+
+type commitInfo struct {
+	Sha string `json:"sha"`
```
Hexcles: Nit: SHA instead of Sha.

api/webhook/taskcluster.go
lukebjerring: Done.

PR #709, opened 2018-10-30
Add flag for structured queries

mdittmer:
The flag is a no-op for now; I defined methods in the `test-runs` component for diverging query processing behaviour, but they contain a `TODO` and delegate to the unstructured query behaviour for now.

### DIFF
--- a/util/populate_dev_data.go
+++ b/util/populate_dev_data.go
@@ -173,7 +173,7 @@ func main() {
 	log.Print("Adding flag defaults...")
 	addFlag(ctx, "queryBuilder", enabledFlag)
 	addFlag(ctx, "diffFilter", enabledFlag)
-  addFlag(ctx, "diffFromAPI", enabledFlag)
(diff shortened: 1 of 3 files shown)


util/populate_dev_data.go
```diff
@@ -173,9 +173,10 @@ func main() {
 	log.Print("Adding flag defaults...")
 	addFlag(ctx, "queryBuilder", enabledFlag)
 	addFlag(ctx, "diffFilter", enabledFlag)
-	addFlag(ctx, "diffFromAPI", enabledFlag)
+  addFlag(ctx, "diffFromAPI", enabledFlag)
```
lukebjerring:
```suggestion
	addFlag(ctx, "diffFromAPI", enabledFlag)
```

mdittmer:
Per offline conversation, we will eventually want the UI to be able to reflect whether or not a query is valid without consulting the server. To avoid writing the parser twice, we'll parse on the client and send a structured query to the server (rather than query string).

PR #713, opened 2018-10-31
Structured queries: Server-side parsing

mdittmer:
Test and implement server-side parsing of structured search queries. Client-side string parsing and structured query generation will be in a follow-up PR.


api/query/query.go
```diff
@@ -17,6 +20,244 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+var browsers = []string{
```
lukebjerring: This is available as `shared.GetDefaultBrowserNames()` or something similar

api/query/search.go
```diff
@@ -55,23 +59,77 @@ func (r byName) Len() int           { return len(r) }
...
+	queryHandler
+}
+
+type structuredSearchHandler struct {
 	queryHandler
 }
 
 func apiSearchHandler(w http.ResponseWriter, r *http.Request) {
-	// Parse query params.
+	if r.Method != "GET" && r.Method != "PUT" {
```
lukebjerring: `PUT`? I think you mean `POST`?

api/query/query.go
```diff
@@ -17,6 +20,244 @@ import (
...
+	browserName string
+	status      int64
+}
+
+func (tsc testStatusConstraint) toPlan(runIDs []int64) plan {
+	return nil
+}
+
+type not struct {
+	not query
```
lukebjerring: `not{not: testNamePattern{pattern: "abc"}}` not not readable

I think we'll have to ignore lint saying we should name the params in Golang, and just write
`not{testNamePattern{"abc"}}`

PR #715, opened 2018-10-31
Implement and integrate client-side structured query parser/generator

mdittmer:
This changes uses [ohm](https://github.com/harc/ohm) to define structured queries compatible with server-side parsing articulated in #713. It also integrates the queries by using `POST`-with-body rather than `GET`-with-query-params on `/api/search`.

### DIFF
--- a/webapp/components/test-search.html
+++ b/webapp/components/test-search.html
@@ -161,6 +161,12 @@ found in the LICENSE file.
 
     /* global WPTFlags */
     class TestSearch extends WPTFlags(window.Polymer.Element) {
+      static get QUERY_GRAMMAR() {
+        return QUERY_GRAMMAR;
+      }
+      static get QUERY_SEMANTICS() {
+        return QUERY_SEMANTICS;
+      }
       static get is() {
         return 'test-search';
       }
--- a/webapp/components/test-search.html
+++ b/webapp/components/test-search.html
@@ -38,7 +38,126 @@ found in the LICENSE file.
     </div>
   </template>
 
+  <script type="text/javascript" src="https://unpkg.com/ohm-js@latest/dist/ohm.js"></script>
+  <script type="text/ohm-js">
+
+  </script>
   <script>
+    const QUERY_GRAMMAR = ohm.grammar(`
+      Query {
+        Q = ListOf<Exp, space*>
+
+        Exp = NonemptyListOf<OrPart, or>
+
+        OrPart = NonemptyListOf<AndPart, and>
+
+        AndPart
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-results.html
```diff
@@ -776,10 +774,12 @@ <h3>
         this.navigateToPath(e.detail.path);
       }
 
-      handleSearchCommit() {
+      handleSearchCommit(e) {
+        const detail = e.detail;
         // Fetch search results when test-search signals that user has committed
         // to search string (by pressing <Enter>).
-        this.fetchResults();
+        this.fetchResults(this.structuredQueries ?
```
lukebjerring: Nit: wrap like

        this.fetchResults(this.structuredQueries
          ? detail.structuredQuery
          : detail.query);

webapp/components/test-search.html
```diff
@@ -38,7 +38,128 @@
     </div>
   </template>
 
+  <script type="text/javascript" src="https://unpkg.com/ohm-js@latest/dist/ohm.min.js"></script>
+  <script type="text/ohm-js">
```
lukebjerring: What's with the extra `script` tags?

mdittmer: OHM has a mechanism for scraping the text of a script tag. I originally was using it, but it doesn't play well with web components, so I switched to a multiline string. This is just leftover; I'll delete it.

mdittmer:
Fix data binding bug found thanks to webdriver tests (yay tests!). @lukebjerring PTAL again.

lukebjerring:
Empty search doesn't work (i.e. throws a 400 and shows no results) when the feature flag is enabled (but, a search like `2dcontext` does show results).

mdittmer:
This appears to be a staging caching issue. Here's the curl version of what staging currently does under the hood with the flag turned on:

```
curl -v -X POST --header "Content-Type: application/json" --data '{"run_ids":[5088567629971456,5874471550320640,6316346107232256,5198267470053376],"query":{"pattern":""}}' 'https://sq-cs-parse-dot-wptdashboard-staging.appspot.com/api/search'
```

This yields:

```
Failed to parse query fragment as test name pattern, test status constraint, negation, disjunction, or conjunction
```

However, when I run this branch locally:

```
curl -v -X POST --header "Content-Type: application/json" --data '{"run_ids":[5088567629971456,5874471550320640,6316346107232256,5198267470053376],"query":{"pattern":""}}' 'http://0.0.0.0:8080/api/search'
```

This yields a whole bunch of results.

PR #717, opened 2018-11-02
Fix markup structure in wpt-flags.html

foolip:
## Description
Introduced in https://github.com/web-platform-tests/wpt.fyi/pull/709
and causes a visual misalignment in /flags.

## Review Information
Check /flags in test deployment.


foolip:
https://wpt-flags-markup-dot-wptdashboard-staging.appspot.com/flags approves!

PR #718, opened 2018-11-02
Use go-github protos instead of local structs

lukebjerring:
## Description
go-github defines these protos already, so can reduce our code lines, and future code-churn.

### DIFF
--- a/api/webhook/taskcluster.go
+++ b/api/webhook/taskcluster.go
@@ -10,6 +10,7 @@ import (
 	"crypto/sha1"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


api/webhook/taskcluster.go
```diff
@@ -81,24 +83,20 @@ func tcWebhookHandler(w http.ResponseWriter, r *http.Request) {
 
 // https://developer.github.com/v3/activity/events/types/#statusevent
 type statusEventPayload struct {
-	SHA       string      `json:"sha"`
-	State     string      `json:"state"`
-	Context   string      `json:"context"`
-	TargetURL string      `json:"target_url"`
-	Branches  branchInfos `json:"branches"`
+	github.StatusEvent
```
Hexcles: Instead of embedding a single field, let's make `statusEventPayload` an alias of `github.StatusEvent`.

lukebjerring: Done.

Issue #712, opened 2018-10-31
Implement a custom check (in wpt) for wpt.fyi


lukebjerring:
Now that we're consuming the output of TaskCluster consistently, we should be building a custom status check for wpt.fyi that is resolved once the results of the run are available for browsing on wpt.fyi.

At a later date, this will tie into a failed status check for any detected regressions.

lukebjerring:
![image](https://user-images.githubusercontent.com/777956/47804463-4530b980-dd0b-11e8-927f-9215fd10feb7.png)

This diagram loosely describes the flow; currently:
- GitHub pushes status events to `/api/webhooks/taskcluster`
- TaskCluster eventually pushes its success to GitHub, which in turn notifies wpt.fyi
- wpt.fyi fetches the results from taskcluster

In the future, we'd like the same model for Azure Pipelines, which will involve generalizing `/api/webhooks/taskcluster` to `/api/webhooks/status`.

wpt.fyi will create a `check_suite` for each SHA, and a `check_run` in the suite for each product run that we learn about (via status updates). The 'create run' endpoint used by the processor will be responsible for following up on the `check_run` completion.

lukebjerring:
![image](https://user-images.githubusercontent.com/777956/47804717-e3bd1a80-dd0b-11e8-87f0-e99aaeeed016.png)

@Hexcles drew a "timeline" of the to-and-fro, where dashed/dotted lines represent features not yet implemented.

foolip:
I'd like to add a suggestion here. When full runs from master for the merged commit and the previous merge_pr_* commit are available, a status check on the master commit linking to a diff view would be very useful. Here are examples of where I wanted that view today:
https://github.com/web-platform-tests/wpt/pull/14068#issuecomment-441183248
https://github.com/web-platform-tests/wpt/pull/14087#issuecomment-441185673
https://github.com/web-platform-tests/wpt/pull/14088#issuecomment-441185059

Much of the infrastructure would be the same as we'd use to notice large regressions on master.

@lukebjerring, if you think that's too much scope creep, I'd be happy to file a separate issue.

foolip:
https://github.com/web-platform-tests/wpt/pull/14094#issuecomment-441243918 too.

lukebjerring:
Custom checks are now implemented.

PR #719, opened 2018-11-02
Process non-master branch Taskcluster results (with flag)

lukebjerring:
## Description
Changes the `/api/webhook/taskcluster` handler to not ignore non-master branches when `taskclusterAllBranches` flag is enabled.

Working towards #712 

This requires a tweak to the logic for task names (we want to ignore `stability`, but process `results`).
It also requires us to override what Taskcluster's wpt-report says was the SHA, since that SHA is a squash-commit over `master` and doesn't match the GitHub SHA for the PR results.

## Review Information
While difficult to confirm this works locally, it's a copy of the code from my [prototype](https://github.com/lukebjerring/web-platform-tests/pull/8/checks) for which you can see `change` branch results have made it into staging: https://staging.wpt.fyi/test-runs?label=change

### DIFF
--- a/api/webhook/taskcluster.go
+++ b/api/webhook/taskcluster.go
@@ -323,8 +323,9 @@ func createAllRuns(log shared.Logger,
 func createRun(client *http.Client, sha, api string, username string, password string, reportURLs []string, labels []string) error {
 	// https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md#url-payload
 	payload := make(url.Values)
-	payload.Add("revision", sha[:10])
-	payload.Add("full_revision_hash", sha)
+	// Not to be confused with `revision` in the wpt.fyi TestRun model, this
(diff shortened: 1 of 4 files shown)


api/webhook/taskcluster.go
```diff
@@ -128,7 +130,7 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 		return false, err
 	}
 
-	if !shouldProcessStatus(&status) {
+	if !shouldProcessStatus(shared.IsFeatureEnabled(ctx, flagTaskclusterAllBranches), &status) {
```
Hexcles: The first argument is `masterOnly`, so this says `IsEnabled(taskclusterAllBranches) == masterOnly`? Shouldn't it be the other way around?

lukebjerring:
Failure is from Travis having incorrectly gone green in #715 - and already fixed on master.

PR #721, opened 2018-11-02
Refactor verifySignature for re-use

lukebjerring:
## Description
Extract the "load secret + verify payload" behaviour into a reusable method, in prep for adding another GitHub webhook for custom checks.

Working towards #712

### DIFF
--- a/api/webhook/verify.go
+++ b/api/webhook/verify.go
@@ -12,7 +12,7 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
-func verifyRequestSignature(r *http.Request) ([]byte, error) {
+func verifyAndGetPayload(r *http.Request) ([]byte, error) {
 	ctx := shared.NewAppEngineContext(r)
 	log := shared.GetLogger(ctx)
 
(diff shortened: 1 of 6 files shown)


api/webhook/verify.go
```diff
@@ -0,0 +1,49 @@
...
+	log := shared.GetLogger(ctx)
+
+	payload, err := ioutil.ReadAll(r.Body)
+	r.Body.Close()
+	if err != nil {
+		log.Errorf("Failed to read request body: %s", err.Error())
+		return nil, errors.New("Failed to read request body")
+	}
+
+	secret, err := getSecret(ctx)
```
Hexcles: We should move `getSecret` to this file as well. Besides, it might be worthwhile to generalize the `getSecret` (and this function as well) to take in the name for the token, so that we can use different tokens in different webhooks.

lukebjerring: Agreed (and actually already in flight) - copy-pasted the changes from https://github.com/web-platform-tests/wpt.fyi/pull/732/files#diff-04b8cc3bdfca7285a8fb0ee07b8069ccR283

api/webhook/verify.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/hex"
+	"errors"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+func verifyAndGetPayload(r *http.Request) ([]byte, error) {
```
Hexcles: Optional: this is fine for now, but I imagine you'd like to reuse this method in checks to verify the payload? (I saw your TODO in that PR.)

When that happens, we'd need to export (and perhaps move) this function, and add `tokenName` as an argument.

lukebjerring: Will do :)

PR #722, opened 2018-11-02
Plumb diff filter params through the UI

lukebjerring:
## Description
Splits `path` param from `filter` for parsing, and plumbs the value to the webcomponent.
We'll need this to make the UI ignore tests that didn't run on a PR.

Also changes the `/api/diff` endpoint to allow anything that produces exactly 2 results, e.g. `?product=chrome&max-count=2`

## Review Information
Load some filtered view, e.g. `/results?products=chrome[experimental],chrome[stable]&diff&filter=U` will only show unchanged things (kinda useless, but easy to tell it works).

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -213,7 +213,7 @@ found in the LICENSE file.
               <template is="dom-if" if="[[diffShown]]">
                 <th>
                   <test-run test-run="[[diffRun]]"></test-run>
-                  <template is="dom-if" if="[[diffFilter]]">
+                  <template is="dom-if" if="[[diffFilterUIToggle]]">
                     <paper-icon-button icon="filter-list" onclick="[[toggleDiffFilter]]"></paper-icon-button>
                   </template>
                 </th>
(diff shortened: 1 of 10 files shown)


webapp/components/wpt-results.html
```diff
@@ -445,11 +445,15 @@ <h3>
           return;
         }
         let url = new URL('/api/diff', window.location);
-        url.search = query
-          .replace('product', 'before')
-          .replace('product', 'after');
+        url.search = query;
+        const productParams = url.searchParams.getAll('product');
```
foolip: This is nicer now :)

lukebjerring:
Yeah, as discussed offline,

Take `beacon-basic-blob.html` for example. The UI shows it as 0 / 0 / 0, but if you see the /api/diff fetch:
https://diff-filter-dot-wptdashboard-staging.appspot.com/api/diff?before=chrome%5Bexperimental%5D&after=chrome%5Bstable%5D&filter=C you don't see `beacon-basic-blob.html` in the result.

So, as you say, I'll be improving the client-side filtering to match the same behaviour in the UI.

PR #727, opened 2018-11-05
Add feature flag for fetching master branch runs only

lukebjerring:
## Description
Adds a field to the query builder, and uses the `label=master` in the default query.

## Review Information
- Head to /flags, enable the feature
- Open the query builder, tick the box, reload, see `master` label is used

### DIFF
--- a/webapp/components/test-runs-query.html
+++ b/webapp/components/test-runs-query.html
@@ -73,16 +73,20 @@ found in the LICENSE file.
           this.productSpecs = (products || []).map(p => this.getSpec(p));
         }
 
+        /**
+        * Convert the UI property values into their equivalent URI query params.
+        */
         computeTestRunQueryParams(sha, aligned, master, labels, productSpecs, maxCount, from, to) {
           const params = {};
           if (!this.computeIsLatest(sha)) {
             params.sha = sha;
           }
-          if (this.masterRunsOnly && master
-              && (!labels || !labels.includes('master'))) {
-            labels = (labels || []).concat('master');
+          // Convert bool master to label 'master'
+          labels = labels && Array.from(labels) || [];
+          if (this.masterRunsOnly && master && !labels.includes('master')) {
+            labels.push('master');
           }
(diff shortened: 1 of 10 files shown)


webapp/components/test-runs-query.html
```diff
@@ -72,11 +73,15 @@
...
 
-        computeTestRunQueryParams(sha, aligned, labels, productSpecs, maxCount, from, to) {
+        computeTestRunQueryParams(sha, aligned, master, labels, productSpecs, maxCount, from, to) {
           const params = {};
           if (!this.computeIsLatest(sha)) {
             params.sha = sha;
           }
+          if (this.masterRunsOnly && master
+              && (!labels || !labels.includes('master'))) {
+            labels = (labels || []).concat('master');
```
mdittmer: This should be `labels = (labels || []).concat(['master']);` or `labels = labels || []; labels.push('master');`

webapp/components/test-runs-query.html
```diff
@@ -182,6 +187,13 @@
           if ('aligned' in params) {
             batchUpdate.aligned = params.aligned;
           }
+          if (this.masterRunsOnly) {
```
mdittmer: I'm confused. Cica line 81 we model "master only" as a label, `master`. Here, we're modeling it as a property, `master:true`, and removing its corresponding label. Why don't we model it consistently one way or the other?

webapp/components/test-runs-query.html
lukebjerring: We don't want to mutate the UI property (potentially there was a bug before this PR). Your code is more readable though; changed.

webapp/components/test-runs-query.html
lukebjerring: Server-side URI params vs "less confusing UI" properties.

lukebjerring:
No, the changes were the same treatment of labels as a copy of the array (and always existing now).

PR #728, opened 2018-11-06
Handle diff filter on the client

lukebjerring:
## Description
Follow up for https://github.com/web-platform-tests/wpt.fyi/pull/722, which applies the filtering on the client.

## Review Information
- Enable the diff filter toggle, but don't enable the API endpoint
- Load a diff view, with a filter, e.g. `filter=AD`
- See that only adds (new files) and deletes (removed files) are shown.

### DIFF
--- a/webapp/components/wpt-results.html
+++ b/webapp/components/wpt-results.html
@@ -632,16 +632,27 @@ found in the LICENSE file.
                 row.results[i].total += rs[i].total;
               }
               if (this.diff && rs.length === 2) {
-                let diff = [];
+                let diff;
                 if (this.diffResults) {
                   diff = this.diffResults[r.test] || [0, 0, 0];
                 } else {
                   const [before, after] = rs;
+                  const deleted = before.total > 0 && after.total === 0;
+                  const added = after.total > 0 && before.total === 0;
+                  if (deleted && !this.diffFilter.includes('D')
+                      || added && !this.diffFilter.includes('A')) {
+                    return acc;
+                  }
                   const failingBefore = before.total - before.passes;
                   const failingAfter = after.total - after.passes;
-                  diff.push(Math.max(after.passes - before.passes, 0)); // passes
-                  diff.push(Math.max(failingAfter - failingBefore, 0)); // regressions
-                  diff.push(after.total - before.total); // total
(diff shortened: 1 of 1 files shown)


lukebjerring:
So, a great test-case is the regression we saw recently in https://github.com/web-platform-tests/wpt/pull/13878

https://diff-client-dot-wptdashboard-staging.appspot.com/results/?product=chrome%5Bmaster%5D&product=chrome%40340b29725f&diff&filter=ACU
https://diff-client-dot-wptdashboard-staging.appspot.com/results/?product=firefox%5Bmaster%5D&product=firefox%40340b29725f&diff&filter=ACU

And, the post-fix results:
https://diff-client-dot-wptdashboard-staging.appspot.com/results/?product=chrome%5Bmaster%5D&product=chrome%40c4f9ad65b2&diff&filter=ACU
https://diff-client-dot-wptdashboard-staging.appspot.com/results/?product=firefox%5Bmaster%5D&product=firefox%40c4f9ad65b2&diff&filter=ACU

foolip:
With the comparisons in https://github.com/web-platform-tests/wpt.fyi/pull/728#issuecomment-436264315 there are just a lot of subtests that have gone missing, so I couldn't check what the effect is for regressions or improvements.

However, starting at https://diff-client-dot-wptdashboard-staging.appspot.com/results/?product=chrome%5Bstable%5D&product=chrome%5Bbeta%5D&aligned&diff and adding various `filter=` values, it doesn't seem quite right. `filter=A`, `filter=D` and `filter=AD` all show "0 / 0 / 0" for every row. Given that `filter=AD` was mentioned in description this seems like it's probably a bug, so I'll stand by for reactions.

Adding more API documentation for what effect the filter should have might make it easier for me to try testing it in the right ways :)

lukebjerring:
Sure - you're comparing the same SHA across 2 browsers. In the best case (no harness errors), no tests are added or deleted (so A and D are both empty). C should show the same as the default, since it's showing everything that's changed in some way. Since the UI treats no diff as 0/0/0, the U param on the API is sometimes useful (smaller response) but has no effect on the UI.

I suspect you expected D to be regressions? Maybe R as a feature request then :)

PR #617, opened 2018-10-03
Handle the standard test run query params in /api/diff

lukebjerring:
## Description
Ensures that params like `aligned` and `labels` are respected.

Also drops the base64 behaviour for the `before` and `after` params (a speculative, unused feature).

## Review Information
- Visit https://diff-dot-wptdashboard-staging.appspot.com/results?before=chrome[stable]&after=chrome[experimental]&aligned
- In another tab, visit https://diff-dot-wptdashboard-staging.appspot.com/results?product=chrome[stable]&product=chrome[experimental]&diff
- Find a row with a diff count that differs between the 2 diffs
- Ensure the same test's diff value is correctly calculated by https://diff-dot-wptdashboard-staging.appspot.com/api/diff (for the same params) for the row that would otherwise be wrong.
  - https://diff-dot-wptdashboard-staging.appspot.com/api/diff?before=chrome[stable]&after=chrome[experimental]&aligned
  - vs https://diff-dot-wptdashboard-staging.appspot.com/api/diff?before=chrome[stable]&after=chrome[experimental]

PR #729, opened 2018-11-06
Support fetching runs by their IDs

lukebjerring:
## Description
Follow-up from https://github.com/web-platform-tests/wpt.fyi/pull/617 

On `/api/diff`, this prevents the (unlikely) race-condition, and any potential discrepancies, between executing the same query on 2 different api endpoints.

### DIFF
--- a/api/diff.go
+++ b/api/diff.go
@@ -65,19 +65,16 @@ func handleAPIDiffGet(w http.ResponseWriter, r *http.Request) {
 	} else {
 		// NOTE: We use the same params as /results, but also support
 		// 'before' and 'after' and 'filter'.
-		runFilter, err := shared.ParseTestRunFilterParams(r)
-		if err != nil {
+		runFilter, parseErr := shared.ParseTestRunFilterParams(r)
+		if parseErr != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
 			return
 		}
-
-		var beforeAndAfter shared.ProductSpecs
-		beforeAndAfter, err = shared.ParseBeforeAndAfterParams(r)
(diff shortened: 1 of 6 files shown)


api/diff.go
```diff
@@ -45,18 +45,46 @@ func handleAPIDiffGet(w http.ResponseWriter, r *http.Request) {
...
+		}
+		runs, err = LoadTestRunsForFilters(ctx, runFilter)
 	}
 
-	if len(beforeAndAfter) > 0 {
-		runFilter.Products = beforeAndAfter
-	}
-	var runs shared.TestRuns
-	if runs, err = LoadTestRunsForFilters(ctx, runFilter); err != nil {
+	if err != nil {
```
mdittmer: nit: Let's move this up to the end of the `if len(runIDs) > 0 { ... }` block, since all `err` statements in its `else` use `..., err := ` (living inside the `else`'s block scope). Just this block can move; IIUC, the `else if len(runs) != 2` can become ` len(runs) != 2`.

webapp/components/wpt-results.html
```diff
@@ -343,7 +343,7 @@ <h3>
           },
           diffURL: {
             type: String,
-            computed: 'computeDiffURL(testRuns, query)',
+            computed: 'computeDiffURL(testRuns)',
```
mdittmer: Are changes in this file a bug fix? Change from `this.diffFilter` to `this.filter` without corresponding property changes seems non-trivial.

api/diff.go
lukebjerring: Renamed the local vars to `parseErr` to be more readable.

PR #739, opened 2018-11-08
Complete any check_suites when a run arrives

lukebjerring:
## Description
Leverages the `checks` package's API to trigger a `completed` status for any `CheckSuite` for the SHA of the newly-uploaded run.

Currently, not doing this, we end up with infinitely `in_progress` check_runs.

### DIFF
--- a/api/receiver/create_run.go
+++ b/api/receiver/create_run.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/web-platform-tests/wpt.fyi/api/checks"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
@@ -18,7 +19,7 @@ import (
 const InternalUsername = "_processor"
 
 // HandleResultsCreate handles the POST requests for creating test runs.
-func HandleResultsCreate(a AppEngineAPI, w http.ResponseWriter, r *http.Request) {
+func HandleResultsCreate(a AppEngineAPI, s checks.SuitesAPI, w http.ResponseWriter, r *http.Request) {
 	username, password, ok := r.BasicAuth()
 	if !ok || username != InternalUsername || !a.authenticateUploader(username, password) {
 		http.Error(w, "Authentication error", http.StatusUnauthorized)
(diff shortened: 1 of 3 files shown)


api/receiver/create_run_test.go
```diff
@@ -44,12 +45,14 @@ func TestHandleResultsCreate(t *testing.T) {
 	req.SetBasicAuth("_processor", "secret-token")
 	w := httptest.NewRecorder()
 	mockAE := NewMockAppEngineAPI(mockCtrl)
+	mockS := checks.NewMockSuitesAPI(mockCtrl)
 	gomock.InOrder(
 		mockAE.EXPECT().authenticateUploader("_processor", "secret-token").Return(true),
 		mockAE.EXPECT().addTestRun(gomock.Any()).Return(testDatastoreKey, nil),
+		mockS.EXPECT().CompleteCheckRun(gomock.Any(), gomock.Any()).Return(true, nil),
```
Hexcles: Let's be a little more rigorous here. We know the values of the two arguments and let's just expect them explicitly instead of using `gomock.Any`.

lukebjerring: Done.

api/receiver/create_run_test.go
```diff
@@ -26,15 +26,17 @@ func TestHandleResultsCreate(t *testing.T) {
...
+		"browser_name":       "firefox",
+		"browser_version":    "59.0",
+		"os_name":            "linux",
+		"os_version":         "4.4",
+		"revision":           sha[:10],
+		"full_revision_hash": sha,
+		"labels":             []string{"foo", "bar"},
+		"time_start":         "2018-06-21T18:39:54.218000+00:00",
+		"time_end":           "2018-06-21T20:03:49Z",
 		// Intentionally missing full_revision_hash; no error should be raised.
```
Hexcles: Please remove this line.

api/receiver/create_run.go
```diff
@@ -46,6 +48,17 @@ func HandleResultsCreate(a AppEngineAPI, s checks.SuitesAPI, w http.ResponseWrit
...
+	if len(testRun.FullRevisionHash) != 40 {
+		http.Error(w, "full_revision_hash must be the full SHA (40 chars)", http.StatusBadRequest)
+		return
+	} else if testRun.Revision != "" && strings.Index(testRun.FullRevisionHash, testRun.Revision) != 0 {
+		http.Error(w,
+			fmt.Sprintf("Mismatch of full_revision_hash and revision fields: %s vs %s", testRun.FullRevisionHash, testRun.Revision),
+			http.StatusBadRequest)
+		return
+	}
+	testRun.Revision = testRun.FullRevisionHash[:10]
```
Hexcles: Is this needed? You've already checked the consistency in the if condition above.

lukebjerring: In the case that it's empty, or a matching substring that isn't 10 chars long, this will correct it.

PR #742, opened 2018-11-08
Create pending checks when processing starts

lukebjerring:
## Description
Adds a hook to the `checks.SuitesAPI` for a pending run for the given product + sha when the taskcluster webhook successfully kicks off the processor for it.


api/webhook/taskcluster.go
```diff
@@ -152,7 +153,15 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
...
-	err = createAllRuns(log, urlfetch.Client(slowCtx), api, *status.SHA, username, password, urlsByBrowser, labels)
+	suitesAPI := checks.NewSuitesAPI(ctx)
+	err = createAllRuns(log,
+		urlfetch.Client(slowCtx),
+		suitesAPI,
+		uploadURL,
+		*status.SHA,
+		username,
+		password,
+		urlsByBrowser, labels)
```
Hexcles: Nit: be consistent about the styling; put `labels` into a new line.

api/webhook/taskcluster.go
```diff
@@ -264,9 +274,12 @@ func createAllRuns(log shared.Logger,
 		go func(browser string, urls []string) {
 			defer wg.Done()
 			log.Infof("Reports for %s: %v", browser, urls)
-			err := createRun(client, sha, api, username, password, urls, labels)
+			err := createRun(client, sha, uploadURL, username, password, urls, labels)
 			if err != nil {
 				errors <- err
+			} else {
+				browserName := strings.Split(browser, "-")[0] // chrome-dev => chrome
```
Hexcles: Why do you want to extract just the browser name and leave the channel aside? We could have chrome-dev & chrome-stable on the same commit.

api/webhook/taskcluster.go
```diff
@@ -264,9 +274,12 @@ func createAllRuns(log shared.Logger,
...
+				suitesAPI.PendingCheckRun(sha, browserName)
```
Hexcles: Do you have a rough idea how long this API call may take?

lukebjerring: Anecdotally, requests are quick (mostly <100ms, no total request times for requests that send these are over 1s)

api/checks/webhook.go
```diff
@@ -108,15 +108,17 @@ func handleCheckRunEvent(ctx context.Context, payload []byte) (bool, error) {
...
-	if checkRun.Action != nil &&
-		(*checkRun.Action == "created" || *checkRun.Action == "rerequested") {
-		name, sha := *checkRun.CheckRun.Name, *checkRun.CheckRun.HeadSHA
-		log.Debugf("Check run %s @ %s %s", name, sha[:7], *checkRun.Action)
-		spec, err := shared.ParseProductSpec(*checkRun.CheckRun.Name)
-		if err != nil {
-			log.Errorf("Failed to parse \"%s\" as product spec")
+	if checkRun.Action != nil {
+		if (*checkRun.Action == "created" && *checkRun.CheckRun.Status != "completed") ||
+			*checkRun.Action == "rerequested" {
```
Hexcles: FWIW, I don't actually know the semantics of these actions/statuses, so I'll trust your investigation.

api/checks/webhook.go
```diff
@@ -108,15 +108,17 @@ func handleCheckRunEvent(ctx context.Context, payload []byte) (bool, error) {
...
+			name, sha := *checkRun.CheckRun.Name, *checkRun.CheckRun.HeadSHA
+			log.Debugf("Check run %s @ %s %s", name, sha[:7], *checkRun.Action)
```
Hexcles: Nit: let's just log the full hash here.

api/checks/webhook.go
```diff
@@ -108,15 +108,17 @@ func handleCheckRunEvent(ctx context.Context, payload []byte) (bool, error) {
...
 	}
 
-	if checkRun.Action != nil &&
-		(*checkRun.Action == "created" || *checkRun.Action == "rerequested") {
-		name, sha := *checkRun.CheckRun.Name, *checkRun.CheckRun.HeadSHA
-		log.Debugf("Check run %s @ %s %s", name, sha[:7], *checkRun.Action)
-		spec, err := shared.ParseProductSpec(*checkRun.CheckRun.Name)
-		if err != nil {
-			log.Errorf("Failed to parse \"%s\" as product spec")
+	if checkRun.Action != nil {
```
Hexcles: Nit: we have too many levels of nesting here and this one feels like an error check, so let's convert this top-level if into an early return.

PR #746, opened 2018-11-09
Refactor results-processor & make it (almost) idempotent

### LINKED ISSUE #247: Make results receiver idempotent
Requirement: if a test runner uploads the same result or TaskQueue dispatches the same task multiple times, we shouldn't create any duplicated data in GCS or Datastore.

The GCS part is already satisfied. The paths for both the full report and the sharded JSONs are deterministic and unique (keyed by \<revision, product, SHA of the report\> since #233 ). And the transformation (splitting) process is also deterministic. Hence, even if the processor processes the same result multiple times, the final data in GCS will still be the same.

Datastore part isn't satisfied yet, because processor always invoke the `/api/run` API to create test runs and neither the processor or the test run API has any duplication detection. We can similarly leverage #233 and use \<revision, product, SHA of the report\> as a unique key in Datastore. The test run API should not create another run if the key already exists.

Lastly, an optional optimization is to skip a task altogether if the API says the key already exists.

Hexcles:
Please see each individual commit (and review them one by one).

Fixes #247 .

### DIFF
--- a/results-processor/gsutil.py
+++ b/results-processor/gsutil.py
@@ -22,6 +22,7 @@ def _call(command, quiet=False):
 
 
 def gs_to_public_url(gcs_path):
+    """Converts a gs:// URI to a HTTP URL."""
     assert gcs_path.startswith('gs://')
     return gcs_path.replace('gs://', 'https://storage.googleapis.com/', 1)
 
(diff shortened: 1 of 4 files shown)


Hexcles:
@lukebjerring thanks for the review. Yes your understanding is correct. I'll rebase-merge this PR.

PR #748, opened 2018-11-12
Only show 0/0/0 on filter=U

lukebjerring:
## Description
This improves the support of the `filter` param for filtering differences in the UI.

Changes 0 / 0 / 0 from being the default to not showing anything when "Unchanged" is not requested. For diffs which have a large number of deletes, e.g. `master` vs a PR, this improves the output when omitting the `D` filter.

## Review Information
Visit a PR diff, e.g. https://filter-dot-wptdashboard-staging.appspot.com/results/?diff=&product=chrome%5Bmaster%5D&product=chrome%407831c217d50a7e50a8c0d0323bb86e9a91d6e31e

Add `&filter=ACU` (no `D`) and see the improvement.


foolip:
Ran into https://github.com/web-platform-tests/wpt.fyi/issues/750 trying to compare two runs with more differences.

foolip:
Using https://wpt.fyi/results/?products=edge@a516e7edca,edge@b3ca0cee4b&diff instead.

foolip:
Oh, one of those runs wasn't submitted to staging. Trying again :)

foolip:
https://filter-dot-wptdashboard-staging.appspot.com/results/?products=firefox@a516e7edca,firefox@4176e9ef5f&diff works! Adding just `filter=A` doesn't work, will comment in code.

webapp/components/wpt-results.html
```diff
@@ -630,38 +627,58 @@ <h3>
...
+        }
+        const failingBefore = before.total - before.passes;
+        const failingAfter = after.total - after.passes;
+        const diff = [
+          Math.max(after.passes - before.passes, 0), // passes
+          Math.max(failingAfter - failingBefore, 0), // regressions
+          after.total - before.total // total
+        ];
+        const hasChanges = diff.some(v => v !== 0);
+        if (!this.diffFilter.includes('C') && hasChanges) {
```
foolip: I think this is the reason that just `filter=A` doesn't work now, where it did something before.

Can you add test coverage to make sure this doesn't regress?

foolip:
A thought that came up again while looking at this is that all of the counting will have to be replaced by deep diffing of results to make this correct. Hopefully not too much of this code will be invalidated by that change, but I suspect most of it will? That's not a reason to stop and wait, but what things have to fall into place before we can do more correct diffing that isn't based on counting/comparing subtest passes/totals?

lukebjerring:
At that stage, diffing would become an extra column in the results (/api/search) output, and its value would match the current format either way, so I don't envision incompatibility problems.

webapp/components/wpt-results.html
```diff
@@ -650,38 +647,59 @@ <h3>
...
+          after.total - before.total // total
+        ];
+        const hasChanges = diff.some(v => v !== 0);
+        if ((this.diffFilter.includes('A') && added)
+          || (this.diffFilter.includes('D') && deleted)
+          || (this.diffFilter.includes('C') && hasChanges)
+          || (this.diffFilter.includes('U') && !hasChanges)) {
+          return diff;
+        }
+        return undefined;
```
CYBAI: Maybe we might not need to write `undefined` explicity when `return` 👀?

lukebjerring:
```suggestion
        return;
```

foolip:
I see, so diff would move to the server side. That makes sense, and just means replacing a bunch of code, not some intricate changes to it.

webapp/components/wpt-results.html

PR #749, opened 2018-11-12
Ignore other check_suite apps

lukebjerring:
## Description
Currently we're trying to parse all check_run names as a product and erroring out, when we should be ignoring other App IDs.

### DIFF
--- a/api/checks/webhook.go
+++ b/api/checks/webhook.go
@@ -9,7 +9,6 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


api/checks/webhook.go
```diff
@@ -84,6 +87,12 @@ func handleCheckSuiteEvent(ctx context.Context, payload []byte) (bool, error) {
 		return false, err
 	}
 
+	appID := checkSuite.GetCheckSuite().GetApp().GetID()
```
Hexcles: Actually, one question: what if `App` is nil? Will this be deferencing a null pointer?

api/checks/webhook.go
```diff
@@ -128,13 +137,17 @@ func handleCheckRunEvent(ctx context.Context, payload []byte) (bool, error) {
 		return false, err
 	}
 
-	if checkRun.Action == nil {
-		return false, errors.New("No action present on the check_run event")
+	appID := checkRun.GetCheckRun().GetApp().GetID()
```
Hexcles: ditto

api/checks/webhook.go
lukebjerring: Designed to avoid nil checks:

`// GetApp returns the App field.
func (c *CheckRun) GetApp() *App {
	if c == nil {
		return nil
	}
	return c.App
}`

PR #751, opened 2018-11-13
Pass around specs in checks package

lukebjerring:
## Description
Changes from passing around a browser name to passing around a spec, to improve flexibility in the future.


shared/sharedtest/util.go
```diff
@@ -47,3 +48,26 @@ func NewTestContext() context.Context {
...
+	spec string
+}
+
+func (s sameProductSpec) Matches(x interface{}) bool {
+	if p, ok := x.(shared.ProductSpec); ok && p.String() == s.spec {
+		return true
+	} else if str, ok := x.(string); ok && str == s.spec {
+		return true
+	}
+	return false
```
foolip: Aren't there equivalent forms that stringify to different things that would be treated as different here? Whether that's ok or not, can you add a comment about what the consequences are when it happens?

Maybe tests have to be authored to avoid that case, simply?

lukebjerring: No; Some equivalent _queries_ might collapse to different specs, but that's fine. This is just verifying the spec value that gets piped through a mock. `spec.String()` is deterministic. Think of this as a deep-equals for a product spec, but the asserter (test author) doesn't need to build a spec, they can give the stringified version.

foolip: Alright, thanks :)

PR #754, opened 2018-11-13
When a check suite is created, detect PRs across forks

lukebjerring:
## Description
Detects when a check_suite event occurs for a SHA that's part of a PR that originates from another fork (but is created on the web-platform-tests/wpt repo), and requests a check-suite for the wpt repo too.

The one (other) case that we need to be wary of is the initial creation of the PR. A check_suite is created by GitHub when a new commit is pushed. This, of course, occurs on the forked repository, and there is also initially no associated PR at the time. Thus, when a PR is opened, we should ensure a check_suite on wpt for the PR's HeadSHA.  

## Review Information
Example check that used this code:
https://github.com/web-platform-tests/wpt/pull/14037/checks?sha=2ececcbafaa0b19cb5fac4e912c8709a7e295642

### DIFF
--- a/api/checks/webhook.go
+++ b/api/checks/webhook.go
@@ -28,6 +28,8 @@ import (
 
 // NOTE(lukebjerring): This is https://github.com/apps/staging-wpt-fyi-status-check
 const wptfyiCheckAppID = 19965
+const wptRepoID = 3618133
+const wptRepoInstallationID = 449270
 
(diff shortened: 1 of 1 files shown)


lukebjerring:
Ping. Regardless of whether this is WAI, or gets changed by GitHub, it's a necessary change for the check to work in the interim.

PR #758, opened 2018-11-14
Handle fetch errors a little better

lukebjerring:
## Description
Makes sure that we handle results fetching (from /api/search) failures in the UI.
Also tweaks the error message to include which run is being a problem.

### DIFF
--- a/webapp/components/loading-state.html
+++ b/webapp/components/loading-state.html
@@ -37,10 +37,14 @@ still being loaded (generally, fetched).
         }
       }
 
-      async load(promise) {
+      async load(promise, opt_errHandler) {
         this.loadingCount++;
         try {
           return await promise;
+        } catch (e) {
+          if (opt_errHandler) {
+            opt_errHandler(e);
+          }
         } finally {
           this.loadingCount--;
         }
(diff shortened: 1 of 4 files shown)


webapp/components/loading-state.html
```diff
@@ -37,10 +37,14 @@
         }
       }
 
-      async load(promise) {
+      async load(promise, opt_errHandler) {
```
LukeZielinski: maybe a naive question, but should there be a change to calling code to pass in opt_errHandler?

LukeZielinski: Cool thanks for pointing that out.

webapp/components/wpt-results.html
```diff
@@ -527,15 +530,20 @@ <h3>
           }
         }
 
-        this.load(window.fetch(url, fetchOpts).then(r => {
-          if (!r.ok || r.status !== 200) {
-            return Promise.reject('Failed to fetch results data.');
+        this.load(
+          window.fetch(url, fetchOpts).then(r => {
+            if (!r.ok || r.status !== 200) {
```
Hexcles: @lukebjerring Why does this throw on non-200 responses? `r.ok` means status is in [200,299], all of which should be considered successful.

PR #762, opened 2018-11-14
Add an admin-only endpoint to flush cache

Hexcles:
/admin/cache/flush


Hexcles:
Verified in staging that this is indeed admin-only. Merging to master.

I'll push a (non-default) version to prod so that we can flush the cache in prod.

PR #763, opened 2018-11-14
Add admin-only endpoint for toggling features environment-wide

lukebjerring:
## Description
This simplifies the current need to head to appengine and manually create an entity, into a simple "tick the box" (same style as /flags, but posts to the /admin/flags endpoint and affects datastore, not local storage.).

## Review Information
Head to /admin/flags, toggle a feature.

NOTE: This affects the whole staging environment, so revert your changes :)

### DIFF
--- a/webapp/admin_handler.go
+++ b/webapp/admin_handler.go
@@ -5,6 +5,9 @@
 package webapp
 
 import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
 	"net/http"
 
 	"github.com/web-platform-tests/wpt.fyi/api/receiver"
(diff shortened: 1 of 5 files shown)


webapp/admin_handler.go
```diff
@@ -20,8 +23,48 @@ func adminUploadHandler(w http.ResponseWriter, r *http.Request) {
...
+	} else if r.Method == "POST" {
+		if !a.IsAdmin() {
+			http.Error(w, "Admin only", http.StatusUnauthorized)
+			return
+		}
+		var flag shared.Flag
+		if bytes, err := ioutil.ReadAll(r.Body); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		} else if err = json.Unmarshal(bytes, &flag); err != nil {
```
Hexcles: Nit: all these if blocks have early return. I'd personally prefer getting rid of the redundant `else`s.

lukebjerring: So `bytes` won't exist outside the if unless I declare it first, which means it's like, an extra 4 lines to do it that way.

PR #766, opened 2018-11-15
Email on build failure

lukebjerring:
## Description
`master` build has failed a couple of times. Emails to that effect would be good (so the PR author can follow up).


lukebjerring:
Apparently this is roughly the default; @Hexcles did you get an email that master had failed?

Hexcles:
Yeah I've always been getting the emails, long before the change. I'm pretty sure this is just the default.

Note that the emails go to my personal account, so check yours.

PR #768, opened 2018-11-15
Perform sanity check before caching /api/search response

mdittmer:
## Description

- `/api/search`: Do not cache `{..., "results": []}`
- Consistent logging required adding `context.Context` to "should cache" decision function
- Bug fix: "is cacheable" should behave differently on `GET` (with query params) and `POST` (with JSON body) requests


mdittmer:
So that's fun: https://travis-ci.org/web-platform-tests/wpt.fyi/jobs/455566635#L2437

CC: @Hexcles

mdittmer:
@lukebjerring authored #805 to factor out logic and tests around reading run IDs.

PR #769, opened 2018-11-15
Add a query-builder to the /runs UI

lukebjerring:
## Description
Also adds support for `to` and `from` (required for the default query on that tab).

For picking the date, leverages https://vaadin.com/directory/component/vaadinvaadin-date-picker

## Review Information
Visit /runs, try a few different queries.


foolip:
Haven't looked at code, but I tried https://runs-builder-dot-wptdashboard-staging.appspot.com/runs a bit. I noticed that the URL isn't updated when submitting, which would be great for sharing links. Trivial to fix, I presume?

foolip:
On the nitpicky side, I see that "EDIT" isn't vertically aligned in the box. Needs line-height?

foolip:
An unrelated thing I noticed when looking for very old runs is that we don't show a year, so runs from 2017 might seem like they're from 2018 if one doesn't click through and check.

lukebjerring:
Resolved the year by including it when it's not the current year.

lukebjerring:
Regarding virtual nagivation: that's actually a little complicated and should be its own follow-up PR.

lukebjerring:
EDIT button is vertically aligned for me (ff / chrome both). What browser? Screenshot?

foolip:
I'm running Chrome on Mac. Screenshot:
screen shot 2018-11-16 at 5 59 27 pm

foolip:
> Resolved the year by including it when it's not the current year.

Looks like this also causes the date to not be shown? Did it take too much space? (Fine if that's the reason and not accidental.)

lukebjerring:
Try a cache-refresh?

https://runs-builder-dot-wptdashboard-staging.appspot.com/runs?from=2017-11-13T05%3A00%3A00.000Z&to=2018-01-13T05%3A00%3A00.000Z

foolip:
https://runs-builder-dot-wptdashboard-staging.appspot.com/runs?from=2017-11-13T05%3A00%3A00.000Z&to=2018-01-13T05%3A00%3A00.000Z works for me, looks like year is shown the first time it changes, like days.

.travis.yml
```diff
@@ -2,11 +2,6 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-notifications:
```
foolip: Accidental change? Shouldn't someone get email for failures?

webapp/components/test-runs-query.html
```diff
@@ -205,10 +211,17 @@
...
-            var diff = Math.floor((Date.now() - from) / 86400000);
-            return `Showing test runs from the last ${diff} days`;
+            const max = to || Date.now();
+            var diff = Math.floor((max - from) / 86400000);
+            const shortDate = date => date.toLocaleDateString(navigator.language, {
+              month: 'short',
+              day: 'numeric',
+            });
+            return to
+              ? `Showing test runs from ${shortDate(from)} to ${shortDate(to)}`
```
foolip: Including the year if different or not the current year here would be good. I got "Showing test runs from Dec 15 to Nov 21" where the start was 2016 and end 2018. In that case I'd expect the year to be shown for both, so for simplicity maybe always show year for both dates or neither.

webapp/components/test-runs-query.html
```diff
@@ -205,10 +211,17 @@
           this.setProperties(batchUpdate);
         }
 
-        computeResultsRangeMessage(from, maxCount) {
+        computeResultsRangeMessage(from, to, maxCount) {
           if (from) {
-            var diff = Math.floor((Date.now() - from) / 86400000);
-            return `Showing test runs from the last ${diff} days`;
+            const max = to || Date.now();
```
foolip: `max` is only used to compute `diff` and diff is only used if `to` is falsy, so the purpose of this line is hard to parse. Seems like it could be replaced with `const max = Date.now()`. Maybe `if (to) { /* compute and return stuff */ } else { /* other stuff */ }` to make this bit more clear?

webapp/components/wpt-results.html
```diff
@@ -628,6 +628,11 @@ <h3>
                     total: 0,
                   })),
                 };
+                row.diff = {
```
foolip: These and some following changes seem unrelated to this PR. Can you split out?

webapp/components/test/wpt-results.html
```diff
@@ -64,82 +64,6 @@
             });
           });
         });
-
-        suite('computeDifferences(before, after)', () => {
```
foolip: Why is this bunch of tests deleted? This isn't the component this PR is changing.

.travis.yml
lukebjerring: Bad merge, maybe. Reverted.

webapp/components/test-runs-query.html
lukebjerring: Done.

webapp/components/test/wpt-results.html
lukebjerring: Bad merge. reverted.

webapp/components/wpt-results.html
lukebjerring: Bad merge. reverted.

lukebjerring:
Something strange was going on so I reset to master and reverted the weirdness and forcepushed a squashed commit... PTAL

webapp/components/test-runs-query.html
lukebjerring: Done (when they differ)

lukebjerring:
I can't repro the bug you describe. There's a chance that you're hitting public-cache issues, so make sure you give it a cache clear and see if it's still an issue.

foolip:
Taking another look. Note that the alignment bug in https://github.com/web-platform-tests/wpt.fyi/pull/769#issuecomment-439458356 is still present.

webapp/components/test-runs-query.html
```diff
@@ -205,10 +211,23 @@
...
+              };
+              if (from.getYear() !== to.getYear()) {
+                opts.year = 'numeric';
+              }
+              return date.toLocaleDateString(navigator.language, opts);
+            };
+            if (to) {
+              return `Showing test runs from ${shortDate(from)} to ${shortDate(to)}`;
+            }
+            const days = Math.floor((Date.now() - from) / 86400000);
```
foolip: Thanks for changing this, much clearer!

foolip:
I can still repro the bug with devtools open and "Disable cache" checked. Exact steps:
1. Visit https://runs-builder-dot-wptdashboard-staging.appspot.com/runs, see both Chrome Dev and Chrome stable logos, both Firefox Nightly and Firefox stable, and so on.
1. Click "EDIT", add "stable" to the labels field and click submit.
1. Now you see fewer runs, just stable browser logos.
1. Click "EDIT", note that the labels field is empty, change nothing, and submit
1. Now you see the original view.

lukebjerring:
Bug should be fixed by #794

PR #771, opened 2018-11-16
Whitelist chromium-wpt-export-bot for wpt.fyi checks

lukebjerring:
## Description
The bot produces a bunch of exports from some CLs that worked in Chromium™ and are a good candidate for regression checks.


lukebjerring:
Eh, merging this. If either of you object, we can roll back.

lukebjerring:
Not yet, we don't report failure at all yet :)

foolip:
An even better worst case then :)

Issue #588, opened 2018-09-26
Allow /api/runs to enumerate runs for at least past 3 months


foolip:
[ecosystem-infra-stats](https://github.com/foolip/ecosystem-infra-stats/blob/bce9e3cd153eb2a7427abec1e05bc5878f093499/wpt-dashboard-stats.py#L10) uses /api/runs to compute latency stats for run uploads to wpt.fyi.

In https://github.com/web-platform-tests/wpt.fyi/pull/584 the limit is being lowered to 500 (or lower) and with Taskcluster runs soon producing results per push to master, this will not be enough to get all runs in the previous 3 months.

Ideally there should be an API to enumerate *all* runs ever, even if paginated, to make it easier to perform ad-hoc analysis by grabbing a bunch of runs. Failing that, slightly more than 3 months should suffice to compute latency quarterly, if the computation is done soon after the end of each quarter.

lukebjerring:
I believe this is now possible after #673, though it will still be up to a max of 500 of each product I think.

e.g. https://wpt.fyi/api/runs?from=2018-06-01T00%3A00%3A00.000Z

PR #772, opened 2018-11-16
Add `max-count` pagination ability to /api/runs

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/588

Hides the token behind a feature, but supports base64 encoded tokens of a TestRunFilter.

Currently, the next page only works for `MaxCount` (using an offset). Will follow up with some time-range based behaviour. 

## Review Information
- Head to /admin/flags, turn on the feature
- Visit /api/runs?max-count=1, observe the `x-wpt-next-page` header in the response headers
- Fetch /api/runs?page=[the token], see that you get the next runs (1 per product).

As for code review, I recommend you eyeball and overlook the first commit (a refactor that'll be needed later when combining pagination of time + max-count), and scrutinize the 2nd commit.

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -27,6 +27,7 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	var testRuns shared.TestRuns
+	var nextPageToken string
 	if len(ids) > 0 {
 		testRuns, err = ids.LoadTestRuns(ctx)
 		if multiError, ok := err.(appengine.MultiError); ok {
@@ -49,6 +50,15 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 		testRuns, err = LoadTestRunsForFilters(ctx, filters)
+
+		if err == nil {
+			if shared.IsFeatureEnabled(ctx, "paginationTokens") {
+				nextPage := filters.NextPage(testRuns)
+				if nextPage != nil {
+					nextPageToken, _ = nextPage.Token()
+				}
+			}
+		}
 	}
 
 	if err != nil {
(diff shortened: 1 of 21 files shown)


lukebjerring:
Yes, that's why I was refactoring. We need to know what was loaded; if you asked for 30 runs and returned less than 30 runs we can assume there's no more pages. 
I will add that logic as part of this PR then - I'd intended to do it separately

lukebjerring:
Added logic to stop returning tokens when result count(s) are less than the capped amount. PTAL.

foolip:
Trying to paginate to the end again I ran into a 500 at https://pagination-dot-wptdashboard-staging.appspot.com/api/runs?page=eyJzaGEiOiJsYXRlc3QiLCJtYXhjb3VudCI6NTAwLCJvZmZzZXQiOjI1MDB9

api/test_runs.go
```diff
@@ -14,6 +14,9 @@ import (
 	"google.golang.org/appengine/datastore"
 )
 
+const nextPageTokenHeaderName = "x-wpt-next-page"
```
foolip: https://stackoverflow.com/a/3561399 suggests no "x-" prefix.

shared/params.go
```diff
@@ -670,3 +676,20 @@ func ParseBeforeAndAfterParams(r *http.Request) (ProductSpecs, error) {
...
+	}
+	decoded, err := base64.URLEncoding.DecodeString(token)
+	if err != nil {
+		return nil, err
+	}
+	var filter TestRunFilter
+	if err := json.Unmarshal([]byte(decoded), &filter); err != nil {
+		return nil, err
+	}
+	return &filter, nil
```
foolip: Is there any malicious input which is valid JSON but which could cause trouble? Negative `MaxCount`, start times that come after end times, anything like that? I suspect that the worst that will happen is a 500? Do you think it's worth adding a comment? Also, does this code have write access in case of an exploit?

lukebjerring: It'd just 500; we aren't running an arbitrary user input as a full query, so no `-- drop table USER;` madness.

api/test_runs.go
lukebjerring: Done.

lukebjerring:
That last page is just the final boss. Very hard to unmarshal.

lukebjerring:
Turns out we had a couple of very old entities which were using a `name` ID key. Deleted, now things are happy.

foolip:
Neat! Very much looking forward to see this deployed to prod :)

PR #774, opened 2018-11-19
Avoid HTML escaping in md template

lukebjerring:
## Description
We're currently escaping when we shouldn't be.
(e.g. https://github.com/web-platform-tests/wpt/pull/14119/checks?check_run_id=33055714)

### DIFF
--- a/api/checks/summaries/compile_test.go
+++ b/api/checks/summaries/compile_test.go
@@ -16,7 +16,7 @@ func TestCompile_Completed(t *testing.T) {
 	foo := Completed{
 		HostName: "foo.com",
 		HostURL:  "foo.com/results/",
-		DiffURL:  "foo.com/diff/",
+		DiffURL:  "foo.com/diff?before=chrome[master]&after=chrome@0123456789",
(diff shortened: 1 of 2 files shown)


api/checks/summaries/compile_test.go
```diff
@@ -16,7 +16,7 @@ func TestCompile_Completed(t *testing.T) {
 	foo := Completed{
 		HostName: "foo.com",
 		HostURL:  "foo.com/results/",
-		DiffURL:  "foo.com/diff/",
+		DiffURL:  "foo.com/diff?before=chrome[master]&after=chrome@0123456789",
```
LukeZielinski: nit: can these be in alphabetical order?

Hexcles:
@lukebjerring sorry for the late response, but I wonder if we can just use `text/template` instead of `html/template` so that there's no auto escaping at all? Especially if we trust all the fields.

lukebjerring:
I am clearly not very effective at searching for golang packages...
https://github.com/web-platform-tests/wpt.fyi/pull/780

PR #776, opened 2018-11-19
Add an endpoint for computing the wpt.fyi check

lukebjerring:
## Description
Actually performs a diff and detects whether anything has regressed, giving a failure outcome in the case that it has.

This endpoint will be scheduled on a TaskQueue by the `create_run` handler, so that we don't block waiting for the results fetch and computation.


api/checks/api.go
```diff
@@ -49,3 +53,7 @@ func (s suitesAPIImpl) PendingCheckRun(sha string, product shared.ProductSpec) (
 func (s suitesAPIImpl) CompleteCheckRun(sha string, product shared.ProductSpec) (bool, error) {
 	return completeCheckRun(s.ctx, sha, product)
 }
+
+func (s suitesAPIImpl) ScheduleResultsProcessing(sha string, product shared.ProductSpec) error {
+	return scheduleResultsProcessing(s.ctx, sha, product)
```
Hexcles: I don't see `scheduleResultsProcessing` itself being used anywhere else. Can we just embed the implementation here in this exported method?

api/checks/routes.go
```diff
@@ -10,4 +10,7 @@ import "github.com/web-platform-tests/wpt.fyi/shared"
 func RegisterRoutes() {
 	// GitHub webhook for creating custom status checks.
 	shared.AddRoute("/api/webhook/check", "api-webhook-check", checkWebhookHandler)
+
+	// Endpoint for computing outcome and updating any checks for the given commit.
```
Hexcles: Can we make it clear that this is the endpoint of the TaskQueue?

api/checks/update.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"google.golang.org/appengine"
+)
+
+// updateCheckHandler handles /api/checks/[commit] POST requests.
+func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	sha := vars["commit"]
+	if len(sha) != 40 {
+		http.Error(w, fmt.Sprintf("Invalid commit: %s", sha), http.StatusBadRequest)
```
Hexcles: So one tip about designing the endpoint of TaskQueue: if the error can't be recovered by retrying (e.g. invalid parameters), you'd need to return a 2XX status code to "pretend" a success otherwise the task will be retried forever and gets piled up.

api/checks/suites.go
```diff
@@ -121,3 +129,15 @@ func completeCheckRun(ctx context.Context, sha string, product shared.ProductSpe
 	}
 	return true, nil
 }
+
+func scheduleResultsProcessing(ctx context.Context, sha string, product shared.ProductSpec) error {
+	host := shared.GetHostname(ctx)
+	target, _ := url.Parse(fmt.Sprintf("https://%s/api/checks/%s", host, sha))
+	q := target.Query()
+	q.Set("product", url.QueryEscape(product.String()))
```
Hexcles: IIRC `Values.Encode` will take care of encoding the values, so this is a double-encoding.

Please add a test case to verify.

api/checks/api.go
```diff
@@ -25,6 +25,10 @@ type SuitesAPI interface {
 	// Returns true if any check_runs were created (i.e. any CheckSuite entities were
 	// found, and the create succeeded).
 	CompleteCheckRun(sha string, browser shared.ProductSpec) (bool, error)
+
+	// ScheduleResultsProcessing adds a webhook to TaskQueue for the given sha and
```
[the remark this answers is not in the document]
lukebjerring:
```suggestion
	// ScheduleResultsProcessing adds a URL callback to TaskQueue for the given sha and
```

api/checks/update.go
lukebjerring: We can also explicitly set a reasonable `task_retry_limit`, e.g. 3?

api/checks/suites.go
lukebjerring: I was using this quite wrong anyway; cleanup makes it redundant.

api/receiver/appengine.go
```diff
@@ -48,26 +44,22 @@ type AppEngineAPI interface {
 
 // appEngineAPIImpl is backed by real AppEngine APIs.
 type appEngineAPIImpl struct {
-	ctx   context.Context
+	shared.AppEngineAPI
```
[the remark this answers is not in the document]
lukebjerring: Done.

Hexcles:
@lukebjerring can you fix the merge conflicts and Travis errors?

api/checks/api_test.go
```diff
@@ -0,0 +1,40 @@
+//+build small
```
Hexcles: This should be a medium test because of the use of dev_appserver. And the file needs to be renamed.

webapp/queue.yaml
```diff
@@ -3,3 +3,8 @@ queue:
   target: processor
   max_concurrent_requests: 2
   rate: 1/m
+queue:
+- name: check-processing
+  target: processor
+  retry_parameters:
+    task_retry_limit: 3
```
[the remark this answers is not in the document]
lukebjerring: Done.

shared/util.go
```diff
@@ -27,6 +27,9 @@ const StableLabel = "stable"
 // i.e. run from the master branch.
 const MasterLabel = "master"
 
+// CreateTestRunPath is the path for POST creation on a TestRun.
+const CreateTestRunPath = "/api/results/create"
```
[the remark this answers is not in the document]
lukebjerring: Nowhere - reverted.

api/checks/api.go
```diff
@@ -25,19 +29,29 @@ type SuitesAPI interface {
...
+
+type suitesAPIImpl struct {
+	ctx   context.Context
+	queue string
 }
 
 // NewSuitesAPI returns a real implementation of the SuitesAPI
 func NewSuitesAPI(ctx context.Context) SuitesAPI {
+	queue := ""
+	if !appengine.IsDevAppServer() {
```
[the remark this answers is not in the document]
lukebjerring: Done.

api/checks/api_test.go
lukebjerring: Done.

api/checks/update.go
```diff
@@ -0,0 +1,150 @@
...
+		return
+	}
+
+	product, _ := shared.ParseProductSpec(runs[0].Product.String())
+	checkState := summaries.CheckState{
+		Product:    product,
+		HeadSHA:    sha,
+		Title:      getCheckTitle(product),
+		DetailsURL: getMasterDiffURL(ctx, sha, product),
+		Status:     "in_progress",
```
Hexcles: @lukebjerring stumbled upon this again and I'm wondering shouldn't this be "completed"?

PR #777, opened 2018-11-20
Log reason for ignoring taskcluster

lukebjerring:
## Description
Trying to diagnose why checks aren't being finished.


api/taskcluster/webhook.go
```diff
@@ -74,10 +74,12 @@ func tcWebhookHandler(w http.ResponseWriter, r *http.Request) {
 }
 
 // https://developer.github.com/v3/activity/events/types/#statusevent
-type statusEventPayload github.StatusEvent
+type statusEventPayload struct {
```
lukebjerring: type aliasing causes the struct's methods to get dropped, yet encapsulation does not...

PR #780, opened 2018-11-20
Avoid HTML escaping in md template properly

lukebjerring:
Reverts web-platform-tests/wpt.fyi#774


Hexcles:
Travis failed because of nondeterministic product ordering (being fixed in #781 ). Force-merge.

PR #781, opened 2018-11-20
Fix product order preservation

lukebjerring:
## Description
Tests are flaky from the non-deterministic ranging over a map, mistakenly done in #772 

Luckily, most callers were using `AllRuns` already, so this isn't too bad of a follow-up fix.


lukebjerring:
Given that this resolves a broken `master` branch, I'm going to merge as TBR

Issue #784, opened 2018-11-20
Support listing runs by PR number


lukebjerring:
Because PRs are transient, and GitHub is the source of truth, it would be inherently problematic to store PR metadata ourselves.

However, getting a list of runs associated with a PR would often be quite helpful.

Hexcles:
It'd be great to also have the reverse lookup: finding out which PR a commit belongs to (if it is part of a PR) so that we can provide a back link to GitHub PRs from wpt.fyi when showing the regression view.

lukebjerring:
That latter case is covered by linking to the commit on GitHub, which surfaces a link to any PR on the commit's page.

lukebjerring:
Getting a PR from a commit is covered by #900 (head to the commit on GitHub, which will surface the PR).

PR #785, opened 2018-11-20
Add feature for listing runs for a PR

lukebjerring:
## Description
First steps towards https://github.com/web-platform-tests/wpt.fyi/issues/784

Adds `runsByPRNumber` feature, which will run a GitHub API fetch for commits for the PR number and list all runs pertaining to all SHAs in the PR.

## Review Information
NOTE - won't work on the UI as a result of this PR, needs a follow-up.

- Head to /admin/flags, ensure feature is enabled
- Load /api/runs?pr=[PR number]
  - Needs to be a whitelisted user, e.g. `autofoolip`
  - e.g. [/api/runs?pr=14143](https://prs-dot-wptdashboard-staging.appspot.com/api/runs?pr=14143)

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -9,7 +9,9 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"github.com/google/go-github/github"
 	"github.com/web-platform-tests/wpt.fyi/shared"
+	"golang.org/x/oauth2"
 	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
 )
@@ -29,6 +31,11 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	pr, err := shared.ParsePRParam(r)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+	}
+
 	var testRuns shared.TestRuns
 	var nextPageToken string
 	if len(ids) > 0 {
@@ -45,6 +52,9 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 				err = nil
 			}
 		}
+	} else if pr != nil && shared.IsFeatureEnabled(ctx, "runsByPRNumber") {
+		commits := getPRCommits(ctx, *pr)
+		testRuns, err = shared.LoadTestRunsBySHAs(ctx, commits...)
 	} else {
 		var filters shared.TestRunFilter
 		filters, err = shared.ParseTestRunFilterParams(r)
@@ -129,4 +139,27 @@ func LoadTestRunsForFilters(ctx context.Context, filters shared.TestRunFilter) (
 		return nil, err
 	}
 	return shared.LoadTestRunsByKeys(ctx, keys)
+}
+
+func getPRCommits(ctx context.Context, pr int) []string {
+	log := shared.GetLogger(ctx)
+	secret, err := shared.GetSecret(ctx, "github-api-token")
+	if err != nil {
+		log.Debugf("Failed to load github-api-token: %s", err.Error())
+		return nil
+	}
+	oauthClient := oauth2.NewClient(ctx, oauth2.StaticTokenSource(&oauth2.Token{
+		AccessToken: secret,
+	}))
+	githubClient := github.NewClient(oauthClient)
+	commits, _, err := githubClient.PullRequests.ListCommits(ctx, "web-platform-tests", "wpt", pr, nil)
+	if err != nil || commits == nil {
+		log.Errorf("Failed to fetch PR #%v: %s", pr, err.Error())
+		return nil
+	}
+	shas := make([]string, len(commits))
+	for i := range commits {
+		shas[i] = commits[i].GetSHA()
+	}
+	return shas
 }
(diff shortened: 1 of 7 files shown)


lukebjerring:
Failing tests are covered in #781

api/test_runs.go
```diff
@@ -130,3 +140,26 @@ func LoadTestRunsForFilters(ctx context.Context, filters shared.TestRunFilter) (
...
+	secret, err := shared.GetSecret(ctx, "github-api-token")
+	if err != nil {
+		log.Debugf("Failed to load github-api-token: %s", err.Error())
+		return nil
+	}
+	oauthClient := oauth2.NewClient(ctx, oauth2.StaticTokenSource(&oauth2.Token{
+		AccessToken: secret,
+	}))
+	githubClient := github.NewClient(oauthClient)
+	commits, _, err := githubClient.PullRequests.ListCommits(ctx, "web-platform-tests", "wpt", pr, nil)
```
foolip: What's the default limit on number of commits returned? The last returned is the most recent and interesting, so if it falls outside the window that'd be bad. Maybe just set the limit to 100 and fail if the number of returned commits is 100 so we notice it? It's really only when people have accidentally pushed a rebase of master or something that you get that many commits.

Alternatively, paginate and get *all* commits, however many they are.

api/test_runs.go
```diff
@@ -130,3 +140,26 @@ func LoadTestRunsForFilters(ctx context.Context, filters shared.TestRunFilter) (
...
+	if err != nil || commits == nil {
+		log.Errorf("Failed to fetch PR #%v: %s", pr, err.Error())
+		return nil
+	}
+	shas := make([]string, len(commits))
+	for i := range commits {
+		shas[i] = commits[i].GetSHA()
```
foolip: Here I wondered https://www.reddit.com/r/golang/comments/1m25a1/map_reduce_and_filter_in_go/, guess not :)

shared/datastore.go
```diff
@@ -25,6 +25,23 @@ func LoadTestRun(ctx context.Context, id int64) (*TestRun, error) {
 	return &testRun, nil
 }
 
+// LoadTestRunsBySHAs loads all test runs that belong to any of the given revisions (SHAs).
+func LoadTestRunsBySHAs(ctx context.Context, shas ...string) (runs TestRuns, err error) {
+	for _, sha := range shas {
+		if len(sha) > 10 {
+			sha = sha[:10]
```
foolip: Aren't we storing the full sha in datastore now? Can we avoid the shortening? (Not because ambiguity is a high risk, but because it'd be less code.)

shared/datastore.go
```diff
@@ -25,6 +25,23 @@ func LoadTestRun(ctx context.Context, id int64) (*TestRun, error) {
...
+		}
+		var shaRuns TestRuns
+		keys, err := datastore.NewQuery("TestRun").Filter("Revision =", sha).GetAll(ctx, &shaRuns)
+		if err != nil {
+			return runs, err
```
foolip: Should the partial list of runs be returned here? Just because this isn't nullable, or what's the convention? (Sorry still a golang noob.)

shared/datastore.go
```diff
@@ -25,6 +25,23 @@ func LoadTestRun(ctx context.Context, id int64) (*TestRun, error) {
...
+		}
+		shaRuns.SetTestRunIDs(keys)
+		runs = append(runs, shaRuns...)
+	}
+	return runs, err
```
foolip: Same sort of style question here, should we return err when we know it's nil?

lukebjerring: Nothing wrong with doing that (and it's preferred to return values (including nil) for all params)

shared/datastore.go
lukebjerring: A partial result _should_ be ignored by the caller, but sometimes it's useful to pass the results that actually loaded prior to the error.

shared/datastore.go
lukebjerring: We are, but it's not indexed. See #49

api/test_runs.go
lukebjerring: Eh, for simplicity, I'll wait until we actually see this as a real-world problem.

shared/datastore.go
foolip: Ah, thanks.

api/test_runs.go
foolip: If the limit is unknown, waiting until a bug happens and is reported by a user seems too far in the iterative direction. When bugs happen, getting the fix into prod can take time even when the fix is trivial. At least let's understand the limit and fail in an explicit way so that the problem is trivially debugged when it happens.

https://developer.github.com/v3/pulls/#list-commits-on-a-pull-request claims that up to 250 commits can be listed. https://github.com/foolip/wpt/pull/5 had 68 commits, but https://api.github.com/repos/foolip/wpt/pulls/5/commits only lists 30. I don't know where the canonical documentation is, but https://octokit.github.io/rest.js/#api-Pulls-listCommits also says default 30, max 100.

I tweaked a script I had to checking on my Azure Pipelines statuses, and 30 is low enough that we have plenty of PRs exceeding it: 13499, 12026, 6467, 1242. That was for open PRs only, presumably lots of merged ones have >30 too.

Can you file an issue to follow up?

lukebjerring: As per https://developer.github.com/v3/#pagination, setting the limit to 100.
I'll also file an issue for paginating; we can paginate on wpt.fyi and plumb "Offset: 100" to mean "GitHub Page 2".
https://github.com/web-platform-tests/wpt.fyi/pull/792
https://github.com/web-platform-tests/wpt.fyi/issues/793

PR #788, opened 2018-11-21
Support callback_url from processor

lukebjerring:
## Description
Adds an optional `callback_url` field for the /api/results/upload endpoint, which dictates where we post the `TestRun` to. This is necessary when we're integration-testing a staging instance with modified behaviour of the `/api/results/create` endpoint (i.e. for the wpt.fyi checks).


api/taskcluster/webhook.go
```diff
@@ -326,6 +337,9 @@ func createRun(client *http.Client, sha, api string, username string, password s
 	if labels != nil {
 		payload.Add("labels", strings.Join(labels, ","))
 	}
+	// Ensure we call back to this appengine version instance.
+	host := appengine.DefaultVersionHostname(ctx)
```
Hexcles: I don't think this does what your comment says: it's the *default* version, not this (the current) version.

Hexcles: What about `shared.GetHostname`?

lukebjerring: That's what I had intended to call, done.

shared/datastore.go
```diff
@@ -1,14 +1,14 @@
 package shared
 
 import (
+	"context"
```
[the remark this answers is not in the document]
Hexcles: These are good to have. `context` was experimental but has graduated to stable long ago; I always change `golang.org/x/net/context` to `context` whenever I see one.

PR #794, opened 2018-11-23
QueryBuilder: Handle user-entry semantic labels on submit

lukebjerring:
## Description
Handles an edge-case where the user enters a semantic label (e.g. `stable`) manually as a label, rather than clicking the dropdowns 4 times.

## Review Information
As per https://github.com/web-platform-tests/wpt.fyi/pull/769#issuecomment-440844710 
- Visit /runs
- Edit the query, type a semantic label (e.g. a channel, like `stable`)
- Submit, see the expected results
- Edit the query, type a different semantic label (e.g. `experimental`)

### DIFF
--- a/webapp/components/test-runs-query-builder.html
+++ b/webapp/components/test-runs-query-builder.html
@@ -162,9 +162,7 @@ found in the LICENSE file.
           this.handleAddProduct();
         };
         this.clearAll = this.handleClearAll.bind(this);
-        this.submit = () => {
-          this.onSubmit && this.onSubmit();
-        };
+        this.submit = this.handleSubmit.bind(this);
         this._createMethodObserver('labelsUpdated(labels, labels.*)');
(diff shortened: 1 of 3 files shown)


webapp/components/test-runs-query-builder.html
```diff
@@ -181,6 +179,22 @@ <h3>
...
 
+      handleSubmit() {
+        // Handle the edge-case that the user typed a label for channel or source, etc.
+        const productBuilders = this.shadowRoot.querySelectorAll('product-builder');
+        for (const semantic of window.wpt.SemanticLabels) {
+          for (const label of semantic.values) {
+            if (this.labels.includes(label)) {
+              this.labels = this.labels.filter(l => l !== label);
+              for (const p of productBuilders) {
+                p[semantic.property] = label;
```
foolip: Nice, a pretty deep loop, but all makes sense without comments to explain it :)

PR #799, opened 2018-11-23
TBR: Use get to avoid ValueError

lukebjerring:
> From Werkzeug 0.3 onwards, the KeyError raised by this class is also a subclass of the BadRequest HTTP exception and will render a page for a 400 BAD REQUEST if caught in a catch-all for HTTP exceptions.

For the callers to the processor that aren't providing the value, this is causing 400s.


lukebjerring:
Merged as we have 24h of 400s backlogged on the processor queue.

PR #803, opened 2018-11-24
TBR: Parse /api/checks/{commit} body as Form

lukebjerring:
## Description
The more simple API for a TaskQueue POST dumps given params in the payload; we were trying to parse them from the query params.

This PR refactors the `Parse*Param` helpers to take the `url.Values` objects directly, instead of the `*http.Request`. It also uses `ParseForm()` and passes that for /api/checks/{commit}, so that non-empty POST body as url params in the same manner.


lukebjerring:
I've verified this stops the 400's for all the `/api/checks/{commit}` requests we were seeing, so I'm going to TBR it and merge.

PR #808, opened 2018-11-26
Script + code boilerplate for /api/search/cache

mdittmer:
This change adds boilerplate for a new service for handling structured test result queries. For now, the new handler simply returns `Not implemented`.

This API is intended for internal use, as a delegate to the public `/api/search` API.

### DIFF
--- a/api/query/cache/service/app.yaml
+++ b/api/query/cache/service/app.yaml
@@ -4,9 +4,9 @@ service: searchcache
 
 # Should match something at https://cloud.google.com/compute/docs/machine-types.
 resources:
-  # n1-standard-64 (TODO: Would prefer n1-highmem-96 or larger ultramem)
-  cpu: 64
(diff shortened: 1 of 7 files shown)


api/query/cache/service/app.yaml
```diff
@@ -0,0 +1,15 @@
+runtime: custom
+env: flex
+service: searchcache
+
+# Should match something at https://cloud.google.com/compute/docs/machine-types.
+resources:
+  # n1-standard-96 (TODO: Would prefer n1-highmem-96)
+  cpu: 96
+  memory_gb: 360
```
Hexcles: Just a sanity check: you intend to use automatic scaling (by CPU utilization), right?

mdittmer: Good point. We should probably manually scale to 2 instances to smooth over restarts.

PR #809, opened 2018-11-26
Handle rerequested checks by scheduling computation again

lukebjerring:
## Description
Cleans up and collapses two code branches that (should) do the same thing.
Now when a check_suite (or run) is rerequested, we add the check(s) to the TaskQueue for computing the outcome.


api/checks/webhook.go
```diff
@@ -212,18 +212,19 @@ func handlePullRequestEvent(ctx context.Context, payload []byte) (bool, error) {
...
 		return false, fmt.Errorf("Failed to load test runs: %s", err.Error())
 	}
 	createdSome := false
+	api := NewSuitesAPI(ctx)
 	for _, rbp := range runsByProduct {
 		if len(rbp.TestRuns) > 0 {
-			created, err := completeCheckRun(ctx, sha, rbp.Product)
-			createdSome = createdSome || created
+			err := api.ScheduleResultsProcessing(sha, rbp.Product)
+			createdSome = createdSome || err == nil
```
Hexcles: Hmm, I don't think you need `createdSome` any more. In fact, this method probably only need to return `error` now that you defer the processing to TaskQueue.

api/checks/webhook.go
```diff
@@ -266,19 +268,10 @@ func createCheckRun(ctx context.Context, suite shared.CheckSuite, opts github.Cr
...
-	}
-	req.Header.Set("Accept", "application/vnd.github.antiope-preview+json")
-	checkRun := new(github.CheckRun)
-	resp, err := client.Do(ctx, req, checkRun)
 
+	checkRun, resp, err := client.Checks.CreateCheckRun(ctx, suite.Owner, suite.Repo, opts)
 	if err != nil {
 		log.Warningf("Failed to create check_run: %s", resp.Status)
-		body, _ := ioutil.ReadAll(resp.Body)
-		log.Warningf(string(body))
```
Hexcles: Is this logging not useful?

api/taskcluster/webhook_test.go
```diff
@@ -130,6 +130,9 @@ func TestCreateAllRuns_success(t *testing.T) {
 
 	mockC := gomock.NewController(t)
 	suitesAPI := checks.NewMockSuitesAPI(mockC)
+	suitesAPI.EXPECT().GetSuitesForSHA(sha).Return([]shared.CheckSuite{
+		shared.CheckSuite{SHA: sha},
+	}, nil)
 	suitesAPI.EXPECT().PendingCheckRun(sha, "chrome")
```
Hexcles: Hmm, how did this pass? `PendingCheckRun` now takes a suite. Did you forget to regenerate the mock?

lukebjerring: Actually, it takes an array of suites to update... so "this is fine". I'll update the test :)

api/checks/webhook.go
lukebjerring: No, it was a debugging effort that didn't get cleaned up.

api/checks/webhook.go
lukebjerring: It affects the content of the HTTP response (as to whether anything was scheduled or not).

PR #810, opened 2018-11-26
In-memory results cache: Test & implement monitoring for in-memory index

mdittmer:


api/query/cache/index/index.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+var (
+	errNoRuns = errors.New("No runs")
+	errNilRun = errors.New("Test run is nil")
+)
+
+type byTimeStart []*shared.TestRun
```
lukebjerring: nit: Dump this in `shared` as `type ByTimeStart TestRuns`

api/query/cache/index/index.go
```diff
@@ -0,0 +1,67 @@
...
+	// IngestRun loads the test run results associated with the input test run
+	// into the index.
+	IngestRun(*shared.TestRun) error
+	// EvictAnyRun reduces memory pressure by evicting the cache's choice of run
+	// from memory.
+	EvictAnyRun() error
+}
+
+type wptIndex struct {
+	runs []*shared.TestRun
```
lukebjerring:
```suggestion
	runs shared.TestRuns
```

api/query/cache/index/index.go
```diff
@@ -0,0 +1,67 @@
...
+	if r == nil {
+		return errNilRun
+	}
+
+	i.runs = append(i.runs, r)
+	sort.Sort(byTimeStart(i.runs))
+	return nil
+}
+
+func (i *wptIndex) EvictAnyRun() error {
```
lukebjerring: Maybe this should be EvictFirstRun (based on impl?), or randomized?
```suggestion
func (i *wptIndex) EvictFirstRun() error {
```

api/query/cache/monitor/monitor_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+func getTestHarness(t *testing.T) (*index.MockIndex, *MockRuntime, Monitor) {
+	mockCtrl := gomock.NewController(t)
+	idx := index.NewMockIndex(mockCtrl)
+	rt := NewMockRuntime(mockCtrl)
+	mon := NewIndexMonitor(shared.NewNilLogger(), rt, time.Microsecond, 1, idx)
+	return idx, rt, mon
```
lukebjerring: I think you're missing a call to `mockCtrl.Finish`; you may need to return `mockCtrl` to the caller.

api/query/cache/index/index.go
mdittmer: Could do. My intent was to make it clear that the client doesn't care what run gets evicted, just that a run _is_ evicted to relieve memory pressure.

mdittmer: Didn't address this comment yet. @lukebjerring WDYT of this rationale?

api/query/cache/index/index.go
mdittmer: Turns out `shared.TestRuns` orders by `TimeStart` by default. I'll just use that.

api/query/cache/index/index.go
lukebjerring: Yeah, agreed that `Any` is fine.

PR #811, opened 2018-11-26
Document expected downtime

jugglinmike:
wpt.fyi is expected to occasionally go offline. Users should be informed
of this expectation so they can respond appropriately.


jugglinmike:
It's been a while since I contributed to this codebase. Thanks for setting me straight! I've moved the note to the relevant section of this document.

PR #813, opened 2018-11-27
Refactor: Add DiffAPI for testability

lukebjerring:
## Description
No-op refactor that introduces a `DiffAPI` abstraction to supporting adding a range of tests to various currently-untested areas of the Golang codebase.

## Review Information
For the most-part, just verifying that the diff views are still working should suffice for smoke-testing.

### DIFF
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -113,7 +113,7 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, before, a
 		HeadSHA:    before.FullRevisionHash,
 		Title:      getCheckTitle(product),
 		DetailsURL: diffAPI.GetMasterDiffURL(before.FullRevisionHash, product),
-		Status:     "in_progress",
+		Status:     "completed",
 	}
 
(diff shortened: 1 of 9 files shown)


api/checks/update.go
```diff
@@ -115,30 +100,42 @@ func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
...
+		return nil, err
+	}
+
+	product, _ := shared.ParseProductSpec(before.Product.String())
+	checkState := summaries.CheckState{
+		Product:    product,
+		HeadSHA:    before.FullRevisionHash,
+		Title:      getCheckTitle(product),
+		DetailsURL: diffAPI.GetMasterDiffURL(before.FullRevisionHash, product),
+		Status:     "in_progress",
```
Hexcles: IIUC, this is the final step of this custom check, where we surface regression to GitHub, so should this be "completed" instead of "in_progress"?

lukebjerring: Ha, good catch ( must have been like this already ). 
```suggestion
		Status:     "completed",
```

Issue #734, opened 2018-11-07
Unclear how to paginate /api/runs?aligned


foolip:
When using the `aligned` parameter, such as with https://wpt.fyi/api/runs?label=stable&products=chrome,edge,firefox,safari&aligned&max-count=50, it's not clear how one would paginate to get the next set of runs at an earlier time.

Using the earliest `time_start` of all the runs returns as the next `to` value probably roughly works, but if the start times of a set of aligned runs happens to straddle that time it seems like it would be skipped. It seems like one would have to add a "buffer" amount of time like 1 week to avoid this, and then remove duplicates. However, one then also assumes that no more than `max-count` runs could happen in that buffer time, and would probably have a serious bug (infinite loop) if it ever did.

foolip:
I think a fix for this would be to use the commit time, since then the time would be the same for all runs for the same commit, and there is no time window issue. @Hexcles?

s3ththompson:
Rather than move pagination to userspace, is it possible that the API could just share some pagination middleware among all Datastore resource endpoints? E.g. each response would have the next URL to request in the response body?

lukebjerring:
That's pretty much exactly my plan.. lol. 

While a little unusual, I'm thinking that having it (it being the next page token) as an HTTP header on the response will help avoid a backward incompatible change to the payload.

lukebjerring:
So, while max-count pagination was introduced in #772, there's still a bug with fetching aligned runs (it caps the number of iterated keys, not the number of results).

lukebjerring:
Apart from a know issue #819 this is now using the standard page-token flow.

PR #815, opened 2018-11-27
Fix max-count behaviour for aligned runs

lukebjerring:
## Description
Changes the capping/limit from key iteration to results, and adds support for the offset param (which will cause pagination to behave correctly too).

Needed for https://github.com/web-platform-tests/wpt.fyi/issues/734

## Review Information
- Visit https://aligned-dot-wptdashboard-staging.appspot.com/runs?aligned&max-count=500
- See more than (the current, buggy outcome of) 5 results
- Be impressed by how many aligned runs we actually have

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -224,10 +224,14 @@ func GetAlignedRunSHAs(
 	labels mapset.Set,
 	from,
 	to *time.Time,
-	limit *int) (shas []string, keys map[string]KeysByProduct, err error) {
+	limit *int,
+	offset *int) (shas []string, keys map[string]KeysByProduct, err error) {
+	if limit == nil {
+		maxMax := MaxCountMaxValue
+		limit = &maxMax
+	}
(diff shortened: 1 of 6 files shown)


foolip:
> Be impressed by how many aligned runs we actually have

We're doing pretty OK lately. Not reliably one a day, but usually :)

shared/datastore.go
```diff
@@ -224,10 +224,14 @@ func GetAlignedRunSHAs(
 	labels mapset.Set,
 	from,
 	to *time.Time,
-	limit *int) (shas []string, keys map[string]KeysByProduct, err error) {
+	limit *int,
+	offset *int) (shas []string, keys map[string]KeysByProduct, err error) {
+	if limit == nil {
+		maxMax := MaxCountMaxValue
```
foolip: Who said naming was hard? Ha! (I enjoyed this name.)

Issue #636, opened 2018-10-11
Expose full datastore contents publicly (via API or static dump)


foolip:
In addition to the data in our GCS buckets, entries in datastore are needed by the wpt.fyi frontend.

In order to facilitate offline data analysis, this data should be made readily available, perhaps via https://wpt.fyi/api/runs, or perhaps in some other form. As a secondary motivation, it's also important that it's possible to fork wpt.fyi with the full historical data.

Context: @s3ththompson will be doing some data analysis for Bocoup, generating cool graphs and inspiration for us all.

Hexcles:
Whenever a "run" is returned by any of our APIs, it should already contain all the information (all the fields) in Datastore.

@s3ththompson what information are you looking for?

foolip:
> Does /api/runs not provide the necessary information already?

Possibly, if it can be paginated without missing anything due to changes in the interim. Is that possible?

lukebjerring:
Hmmmm - yes, "paginate" with the `to` param being set to the timestamp of the last item in the previously returned result.

foolip:
OK, that sort of pagination sounds like it should work. Can we document how to do it in this repo for forkability?

s3ththompson:
@foolip @lukebjerring that should be enough for me.

So to clarify, all wpt.fyi data for runs is accessible through https://wpt.fyi/api/runs & https://wpt.fyi/api/interop and consists of 4 separate "sources":
1. Datastore metadata that includes test run time information, labels, and a unique run ID (am I missing any other metadata?)
2. The full report for the run, including test and subtest results (>100mb), stored at `gs://wptd-results`
3. An offline-generated summary for the run including just pass/fail numbers for subtests (~2mb), stored at `gs://wptd`
4. An offline-generated summary of interop rates among browsers (~4mb), stored at `gs://wptd-metrics`

It looks like paging through the runs & interop api is the easiest way to download the data for offline analysis (rather than rsyncing the `gs://` buckets)...

One more question: I see that summaries are generated here: https://github.com/web-platform-tests/wpt.fyi/blob/master/results-processor/wptreport.py#L243 are interop rates still computed from https://github.com/web-platform-tests/results-analysis? Is there a benefit to re-computing historical interop rates based on the latest results-analysis image?

s3ththompson:
Also, I'm planning to contribute some documentation once I successfully scrape the right data.

foolip:
[...]

PR #816, opened 2018-11-27
Add time-based pagination

lukebjerring:
## Description
Recognizes the `from` param as pageable, and thus returns a `wpt-next-page` header on /api/runs?from=[RFC3339]`

Part of https://github.com/web-platform-tests/wpt.fyi/issues/636

## Review Information
- Visit /runs with the devtools open
- See that the request to /api/runs has a `wpt-next-page` header (token)
- Load /api/runs?page=[token]

### DIFF
--- a/shared/test_run_filter_test.go
+++ b/shared/test_run_filter_test.go
@@ -1,0 +1,54 @@
+// +build small
+
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTestRunFilter_NextPage_MaxCount(t *testing.T) {
+	ten := 10
+	filter := TestRunFilter{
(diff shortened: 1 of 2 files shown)


shared/test_run_filter_test.go
```diff
@@ -0,0 +1,54 @@
...
+			TestRuns: make(TestRuns, 10),
+		},
+	}
+	assert.Equal(t, &TestRunFilter{
+		MaxCount: &ten,
+		Offset:   &ten,
+	}, filter.NextPage(loadedRuns))
+}
+
+func TestTestRunFilter_NextPage_From(t *testing.T) {
```
foolip: Can you also add a test for paginating aligned runs where the range of times of the individual runs in each aligned group overlap? I expect that will not work without some extra work, because some of the runs that will be needed to discover the next aligned run are excluded by the filtering. Given how slow (>24 h) Edge runs have sometimes been I expect this has happened and that some aligned runs would be missed in pagination because of it.

lukebjerring: I think the simplest solution here is to include the run in the time-range page that does include the first run for the SHA. I'll add a test and implement that.

PR #817, opened 2018-11-27
Add load more button

lukebjerring:
## Description
Adds a "load more" button to the bottom of the /runs list when a `wpt-next-page` header is returned from the /api/runs fetch.

## Review Information
- Visit https://more-dot-wptdashboard-staging.appspot.com/runs?max-count=10
- Click "Load more"
- See another 10 runs per-product

NOTE: The UI currently merges `max-count` with the default `from` param of 1 week, so it won't load more than one week's worth of runs.

### DIFF
--- a/webapp/components/test-runs.html
+++ b/webapp/components/test-runs.html
@@ -35,6 +35,7 @@ found in the LICENSE file.
               type: Boolean,
               computed: 'computePathIsATestFile(path)'
             },
+            nextPageToken: String,
           };
         }
 
@@ -65,8 +66,13 @@ found in the LICENSE file.
           }
           // Fetch by products.
           if (this.productSpecs && this.productSpecs.length) {
-            runs.push(fetch(`/api/runs${this.query}`)
-              .then(r => r.ok && r.json()));
+            runs.push(
+              fetch(`/api/runs${this.query}`)
+                .then(r => r.ok && r.json().then(runs => {
+                  this.nextPageToken = r.headers.get('wpt-next-page');
+                  return runs;
+                }))
+            );
           }
           const fetches = await Promise.all(runs);
 
@@ -77,6 +83,22 @@ found in the LICENSE file.
           }, []);
           this.testRuns = flattened;
           return flattened;
+        }
+
+        /**
+         * Fetch the next page of runs, using nextPageToken, if applicable.
+         */
+        loadMoreRuns() {
+          if (!this.nextPageToken) {
+            return;
+          }
+          const url = new URL('/api/runs', window.location);
+          url.searchParams.set('page', this.nextPageToken);
+          this.nextPageToken = null;
+          return fetch(url).then(r => r.ok && r.json().then(runs => {
+            this.splice('testRuns', this.testRuns.length - 1, 0, ...runs);
+            this.nextPageToken = r.headers.get('wpt-next-page');
+          }));
         }
 
         splitPathIntoLinkedParts(inputPath) {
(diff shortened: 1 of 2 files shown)


webapp/components/test-runs.html
```diff
@@ -65,8 +66,13 @@
           }
           // Fetch by products.
           if (this.productSpecs && this.productSpecs.length) {
-            runs.push(fetch(`/api/runs${this.query}`)
-              .then(r => r.ok && r.json()));
+            runs.push(
+              fetch(`/api/runs${this.query}`)
+                .then(r => r.ok && r.json().then(runs => {
```
Hexcles: I saw your earlier change #758 which throws an exception if fetch fails. Shall we modify this (and `loadMoreRuns`) to throw/reject if `!r.ok` by the way?

webapp/components/wpt-runs.html
```diff
@@ -218,58 +232,61 @@
...
-                return oldest;
-              }, new Date()); // Existing runs should be historical...
-          };
-          const flattened = Object.entries(testRunsBySha)
-            .map(([sha, runs]) => ({ sha, runs, firstRunDate: firstRunDate(runs) }))
-            .sort((a, b) => b.firstRunDate.getTime() - a.firstRunDate.getTime());
+      testRunsLoaded(testRuns) {
+        let browsers = new Set();
+        // Group the runs by their revision/SHA
+        let testRunsBySha = testRuns.reduce((accum, results) => {
```
Hexcles: Now that `this.testRuns` is renamed to `this.testRunsBySha`, we should rename this local variable to avoid confusion as they are not the same (`this.testRunsBySha` is the flattened `testRunsBySha`).

webapp/components/wpt-runs.html
```diff
@@ -218,58 +232,61 @@
...
-            if (!accum[results.revision]) {
-              accum[results.revision] = {};
-            }
-            if (!accum[results.revision][results.browser_name]) {
-              accum[results.revision][results.browser_name] = [];
-            }
-            accum[results.revision][results.browser_name].push(results);
-            return accum;
-          }, {});
+        return this.load(this.loadRuns());
```
Hexcles: I don't see much benefit keeping this method now. In fact, it's probably too much abstraction/indirection (especially too many `load*` functions around here). We can just replace all `this.loadData` with `this.load(this.loadRuns())`, which is more consistent with `this.load(this.loadMoreRuns())`.

webapp/components/test-runs.html
lukebjerring: No; a 404 is legitimate since the query can be user-input. We combine things in the `Promise.all` below, and I wouldn't want multiple sources to cancel each other because one 404s.

webapp/components/wpt-runs.html
lukebjerring: Renamed.

webapp/components/wpt-runs.html
lukebjerring: Done.

lukebjerring:
@LukeZielinski - while it seems strange, that is actually the expected behaviour because our Edge/Safari runs are _far_ less frequent. The max-count param is applied per-product, and as noted in the description,

> NOTE: The UI currently merges max-count with the default from param of 1 week, so it won't load more than one week's worth of runs.

For edge/safari, this is roughly 7 runs, while the other products have many more runs to load inside the 1 week timespan.

PR #822, opened 2018-11-28
Add "Tests" part of search cache index

mdittmer:
The search cache contains shards that service queries in parallel. Each shard contains with two main parts:

1. A maping of test IDs and test names
1. A mapping of test run IDs and test IDs and result IDs

This change implements the first of these. It is a port of [this file](https://github.com/web-platform-tests/data-migration/blob/mem/grid/mem/tests.go) from the proof-of-concept implementation.

CC @KyleJu, who may end up working on a directory tree-based implementation of this and other components.

### DIFF
--- a/api/query/cache/index/tests_test.go
+++ b/api/query/cache/index/tests_test.go
@@ -21,16 +21,15 @@ func TestGetName_fail(t *testing.T) {
 func TestAddGetName(t *testing.T) {
 	ts := NewTests()
 	name := "/a/b/c"
-	var subName *string
-	id, err := ts.Add(name, subName)
+	id, err := ts.Add(name, nil)
 	assert.Nil(t, err)
 	actualName, actualSubName, err := ts.GetName(id)
 	assert.Nil(t, err)
 	assert.Equal(t, name, actualName)
-	assert.Equal(t, subName, actualSubName)
+	assert.Nil(t, actualSubName)
 
 	subNameValue := "some sub name"
-	subName = &subNameValue
+	subName := &subNameValue
 	id, err = ts.Add(name, subName)
 	assert.Nil(t, err)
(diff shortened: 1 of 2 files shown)


api/query/cache/index/tests_test.go
```diff
@@ -0,0 +1,40 @@
...
+func TestGetName_fail(t *testing.T) {
+	ts := NewTests()
+	_, _, err := ts.GetName(TestID{})
+	assert.NotNil(t, err)
+}
+
+func TestAddGetName(t *testing.T) {
+	ts := NewTests()
+	name := "/a/b/c"
+	var subName *string
```
lukebjerring:
```suggestion
```

api/query/cache/index/tests_test.go
```diff
@@ -0,0 +1,40 @@
...
+	id, err := ts.Add(name, subName)
+	assert.Nil(t, err)
+	actualName, actualSubName, err := ts.GetName(id)
+	assert.Nil(t, err)
+	assert.Equal(t, name, actualName)
+	assert.Equal(t, subName, actualSubName)
+
+	subNameValue := "some sub name"
+	subName = &subNameValue
```
lukebjerring:
```suggestion
	subName := &subNameValue
```

api/query/cache/index/tests_test.go
```diff
@@ -0,0 +1,40 @@
...
+	ts := NewTests()
+	_, _, err := ts.GetName(TestID{})
+	assert.NotNil(t, err)
+}
+
+func TestAddGetName(t *testing.T) {
+	ts := NewTests()
+	name := "/a/b/c"
+	var subName *string
+	id, err := ts.Add(name, subName)
```
lukebjerring:
```suggestion
	id, err := ts.Add(name, nil)
```

PR #823, opened 2018-11-28
Fix wpt check_suite creation logic

lukebjerring:
## Description
Fixes a bug with checking that the `err` is nil (wrong comparator).


lukebjerring:
TBR; minor bugfix.

PR #827, opened 2018-11-29
Use only browser name in check name

lukebjerring:
## Description
We name the pending check after the browser name, but the completed check is being named after the full product spec for the TestRun, which leaves the differently-named pending runs hanging.


lukebjerring:
TBR'ing this simple change.

PR #829, opened 2018-11-29
Add actions to regressed results

lukebjerring:
## Description
Exposes a (currently probably ignored) action button for regression results.
GitHub documents that these fields are accepted, but not how they manifest when the user actually clicks them.

### DIFF
--- a/api/checks/summaries/compile.go
+++ b/api/checks/summaries/compile.go
@@ -7,6 +7,7 @@ import (
 	"runtime"
 	"text/template"
 
+	"github.com/lukebjerring/go-github/github"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
@@ -23,6 +24,9 @@ type Summary interface {
 	// GetCheckState returns the info needed to update a check.
 	GetCheckState() CheckState
 
+	// GetActions returns the actions that can be taken by the user.
+	GetActions() []*github.CheckRunAction
+
 	// GetSummary compiles the summary markdown template.
 	GetSummary() (string, error)
 }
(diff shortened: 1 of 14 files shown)


lukebjerring:
An example of the buttons showing in the UI:
https://github.com/lukebjerring/wpt/pull/8/checks?check_run_id=36318705

foolip:
I can't see the buttons in https://github.com/lukebjerring/wpt/pull/8/checks?check_run_id=36318705 and that's probably because I don't have write access to the repo. I'll just imagine what they look like, I've seen them before :)

api/checks/summaries/compile.go
```diff
@@ -114,3 +129,19 @@ func (r Regressed) GetCheckState() CheckState {
...
+	return []*github.CheckRunAction{
+		&github.CheckRunAction{
+			Identifier:  "recompute",
+			Label:       "Recompute",
+			Description: "Recompute against the latest master run",
+		},
+		&github.CheckRunAction{
+			Identifier:  "ignore",
+			Label:       "Ignore",
+			Description: "Mark these results as expected (passing)",
```
foolip: Not sure I understand the "(passing)" bit means. It will be for failing tests that this button even appears, right?

api/checks/webhook.go
```diff
@@ -177,18 +179,32 @@ func handleCheckRunEvent(aeAPI shared.AppEngineAPI, suitesAPI SuitesAPI, payload
...
-		if err != nil {
-			log.Errorf("Failed to parse \"%s\" as product spec", checkRun.GetCheckRun().GetName())
-			return false, err
+		shouldSchedule = true
+	} else if action == "requested_action" {
+		actionID := checkRun.GetRequestedAction().Identifier
+		if actionID == "recompute" {
+			shouldSchedule = true
+		} else if actionID == "ignore" {
+			// TODO(lukebjerring): Created IgnoredRegression summary.
```
foolip: It will be glorious!

api/checks/summaries/compile.go
lukebjerring: If you don't, others might not either. Yes, these buttons are only shown for a "Regressed" outcome.

It's supposed to imply that the currently-failing check would then be marked as passing.

PR #830, opened 2018-11-29
Test & implement search cache index stores all run+test+result data

mdittmer:
- `Index` stores `Tests` and shards of `Results`
- `Index` carefully manages repeated and/or concurrent attempts to add same run
- Test new `Index` functionality + medium test for "run eviction relieves memory pressure"

CC @KyleJu

### DIFF
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -5,15 +5,27 @@
 package index
 
 import (
+	"encoding/json"
 	"errors"
+	"fmt"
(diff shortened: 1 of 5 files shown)


mdittmer:
Re: sharding maps

This allows each shard to iterate over the tests that it is responsible for by construction, without having to keep an extra copy of all test IDs in memory. When queries are serviced, each shard will be responsible for its own goroutine that will iterate over the maps owned by that shard.

PR #833, opened 2018-11-29
Add AppID field

lukebjerring:
## Description
We will need to extract the AppID instead of fixing it to the staging.wpt.fyi App once we deploy to prod, but also needed for a testing copy of the App used in #829.


lukebjerring:
TBR'ing to stop the 500's in staging.

PR #834, opened 2018-11-29
Tweak logging for JWT client creation failure

lukebjerring:
## Description
The creation of JWT client is failing in some instances with a different AppID, and this will help diagnose those failures.


api/checks/webhook.go
```diff
@@ -267,7 +264,9 @@ func createWPTCheckSuite(ctx context.Context, appID, installationID int64, sha s
 		HeadSHA: sha,
 	}
 	suite, _, err := client.Checks.CreateCheckSuite(ctx, wptRepoOwner, wptRepoName, opts)
-	if err == nil && suite != nil {
+	if err != nil {
+		log.Errorf("Failed to create GitHub check suite: %s", err.Error())
+	} else if suite != nil {
```
LukeZielinski: nit: I'm assuming you can't have both err and suite be set at the same time? Although removing the "else" and just having two separate "ifs" might be more resilient if that assumption ever changes.

Issue #835, opened 2018-11-29
One subtask failure on Taskcluster causes all results to be dropped


lukebjerring:
When we process Taskcluster status, we ignore any non-success status.

However, subtasks can be successful in some cases.
e.g. https://tools.taskcluster.net/groups/Gr1qsGRiR6mOu4vVSV-Xvg
chrome failed, but FF was fine. We shouldn't drop FF on the floor just because Chrome failed.

lukebjerring:
For more debug info, this was from lukebjerring/wpt @ 1f006b18b336eef01f8616c5576763bacd6a4bad
https://github.com/lukebjerring/wpt/pull/8/commits/1f006b18b336eef01f8616c5576763bacd6a4bad

PR #836, opened 2018-11-29
Log task group id

lukebjerring:
## Description
Will help with debugging for https://github.com/web-platform-tests/wpt.fyi/issues/835


Hexcles:
And in fact, I'd prefer to still have one logging message that once there's a match, the processing should happen in this request.

lukebjerring:
We log any case that we choose to skip, and any case that we error; I really don't think we need to also log the fact that non-return-cases weren't hit.

PR #837, opened 2018-11-29
Add an instance to the processor

lukebjerring:
## Description
We've had a drastic increase in processing needs now that we handle both master and PRs.


lukebjerring:
Redundant after #838

PR #838, opened 2018-11-29
Speed up results processor by using cp instead of rsync

Hexcles:
## Description

We don't really need to synchronize the directories. `rsync` used to save us time when the results already exist, but such scenario is becoming less likely following #746. Therefore, we can switch to `cp` which uses less CPU in general, and is easier to parallelize.

Based on monitoring, we only use ~1MB/s network bandwidth during the recursive upload, which is far less than the available bandwidth between GCE & GCS, while on the other hand we use almost 100% CPU. In other words, we are CPU-bound because of the sheer amount of small files. With this PR, we are able to bring down the average upload time (for a full run) from ~30min to ~5min.

Additional drive-by changes:
1. Fix the logging setup that's broken by google.cloud imports.
2. Use logging.warning instead of the deprecated logging.warn.

## Review

This change has been tested online in staging with careful monitoring.

### DIFF
--- a/results-processor/gsutil.py
+++ b/results-processor/gsutil.py
@@ -9,42 +9,15 @@ import subprocess
 _log = logging.getLogger(__name__)
 
 
-def _call(command, quiet=False):
-    _log.info('EXEC%s: %s',
-              '(quiet)' if quiet else '',
-              ' '.join(command))
-    if quiet:
-        subprocess.check_call(command,
-                              stdout=subprocess.DEVNULL,
-                              stderr=subprocess.DEVNULL)
-    else:
-        subprocess.check_call(command)
+def _call(command):
+    _log.info('EXEC: %s', ' '.join(command))
+    subprocess.check_call(command)
 
 
 def gs_to_public_url(gcs_path):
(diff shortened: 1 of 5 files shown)


results-processor/gsutil.py
```diff
@@ -55,7 +35,7 @@ def copy(path1, path2, gzipped=False, quiet=False):
         gzipped: Whether path1 is gzipped (if True, 'Content-Encoding:gzip'
             will be added to the headers).
     """
-    command = ['gsutil', '-m']
+    command = ['gsutil', '-q', '-m']
```
lukebjerring: Shouldn't this be conditional on `quiet` param?

Hexcles:
Sorry I'm doing manual testing. The PR isn't really ready yet. (Also logging isn't quite behaving as intended.)

Hexcles:
@lukebjerring this is now finally completed. I'm pretty confident from testing and monitoring the change in staging for a couple of days. Processing time goes from 30min down to 5min, not quite 10x, but it's still a pretty good improvement :)

PR #840, opened 2018-11-30
Implement the checks "ignore" user-action

lukebjerring:
## Description
Following up from #829 

Adds behaviour for the ignore action, which prepends a message about who marked it passing, and changes the status to `success`.


api/checks/webhook.go
```diff
@@ -26,8 +26,7 @@ const (
 
 func isKnownAppID(appID int64) bool {
 	switch appID {
-	case wptfyiCheckAppID:
-	case checksStagingAppID:
+	case wptfyiCheckAppID, checksStagingAppID:
```
lukebjerring: @Hexcles @foolip @mdittmer 
A cautionary Golang tale is told in this little change. Fallthrough is not a thing, and `break` isn't needed.

    switch foo {
    case "bar", "baz":
        // "bar" or "baz" will do these lines
    case "qux":
    case "quux":
        // "quux" hits this line, but "qux" is actually an empty clause.

Issue #611, opened 2018-10-02
Maybe: add dev and nightly labels to Chrome Dev and Firefox Nightly runs respectively


foolip:
Looking at https://wpt.fyi/test-runs I trying to create URL like this one:
https://wpt.fyi/results/?products=chrome[stable],chrome[dev]&aligned=true

That's because I see "dev" right there in the logo. But we only add the experimental label.

It would be great if the dev label was also applied so that this works. Similarly for Firefox.

Hexcles:
First of all, I don't think we support multiple advanced product specs in `products`, i.e. if you have a bunch of `browser[label]`, you'd need to specify them via multiple `product` parameters.

Secondly, there's some confusion about the labels here. We (tentatively) define the "experimental" channel as cutting edge + experimental features, so all the Chrome Dev and Firefox Nightly runs we have right now are experimental, as they all have experimental features enabled.

If we are to add the "dev" label to a run, the run will need to be cutting edge **without** experimental features. See #538 cc @lukebjerring .

foolip:
> First of all, I don't think we support multiple advanced product specs in `products`, i.e. if you have a bunch of `browser[label]`, you'd need to specify them via multiple `product` parameters.

That does seem to work, though. Just by accident?

> See #538

Commented there.

lukebjerring:
The issue was actually with nested commas - a single label was fine. Both now work, though, since #669

lukebjerring:
@Hexcles - the proposal from @foolip is to include extra alias-labels for the `dev` channel which match how the products name their own channel-equivalents.

Hexcles:
Re-opening because we might also want to re-label all the existing runs.

foolip:
I guess this also hasn't been deployed yet?

Hexcles:
No & no.

foolip:
I tried that, but like your link it showed no results.

foolip:
@Hexcles you reopened once before to do that, do you still think we should?

Hexcles:
Sorry for late response. I didn't patch the historic data because the change to add these new labels wasn't pushed to prod yet. And now that we had a prod push, I'll batch label all previous runs.

Hexcles:
Retroactive tagging in progress on staging; will continue to prod after verification.

Hexcles:
This is done.

foolip:
Thanks @Hexcles!

PR #841, opened 2018-11-30
Include channel in checks

lukebjerring:
## Description
There [can be cases](https://github.com/lukebjerring/wpt/pull/8/checks?check_run_id=36553443) where the latest master run was `stable` - we should avoid making that comparison.

Mapping list is copied from the results-processor one, which is in python code.


Hexcles:
And please adjust `TestCreateAllRuns_success` to make sure this is tested.

shared/util.go
```diff
@@ -30,6 +30,24 @@ const StableLabel = "stable"
 // i.e. run from the master branch.
 const MasterLabel = "master"
 
+// ProductChannelToLabel maps known product-specific channel names
+// to the wpt.fyi model's equivalent.
+func ProductChannelToLabel(channel string) string {
```
Hexcles: BTW, I just sent out #844 , which will make this mapping no longer necessary.

lukebjerring: More accurately, not necessary after https://github.com/web-platform-tests/wpt.fyi/issues/611 is closed (because the data is backfilled).

PR #842, opened 2018-11-30
Don't create a pending check unless create succeeds

lukebjerring:
## Description
Prevents checks hanging in a pending status, when they were never going to succeed.
e.g. https://github.com/web-platform-tests/wpt/pull/14309

### DIFF
--- a/api/taskcluster/webhook_test.go
+++ b/api/taskcluster/webhook_test.go
@@ -176,7 +176,7 @@ func TestCreateAllRuns_one_error(t *testing.T) {
 	suitesAPI := checks.NewMockSuitesAPI(mockC)
 	suite := shared.CheckSuite{SHA: sha}
 	suitesAPI.EXPECT().GetSuitesForSHA(sha).Return([]shared.CheckSuite{suite}, nil)
-	suitesAPI.EXPECT().PendingCheckRun(suite, sharedtest.SameProductSpec("firefox"))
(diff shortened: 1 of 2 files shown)


api/taskcluster/webhook_test.go
```diff
@@ -176,7 +176,6 @@ func TestCreateAllRuns_one_error(t *testing.T) {
 	suitesAPI := checks.NewMockSuitesAPI(mockC)
 	suite := shared.CheckSuite{SHA: sha}
 	suitesAPI.EXPECT().GetSuitesForSHA(sha).Return([]shared.CheckSuite{suite}, nil)
-	suitesAPI.EXPECT().PendingCheckRun(suite, sharedtest.SameProductSpec("chrome"))
 	suitesAPI.EXPECT().PendingCheckRun(suite, sharedtest.SameProductSpec("firefox"))
```
Hexcles: Sadly, this is non-deterministic. The mock server is set up to respond OK to the first request, but we don't know which goroutine runs first and hence we can't guarantee whether the first one would be chrome or firefox. I think it's a reasonable compromise to use `gomock.Any` here.

lukebjerring:
```suggestion
	suitesAPI.EXPECT().PendingCheckRun(suite, gomock.Any())
```

lukebjerring:
Merging under the assumption that no other changes were needed, since this is actively causing pending checks.

PR #843, opened 2018-11-30
Implement search cache backfilling strategy

mdittmer:
- `FillIndex(...)` fills index until either:
  - `EvictAnyRun()` is invoked on `Index` due to memory pressure, or
  - `Stop()` is invoked on index's `Monitor`


api/query/cache/backfill/backfill.go
```diff
@@ -0,0 +1,149 @@
...
+	backfilling bool
+}
+
+type backfillMonitor struct {
+	monitor.ProxyMonitor
+
+	idx *backfillIndex
+}
+
+const bytesPerRun = uint64(1.45e+8)
```
lukebjerring:
```suggestion
// bytesPerRun is [EDIT ME: a magically conjured number meaning something]
const bytesPerRun = uint64(1.45e+8)
```

api/query/cache/index/index.go
```diff
@@ -39,6 +39,25 @@ type Index interface {
 	EvictAnyRun() error
 }
 
+// ProxyIndex is a proxy implementation of the Index interface.
```
lukebjerring: This comment doesn't really explain anything. Proxy of/to what?

PR #846, opened 2018-12-03
Add cancel action for pending checks

lukebjerring:
## Description
In some error cases, we might get stuck with a forever pending check. This supports bailing out of those cases, or abandoning a check early if we know no tests are affected.

Example of a forever-pending check:
https://github.com/web-platform-tests/wpt/runs/36396227


lukebjerring:
Will follow up with a timeout in another PR.

PR #848, opened 2018-12-03
Search API: Test & implement query interpretation

mdittmer:
- Interpret `query.AbstractQuery` (formerly `query.query) as `query.ConcreteQuery`
  - This constitutes binding an unmarshaled query to a set of runs
- Export `ConcreteQuery` implementations so that query service strategies can reason about their constituent parts

This work is towards the search cache, but multiple cache search indexing strategies (in different packages) may be implemented to service instances of `ConcreteQuery`. Hence the code lives in `query`, not `cache`, and must be exported.

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -29,10 +29,13 @@ type RunQuery struct {
 	AbstractQuery
 }
 
+// TestNamePattern is a query atom that matches test names to a pattern string.
 type TestNamePattern struct {
 	Pattern string
 }
 
+// BindToRuns for TestNamePattern is a no-op: TestNamePattern implements both
+// AbstractQuery and ConcreteQuery because it is independent of test runs.
 func (tnp TestNamePattern) BindToRuns(runs []shared.TestRun) ConcreteQuery {
 	return tnp
 }
@@ -149,6 +152,8 @@ func (rq *RunQuery) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// UnmarshalJSON for TestNamePattern attempts to interpret a query atom as as
+// {"pattern":<test name pattern string>}.
 func (tnp *TestNamePattern) UnmarshalJSON(b []byte) error {
 	var data struct {
 		Pattern string `json:"pattern"`
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -15,25 +15,26 @@ import (
 
 var browsers = shared.GetDefaultBrowserNames()
 
-type query interface {
-	toPlan(runIDs []int64) plan
+// AbstractQuery is an intermetidate representation of a test results query that
+//  has not been bound to specific shared.TestRun specs for processing.
+type AbstractQuery interface {
+	BindToRuns(runs []shared.TestRun) ConcreteQuery
 }
 
-type runQuery struct {
+// RunQuery is the internal representation of a query recieved from an HTTP
+// client, including the IDs of the test runs to query, and the structured query
+// to run.
+type RunQuery struct {
 	runIDs []int64
-	query
+	AbstractQuery
 }
 
-func (rq runQuery) toPlan() plan {
-	return rq.query.toPlan(rq.runIDs)
+type TestNamePattern struct {
+	Pattern string
 }
 
-type testNamePattern struct {
-	pattern string
-}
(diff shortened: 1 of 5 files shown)


api/query/atoms.go
```diff
@@ -15,61 +15,118 @@ import (
...
 
 type testStatusConstraint struct {
 	browserName string
 	status      int64
 }
 
-func (tsc testStatusConstraint) toPlan(runIDs []int64) plan {
-	return nil
+func (tsc testStatusConstraint) BindToRuns(runs []shared.TestRun) ConcreteQuery {
+	ids := make([]int64, 0, len(runs))
```
lukebjerring: Nits:
- 1) Remind me whether there was a reason not to `make([]int64, len(runs))` and just set `ids[i] = run.ID`
- 2) (makes 1) redundant) there's a method on `shared.TestRuns` for getting the IDs. Might be called `TestRunIDs()` ?

api/query/atoms.go
```diff
@@ -15,61 +15,118 @@ import (
...
 
 type and struct {
-	and []query
+	and []AbstractQuery
 }
 
-func (a and) toPlan(runIDs []int64) plan {
-	return nil
+func (a and) BindToRuns(runs []shared.TestRun) ConcreteQuery {
+	q := And{make([]ConcreteQuery, 0, len(a.and))}
```
lukebjerring: I feel this is a bit hard to read; can you name the field in the `And` struct that you're make-ing?

api/query/atoms_test.go
```diff
@@ -185,18 +185,343 @@ func TestStructuredQuery_nested(t *testing.T) {
...
+
+func TestStructuredQuery_bindStatusNoRuns(t *testing.T) {
+	assert.Equal(t, True{}, testStatusConstraint{
+		browserName: "Chrome",
+		status:      1,
+	}.BindToRuns([]shared.TestRun{}))
+}
+
+func TestStructuredQuery_bindStatusSingleRun(t *testing.T) {
+	assert.Equal(t, RunTestStatusConstraint{
```
lukebjerring: nit: assign `expected` and `actual` params to locally well-named variables so the reader can tell what should be equal.
I think (?) that `firstRun` and `firefox` make sense. But I'm not sure.. hence the request ;)

api/query/atoms_test.go
```diff
@@ -185,18 +185,343 @@ func TestStructuredQuery_nested(t *testing.T) {
...
+				Product: shared.Product{
+					BrowserName: "Chrome",
+				},
+			},
+		},
+	}))
+}
+
+func TestStructuredQuery_bindStatusSomeRuns(t *testing.T) {
+	assert.Equal(t, Or{
```
lukebjerring: Same here; maybe `firstAndThird`, `firefoxRuns` ?

api/query/atoms_test.go
```diff
@@ -185,18 +185,343 @@ func TestStructuredQuery_nested(t *testing.T) {
...
+				Product: shared.Product{
+					BrowserName: "Firefox",
+				},
+			},
+		},
+	}))
+}
+
+func TestStructuredQuery_bindAnd(t *testing.T) {
+	assert.Equal(t, And{
```
lukebjerring: `path` and `edge`

api/query/atoms_test.go
```diff
@@ -185,18 +185,343 @@ func TestStructuredQuery_nested(t *testing.T) {
...
+				Product: shared.Product{
+					BrowserName: "Edge",
+				},
+			},
+		},
+	}))
+}
+
+func TestStructuredQuery_bindOr(t *testing.T) {
+	assert.Equal(t, Or{
```
lukebjerring: ... and the remaining tests too

api/query/atoms.go
mdittmer: We do this because we only grow the slice when `run.BrowserName == tsc.browserName` (not unconditionally).

api/query/atoms.go
mdittmer: Sure. We may not see it (see `len()` guards at end of method), so I'll just make the slice here and apply it (with field name) in the case that we return an `And`.

PR #852, opened 2018-12-04
Pass the required title field for Cancel action

lukebjerring:
## Description
Fixes the current 500'ing cancel requests.


lukebjerring:
@Hexcles TBR'd to fix some permanently-pending status checks.

PR #853, opened 2018-12-04
Add test for regressions on check update

lukebjerring:
## Description
Adds a few tests to ensure regressions produce the regressions summary, and no regressions does not.


shared/run_diff.go
```diff
@@ -150,15 +150,18 @@ func NewTestDiff(before, after []int, filter DiffFilterParam) TestDiff {
 	}
 }
 
+// ResultsDiff is a collection of test diffs, mapped by the test path.
```
LukeZielinski: nit: mapped->keyed?

api/checks/update_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/web-platform-tests/wpt.fyi/api/checks/summaries"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"github.com/web-platform-tests/wpt.fyi/shared/sharedtest"
+)
+
+func TestGetDiffSummary_Regressed(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	before := shared.TestRun{}
```
LukeZielinski: Can you add helper functions for some of the common code? Might make it easier to tell what's changing between tests. AFAICT it's just the TestDiff and the summary lines, right?

api/checks/update_test.go
```diff
@@ -0,0 +1,76 @@
...
+	before.FullRevisionHash = strings.Repeat("0", 40)
+	before.BrowserName = "chrome"
+	after := shared.TestRun{}
+	after.FullRevisionHash = strings.Repeat("1", 40)
+	after.BrowserName = "chrome"
+	runDiff := shared.RunDiff{
+		Differences: shared.ResultsDiff{"/foo.html": shared.TestDiff{0, 1, 0}},
```
LukeZielinski: nit: probably can't hurt to comment why a TestDiff of 0,1,0 is a regression while 1,0,1 is a completion

shared/run_diff_test.go
```diff
@@ -158,16 +158,19 @@ func getDeltaResultsMaps(before []int, after []int) (map[string][]int, map[strin
...
+		"/foo.html": TestDiff{0, 1, 0},
 	}
 	regressions := diff.Regressions()
 	assert.Equal(t, 1, regressions.Cardinality())
 	assert.True(t, regressions.Contains("/foo.html"))
 
-	diff.Differences["/foo.html"] = TestDiff{1, 0, 1}
+	diff = ResultsDiff{"/bar.html": TestDiff{1, 0, 1}}
 	regressions = diff.Regressions()
 	assert.Equal(t, 0, regressions.Cardinality())
```
LukeZielinski: also assert that regressions does not contain bar.html?

shared/run_diff.go
lukebjerring: Done.

api/checks/update_test.go
lukebjerring: Done via a little overkill assertion.

shared/run_diff_test.go
lukebjerring: Done.

PR #855, opened 2018-12-05
Export symbols from query for use in other packages

mdittmer:
Abstract queries need to be instantiated from other packages for testing purposes. Exporting them is cleaner than requiring such tests to unmarshal JSON to instantiate an opaque query object.


lukebjerring:
Merge on green.

PR #857, opened 2018-12-05
Fix cache index synchronization

mdittmer:
Drop use of sync.Map in favour of critical sections that ensure a stable state when reading from and writing to index.

### DIFF
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"sync"
 
+	mapset "github.com/deckarep/golang-set"
 	"github.com/web-platform-tests/results-analysis/metrics"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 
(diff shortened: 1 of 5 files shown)


api/query/cache/index/index.go
```diff
@@ -112,21 +117,30 @@ func (i *shardedWPTIndex) IngestRun(r shared.TestRun) error {
 	// top-level test (i.e., not subtests) integral ID of each test in the report.
 	//
 	// Create RunResults for each shard's partition of this run's results.
-	numShards := uint64(len(i.shards))
-	runResults := make([]RunResults, numShards)
-	for j := uint64(0); j < numShards; j++ {
-		runResults[j] = NewRunResults()
+	numShards := len(i.shards)
```
Hexcles: Will the length of `i.shards` change? (I don't see anything wrong, but just want to confirm.)

mdittmer: The design (which I don't think is violated by any of the code) is that the number of shards is fixed at creation time.

mdittmer:
Excellent point regarding `RWLock`. I erroneously believed that it would permit readers during a write operation. [That is not the case](https://golang.org/pkg/sync/#RWMutex):

> A RWMutex is a reader/writer mutual exclusion lock. The lock can be held by an arbitrary number of readers or a single writer.

PR #858, opened 2018-12-05
Search cache: Define Binder and Plan interfaces for abstract Bind and Execute

mdittmer:


lukebjerring:
I'm questioning the value of abstracting without actually having a second implementation just yet; your call.

mdittmer:
The abstraction is based on the speculation that we'll return some other type from some other plan strategy.

I now remember a second reason too: I don't want this `query` interface to depend on a `cache` type (`TestID`). I'm going to leave this for now. If we keep one implementation and refactor the `*ID` types out of `cache`, then this interface can be made more particular at that time.

PR #859, opened 2018-12-05
Search cache: Add filter functions implementations of Plan

mdittmer:
Plans on a sharded index are implemented by executing a slice of filters, or a ShardedFilter, in parallel. This change introduces filter functions that mirror query atoms. A sharded index creates a slice of filters, or ShardedFilter, that correspond to its shards during the binding phase, then the ShardedFilter can be executed during the execution phase.

This change also creates a placeholder file for forthcoming index+filter bind+execute tests.

### DIFF
--- a/api/query/cache/index/filter.go
+++ b/api/query/cache/index/filter.go
@@ -141,7 +141,7 @@ func (fs ShardedFilter) Execute() interface{} {
 	for _, f := range fs {
 		go func(f filter) {
 			log.Infof("Starting shard iteration; ranging over f.Tests...")
-			ts := make([]TestID, 0)
+			ts := []TestID{}
 			f.Tests().Range(func(t TestID) bool {
 				log.Infof("Processing %v", t)
 				if f.Filter(t) {
--- a/api/query/cache/index/filter.go
+++ b/api/query/cache/index/filter.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


api/query/cache/index/filter.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, errUnknownConcreteQuery
+	}
+}
+
+// Execute runs each filter in a ShardedFilter in parallel, returning a slice of
+// TestIDs as the result. Note that TestIDs are not deduplicated; the assumption
+// is that each filter is bound to a different shard, sharded by TestID.
+func (fs ShardedFilter) Execute() interface{} {
+	res := make(chan []TestID, len(fs))
+	log.Infof("Executing plan over %d shards", len(fs))
```
lukebjerring: Nit: Debugf

api/query/cache/index/filter.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, f := range fs {
+		go func(f filter) {
+			log.Infof("Starting shard iteration; ranging over f.Tests...")
+			ts := make([]TestID, 0)
```
lukebjerring:
```suggestion
			ts := []TestID{}
```

api/query/cache/index/filter.go
mdittmer: Yeah. This was never supposed to go up for review. It's from test debugging.

PR #860, opened 2018-12-05
Search cache: Make Indexes implement Binder

mdittmer:
This (along with existing open PRs) completes the Bind+Execute implementation a sharded index, allowing for landing integration tests labeled as `TODO` in previous PRs.

### DIFF
--- a/api/query/cache/backfill/backfill_medium_test.go
+++ b/api/query/cache/backfill/backfill_medium_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
+	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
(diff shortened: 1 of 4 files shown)


api/query/cache/backfill/backfill_medium_test.go
```diff
@@ -43,6 +44,10 @@ func (i *countingIndex) EvictAnyRun() error {
 	return nil
 }
 
+func (*countingIndex) Bind([]shared.TestRun, query.AbstractQuery) (query.Plan, error) {
+	return nil, nil
```
lukebjerring: Should this return a not-implemented-error of sorts?

api/query/cache/index/index.go
```diff
@@ -194,6 +196,35 @@ func (i *shardedWPTIndex) EvictAnyRun() error {
 	return nil
 }
 
+func (i *shardedWPTIndex) Bind(runs []shared.TestRun, aq query.AbstractQuery) (query.Plan, error) {
+	if len(runs) == 0 {
+		return nil, errNoRuns
+	}
+	if aq == nil {
```
lukebjerring: nit: `else if` and save a line

PR #861, opened 2018-12-05
Where applicable, update existing check-runs

lukebjerring:
## Description
Attempts to prevent check run name mismatching with a two-pronged approach:
- Include the experimental label when converting a TestRun into a product spec
- Pass the existing TestRun to the update method, and compare it to (any) existing run's name-specs

### DIFF
--- a/api/checks/runs.go
+++ b/api/checks/runs.go
@@ -26,6 +26,7 @@ func updateCheckRunSummary(ctx context.Context, summary summaries.Summary, suite
 			for _, run := range runs {
 				if spec, _ := shared.ParseProductSpec(run.GetName()); spec.Matches(*testRun) {
 					existing = run
+					break
 				}
 			}
 		}
--- a/api/checks/runs.go
+++ b/api/checks/runs.go
@@ -13,8 +13,79 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
-func updateCheckRun(ctx context.Context, summary summaries.Summary, suites ...shared.CheckSuite) (bool, error) {
+func updateCheckRunSummary(ctx context.Context, summary summaries.Summary, suites ...shared.CheckSuite) (bool, error) {
 	log := shared.GetLogger(ctx)
+	product := summary.GetCheckState().Product
+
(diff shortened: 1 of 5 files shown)


lukebjerring:
You can see the effects of current (`staging.wpt.fyi status check`) vs this PR (`checks staging instance`) here:
https://github.com/lukebjerring/wpt/pull/8/checks

api/checks/runs.go
```diff
@@ -42,12 +113,11 @@ func updateCheckRun(ctx context.Context, summary summaries.Summary, suites ...sh
...
-			return false, err
-		}
-		log.Debugf("Check for %s/%s @ %s (%s) updated", suite.Owner, suite.Repo, suite.SHA[:7], state.Product.String())
+	client, err := getGitHubClient(ctx, suite.AppID, suite.InstallationID)
+	if err != nil {
+		return false, err
 	}
-	return true, nil
+
+	_, _, err = client.Checks.UpdateCheckRun(ctx, suite.Owner, suite.Repo, run.GetID(), opts)
```
Hexcles: It might be good to extract this part into `updateCheckRun` and place it next to `createCheckRun`, which is currently in `api/checks/webhook.go` (you may also move them together into `api.go`).

api/checks/runs.go
```diff
@@ -13,8 +13,79 @@ import (
...
+
+	testRun := summary.GetCheckState().TestRun
+	for _, suite := range suites {
+		// Update, not create, if a run name matches this completed TestRun.
+		var existing *github.CheckRun
+		if testRun != nil {
+			runs, _ := getExistingCheckRuns(ctx, suite)
+			for _, run := range runs {
+				if spec, _ := shared.ParseProductSpec(run.GetName()); spec.Matches(*testRun) {
+					existing = run
```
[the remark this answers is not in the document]
lukebjerring: Yes, it's possible, no, it doesn't matter. There's like, 3 or less checks to iterate through.

api/checks/runs.go
lukebjerring: I'll follow "rules of three" on this one, and do it when I get another caller.

Issue #714, opened 2018-10-31
Upload 'affected tests' results from Azure Pipelines


lukebjerring:
Azure version of #707

foolip:
The Safari results are being done in https://github.com/web-platform-tests/wpt/pull/14156. The API for getting the results out will be https://dev.azure.com/web-platform-tests/wpt/_apis/build/builds/1040/artifacts?artifactName=drop&api-version=5.0-preview.5&%24format=zip.

@Hexcles @lukebjerring, I think using that PR you should be able to experiment with the results receiver parts to this right now. Note that I'll want to change `artifactName` though, probably to `results`.

lukebjerring:
This is now up and running! https://staging.wpt.fyi/runs?product=safari

foolip:
Sweet, with an azure label too!

lukebjerring:
> Logos are the new labels

https://github.com/web-platform-tests/wpt.fyi/pull/910

PR #863, opened 2018-12-05
Handle azure completed events

lukebjerring:
## Description
Step towards #714 by currying an artifact URL from the details URL.

Unfortunately I've had to wrap the `CheckRunEvent` to get the `detailts_url` field, until I land https://github.com/google/go-github/pull/1065

### DIFF
--- a/api/checks/webhook.go
+++ b/api/checks/webhook.go
@@ -28,12 +28,7 @@ const (
 )
 
 func isWPTFYIApp(appID int64) bool {
-	switch appID {
-	case wptfyiCheckAppID,
-		checksStagingAppID:
-		return true
-	}
-	return false
+	return appID == wptfyiCheckAppID || appID == checksStagingAppID
 }
 
 // checkWebhookHandler listens for check_suite and check_run events,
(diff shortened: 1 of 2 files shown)


api/checks/webhook.go
```diff
@@ -26,11 +27,10 @@ const (
 	checksForAllUsersFeature = "checksAllUsers"
 )
 
-func isKnownAppID(appID int64) bool {
+func isWPTFYIApp(appID int64) bool {
```
Hexcles: If this function only checks the wpt.fyi app now, we don't need to have `switch` any more; in fact, we might as well just get rid of the whole function.

api/checks/webhook.go
```diff
@@ -204,20 +231,50 @@ func handleCheckRunEvent(aeAPI shared.AppEngineAPI, checksAPI API, payload []byt
...
 		return false, nil
 	}
+	detailsURL := event.GetCheckRun().GetDetailsURL()
+	buildID := extractAzureBuildID(detailsURL)
+	if buildID == "" {
+		log.Errorf("Failed to extract build ID from details_url \"%s\"", detailsURL)
+		return false, nil
+	}
+	artifact := fmt.Sprintf(
+		"https://dev.azure.com/%s/%s/_apis/build/builds/%s/artifacts?artifactName=drop&api-version=5.0-preview.5&%%24format=zip",
```
Hexcles: This URL schema is kind of ugly (which isn't our fault). Let's add a link to the Azure Pipeline docs.

lukebjerring:
Failed job was a timeout (flake)

PR #864, opened 2018-12-05
Plumb the callback_url from api/results/upload to TaskQueue

lukebjerring:
## Description
We tried to support `callback_url` in #788, but I missed a request-layer in the conceptual model.

### DIFF
--- a/api/receiver/receive_results.go
+++ b/api/receiver/receive_results.go
@@ -47,7 +47,8 @@ func HandleResultsUpload(a AppEngineAPI, w http.ResponseWriter, r *http.Request)
 
 	// Most form methods (e.g. FormValue, FormFile) will call
 	// ParseMultipartForm and ParseForm if necessary; forms with either
-	// enctype can be parsed.
+	// enctype can be parsed. FormValue gets either query params or form
+	// body entries, favoring the latter.
 	// The default maximum form size is 32MB, which is also the max request
 	// size on AppEngine.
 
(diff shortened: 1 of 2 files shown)


api/receiver/receive_results.go
```diff
@@ -45,14 +45,14 @@ func HandleResultsUpload(a AppEngineAPI, w http.ResponseWriter, r *http.Request)
 		uploader = username
 	}
 
-	// Most form methods (e.g. PostFormValue, FormFile) will call
+	// Most form methods (e.g. FormValue, FormFile) will call
```
lukebjerring: Drive-by: PostFormValue requires the param to be in the POST body, but FormValue lets it be in the URL params _or_ body, with precendence for the body.

This can be quite useful when you're typing a CURL from the command-line.

Hexcles: I'm slightly hesitant about this change. I usually prefer to have all params either in the body (for POST) or in the URL (for GET). My personal experience is that mixing them is error-prone. For local testing of POST, I often use [httpie](https://httpie.org/) instead of curl for cleaner syntax. But if you think this helps your development (we are a small team, after all), then please feel free to go ahead, but remember to add a comment regarding the precedence.

api/receiver/receive_results.go
```diff
@@ -63,13 +63,17 @@ func HandleResultsUpload(a AppEngineAPI, w http.ResponseWriter, r *http.Request)
...
-		"os_name":         r.PostFormValue("os_name"),
-		"os_version":      r.PostFormValue("os_version"),
+		"revision":        r.FormValue("revision"),
+		"browser_name":    r.FormValue("browser_name"),
+		"browser_version": r.FormValue("browser_version"),
+		"os_name":         r.FormValue("os_name"),
+		"os_version":      r.FormValue("os_version"),
+	}
+	callbackURL := r.FormValue("callback_url")
+	if callbackURL != "" {
```
Hexcles: No need to have this check. This is already done in https://github.com/web-platform-tests/wpt.fyi/blob/3b1fb7aaa4517c1ea7e02e486c760ee2fc4b40e7/api/receiver/appengine.go#L131

api/receiver/receive_results.go
lukebjerring: Added a comment. I'm not proposing we mix them for real callers; just that supporting both is better for ad-hoc debugging work.

api/receiver/receive_results.go
lukebjerring: Removed check.

Issue #708, opened 2018-10-30
Execute affected tests both with and without a PR's changes


lukebjerring:
This will allow a more apples-to-apples diff than comparing to a recent full run of the same product (see #707).

lukebjerring:
This (executing both with and without) is a better solution to a problem that was mentioned in https://github.com/web-platform-tests/wpt.fyi/pull/745 - namely, as an interim we are comparing a PR's affected tests againts "the latest master run".
This is done with a link that points to `chrome[master]`, for example - a link that will change over time (as new master runs arrive). 

cc @foolip

foolip:
@lukebjerring, can you send a PR for the Taskcluster changes? Here's what I'd planned to do for Azure Pipelines:
- In a "decision task" make note of what HEAD^1 and HEAD^2 of the synthetic merge commit (refs/pull/1234/merge) are, call them SAVED_MASTER and SAVED_PR_HEAD. In particular the first parent is a commit on master, and master may change.
- In the "with changes" task, check out SAVED_MASTER and merge in SAVED_PR_HEAD. Affected tests are `./wpt tests-affected $SAVED_MASTER`.
- In the "without changes" task check out SAVED_MASTER and merge in SAVED_PR_HEAD, call that commit TMP_MERGE. Then check out SAVED_MASTER and test that. Affected tests are `./wpt tests-affected $TMP_MERGE`.

This is tragically convoluted, but I don't see how else to ensure differences between the two states of the trees were tested are only the changes from the PRs, and not also some random contribution of changes to master that happened as the tasks were queuing.

@jgraham may have thouhts.

foolip:
I sent https://github.com/web-platform-tests/wpt/pull/14093 for a thing I found confusing about the affected tests logic.

I also just conducted an experiment on foolip/wpt to fetch `git fetch https://github.com/foolip/wpt.git refs/pull/12/merge && git rev-parse FETCH_HEAD`, push commits to master, and fetch it again. FETCH_HEAD did change just seconds after. So there's nothing that GitHub does to provide stability here, and honestly I can't imagine what that could be.

Hexcles:
[...]

PR #865, opened 2018-12-06
Harvest the "without changes" runs from Taskcluster

Hexcles:
This change extends the set of valid task names to accept the
"*-results-without-changes" tasks that are being added in
https://github.com/web-platform-tests/wpt/pull/14382 . The change also
adds a "pr_base" label to these special runs, and "pr_head" to the PR
runs with the changes to prepare for the next step: using these runs
to calculate PR regressions (not implemented in this change).

Related to #708 .

### DIFF
--- a/shared/util.go
+++ b/shared/util.go
@@ -33,9 +33,13 @@ const BetaLabel = "beta"
 // i.e. run from the master branch.
 const MasterLabel = "master"
 
-// WithoutPatchLabel is the label for running just the affected tests on a PR
-// but without the patch.
-const WithoutPatchLabel = "without_patch"
+// PRBaseLabel is the implicit label for running just the affected tests on a
+// PR but without the changes (i.e. against the base branch).
+const PRBaseLabel = "pr_base"
+
+// PRHeadLabel is the implicit label for running just the affected tests on the
+// head of a PR (with the changes).
+const PRHeadLabel = "pr_head"
 
 // ProductChannelToLabel maps known product-specific channel names
 // to the wpt.fyi model's equivalent.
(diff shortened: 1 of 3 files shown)


shared/util.go
```diff
@@ -33,6 +33,10 @@ const BetaLabel = "beta"
 // i.e. run from the master branch.
 const MasterLabel = "master"
 
+// WithoutPatchLabel is the label for running just the affected tests on a PR
+// but without the patch.
+const WithoutPatchLabel = "without_patch"
```
lukebjerring: So... I did say that eventually `master` might stop implying a full run, and this is that day.
I'm wondering if labelling the runs that aren't just the affected tests as `full` is a better route for our default homepage.

Alternatively (or even in conjunction with `full`) we can pick a more succinct name for this label. How about we go the science route, and label them as `control`? Or `baseline`? `normal`? `prior`? 
... I think `baseline` might be best, since we use that term with the same semantic meaning in other contexts.

lukebjerring: @foolip suggested `base` (without_patch) and `head` (with patch) offline. They work IMHO, WDYT?

foolip: Then I threw `pr_base` and `pr_head` into the mix. Either is fine, only possible misreading is that `head` could mean HEAD, i.e. master.

Hexcles: Back to your first comment, Luke. I understand `master` eventually won't be full runs in the future, but this PR doesn't change the current status that master runs are always full runs, right?

I like pr_base and pr_head, for the overuse of the term "head" itself.

Hexcles:
@lukebjerring PTAL again

PR #867, opened 2018-12-06
Remove unneeded CheckRunEvent wrapper

lukebjerring:
## Description
I'm not sure the wrapper worked anyway, given that we're seeing nil pointers...


lukebjerring:
TBR'ing to fix 500ing requests

PR #870, opened 2018-12-06
Search cache: Simple bind+execute tests

mdittmer:


api/query/cache/index/index_filter_test.go
```diff
@@ -1,9 +1,249 @@
...
+						},
+					},
+				},
+			},
+		},
+		//
+		// Safari test run: Several result values differ or are missing. One test
+		// does not appear in Chrome, but does appear here.
+		//
+		testRunData{
```
lukebjerring: For nested stuff like this, I find declaring the `testRunData{}` and using foo.Bar = baz uses much less lines than gofmt's wrapping + comma requirements.

mdittmer: I have to admit that when it comes to editing the code later I strongly disagree. This is _a lot_ of lines, and I don't like it. What I like even less is debugging a lightly modified test because I missed some imperative `tempVar.foo.bar = tempBarValue` in a 20-line list of assignments.

lukebjerring: Sure; feel free to merge as-is.

mdittmer: After offline discussion, I think this makes sense for coping with multi-layered type embeddings.

PR #871, opened 2018-12-06
Search cache: Add test aimed at catching index read/write race conditions

mdittmer:


api/query/cache/index/index_test.go
```diff
@@ -148,5 +151,126 @@ func TestEvictNonEmpty(t *testing.T) {
...
+	i.IngestRun(makeRun(2))
+	i.IngestRun(makeRun(3))
+	i.IngestRun(makeRun(4))
+	i.IngestRun(makeRun(5))
+	i.IngestRun(makeRun(6))
+	i.IngestRun(makeRun(7))
+	i.IngestRun(makeRun(8))
+
+	var wg sync.WaitGroup
+	for j := 9; j <= 16; j++ {
```
lukebjerring: Some explanatory comments are needed for what this is actually doing async and what failure would look like.

mdittmer: Added comments. @lukebjerring PTAL again.

api/query/cache/index/index_test.go
```diff
@@ -148,5 +151,146 @@ func TestEvictNonEmpty(t *testing.T) {
...
+
+	// Total number of results is 8 runs * 2 results per run = 16.
+	assert.Equal(t, 16, numResults)
+}
+
+var browsers = []string{
+	"Chrome",
+	"Edge",
+	"Firefox",
+	"Safari",
```
Hexcles: These names should technically be lower-case, but it doesn't really matter in this test.

Issue #866, opened 2018-12-06
Custom checks showing up on all commits on master, comparing against master


foolip:
In https://github.com/web-platform-tests/wpt/commits/master, every commit gets wpt.fyi "chrome[experimental] and "firefox[experimental]" checks. Example:
https://github.com/web-platform-tests/wpt/runs/38327762
https://github.com/web-platform-tests/wpt/runs/38327768

This could be very useful if the comparison was against the previous commit on master.

foolip:
A comparison against the previous run of the same product would in particular come in handy after landing a change like https://github.com/web-platform-tests/wpt/pull/14309 where some regressions might have slipped through.

lukebjerring:
I think the checks you're pointing at are (incorrectly) comparing the same run against itself, since the "latest master run" is the run itself. I'll use this issue to follow up on ensuring that we find the previous run in those cases.

As for making it show up for everyone, you'll just need to enable `checksAllUsers`. This isn't added in the /admin/flags view just yet - will add it.

Hexcles:
By the way, the user whitelist seems to be not working. All commits from all users on master have the wpt.fyi custom checks.

lukebjerring:
It was that way a couple days ago, so you might just be seeing leftovers from that bug, but if you scroll the list a few don't have the check (e.g. commits from youenff)

foolip:
For commit 226f3b5 3 checks are showing up:
![image](https://user-images.githubusercontent.com/498917/49724887-b6e51700-fc6a-11e8-9819-769348513469.png)

But for the preceding commit no checks. Not sure why :)

lukebjerring:
The pr author is whitelisted, so their merge triggers a check suite

gsnedders:
> The pr author is whitelisted, so their merge triggers a check suite

Why does authorship matter when the commit is on the wpt repo itself? If a commit has been pushed to the WPT repo itself, it's been pushed by someone we should consider trustworthy.

foolip:
The list was just to try it out with a small number of people, not for security reasons.

lukebjerring:
Yah, temporary measure to reduce noise during unstable development.

lukebjerring:
This was fixed by #872

PR #872, opened 2018-12-06
In regression checks, load an earlier master to diff against

lukebjerring:
## Description
Part of #866 

Ensures that the master run we compare to occurred before the run we are comparing it to. This is important for checks on the `master` branch itself, but also for avoiding a master branch that might have run away with other changes while we were processing the PR.

### DIFF
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -5,9 +5,11 @@
 package checks
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"sort"
+	"time"
 
 	"github.com/deckarep/golang-set"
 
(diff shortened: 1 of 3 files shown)


api/checks/update.go
```diff
@@ -118,6 +95,41 @@ func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
...
+		return nil, nil, fmt.Errorf("Expected exactly 1 run, but found %v", len(runs))
+	}
+
+	// Get the most recent, but still earlier, master run to compare.
+	labels := filter.Labels
+	if labels == nil {
+		labels = mapset.NewSet()
+	}
+	labels.Add("master")
+	to := prRun.TimeStart.Add(-time.Millisecond)
```
lukebjerring: While not obvious in the diff, this is the actual change. The rest is a refactor so that I could add a test for it.

foolip: Thanks for pointing that out!

PR #875, opened 2018-12-06
Search cache: Implement query handler with backfilling only

mdittmer:
This change implements the `/api/search/cache` endpoint with a cache that backfills from the time it is started.

Tests for this endpoint and integration for posting new test runs will be implemented in follow-up PRs.

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -25,7 +25,7 @@ type AbstractQuery interface {
 // client, including the IDs of the test runs to query, and the structured query
 // to run.
 type RunQuery struct {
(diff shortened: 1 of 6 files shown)


mdittmer:
[This](https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/163546182#L2603) is a surprising error. Looks to be related to checks templates. @lukebjerring can you please investigate?

lukebjerring:
Seems to not be finding a copy of those files, which is very strange. Is it persistent? Seems to only fail on the deploy for cache-query-handler ?

PR #876, opened 2018-12-06
Support upload for Azure artifacts

lukebjerring:
## Description
Works towards https://github.com/web-platform-tests/wpt.fyi/issues/714

Unfortunately, the only API for an artifact is a zip file, with the report nested inside. This PR aims to support `report_path` which implies this case, and extracts that path as the report file for upload.

## Review Information
This has been deployed as `checks` in staging, where I've verified it actually succeeds:
https://staging.wpt.fyi/runs?sha=9b4d522493a45f24338966d2ba9206cd3be383ff


shared/appengine.go
```diff
@@ -44,6 +54,18 @@ func (a AppEngineAPIImpl) Context() context.Context {
 	return a.ctx
 }
 
+// GetHTTPClient returns an HTTP client for the current context.
+func (a AppEngineAPIImpl) GetHTTPClient() *http.Client {
```
lukebjerring: @mdittmer - here's the client abstraction I mentioned.

api/checks/azure_test.go
```diff
@@ -0,0 +1,118 @@
...
+func TestHandleAzurePipelinesEvent(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	sha := strings.Repeat("0123456789", 4)
+	detailsURL := "https://dev.azure.com/web-platform-tests/b14026b4-9423-4454-858f-bf76cf6d1faa/_build/results?buildId=123"
+	event := getCheckRunCreatedEvent("completed", "lukebjerring", sha)
+	event.CheckRun.DetailsURL = &detailsURL
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
lukebjerring: @mdittmer - here's a "fake" http server (as opposed to a mock one). The calling code just expects "given this http.Client, and this URL, fetch it" so we mock the URL we give out (as the server's URL) and the client we give out (as the server's Client).

api/checks/api.go
```diff
@@ -172,6 +179,61 @@ func (s checksAPIImpl) CreateWPTCheckSuite(appID, installationID int64, sha stri
 	return suite != nil, err
 }
 
+func (s checksAPIImpl) GetAzureArtifactsURL(owner, repo string, buildID int64) string {
```
Hexcles: I'm confused by `golint`. Why doesn't it complain the lack of comment for this exported method?

api/README.md
```diff
@@ -355,7 +355,10 @@ __`callback_url`__: (Optional) A URL that the processor should `POST` when succe
 create the TestRun. Defaults to /api/results/create in the current project's environment (e.g. wpt.fyi for
 wptdashboard, staging.wpt.fyi for wptdashboard-staging).
 
-__`result_file`__: A **gzipped** JSON file produced by `wpt run --log-wptreport`.
+__`result_file`__: A **gzipped** JSON file produced by `wpt run --log-wptreport`, or a zip file containing
+the report. If the file is a zip, `report_path` **must** be specified.
+
+__`report_path`__: (Optional) Path to the report file within the result file, if applicable.
```
Hexcles: This change needs to be reverted.

api/receiver/receive_results.go
```diff
@@ -45,15 +45,14 @@ func HandleResultsUpload(a AppEngineAPI, w http.ResponseWriter, r *http.Request)
 		uploader = username
 	}
 
-	// Most form methods (e.g. FormValue, FormFile) will call
+	// Most form methods (e.g. PostFormValue, FormFile) will call
```
Hexcles: This whole file looks like a bad merge.

api/receiver/receive_results_test.go
```diff
@@ -128,7 +128,6 @@ func TestHandleResultsUpload_extra_params(t *testing.T) {
 		"browser_version": "",
 		"os_name":         "",
 		"os_version":      "",
-		"callback_url":    "",
```
Hexcles: Ditto.

Hexcles:
Another reason why we should have a separate `api/azure` package is because we might need to run this adapter in a separate AppEngine service, because of memory constraints: when we start to run full runs on Azure (e.g. Edge on Win10), the zip->gzip transformation may take hundreds of MB of memory. If we have a separate package for the Azure adapter, we can then easily add a new entry point (URL route) to this code and run it as a webhook independent from our GitHub App (i.e. the same as the setup of our Taskcluster listener now).

api/receiver/receive_results_test.go
lukebjerring: Done.

api/receiver/receive_results.go
lukebjerring: Done.

api/README.md
lukebjerring: Done.

api/checks/api.go
lukebjerring: Not sure either.

lukebjerring:
Moved the code to `/api/azure`, PTAL.

lukebjerring:
FWIW, I manually deployed this as version `checks`, and you can see the results here:
https://github.com/lukebjerring/wpt/pull/13/checks?sha=8d996d10a2f1ce9385921df1a1dc44f9b565d28f

api/azure/webhook.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return err
+	}
+
+	fileField, err := writer.CreateFormFile("result_file", "wpt_report.json")
+	_, err = io.Copy(fileField, bytes.NewReader(data))
+	if err != nil {
+		return err
+	}
+	if err := writer.Close(); err != nil {
```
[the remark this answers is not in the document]
lukebjerring: Done.

api/azure/webhook.go
```diff
@@ -0,0 +1,162 @@
...
+	host := aeAPI.GetHostname()
+	writer.WriteField("callback_url", fmt.Sprintf("https://%s/api/results/create", host))
+
+	data, err := azureAPI.FetchAzureArtifact(artifact)
+	if err != nil {
+		return err
+	}
+
+	fileField, err := writer.CreateFormFile("result_file", "wpt_report.json")
+	_, err = io.Copy(fileField, bytes.NewReader(data))
```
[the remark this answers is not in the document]
lukebjerring: Done.

PR #878, opened 2018-12-07
Try all suites on error, and match App IDs

lukebjerring:
## Description
In the (abnormal) case that we have multiple CheckSuites for different apps, we were erroring trying to patch with the wrong creds. This change stops one suite's error blocking other suites, and also ensures that we don't mismatch App ID.

### DIFF
--- a/api/checks/runs.go
+++ b/api/checks/runs.go
@@ -15,15 +15,28 @@ import (
 
 func updateCheckRunSummary(ctx context.Context, summary summaries.Summary, suites ...shared.CheckSuite) (bool, error) {
 	log := shared.GetLogger(ctx)
+	if len(suites) < 1 {
+		return false, nil
+	}
 	product := summary.GetCheckState().Product
-
 	testRun := summary.GetCheckState().TestRun
+
+	// Attempt to update any existing check runs for this SHA.
+	checkRuns, err := getExistingCheckRuns(ctx, suites[0])
+	if err != nil {
+		log.Warningf("Failed to load existing check runs for %s: %s", suites[0].SHA[:7], err.Error())
+	}
+
+	createdAny := false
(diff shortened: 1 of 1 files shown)


api/checks/runs.go
```diff
@@ -15,15 +15,28 @@ import (
...
 	log := shared.GetLogger(ctx)
+	if len(suites) < 1 {
+		return false, nil
+	}
 	product := summary.GetCheckState().Product
-
 	testRun := summary.GetCheckState().TestRun
+
+	// Attempt to update any existing check runs for this SHA.
+	checkRuns, err := getExistingCheckRuns(ctx, suites[0])
```
lukebjerring: Since all suites have access to `checks:read`, and this method loads all check runs for a given SHA, we can just call it once, for the first suite.

api/checks/runs.go
```diff
@@ -15,15 +15,28 @@ import (
...
+	if err != nil {
+		log.Warningf("Failed to load existing check runs for %s: %s", suites[0].SHA[:7], err.Error())
+	}
+
+	createdAny := false
+	errors := make(chan error, len(suites))
```
Hexcles: I don't see any gorountine here. What's the benefit of using a channel instead of a slice?

api/checks/runs.go
Hexcles: Does `getExistingCheckRuns(suites[0])` give *all* check runs from all suites, or just the check runs for the first suite?

Hexcles: ping @lukebjerring 

If not, it seems like you want to combine all check runs from all suites.

lukebjerring: It's getting _all_ CheckRun objects for a SHA, regardless of suite.

PR #879, opened 2018-12-07
Eliminate api/query -> api -> ... -> api/checks dep chain

mdittmer:
`searchcache` service build is failing because it erroneously depends on `api/checks` and its template files. This change factors a helper function out to the `shared` package to break the dep chain.


Hexcles:
I'm a bit more curious about the api -> ... -> api/checks part? Any idea where that dependency is?

shared/util_test.go
```diff
@@ -52,3 +63,56 @@ func TestProductChannelToLabel(t *testing.T) {
...
+					Revision: shortSHA,
+				},
+				ResultsURL: resultsURL,
+			},
+			"/",
+			resultsURL,
+		})
+}
+
+func checkResult(t *testing.T, c Case) {
```
lukebjerring: I'd drop `Case` and just pass the 3 params to this function. Will shave 10 or more lines of code.

mdittmer:
@Hexcles `api/results_receive_handler.go` depends on `github.com/web-platform-tests/wpt.fyi/api/checks`. I did not investigate that link in the chain, as there seemed to be no strong reason for the pure helper function I'm moving to live in `api`.

Hexcles:
@mdittmer thanks. Yeah this LGTM.

PR #880, opened 2018-12-07
Connect search API frontend to cache backend

mdittmer:
This change tests and implements `/api/search` delegating to `/api/search/cache` when the request is a structured query that cannot be serviced using summary files.

The change required some refactoring of the handlers in `search.go` to ensure that an `AppEngineAPI` instance is attached to the handlers. The `AppEngineAPI` is used to instantiate a `*http.Client`, which is done in a special way in AppEngine Standard, and is mocked in tests.

### DIFF
--- a/api/query/search.go
+++ b/api/query/search.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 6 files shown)


api/query/search.go
```diff
@@ -109,8 +121,26 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+			return
+		}
+
+		data, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			http.Error(w, "Error reading response from search API cache", http.StatusInternalServerError)
+		}
+
+		w.WriteHeader(resp.StatusCode)
+		w.Write(data)
```
lukebjerring: nit: You can collapse a read all + write into a `Copy`, something like

    io.Copy(w, resp.Body)

api/query/search_test.go
```diff
@@ -104,3 +111,64 @@ func TestPrepareSearchResponse_qUC(t *testing.T) {
...
+	}))
+
+	serverURL, err := url.Parse(server.URL)
+	assert.Nil(t, err)
+	hostname := serverURL.Host
+
+	api := sharedtest.NewMockAppEngineAPI(ctrl)
+
+	api.EXPECT().GetHostname().Return(hostname)
+	api.EXPECT().GetHTTPClient().DoAndReturn(func() *http.Client {
```
lukebjerring: Why not just `.Return(server.Client())` ?

mdittmer: I was having problems and this `DoAndReturn()` was one of the things I tried to fix it. I can't remember whether or not the problem actually lay elsewhere. I'll change this to `Return()` if the test behaves correctly with it.

PR #881, opened 2018-12-07
Compare PR results against the results without changes

### LINKED ISSUE #708: Execute affected tests both with and without a PR's changes

Hexcles:
## Description

The biggest changes are in `loadRunsToCompare` and `api/receiver/create_run.go`:
* In `loadRunsToCompare`, we now need to handle the master runs and PR runs separately.
* In `api/receiver/create_run.go`, we no longer schedule the results processing for runs with the label "pr_base"; otherwise, checks on PRs will be scheduled to be updated twice.

Fixes #708 .

## Review Information

Please review commit by commit, as the second one is largely mechanical.


api/checks/update.go
```diff
@@ -95,60 +96,80 @@ func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
...
-func loadRunsToCompare(ctx context.Context, filter shared.TestRunFilter) (prRun, masterRun *shared.TestRun, err error) {
-	log := shared.GetLogger(ctx)
-	one := 1
-	runs, err := shared.LoadTestRuns(ctx, filter.Products, filter.Labels, filter.SHA, filter.From, filter.To, &one, nil)
-	prRun = runs.First()
+func loadRunsToCompare(ctx context.Context, filter shared.TestRunFilter) (headRun, baseRun *shared.TestRun, err error) {
+	// If the run is on master, we will have only one match;
+	// for PRs, we have one for pr_head and one for pr_base.
+	two := 2
+	runs, err := shared.LoadTestRuns(ctx, filter.Products, filter.Labels, filter.SHA, filter.From, filter.To, &two, nil)
```
lukebjerring: In the rare case that we have a master commit land with exactly the same SHA, or in the case of duplicate runs, I'm concerned that we'll 404 here when we probably shouldn't. Given that you're deliberately filtering for `shared.PRHeadLabel` below, why not query for exactly one run _with_ that label? And if it's not found, we try and fetch master @ the same SHA.

So, for say, chrome:
- let head be chrome[pr_head]@sha
- If head is nil, let head be chrome[master]@sha
- If head is still nil, return a 404
- If head is a pr_head, let base be chrome[pr_base]@sha
- If base is nil, let base be the latest chrome[master] _before_ head.TimeStart

The last line means "fall back to diffing against master when pr_base can't be found."

Hexcles: That is a bit too many special conditions and fallbacks. What about making this a bit more generic?

How about we include *all* labels for a test run in the spec? So for master runs we will have `chrome[master,taskcluster,experimental]@sha` and for PR runs we have `chrome[pr_head,taskcluster,experimental]@sha`. Then in this function we can simply test whether labels contain "master" or "pr_head", then use different queries (either the same spec but *earlier than sha*, or the same spec but with "pr_head" replaced by "pr_base").

Hexcles:
@lukebjerring PTAL again (at the first commit). I didn't exactly follow your suggestions; instead, the logic is:

0. Find the first (latest) run matching the spec.
1. If the run has label "master", use it as the head run, find an earlier run as the base run to compare against.
2. If the run has label "pr_base", use it as the base run, find another run matching the spec but also with label "pr_head" as the head run.
3. If the run has label "pr_head", use it as the head run, find another run matching the spec but also with label "pr_base" as the base run.

(1, 2, 3 are mutually exclusive.)

api/checks/update.go
```diff
@@ -95,60 +96,86 @@ func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
...
-		labels = mapset.NewSet()
-	}
-	labels.Add("master")
-	to := prRun.TimeStart.Add(-time.Millisecond)
-	masterRuns, err := shared.LoadTestRuns(ctx, filter.Products, labels, "", nil, &to, &one, nil)
-	masterRun = masterRuns.First()
+	one := 1
+	to := headRun.TimeStart.Add(-time.Millisecond)
+	labels := mapset.NewSetWith(shared.MasterLabel)
+	runs, err := shared.LoadTestRuns(ctx, filter.Products, labels, "", nil, &to, &one, nil)
```
lukebjerring:
```suggestion
	runs, err := shared.LoadTestRuns(ctx, filter.Products, labels, shared.LatestSHA, nil, &to, &one, nil)
```

PR #884, opened 2018-12-10
Search cache: Add load-newer-runs update

mdittmer:
This change implements a polling strategy to load newer runs (runs added after cache backfill started). It also integrates polling into the cache's production setup routine.

Since, for now, this is the only mechanism for receiving new runs, the default polling configuration is relatively aggressive: Load up to 10 new runs every 10 seconds. The frequency can be reduced if/when we implement best effort pushing of runs from the processor to the cache on the processor's critical path. The purpose of implemention such a scheme would be to ensure that runs are cached before they are committed to datastore and begin showing up in the wpt.fyi UI.

### DIFF
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -38,7 +38,7 @@ func ErrRunExists() error {
 	return errRunExists
 }
 
-// ErrRunExists returns the error associated with an attempt to perform
+// ErrRunLoading returns the error associated with an attempt to perform
 // operations on a run currently unknown to an Index when the Index, in fact,
 // is currently loading data associated with the run.
 func ErrRunLoading() error {
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -31,6 +31,20 @@ var (
 	errZeroRun            = errors.New("Cannot ingest run with ID of 0")
 )
 
+// ErrRunExists returns the error associated with an attempt to perform
+// operations on a run currently unknown to an Index when the Index, in fact,
+// already knows about the run.
+func ErrRunExists() error {
+	return errRunExists
+}
+
+// ErrRunExists returns the error associated with an attempt to perform
+// operations on a run currently unknown to an Index when the Index, in fact,
+// is currently loading data associated with the run.
+func ErrRunLoading() error {
+	return errRunLoading
+}
+
 // Index is an index of test run results that can ingest and evict runs.
 type Index interface {
 	query.Binder
--- a/api/query/cache/poll/poll.go
+++ b/api/query/cache/poll/poll.go
@@ -13,13 +13,13 @@ import (
 )
 
 // KeepRunsUpdated implements updates to an index.Index via simple polling every
-// freq duration for at most limit runs loaded from fetcher.
-func KeepRunsUpdated(fetcher backfill.RunFetcher, logger shared.Logger, freq time.Duration, limit int, idx index.Index) {
-	// Start by waiting polling frequency. This reduces the chance of false alarms
+// interval duration for at most limit runs loaded from fetcher.
+func KeepRunsUpdated(fetcher backfill.RunFetcher, logger shared.Logger, interval time.Duration, limit int, idx index.Index) {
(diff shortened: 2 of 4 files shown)


api/query/cache/index/index.go
```diff
@@ -31,6 +31,20 @@ var (
 	errZeroRun            = errors.New("Cannot ingest run with ID of 0")
 )
 
+// ErrRunExists returns the error associated with an attempt to perform
+// operations on a run currently unknown to an Index when the Index, in fact,
+// already knows about the run.
+func ErrRunExists() error {
```
Hexcles: I don't know what is the convention in Go, but what is the benefit of exporting a function that returns the global variable over just exporting the global variable?

api/query/cache/poll/poll.go
```diff
@@ -0,0 +1,71 @@
...
+	"time"
+
+	"github.com/web-platform-tests/wpt.fyi/api/query/cache/backfill"
+	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// KeepRunsUpdated implements updates to an index.Index via simple polling every
+// freq duration for at most limit runs loaded from fetcher.
+func KeepRunsUpdated(fetcher backfill.RunFetcher, logger shared.Logger, freq time.Duration, limit int, idx index.Index) {
```
Hexcles: `freq` (frequency) has the unit of "/s" (Hz). I think you meant interval.

api/query/cache/poll/poll.go
```diff
@@ -0,0 +1,71 @@
...
+	// Start by waiting polling frequency. This reduces the chance of false alarms
+	// from log monitoring when KeepRunsUpdated is invoked around the same time as
+	// index backfilling.
+	logger.Infof("Starting index update via polling; waiting polling frequency first...")
```
Hexcles: Same here

api/query/cache/poll/poll.go
```diff
@@ -0,0 +1,71 @@
...
+				} else if err == index.ErrRunLoading() {
+					logger.Infof("Not updating run (already loading): %v", run)
+				} else {
+					logger.Errorf("Error ingesting run: %v: %v", run, err)
+				}
+			} else {
+				logger.Infof("Updated run index; new run: %v", run)
+
+				if i != 0 && !found {
+					logger.Errorf("Runs loaded out of order: Skipped %d runs: %v, then loaded new run: %v", i, runs[:i], run)
```
Hexcles: This is not necessarily true, right? If runs are out of order, we could see "loaded new loaded new", in which case you can't say the first three runs are skipped.

api/query/cache/poll/poll.go
```diff
@@ -0,0 +1,71 @@
...
+				}
+				found = true
+				lastLoadTime = time.Now()
+			}
+		}
+
+		if !found {
+			logger.Warningf("No runs loaded throughout polling iteration. Last run update was at %v", lastLoadTime)
```
Hexcles: IIUC, this is quite normal, so it should probably be `Debugf`.

api/query/cache/index/index.go
mdittmer: It prevents other packages from executing `index.ErrRunExists = somethingElse`.

api/query/cache/poll/poll.go
mdittmer: The assumption here is that the `backfill.RunFetcher` fetches runs in the order in which they are intended to be loaded, and that any other loaders (like a backfiller) will have "outpaced" this one's limit.

Under these assumptions, we should not load a series of runs' metadata, skip a few, and then find one not loaded. We should, instead, start loading runs and then reach a point at which all the rest of the runs within the limit are already loaded.

Perhaps we should check each other's interpretation offline.

api/query/cache/poll/poll.go
mdittmer: I think the debug logging level is `INFO`. Could we switch from `WARN` to `INFO`?

PR #887, opened 2018-12-11
Search cache: Inject base query to ignore irrelevant tests

mdittmer:
This change filters out tests that are missing from all test runs of interest when serving queries from the search cache. This is achieved by turning every query, q, in to AND(baseQuery, q), where baseQuery ensures that results filter out (sub)tests that do not appear in runs bound to the query from any of the four major browsers.


api/query/cache/service/main.go
```diff
@@ -34,6 +35,39 @@ var (
 
 	idx index.Index
 	mon monitor.Monitor
+
+	// Base query to AND against user's query: only include (sub)tests with
+	// non-UNKNOWN status for at least one of the four major browsers.
+	// I.e.,
+	//     !chrome:UNKNOWN | !edge:UNKOWN | !firefox:UNKNOWN | !safari:UNKNOWN
+	baseQuery = query.AbstractOr{
```
lukebjerring: This is tightly coupled to the default browser set, which 
a) doesn't handle say, uc-browser
b) has more non-empty checks than it needs to on say, chrome v chrome

Can we make this a function of the runs that are being queried, or can we implement an `!any-browser:UNKNOWN` type of thing?

mdittmer: @lukebjerring Just updated code to construct a concrete query over actual run IDs from the request. PTAL.

api/query/cache/service/main.go
```diff
@@ -85,6 +87,55 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
...
+				Run:    runID,
+				Status: shared.TestStatusUnknown,
+			},
+		}
+	}
+
+	// Add baseQuery to existing AND in q=AND(...), or create AND(baseQuery, q).
+	if andQ, ok := q.(query.And); ok {
+		andQ.Args = append([]query.ConcreteQuery{baseQuery}, andQ.Args...)
+		q = andQ
```
lukebjerring: Nit: Do you expect this optimization to matter much? If not, I don't see the harm in a tree that's 1 layer deeper.

mdittmer: It could matter because we will eventually want to optimize the order in which predicate fragments run. This optimization is much easier to do heuristically if we keep filters as "flat" as possible.

The other reason is consistency: Other `AND`/`OR` managing code contains these checks to keep queries "flat".

mdittmer:
As per discussion offline, this will be tested in a follow-up PR.

Issue #538, opened 2018-09-12
Discuss and define the terminologies for release channels of major browsers


Hexcles:
@lukebjerring raised this question earlier. We should investigate what the various release channels of (four) major browsers really mean, and try our best to align them under a set of well-defined terminologies in wpt.fyi. An example would be four channels: stable, beta, dev, experimental (though we need to flesh out the definitions). Based on that, we can create labels for these channels and make sure we are comparing apples to apples on wpt.fyi.

Currently on wpt.fyi we only have two labels, `stable` and `experimental`. With beta results coming from Taskcluster, we'll soon have `beta`, too.

lukebjerring:
We originally jotted a proposal here:
https://docs.google.com/presentation/d/1f5MsezY4TRz4W-JTnqPuW99WFtMp24WNYaPZ6OSdmTM/edit#slide=id.g3ca2f40f94_0_24 

Summary:
- **stable**: The current public release of a browser.
- **beta**: A pre-stable release branch of a browser.
- **dev**: Tip-of-tree branch of a browser, generally updated very regularly.
- **experimental**: `dev`, _including_ experimental features (via enabling flags).

Notes:
- `dev` should be as close to tip of tree as possible.
Picking on Chromium as an example, we actually have more than one channel that fall under wpt.fyi's `dev` definition - `Dev` and `Canary`. For our purposes, we'd just take the most regularly updating of the options.
- `beta` should just be the _next_ branch that will be released as `stable`

foolip:
In https://github.com/web-platform-tests/wpt.fyi/pull/590#issuecomment-426418724 and https://github.com/web-platform-tests/wpt.fyi/issues/611, I think the root problem is that we want mutually exclusive labels for channels, but "experimental" isn't a channel, but a channel + flags. This means we can't accurately label the the channel for these runs.

I think it would probably be beneficial to have one set of labels for channel *only*, which is mutually exclusive, and where we match the nomenclature of each browser. It would mean that we use chrome[dev] and firefox[nightly], but it's possible that we'd add chrome[canary] and firefox[dev] as well, which aren't the same.

One problem is that we already have lots of links to `?label=experimental`, so we probably can't make "experimental" mean exactly "including experimental features (via enabling flags)", but rather just "the most bleeding edge configuration we have for this product".

lukebjerring:
Can we convert that vague proposal into a literal proposal here, @foolip ?

foolip:
Alright :)

[...]

PR #889, opened 2018-12-11
Checks: Plumb channel and improve robustness of diff urls

lukebjerring:
## Description
See https://github.com/web-platform-tests/wpt/issues/14465

This avoids version strings in a couple of cases.


shared/run_diff.go
```diff
@@ -24,7 +24,7 @@ import (
 type DiffAPI interface {
 	GetRunsDiff(before, after TestRun, filter DiffFilterParam, paths mapset.Set) (RunDiff, error)
 	GetDiffURL(before, after TestRun, diffFilter *DiffFilterParam) *url.URL
-	GetMasterDiffURL(sha string, product ProductSpec) *url.URL
+	GetMasterDiffURL(sha string, product ProductSpec, diffFilter *DiffFilterParam) *url.URL
```
Hexcles: Why the change of function signature? Is there (or will there be) any call of `GetMasterDiffURL` that specifies non-nil diff filter param?

shared/run_diff.go
```diff
@@ -42,10 +42,13 @@ func NewDiffAPI(ctx context.Context) DiffAPI {
...
 func (d diffAPIImpl) GetDiffURL(before, after TestRun, diffFilter *DiffFilterParam) *url.URL {
 	filter := TestRunFilter{}
-	filter.Products = ProductSpecs{
-		ProductSpec{ProductAtRevision: before.ProductAtRevision},
-		ProductSpec{ProductAtRevision: after.ProductAtRevision},
-	}
+	filter.Products = make(ProductSpecs, 2)
+	filter.Products[0].BrowserName = before.BrowserName
+	filter.Products[0].Revision = before.Revision
+	filter.Products[0].Labels = mapset.NewSet(before.Channel())
```
Hexcles: What about including all the labels?

lukebjerring: Sigh. Let's land https://github.com/web-platform-tests/wpt.fyi/pull/906 and then it doesn't matter at all.

shared/run_diff.go
lukebjerring: Consistency.

shared/models.go
```diff
@@ -119,6 +119,20 @@ func (r TestRun) IsExperimental() bool {
...
 }
 
+// Channel return the channel label, if any, for the given run.
+func (r TestRun) Channel() string {
+	for _, label := range r.Labels {
+		switch label {
+		case StableLabel,
+			ExperimentalLabel,
+			DevLabel,
+			BetaLabel:
```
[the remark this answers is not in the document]
lukebjerring: Yeah.. ok . Fixed.
A) #538 needs to be concluded
B) the query builders are wrong for a different, similar reason.

api/checks/update.go
```diff
@@ -172,8 +172,12 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, baseRun,
 	checkState := summaries.CheckState{
 		TestRun: &headRun,
 		Product: shared.ProductSpec{
-			ProductAtRevision: headRun.ProductAtRevision,
-			Labels:            labels,
+			// [browser]@[sha] is plenty specific, and avoids bad version strings.
```
Hexcles: This is probably not specific enough in some cases.

Example: when `headRun` is a PR run, we would only pass `browser[channel]@revision` to `GetMasterDiffURL`, which will match to both `pr_base` and `pr_head` runs, and the users would always expect the `pr_head` run.

api/checks/update.go
```diff
@@ -172,8 +172,12 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, baseRun,
 	checkState := summaries.CheckState{
```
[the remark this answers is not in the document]
lukebjerring: Done.

api/receiver/create_run.go
```diff
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/deckarep/golang-set"
```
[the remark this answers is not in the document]
lukebjerring: Done.

shared/params.go
```diff
@@ -113,13 +113,18 @@ func (p ProductSpecs) Strings() []string {
...
-	if p.Labels != nil && p.Labels.Cardinality() > 0 {
-		labels := make([]string, 0, p.Labels.Cardinality())
-		for l := range p.Labels.Iter() {
-			labels = append(labels, l.(string))
+	if p.Labels != nil {
+		p.Labels.Remove("") // Remove the empty label, if present.
+		if p.Labels.Cardinality() > 0 {
+			labels := make([]string, 0, p.Labels.Cardinality())
+			for l := range p.Labels.Iter() {
+				if l != "" {
```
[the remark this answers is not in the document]

api/checks/update.go
lukebjerring: This is deliberately the case. This only needs to identify any _runs_ precisely enough for a given check; i.e. the CheckState's product is specific enough to identify chrome[experimental], and what the means as a check depends on what's been run (e.g. `master` vs an earlier `master`, `pr_base` vs `pr_head`, etc.)

As for below, let's rejig the API to take a run to diff against master, and then we can be less ambiguous :)

lukebjerring:
Double-checking behaviour @ https://github.com/lukebjerring/wpt/pull/8/checks?sha=6d14f4b50f4e9121bd55504a6cc33b471569eb2d

Issue #895, opened 2018-12-12
Link to the commit on GitHub


jgraham:
The SHA1 of different runs is a link but it links to the results for that run. It would be useful to also/instead have a link to the changeset on GitHub so that it would be easy to check what was being tested. In particular with diff links it's helpful to ensure that you got the comparison the right way around.

Hexcles:
Fixed by #900

PR #900, opened 2018-12-12
Add feature for showing github commit link

lukebjerring:
## Description
As per #895, adds a feature (under /flags) for showing a link to GitHub for the commit (revision) used for a test run.

The link is straight to web-platform-tests/wpt/commit/[sha], which has a few problems:
- The jump from there to the PR isn't a very obvious link
- Commits from a fork aren't linked correctly

On the bright side, we don't run an API fetch to GitHub for every single SHA to search for its PR...

cc @jgraham 

## Review Information
Head to /flags, enable the github links, see the links in the header on the homepage.


lukebjerring:
FWIW, looks like this:

image

jgraham:
I wonder if there's some way we can avoid repeating the hash, which is pretty visually noisy. I might even argue that this link is more important than the "take me to some default runs for this revision" link, which is pretty much superceeded by the query builder. But if people don't like removing that, we could consider something like `@c1faeb4eb5 [GitHub]`, where `[Github]` would be the GH logo.

lukebjerring:
Hmmm.. We can probably do without the @sha link to wpt.fyi itself, since we have a query builder with a SHA field. GitHub can just take its place. Let me tweak the feature.

lukebjerring:
I've made the feature swap out the link, and enabled the feature staging-wide, so you should be able to see it @ https://pr-link-dot-wptdashboard-staging.appspot.com/admin/flags once 
https://github.com/web-platform-tests/wpt.fyi/pull/900/commits/f6df9283f2c813178bb94e12df5e25bbeb44958c concludes

webapp/components/test-run.html
```diff
@@ -53,7 +58,15 @@
         <div>{{displayName(testRun.browser_name)}} {{shortVersion(testRun.browser_name, testRun.browser_version)}}</div>
         <template is="dom-if" if="{{ !isDiff(testRun.browser_name) }}">
           <div>{{displayName(testRun.os_name)}} {{testRun.os_version}}</div>
-          <div class="revision">@<a href="?sha={{testRun.revision}}">{{testRun.revision}}</a></div>
+          <template is="dom-if" if="[[githubCommitLinks]]">
+            <div class="github">
+              <img src="/static/github.svg" />
```
foolip: Can you sprinkle some CSS so that this icon is vertically aligned with the text no matter the text size? Seems a few px off now.

PR #906, opened 2018-12-12
Plumb run_id params through the UI

lukebjerring:
## Description
Support loading /results with a `run_id` param (or a few), by plumbing the params through to the component.

## Review Information
- Load /results?run_id=[A run ID]&run_id=[Another run ID]
- Be amazed

This is a precursor to a better fix for #889


lukebjerring:
Verified via https://run-ids-dot-wptdashboard-staging.appspot.com/results/?run_id=6341547230494720

PR #909, opened 2018-12-12
Fix handling and logging of golang-frontend->golang-cache request

mdittmer:


Hexcles:
Essentially when `client.Do` fails the response is not usable at all. When it succeeds, you'd still need to check the status code and log and produce an error yourself if it is not 2XX.

api/query/search.go
```diff
@@ -131,14 +131,26 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+			return
+		}
+		req.Header.Add("Content-Type", "application/json")
+
+		resp, err := client.Do(req)
+		if err != nil {
+			logger.Errorf("Error connecting to search API cache: %v", err)
+			http.Error(w, "Error connecting to search API cache", http.StatusInternalServerError)
+			return
+		} else if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
Hexcles: 1. No need to have the `else` because of early return.
2. More importantly (though not introduced by this PR), `resp.Body` isn't closed.

```suggestion
		}
		defer resp.Body.Close()
		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```

PR #910, opened 2018-12-13
Add azure pipelines source logo

lukebjerring:
## Description
Treats the `azure` label as a source, shows a sweet, sweet rocket.

## Review Information
- Visit /runs?label=azure
- Be amazed


webapp/components/display-logo.html
```diff
@@ -114,9 +116,8 @@
         if (!showSource || !product.labels) {
           return '';
         }
-        const sources = ['taskcluster', 'buildbot', 'msedge'];
-        const labels = new Set(product.labels);
-        return sources.find(s => labels.has(s));
+        const sources = window.wpt.Sources;
+        return product.labels.find(s => sources.has(s));
```
Hexcles: Nit: maybe just combine the two lines.

lukebjerring: I think that would execute the getter, which creates a new set, every loop.

foolip:
I love it, can't wait to see this logo for full runs too!

PR #913, opened 2018-12-13
Collapse results summaries for non-regression output

lukebjerring:
## Description
Iteratively reduce the maximum depth of the paths by collapsing folders until the total number of folders is under the limit desired (or they can't be collapsed further).

This is needed for full runs, where we're showing 10 results from 2dcontext, "And 1000s more..."

### DIFF
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"net/http"
 	"sort"
+	"strings"
 	"time"
 
 	mapset "github.com/deckarep/golang-set"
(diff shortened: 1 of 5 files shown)


api/checks/update.go
```diff
@@ -256,3 +258,77 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, baseRun,
...
+		}
+	}
+	return result
+}
+
+func collapsePaths(keys []string, limit int) mapset.Set {
+	result := shared.NewSetFromStringSlice(keys)
+	// 10 iterations to avoid edge-case infinite looping risk.
+	for i := 0; i < 10 && result.Cardinality() > limit; i++ {
+		sort.Sort(pathKeys(keys))
```
foolip: Do we really need to sort in each iteration? Assert `issorted` or something instead?

lukebjerring: We use a set to track what collapsed paths remain after each loop, but we want to iterate those in longest-first order. The alternative here would be to mutate the array as we go, or do a pre-iteration, but I don't think that code complexity would save us much time (we're only talking sorting 30k items up to 10 times).

Issue #914, opened 2018-12-13
Only show master runs on the recent runs page


Hexcles:
I don't think PR runs are useful on the ["recent runs" page](https://wpt.fyi/test-runs). They are transient by their nature.

lukebjerring:
This is /runs part of #702

PR #916, opened 2018-12-13
Default to master runs in the /runs ui

lukebjerring:
## Description
Adds a label of `master` by default for /runs

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/914

### DIFF
--- a/webapp/routes_test.go
+++ b/webapp/routes_test.go
@@ -46,7 +46,6 @@ func TestInteropAnomaliesBound(t *testing.T) {
 
 func TestRunsBoundHSTS(t *testing.T) {
 	assertHandlerIs(t, "/test-runs", "test-runs")
-	assertHSTS(t, "/test-runs")
 }
 
 func TestApiDiffBoundCORS(t *testing.T) {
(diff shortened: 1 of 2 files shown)


webapp/routes_test.go
```diff
@@ -46,7 +46,6 @@ func TestInteropAnomaliesBound(t *testing.T) {
 
 func TestRunsBoundHSTS(t *testing.T) {
```
Hexcles: Rename the test (remove `HSTS`).

webapp/test_runs_handler.go
```diff
@@ -28,6 +29,9 @@ func testRunsHandler(w http.ResponseWriter, r *http.Request) {
 	if testRunFilter.IsDefaultQuery() {
 		aWeekAgo := time.Now().Truncate(time.Hour*24).AddDate(0, 0, -7)
 		testRunFilter.From = &aWeekAgo
+		if aeAPI.IsFeatureEnabled("masterRunsOnly") {
```
Hexcles: I feel like this should be the default regardless of the feature flag. WDYT?

PR #918, opened 2018-12-13
Show master/merge icons on /runs page

lukebjerring:
## Description
Gives a visual cue for `pr_head` and `master` labels, now that we'll occasionally have both in the same view.


foolip:
https://master-icons-dot-wptdashboard-staging.appspot.com/runs?label=azure doesn't show the master icon for the runs that aren't PRs, is that a bug?

lukebjerring:
@Hexcles - got an icon in mind for `pr_base` ?

lukebjerring:
Not really; runs don't have a concept of PR number.
What's extra strange that the commits should both be the same (the SHA of the commit that caused the run) - is it possible that Azure is misconfigured and not overriding the revision with the commit that started the check ( for `pr_base`, specifically )?

lukebjerring:
Wait; nevermind. Both `pr_head` and `pr_base` are in the group; let me fix the code.

foolip:
Yeah, I didn't do anything to override the commit. Seems like it'd be best to record all 3 commits involved for both merge and base runs, so that we can tell if the comparison is valid?

lukebjerring:
Maybe, but that's tangential to this PR.

PR #922, opened 2018-12-14
Add test status value: "ASSERT"

mdittmer:
According to https://searchfox.org/mozilla-central/source/testing/mozbase/mozlog/mozlog/logtypes.py#184-196 at time of authorship.

```python
class Status(DataType):
    allowed = ["PASS", "FAIL", "OK", "ERROR", "TIMEOUT", "CRASH", "ASSERT", "SKIP"]

    def convert(self, data):
        value = data.upper()
        if value not in self.allowed:
            raise ValueError
        return value


class SubStatus(Status):
    allowed = ["PASS", "FAIL", "ERROR", "TIMEOUT", "ASSERT", "NOTRUN", "SKIP"]
```

According to @jgraham this is not currently used, in practice, by Mozilla, but the intent is that a halting but non-crashing assertion failed.

### DIFF
--- a/shared/spanner.go
+++ b/shared/spanner.go
@@ -38,6 +38,10 @@ const (
 	// TestStatusSkip indicates that the test was disabled for this test run.
 	TestStatusSkip int64 = 8
 
+	// TestStatusAssert indicates that a non-fatal assertion failed. This test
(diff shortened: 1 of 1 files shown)


mdittmer:
wpt_test_status_assert_convo

PR #923, opened 2018-12-14
Add browser-only failure queries to the Insights tab

lukebjerring:
## Description
Adds a card for each of the 4 default browsers with a link to a query for failures occurring in only those browsers.

## Review
- Visit /admin/flags, enable the insights tab
- Visit the insights tab and structured queries
- Verify the queries are as-expected.


webapp/components/wpt-insights.html
```diff
@@ -0,0 +1,68 @@
...
+    class Insights extends ProductInfo(WPTFlags(window.Polymer.Element)) {
+      static get is() {
+        return 'wpt-insights';
+      }
+
+      get insights() {
+        const browsers = ['chrome', 'edge', 'firefox', 'safari'];
+        return browsers.map(b => {
+          const others = browsers.filter(o => o !== b);
+          const othersFailing = others
```
mdittmer: nit: The string being constructed appears to be `othersPassing`.

webapp/components/wpt-insights.html
```diff
@@ -0,0 +1,68 @@
...
+            .map(o => `(${o}:pass|${o}:ok)`)
+            .join(' ');
+          const query = `!${b}:pass !${b}:ok ${othersFailing}`;
+          const url = new URL('/results/', window.location);
+          url.searchParams.set('q', query);
+          return {
+            name: `${this.displayName(b)}-only failures`,
+            url,
+            query,
+            desc: `Tests that are failing in ${this.displayName(b)}, but passing in the other browsers (${others.map(this.displayName).join(', ')})`,
```
mdittmer: nit: Should `this.displayName` be bound before being passed to `map()`?

webapp/components/wpt-insights.html
```diff
@@ -0,0 +1,68 @@
...
+  </template>
+  <script>
+    /* global ProductInfo, WPTFlags */
+    class Insights extends ProductInfo(WPTFlags(window.Polymer.Element)) {
+      static get is() {
+        return 'wpt-insights';
+      }
+
+      get insights() {
+        const browsers = ['chrome', 'edge', 'firefox', 'safari'];
```
mdittmer: nit: Should this be configurable? (I.e., these four strings are the default value on a property.)

mdittmer: ... ditto for the `insights` property. Should we have a helper that generates a default value so that the owner of this component could, instead, bind a value to `insights`?

lukebjerring: Eventually :)

webapp/components/wpt-insights.html
lukebjerring: Nah, it doesn't use `this`, it's just looking up a key in a map.

webapp/components/wpt-insights.html
lukebjerring: Done.

PR #924, opened 2018-12-14
Namespace some more constants

lukebjerring:
## Description
Clean up a few constants, with the benefit of less pollution of `window` and also a fresh copy of things that are collections.

### DIFF
--- a/webapp/components/product-info.html
+++ b/webapp/components/product-info.html
@@ -25,15 +25,21 @@
       m.set('msedge', 'MS Edge');
       return m;
     })();
-    /*eslint-disable no-unused-vars*/
-    const VERSION_PATTERNS = {
-      major: /(\d+)/,
-      majorAndMinor: /(\d+\.\d+)/
-    };
-    const DEFAULT_BROWSER_NAMES = ['chrome', 'edge', 'firefox', 'safari'];
-    const DEFAULT_PRODUCT_SPECS = Array.from(DEFAULT_BROWSER_NAMES);
-    const DEFAULT_PRODUCTS = DEFAULT_PRODUCT_SPECS.map(p => parseProductSpec(p));
-
+    Object.defineProperty(window.wpt, 'VersionPatterns', {
+      value: Object.freeze({
+        Major: /(\d+)/,
+        MajorAndMinor: /(\d+\.\d+)/,
+      }),
+    });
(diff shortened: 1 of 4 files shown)


webapp/components/product-info.html
```diff
@@ -25,15 +25,21 @@
...
+      value: Object.freeze({
+        Major: /(\d+)/,
+        MajorAndMinor: /(\d+\.\d+)/,
+      }),
+    });
+    Object.defineProperty(window.wpt, 'DefaultBrowserNames', {
+      get: () => ['chrome', 'edge', 'firefox', 'safari'],
+    });
+    Object.defineProperty(window.wpt, 'DefaultProductSpecs', {
+      get: () => window.wpt.DefaultBrowserNames,
```
mdittmer: Why do we have both of these? `DefaultProductSpecs` and `DefaultBrowserNames`?

webapp/components/product-info.html
```diff
@@ -25,15 +25,21 @@
...
-    const DEFAULT_PRODUCTS = DEFAULT_PRODUCT_SPECS.map(p => parseProductSpec(p));
-
+    Object.defineProperty(window.wpt, 'VersionPatterns', {
+      value: Object.freeze({
+        Major: /(\d+)/,
+        MajorAndMinor: /(\d+\.\d+)/,
+      }),
+    });
+    Object.defineProperty(window.wpt, 'DefaultBrowserNames', {
+      get: () => ['chrome', 'edge', 'firefox', 'safari'],
```
mdittmer: nit: Why create a new Array every time? `Object.freeze()` should work on an array.

webapp/components/product-info.html
```diff
@@ -25,15 +25,21 @@
...
+    });
+    Object.defineProperty(window.wpt, 'DefaultProductSpecs', {
+      get: () => window.wpt.DefaultBrowserNames,
+    });
+    Object.defineProperty(window.wpt, 'DefaultProducts', {
+      get: () => window.wpt.DefaultProductSpecs.map(p => parseProductSpec(p)),
```
mdittmer: This is pretty weird. Where is`parseProductSpec()` defined? I don't see it in prior lines of this file.

mdittmer: nit: We should freeze this Array too.

PR #929, opened 2018-12-14
Treat sha as a prefix of FullRevisionHash

lukebjerring:
## Description
Changes the datastore treatment of the `sha` param to be "a prefix of the `FullRevisionHash` field".
This steps towards #49 and also supports 7-char SHAs.

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -32,19 +32,21 @@ func LoadTestRunsBySHAs(ctx context.Context, shas ...string) (runs TestRuns, err
 			sha = sha[:10]
 		}
 		q := datastore.NewQuery("TestRun")
-		keys, err := loadKeysForRevision(ctx, q, sha)
+		ids, err := loadKeysForRevision(ctx, q, sha)
 		if err != nil {
 			return runs, err
 		}
-		shaRuns := make(TestRuns, len(keys))
-		for i := range keys {
-			run, err := LoadTestRun(ctx, keys[i].IntID())
+		shaRuns := make(TestRuns, len(ids))
+		for i := range ids {
+			run, err := LoadTestRun(ctx, ids[i])
 			if err != nil {
 				return nil, err
 			}
 			shaRuns[i] = *run
 		}
-		shaRuns.SetTestRunIDs(keys)
+		for i := range ids {
+			shaRuns[i].ID = ids[i]
+		}
 		runs = append(runs, shaRuns...)
 	}
 	return runs, err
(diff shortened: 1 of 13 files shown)


lukebjerring:
One sec, relaxing the 10-char minimum for the query param handling...

webdriver/webapp_server.go
```diff
@@ -28,7 +29,7 @@ var (
 )
 
 // StaticTestDataRevision is the SHA for the local (static) test run summaries.
-const StaticTestDataRevision = "b952881825"
+const StaticTestDataRevision = "b952881825e7d3974f5c513e13e544d525c0a631"
```
Hexcles: If you change this variable, you'll need to either fix this line:
https://github.com/web-platform-tests/wpt.fyi/blob/c6886c69ba71795f11be1feb71ec4c58185feda2/webdriver/webapp_server.go#L252

or rename the static directory.

util/populate_dev_data.go
```diff
@@ -45,7 +45,7 @@ func main() {
 	staticDataTime, _ := time.Parse(time.RFC3339, "2017-10-18T00:00:00Z")
 
 	// Follow pattern established in run/*.py data collection code.
-	const staticRunSHA = "b952881825"
+	const staticRunSHA = "b952881825e7d3974f5c513e13e544d525c0a631"
```
Hexcles: Same.

shared/params.go
```diff
@@ -140,39 +140,22 @@ const MaxCountMaxValue = 500
 // MaxCountMinValue is the minimum allowed value for the max-count param.
 const MaxCountMinValue = 1
 
-// SHARegex is a regex for SHA[0:10] slice of a git hash.
-var SHARegex = regexp.MustCompile("[0-9a-fA-F]{10,40}")
+// SHARegex is a regex for 7 to 40 char prefix of a git hash.
```
Hexcles: Why is the minimum 7? Is it trying to make sure SHA is unique? What'll happen is the prefix isn't unique?

shared/datastore.go
```diff
@@ -181,24 +218,45 @@ func contains(s []string, x string) bool {
 	return false
 }
 
-// Loads any keys for a full string match or a version prefix (Between [version].* and [version].9*)
-func loadKeysForBrowserVersion(ctx context.Context, query *datastore.Query, version string) (result *mapset.Set, err error) {
+// Loads any keys for a revision prefix or full string match
+func loadKeysForRevision(ctx context.Context, query *datastore.Query, sha string) (result []*datastore.Key, err error) {
```
Hexcles: Looks like all the call sites convert the return value to `IntID`. Shall we just return the int ID (also for consistency with the other `load*` function)?

shared/params.go
lukebjerring: Don't have a great reason, but the shorter we get, the more likely to clash. If it's not unique, it'll just match more runs. Do you prefer a different minimum?

shared/datastore.go
lukebjerring: Done.

util/populate_dev_data.go
lukebjerring: Done.

webdriver/webapp_server.go
lukebjerring: Done.

Issue #663, opened 2018-10-16
Detect renames when diffing across revisions


lukebjerring:
Currently, a simple file rename would be detected as N deleted tests and N added tests, rather than what's effectively a no-op in terms of pass rates.

lukebjerring:
Related to https://github.com/web-platform-tests/wpt.fyi/issues/644

lukebjerring:
Example of a (currently) bad diff:
https://wpt.fyi/results/css/css-shapes/shape-outside/values?product=chrome[experimental]@4b370b4c0b&product=chrome[experimental]@bb9c0ce15e&diff=true&q=shape-outside-inset-0

Hexcles:
We can use GitHub API to get a structured output of `git diff`, which has a list of `files` with status (ACDMR...). e.g.

https://api.github.com/repos/web-platform-tests/wpt/compare/4b370b4c0b...bb9c0ce15e

This saves us the trouble of having to checkout the repository and shelling out to git.

The caveat is:

> The response will include a comparison of up to 250 commits. If you are working with a larger commit range, you can use the Commit List API to enumerate all commits in the range.

Hexcles:
I'd assume that this API has the same internal mechanism as `git`, i.e. heuristic-based rename detection. I don't see any parameter (namely similarity index) in the API. I'd assume it's the git default, 50%.

lukebjerring:
A canonical example of exploded renames:
https://staging.wpt.fyi/results/workers?diff&run_id=5969501392732160&run_id=5487233003945984

PR #932, opened 2018-12-17
Explode possible renames for .window.js / .any.js test files

lukebjerring:
## Description
Further work for #663 

Adds a utility for exploding possible renames for `.any.js`, which will improve rename handling for diffs.


lukebjerring:
A merge of this and #933 shows the rename handling working:
https://temp-dot-wptdashboard-staging.appspot.com/results/workers?diff&run_id=5969501392732160&run_id=5487233003945984

shared/manifest.go
```diff
@@ -0,0 +1,57 @@
...
+import "strings"
+
+const (
+	windowJSSuffix                 = ".window.js"
+	windowHTMLSuffix               = ".window.html"
+	anyJSSuffix                    = ".any.js"
+	anyJSWindowHTMLSuffix          = ".any.html"
+	anyJSDedicatedWorkerHTMLSuffix = ".any.worker.html"
+	anyJSServiceWorkerHTMLSuffix   = ".any.serviceworker.html"
+	anyJSSharedWorkerHTMLSuffix    = ".any.sharedworker.html"
```
Hexcles: There are in fact two types of different constants here: `.window.js` and `.any.js` are the keys, while the rest `*.html` are the expanded values, so perhaps we should define a map instead.

Besides, you are missing `.worker.js`.

lukebjerring: Changed to a map + added worker, PTAL

shared/manifest.go
```diff
@@ -0,0 +1,53 @@
...
+			result[eBefore[i]] = eAfter[i]
+		}
+	}
+	return result
+}
+
+// ExplodePossibleFilenames explodes the given single filename into the test names that
+// could be created for it at runtime.
+func ExplodePossibleFilenames(filePath string) []string {
+	// https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests
```
Hexcles:
```suggestion
```

shared/manifest.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import "strings"
+
+// explosions returns a map of the exploded test by filename suffix.
```
Hexcles:
```suggestion
// explosions returns a map of the exploded test by filename suffix.
// https://web-platform-tests.org/writing-tests/testharness.html#multi-global-tests
```

shared/manifest.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// ExplodePossibleRenames returns a map of equivalent renames for the given file rename.
+func ExplodePossibleRenames(before, after string) map[string]string {
+	result := map[string]string{
+		before: after,
+	}
+	eBefore := ExplodePossibleFilenames(before)
+	eAfter := ExplodePossibleFilenames(after)
+	if len(eBefore) == len(eAfter) {
```
Hexcles: Optional: we could use more heuristics here if the lengths are different.

e.g. `.any.html` can be compared against `.window.html`, and `.any.worker.html` can be compared to `.worker.html`.

lukebjerring: Might follow up with other cases.

PR #933, opened 2018-12-17
Diff with HEAD for finding renames in pr_base

lukebjerring:
## Description
Further work on #663 

Note that GitHub uses the triple-dot algorithm when comparing, so even if master runs away from the PR's commit, the differences will be the master-branch point to tip-of-PR-branch, which is exactly what we want.

## Review Information
- Load /api/diff?run_id=5969501392732160&run_id=5487233003945984
- See a rename of /workers/importscripts_mime.tentative.any.js

### DIFF
--- a/shared/run_diff.go
+++ b/shared/run_diff.go
@@ -289,7 +289,11 @@ func (d diffAPIImpl) GetRunsDiff(before, after TestRun, filter DiffFilterParam, 
 
 	var renames map[string]string
 	if IsFeatureEnabled(d.ctx, "diffRenames") {
-		renames = getDiffRenames(d.ctx, before.FullRevisionHash, after.FullRevisionHash)
+		beforeSHA := before.FullRevisionHash
+		if before.FullRevisionHash == after.FullRevisionHash && before.IsPRBase() {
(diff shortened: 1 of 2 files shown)


lukebjerring:
A merge of this and #932 shows the renames working:
https://temp-dot-wptdashboard-staging.appspot.com/results/workers?diff&run_id=5969501392732160&run_id=5487233003945984

shared/run_diff.go
```diff
@@ -289,7 +289,11 @@ func (d diffAPIImpl) GetRunsDiff(before, after TestRun, filter DiffFilterParam,
 
 	var renames map[string]string
 	if IsFeatureEnabled(d.ctx, "diffRenames") {
-		renames = getDiffRenames(d.ctx, before.FullRevisionHash, after.FullRevisionHash)
+		beforeSHA := before.FullRevisionHash
```
lukebjerring:
```suggestion
 		beforeSHA := before.FullRevisionHash
 		// Use HEAD...[sha] for PR results, since PR run results always override the value of 'revision' to the PRs HEAD revision.
```

PR #936, opened 2018-12-17
Fix intrashard concurrency bug

mdittmer:
Presently there is a mutex guarding read and write to the entire searchcache index. However, one of its read operations gives an index-type object shared custody of test+result data, which leads to a concurrency bug when the index-type object attempts to read from the shared structures while the enclosing Index shard is performing a write operation on the same shared structure.

This change introduces a sync.RWMutex for each shard. A pointer to the mutex is passed to index-type objects so they can acquire a read lock. The Index-type implementation is updated to synchronize its access to shard data as too.

### DIFF
--- a/api/query/cache/index/filter.go
+++ b/api/query/cache/index/filter.go
@@ -7,6 +7,7 @@ package index
 import (
 	"errors"
 	"strings"
+	"sync"
 
 	log "github.com/sirupsen/logrus"
 	"github.com/web-platform-tests/wpt.fyi/api/query"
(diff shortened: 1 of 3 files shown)


mdittmer:
It's unclear how to test this without introducing a lot of abstraction. Here's one way:

1. Wrap the offending map in an interface;
1. Wrap the locking mechanism in an interface;
1. Write a malicious interface implementation that deliberately pauses during map reads and writes, and also checks some state shared between the map and locking mechanism that ensures that the lock is properly held when entering a read and/or write operation.

It seems about as likely that this test code would contain bugs and/or race conditions as the critical sections would. Do you agree, @lukebjerring?

lukebjerring:
Yeah I agree. How'd you diagnose it originally?

mdittmer:
Crash in staging :(

PR #938, opened 2018-12-18
Persist /runs query changes in the URL

lukebjerring:
## Description
More builder work (https://github.com/web-platform-tests/wpt.fyi/issues/535)

When the query is submitted on /runs tab, persist it to the URL.

## Review Information
- Visit the /runs page
- Change the query, submit
- See the changes in the URL
- Reload the page, see the same results

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -87,7 +87,7 @@
        * Defaults to the queryParams property.
        */
       navigationQueryParams() {
-        return Object.assign({}, this.queryParams);
+        return this.queryParams && JSON.parse(JSON.stringify(this.queryParams));
       }
 
       bindNavigate() {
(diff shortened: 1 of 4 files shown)


mdittmer:
When trying the staging instance, I noticed the back button doesn't update the list of runs to the former state.

webapp/components/self-navigator.html
```diff
@@ -84,9 +84,10 @@
 
       /**
        * Get query params to persist when creating history.
+       * Defaults to the queryParams property.
        */
       navigationQueryParams() {
-        return {};
+        return Object.assign({}, this.queryParams);
```
mdittmer: Just double-checking: Is `queryParams` a shallow object? Otherwise we may get some wonkiness here.

lukebjerring: Done.

lukebjerring:
Back-button is broken for the /results/ tab too, so I'll fix that issue separately.

PR #939, opened 2018-12-18
Cleanup redundant seconds info in timestamps

lukebjerring:
## Description
Reduces a little bloat in urls that have a timestamp.


lukebjerring:
Probably worth looking at after the Polymer 3 mayhem.

Issue #941, opened 2018-12-18
Back button does not reset query state


lukebjerring:
Found during https://github.com/web-platform-tests/wpt.fyi/pull/938

- Visit staging.wpt.fyi, change the query
- Hit back
- Query is not reset

PR #943, opened 2018-12-18
Include landing state in history stack

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/941

### DIFF
--- a/webapp/components/self-navigator.html
+++ b/webapp/components/self-navigator.html
@@ -142,7 +142,7 @@
       }
 
       getLocation(params) {
-        const url = new URL(location);
+        const url = new URL(window.location);
         url.search = '';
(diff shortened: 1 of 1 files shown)


webapp/components/self-navigator.html
```diff
@@ -138,6 +140,18 @@
         url.pathname = this.navigationPathPrefix() + testPath;
         this.navigateToLocation(url);
       }
+
+      getLocation(params) {
+        const url = new URL(location);
```
mdittmer: nit: if this is `window.location` we should be consistent and qualify it as such.

lukebjerring:
```suggestion
        const url = new URL(window.location);
```

PR #944, opened 2018-12-18
Keep the query message visible between submissions in /runs

lukebjerring:
## Description
Changes the message to be visible at all times, instead of only when runs are loaded.

### DIFF
--- a/webapp/components/test-runs-query-builder.html
+++ b/webapp/components/test-runs-query-builder.html
@@ -455,7 +455,7 @@ found in the LICENSE file.
           },
           channels: {
             type: Array,
-            value: window.wpt.Channels,
+            value: Array.from(window.wpt.Channels),
           },
           versionsURL: {
             type: String,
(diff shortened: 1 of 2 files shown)


webapp/components/test-runs-query-builder.html
```diff
@@ -455,7 +455,7 @@ <h3>
           },
           channels: {
             type: Array,
-            value: window.wpt.Channels,
+            value: Array.from(window.wpt.Channels),
```
lukebjerring: This is a drive-by fix for a Polymer binding warning (Channels is a `Set`)

webapp/components/wpt-runs.html
```diff
@@ -340,6 +347,7 @@
         const queryBefore = this.query;
         const builder = this.shadowRoot.querySelector('test-runs-query-builder');
         this.editingQuery = false;
+        this.nextPageToken = null;
```
lukebjerring: This ensures that the "Load more" button is cleared away when changing query.

PR #949, opened 2018-12-19
Add ignore action to cancelled check

lukebjerring:
## Description
wpt blocks on cancelled checks, so include the ignore action to mark them successful.

Also, drive-by reorganization of the actions to all be in one file.


foolip:
Are there cases where long term manually canceling will still be required? If not, can you file an issue to remove this button once it's no longer needed?

lukebjerring:
As long as we are able to cancel we should be able to ignore cancellation.
Cancelling a check that might take time you don't want to wait is pretty standard.

PR #951, opened 2018-12-19
Fix binding for Not-clause when multiple products match browser name

lukebjerring:
## Description
WIP - currently just a failing unit test.


mdittmer:
This change has been replaced by the new `browser:!STATUS` syntax implemented in #986 and #997

PR #955, opened 2018-12-20
Put pending checks behind a feature

lukebjerring:
## Description
Stops pending checks for now, until we've resolved empty report behaviour.

### DIFF
--- a/webapp/components/wpt-flags.html
+++ b/webapp/components/wpt-flags.html
@@ -44,6 +44,7 @@ found in the LICENSE file.
           'runsByPRNumber',
           'failChecksOnRegression',
           'checksAllUsers',
+          'pendingChecks',
         ];
       }
     });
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-flags.html
```diff
@@ -228,16 +229,22 @@ <h4>Server-side only features</h4>
...
     </paper-item>
-    <paper-item>
+    <paper-item sub-item>
       <paper-checkbox checked="{{checksAllUsers}}">
         Run the wpt.fyi GitHub status check for all users, not just whitelisted ones.
       </paper-checkbox>
     </paper-item>
+    <paper-item sub-item>
+      <paper-checkbox checked="{{pendingChecks}}">
+        Create pending GitHub status check when results first arrive, and are being processed.
```
foolip: Ah, I didn't realize that's when we create it, I assumed we create it even earlier. With this enabled, the time the check is pending should then be a measure of how long the queuing time is. It should be possible to bring that down to just a few seconds, right?

lukebjerring: When enabled, it's the time it takes for the processor to create a summary and call us back with a POST to /api/results/create. This can be as slow as ~5min, depending on the size of the report.

foolip: Oh, limited by the time it takes to upload split results to GCS I guess. Getting rid of those files would be sweet :)

PR #956, opened 2018-12-20
Fix navigation

lukebjerring:
## Description
Broken in #943


lukebjerring:
TBR'ing this since it's a pretty major bug.

PR #961, opened 2018-12-20
Include sender as label for results

lukebjerring:
## Description
Will help authors easily check for their own runs arriving.

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -164,7 +164,7 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 	if status.IsOnMaster() {
 		labels.Add(shared.MasterLabel)
 	} else {
-		sender := status.GetSender().GetLogin()
+		sender := status.GetCommit().GetAuthor().GetLogin()
 		if sender != "" {
 			labels.Add(sender)
 		}
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -160,9 +160,14 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 	// The default timeout is 5s, not enough for the receiver to download the reports.
 	slowCtx, cancel := context.WithTimeout(ctx, resultsReceiverTimeout)
 	defer cancel()
-	var labels []string
+	labels := mapset.NewSet()
 	if status.IsOnMaster() {
-		labels = []string{"master"}
+		labels.Add(shared.MasterLabel)
+	} else {
+		sender := status.GetSender().GetLogin()
+		if sender != "" {
+			labels.Add(sender)
+		}
 	}
 	checksAPI := checks.NewAPI(ctx)
 	err = createAllRuns(
@@ -175,7 +180,7 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 		username,
 		password,
 		urlsByProduct,
-		labels)
+		shared.ToStringSlice(labels))
 	if err != nil {
 		return false, err
 	}
--- a/api/azure/webhook.go
+++ b/api/azure/webhook.go
@@ -16,6 +16,8 @@ import (
 	"strings"
 	"time"
 
+	mapset "github.com/deckarep/golang-set"
+
 	"github.com/google/go-github/github"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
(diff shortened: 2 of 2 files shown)


api/taskcluster/webhook.go
```diff
@@ -160,9 +160,14 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 	// The default timeout is 5s, not enough for the receiver to download the reports.
 	slowCtx, cancel := context.WithTimeout(ctx, resultsReceiverTimeout)
 	defer cancel()
-	var labels []string
+	labels := mapset.NewSet()
 	if status.IsOnMaster() {
```
foolip: Not related to this change, but I'm curious if master on a fork will match this condition? (That might result in confusion.)

api/taskcluster/webhook.go
```diff
@@ -160,9 +160,14 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
...
-		labels = []string{"master"}
+		labels.Add(shared.MasterLabel)
+	} else {
+		sender := status.GetSender().GetLogin()
+		if sender != "" {
+			labels.Add(sender)
```
foolip: Similar to a previous discussion of branch names as labels, doesn't this need some prefix since it's effectively user input? `user_`?

api/taskcluster/webhook.go
```diff
@@ -160,9 +160,14 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 	// The default timeout is 5s, not enough for the receiver to download the reports.
 	slowCtx, cancel := context.WithTimeout(ctx, resultsReceiverTimeout)
 	defer cancel()
-	var labels []string
+	labels := mapset.NewSet()
 	if status.IsOnMaster() {
-		labels = []string{"master"}
+		labels.Add(shared.MasterLabel)
+	} else {
+		sender := status.GetSender().GetLogin()
```
foolip: Where does this information come from in the API? https://api.github.com/repos/web-platform-tests/wpt/commits/7c6c6afac5e1f9ed9cae1159c3f8a58ad27ff9b8/statuses is for https://github.com/web-platform-tests/wpt/pull/14629#issuecomment-449174461 and it doesn't seem to include a sender. I guess it's only for PRs?

api/taskcluster/webhook.go
lukebjerring: Yes, it would, but it'd have to be a PR from master to... something. And I think the logic here being questionable will probably carry over to the config for taskcluster choosing master vs other stuff.

api/taskcluster/webhook.go
lukebjerring: While I agree with the prospect of potential problem, there's already a precedent for the uploader being one of the labels, so until we see handles like `firefox` or `master` in practice, I don't think we need to worry.

api/taskcluster/webhook.go
lukebjerring: Good catch actually; checks attach the sender (who triggered the suite, by pushing the commit).
Changed this to be `status.GetCommit().GetAuthor().GetLogin()` instead.

api/taskcluster/webhook.go
Hexcles: Uploaders are manually created. We try our best to avoid ambiguity. For example, we didn't call the Microsoft uploader edge but msedge instead.

User handles, on the other hand, are completely out of our control. firefox, chrome, master, etc. are all existing users. I'm concerned that it would be hard, if not infeasible, to batch modify existing tags later on when we find out there are already name clashes in our Datastore.

api/taskcluster/webhook.go
foolip: master-to-master PRs are pretty common, it's what happens when people just clone their fork and make some changes.

lukebjerring: Yes, but the task names are different in prs, so we already identify them as pr_head / pr_base by that stage.

Hexcles:
Well, by escaping I mean things like commas (which we use to separate labels), and other chars that are not URL-safe that need to be escaped properly by the frontend especially when we are manually constructing the URL.

foolip:
@Hexcles to make such user names work, shouldn't we just store them verbatim and use URL encoding in the URLs where we try to match the label? Perhaps that is exactly what you're saying?

Hexcles:
@foolip there are two different problems. For the non-URL-safe characters, we just need to make sure we properly encode the params in URLs in frontend (which I'm not sure of); for the commas, it's unfortunately more complicated: commas are URL-safe, but we use them as separators; there is no easy way to tell in the backend whether a comma is a separator or part of a username, unless we do an *additional* layer of encoding/escaping for commas specifically.

foolip:
Ah, right. We could just not add the label if the username contains commas and other internally code points.

PR #962, opened 2018-12-21
Add FullRevisionHash indexes

lukebjerring:
## Description
#929 changed `sha` param behaviour, but forgot to create the indexes (because locally, everything is indexed).

This takes all indexes with `Revision` and duplicates them to the same index with `FullRevisionHash`.


lukebjerring:
Done (via `gcloud datastore indexes create webapp/index.yaml`, both wptdashboard and wptdashboard-staging).

PR #963, opened 2018-12-21
Prefix user, add user param

lukebjerring:
## Description
Follow up some concerns about https://github.com/web-platform-tests/wpt.fyi/pull/961

Prefixes username labelling behaviour with `user:` to ensure we don't clash with existing semantic labels, like `master` or `firefox`. Little bit of paranoid programming here, but beats the headache of trying to retroactively fix the data if we manage to find a username that causes problems.

Treats `?user=lukebjerring` as an alias for `?label=user:lukebjerring`

### DIFF
--- a/api/azure/webhook.go
+++ b/api/azure/webhook.go
@@ -70,7 +70,7 @@ func handleCheckRunEvent(log shared.Logger, azureAPI API, aeAPI shared.AppEngine
 		labels := mapset.NewSet()
 		sender := event.GetSender().GetLogin()
 		if sender != "" {
-			labels.Add("user:" + sender)
+			labels.Add(shared.GetUserLabel(sender))
 		}
 		if artifact.Name == "results" {
 			labels.Add(shared.PRHeadLabel)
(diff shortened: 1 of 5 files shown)


api/taskcluster/webhook.go
```diff
@@ -167,7 +167,7 @@ func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
 	} else {
 		sender := status.GetCommit().GetAuthor().GetLogin()
 		if sender != "" {
-			labels.Add(sender)
+			labels.Add("user:" + sender)
```
Hexcles: There's existing test case for this in `webhook_test.go` (search for "master"). Please adjust the test accordingly to cover this as well.

api/azure/webhook.go
```diff
@@ -70,7 +70,7 @@ func handleCheckRunEvent(log shared.Logger, azureAPI API, aeAPI shared.AppEngine
 		labels := mapset.NewSet()
 		sender := event.GetSender().GetLogin()
 		if sender != "" {
-			labels.Add(sender)
+			labels.Add("user:" + sender)
```
Hexcles: Optional: it would be good to test the Azure webhook, too.

api/taskcluster/webhook.go
lukebjerring: Added `"user:person"` to the labels we pass, and changed the spec for pending checks to not include anything other than the channel.

api/azure/webhook.go
lukebjerring: Will do in a follow-up cleanup.

PR #966, opened 2018-12-21
Tweak the backfill behaviour to get the same number of each product

lukebjerring:
## Description
Refactors the datastore code to support being inside or outside appengine standard, so that we can call the shared library methods for fetching runs with our query params (max-count, etc).

Then, uses the shared library to ensure we get the same number of each product, instead of a disproportionately large number of runs for chrome/firefox (which happen much, much more frequently).

This will allow us to run the flakiness queries against safari / edge.

### DIFF
--- a/api/query/cache/backfill/backfill.go
+++ b/api/query/cache/backfill/backfill.go
@@ -116,7 +116,8 @@ func FillIndex(fetcher RunFetcher, logger shared.Logger, rt monitor.Runtime, int
 }
 
 func startBackfillMonitor(fetcher RunFetcher, logger shared.Logger, maxBytes uint64, m *backfillMonitor) error {
-	runsByProduct, err := fetcher.FetchRuns(int(maxBytes/bytesPerRun) / 4)
+	// FetchRuns will return at most N runs for each product, so divide the upper bound by the number of products.
+	runsByProduct, err := fetcher.FetchRuns(int(maxBytes/bytesPerRun) / len(shared.GetDefaultProducts()))
 	if err != nil {
 		return err
 	}
--- a/api/query/cache/backfill/backfill.go
+++ b/api/query/cache/backfill/backfill.go
@@ -21,7 +21,7 @@ import (
 // RunFetcher provides an interface for loading a limited number of runs
 // suitable for backfilling an index.
 type RunFetcher interface {
-	FetchRuns(limit int) ([]shared.TestRun, error)
+	FetchRuns(limit int) (shared.TestRunsByProduct, error)
 }
 
 type datastoreRunFetcher struct {
@@ -55,7 +55,7 @@ func NewDatastoreRunFetcher(projectID string, gcpCredentialsFile *string, logger
 	return datastoreRunFetcher{projectID, gcpCredentialsFile, logger}
 }
 
-func (f datastoreRunFetcher) FetchRuns(limit int) ([]shared.TestRun, error) {
+func (f datastoreRunFetcher) FetchRuns(limit int) (shared.TestRunsByProduct, error) {
 	ctx := context.Background()
 	var client *datastore.Client
 	var err error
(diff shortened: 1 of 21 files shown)


shared/datastore.go
```diff
@@ -12,9 +16,40 @@ import (
...
+	Distinct() Query
+	Run(Datastore) Iterator
+}
+
+// Datastore abstracts a datastore, hiding the distinctions between cloud and
+// appengine's datastores.
+type Datastore interface {
+	Context() context.Context
+	NewQuery(typeName string) Query
+	GetAll(q Query, dst interface{}) ([]Key, error)
```
lukebjerring: @mdittmer @Hexcles this is where it will be appropriate to follow up by adding methods for fetching runs, and making a mock implementation, to replace the places where we've done that ad-hoc, e.g. 
- https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/query.go#L26
- https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/cache/backfill/backfill.go#L24

shared/datastore.go
```diff
@@ -12,9 +16,40 @@ import (
...
+	Order(order string) Query
+	KeysOnly() Query
+	Distinct() Query
+	Run(Datastore) Iterator
+}
+
+// Datastore abstracts a datastore, hiding the distinctions between cloud and
+// appengine's datastores.
+type Datastore interface {
+	Context() context.Context
```
Hexcles: Regarding Context, if we want to include it in the abstracted API, then we should make it clear that this interface should not be reused and should be constructed in each request.

shared/datastore.go
```diff
@@ -12,9 +16,40 @@ import (
...
+// Query abstracts a datastore.Query
+type Query interface {
+	Filter(filterStr string, value interface{}) Query
+	Project(project string) Query
+	Limit(limit int) Query
+	Offset(offset int) Query
+	Order(order string) Query
+	KeysOnly() Query
+	Distinct() Query
+	Run(Datastore) Iterator
```
Hexcles: Shall we also add a `Run` function on `Datastore` for completeness? (The cloud Datastore package has its `Run` function on `Client` not `Query`.)

shared/datastore_cloud.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return cast, err
+}
+
+type cloudQuery struct {
+	query *datastore.Query
+}
+
+func (q cloudQuery) Filter(filterStr string, value interface{}) Query {
+	return cloudQuery{query: q.query.Filter(filterStr, value)}
```
Hexcles: Nit: specify (or don't specify) the only field name consistently in this file (and `datastore_ae.go`).

api/query/cache/backfill/backfill.go
```diff
@@ -126,31 +116,45 @@ func FillIndex(fetcher RunFetcher, logger shared.Logger, rt monitor.Runtime, int
 }
 
 func startBackfillMonitor(fetcher RunFetcher, logger shared.Logger, maxBytes uint64, m *backfillMonitor) error {
-	runs, err := fetcher.FetchRuns(int(maxBytes / bytesPerRun))
+	runsByProduct, err := fetcher.FetchRuns(int(maxBytes/bytesPerRun) / 4)
```
Hexcles: Too much magic here. Explain a bit in the comment.

Also, what do we do when Edge is gone?

shared/datastore.go
lukebjerring: Well, for the cloud case, it's independent of requests, so I'm not sure it matters (and a context object is created per-request for the `NewAppEngineDatastore` param in any case).

shared/datastore.go
lukebjerring: I originally did, but, no; I'd rather be consistent with all of our datastore code so that copy-paste for similar queries will work, with the only difference being the way the datastore is created.

shared/datastore_cloud.go
lukebjerring: Done.

api/query/cache/backfill/backfill.go
lukebjerring: Made it less magic (`len(shared.GetDefaultProducts())`, and commented.

lukebjerring:
I imagine concerns will be minor and we can retroactively address any - I've got some follow-up consolidation of the mocks to do.

PR #967, opened 2018-12-21
Fix 500s for HEAD rev

lukebjerring:
## Description
We sometimes use "HEAD" for the before in the diff, but are trying to log a 7 char slice of that.


lukebjerring:
TBR'ing since this fixes the 500s.

PR #970, opened 2018-12-22
Fix the length of all action fields

lukebjerring:
## Description
The actions were tweaked in #949 and totally broke things in a not-obvious way. This adds a test that prevents future changes and brings the descriptions back to accepted lengths.


lukebjerring:
TBR'ing this to unblock checks.

PR #972, opened 2018-12-22
Allow shorter shas in /api/checks/[SHA]

lukebjerring:
## Description
Right now the `/api/shas` endpoint serves 10-char SHAs, which makes piping those into /api/checks/[SHA:40] a problem.

I'll change the SHAs endpoint to take a `length` param later, but this makes life easier now.


lukebjerring:
TBR'ing this (have verified it in staging, and it's trivial).

PR #975, opened 2018-12-22
Fix nil-pointer bug when fetching existing runs

lukebjerring:


lukebjerring:
Note that this will very rarely be nil; only if something goes wrong with the request.

Hexcles:
@lukebjerring is this ready for review? (Asking because you haven't requested)

Issue #973, opened 2018-12-22
PR links in checks summaries don't work


lukebjerring:
### Description
/results loads too many runs:
https://staging.wpt.fyi/results/?pr=14655

/runs ignores the ?pr= param entirely and shows the default.

lukebjerring:
I've fixed this temporarily, by merging #976 and #977 and deploying to version `checks`, which is currently serving 100% of traffic for staging.wpt.fyi. 
We'll land the fixes properly after the holidays :)

lukebjerring:
Now fixed on master.

lukebjerring:
The links for `/runs?pr=1234` still don't work

PR #786, opened 2018-11-21
Add plumbing for pr param in UI

lukebjerring:
## Description
Extension of #785 that includes the UI plumbing needed.

## Review Information
Only need to review the last 2 commits.

Find a PR that has loaded results into wpt.fyi, and load ?pr=[PR number]
e.g. [/results/?pr=14143](https://prs-ui-dot-wptdashboard-staging.appspot.com/results/?pr=14143)

PR #976, opened 2018-12-22
Plumb the pr param in /runs view

lukebjerring:
## Description
Part of #973 

Fixes the behaviour of the pr param in /runs, by including parsing + plumbing to the UI (the rest of the code is already present from #786

### DIFF
--- a/webapp/test_runs_handler.go
+++ b/webapp/test_runs_handler.go
@@ -17,36 +17,49 @@ func testRunsHandler(w http.ResponseWriter, r *http.Request) {
 		http.Error(w, "Only GET is supported.", http.StatusMethodNotAllowed)
 		return
 	}
-	ctx := shared.NewAppEngineContext(r)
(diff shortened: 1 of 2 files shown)


webapp/test_runs_handler.go
```diff
@@ -17,36 +17,49 @@ func testRunsHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if err := templates.ExecuteTemplate(w, "test-runs.html", data); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
-		return
+	if pr != nil {
+		filter.PR = pr
+	} else {
+		if testRunFilter.IsDefaultQuery() {
+			// Get runs from a month ago, onward, by default.
```
foolip: This comment doesn't match the code, it seems.

lukebjerring: Done.

PR #977, opened 2018-12-22
Fix pr param for /results

lukebjerring:
## Description
Accompanies #976 for a resolution to #973 

Fixes the behaviour of the /results PR param.
It is fixed by having the server alias `pr` as a request for multiple SHAs, which we ask GitHub for.
So, to facilitate that, this PR refactors the behaviour of our datastore to accept multiple SHAs when loading runs.

This has side effects of:
- Allowing multiple `sha` params in queries. This can be useful for using `/runs` to see all runs across some known shas (a PR, of course, being a great example).
- Allowing other filtering params to be combined with a `pr` filter; e.g. just safari results for the PR.

Example PR with multiple commits: [14655](https://results-pr-dot-wptdashboard-staging.appspot.com/results?pr=14655)

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -59,7 +59,7 @@ func LoadTestRunKeys(
 	ctx context.Context,
 	products []ProductSpec,
 	labels mapset.Set,
-	sha string,
+	revisions []string,
 	from *time.Time,
 	to *time.Time,
 	limit *int,
@@ -76,14 +76,16 @@ func LoadTestRunKeys(
 		}
 	}
 	var globalKeyFilter mapset.Set
-	if !IsLatest(sha) {
-		var ids TestRunIDs
-		if ids, err = loadKeysForRevision(ctx, baseQuery, sha); err != nil {
-			return nil, err
-		}
-		globalKeyFilter = mapset.NewSet()
-		for _, id := range ids {
-			globalKeyFilter.Add(id)
+	if len(revisions) > 1 || len(revisions) == 1 && !IsLatest(revisions[0]) {
+		for _, sha := range revisions {
+			var ids TestRunIDs
+			if ids, err = loadKeysForRevision(ctx, baseQuery, sha); err != nil {
+				return nil, err
+			}
+			globalKeyFilter = mapset.NewSet()
+			for _, id := range ids {
+				globalKeyFilter.Add(id)
+			}
 		}
 	}
 	for i, product := range products {
@@ -163,12 +165,12 @@ func LoadTestRuns(
 	ctx context.Context,
 	products []ProductSpec,
 	labels mapset.Set,
-	sha string,
+	revisions []string,
 	from *time.Time,
 	to *time.Time,
 	limit,
 	offset *int) (result TestRunsByProduct, err error) {
-	keys, err := LoadTestRunKeys(ctx, products, labels, sha, from, to, limit, offset)
+	keys, err := LoadTestRunKeys(ctx, products, labels, revisions, from, to, limit, offset)
 	if err != nil {
 		return nil, err
 	}
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -25,33 +25,6 @@ func LoadTestRun(ctx context.Context, id int64) (*TestRun, error) {
 	return &testRun, nil
 }
 
-// LoadTestRunsBySHAs loads all test runs that belong to any of the given revisions (SHAs).
-func LoadTestRunsBySHAs(ctx context.Context, shas ...string) (runs TestRuns, err error) {
-	for _, sha := range shas {
-		if len(sha) > 10 {
-			sha = sha[:10]
-		}
-		q := datastore.NewQuery("TestRun")
-		ids, err := loadKeysForRevision(ctx, q, sha)
-		if err != nil {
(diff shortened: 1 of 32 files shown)


shared/datastore.go
```diff
@@ -76,14 +76,16 @@ func LoadTestRunKeys(
...
-		globalKeyFilter = mapset.NewSet()
-		for _, id := range ids {
-			globalKeyFilter.Add(id)
+	if len(revisions) > 1 || len(revisions) == 1 && !IsLatest(revisions[0]) {
+		for _, sha := range revisions {
+			var ids TestRunIDs
+			if ids, err = loadKeysForRevision(ctx, baseQuery, sha); err != nil {
+				return nil, err
+			}
+			globalKeyFilter = mapset.NewSet()
```
foolip: Creating a new set in each iteration of the loop seems odd, if it's correct can you add a comment about why? Or not if I'm just being daft. If it is wrong, can a test be written for it?

shared/params.go
```diff
@@ -143,14 +143,17 @@ const MaxCountMinValue = 1
...
 
-// ParseSHAParam parses and validates the 'sha' param for the request.
-// If the param is not present, it returns "latest" by default (and in error cases).
-func ParseSHAParam(v url.Values) (runSHA string, err error) {
-	sha, err := ParseSHA(v.Get("sha"))
-	if err != nil || !SHARegex.MatchString(sha) {
-		return sha, err
+// ParseSHAParam parses and validates any 'sha' param(s) for the request.
+func ParseSHAParam(v url.Values) (SHAs, error) {
+	shas := ParseRepeatedParam(v, "sha", "shas")
```
foolip: This function reads very well, nice parsing helpers :)

shared/test_run_filter.go
```diff
@@ -15,10 +15,27 @@ import (
...
+type SHAs []string
+
+// EmptyOrLatest returns whether the shas slice is empty, or only contains
+// one item, which is the latest keyword.
+func (s SHAs) EmptyOrLatest() bool {
+	return len(s) < 1 || len(s) == 1 && IsLatest(s[0])
+}
+
+// FirstOrLatest returns the first sha in the slice, or the latest keyword.
+func (s SHAs) FirstOrLatest() string {
```
foolip: FWIW, at the call sites I first found this name odd, like "first or last", but I don't know how to improve it without changing what the keyword is.

webapp/components/test-run.html
```diff
@@ -159,7 +159,9 @@
       }
 
       sevenCharSHA(sha) {
-        return sha.substr(0, 7);
+        return sha.length > 7
```
foolip: Doesn't this do the same as before, or what does substr do when asked to make a longer string?

webapp/components/test-runs-query-builder.html
```diff
@@ -239,25 +243,29 @@ <h3>
...
       // Respond to newly fetched shas, or user input, by filtering the autocomplete list.
-      shasUpdated(sha, shas) {
-        if (!shas || !shas.length || !this.queryBuilderSHA) {
+      shasUpdated(sha, matchingSHAs) {
+        if (!matchingSHAs || !matchingSHAs.length || !this.queryBuilderSHA) {
           return;
         }
         if (sha) {
-          shas = shas.filter(s => s.startsWith(sha));
+          matchingSHAs = matchingSHAs.filter(s => s.startsWith(sha));
```
foolip: Are matching shad guaranteed to be 40-char shas? If not the reverse condition seems like it's also needed. If so, maybe a helper to compare shas?

webapp/components/test-runs-query.html
```diff
@@ -74,6 +74,12 @@
           this.updateQueryParams(this.queryParams);
         }
 
+        // sha is a convenience method for getting (the) single sha.
```
foolip: I guess all the call sites for this are potentially buggy with PRs, or is that not a problem?

shared/datastore.go
lukebjerring: No, you're right - that's my bad. Left over from the original single-loop code. Added a unit test.

webapp/components/test-run.html
lukebjerring: Reverted.

webapp/components/test-runs-query-builder.html
lukebjerring: No; they're 10-char SHAs at the moment. This function is basically filtering a complete list of 10-char shas into a filtered list based on what you've typed (prefix matching). I'll be cleaning up the /api/shas endpoint to accept a length param at some point, and we can revise this then; however, nobody would type the 40 char sha. They'd paste it, and having the value not match autocomplete won't be an issue.

webapp/components/test-runs-query.html
lukebjerring: Yes, but since we're expecting one SHA in most cases, it's better to have this as a stop-gap for those call sites I've missed (if there are any) that aren't covered by unit tests, than to risk non-pr case breakages.

foolip: Makes sense. Should the call sites be converted and this eventually removed, or is it fine if it stays?

PR #979, opened 2019-01-03
Add a feature for including a service worker to cache components

lukebjerring:
## Description
Adds a dynamic endpoint for `/service-worker.js` that, for valid releases (versions that are a 7-char SHA), crawls the /static/ and /components/ dirs in webapp and preemptively caches all of those requests.

Also caches /bower_components/ requests, but waits for a caller to have requested them (so that we can filter down to files that are actually in use. The directory tree is huge).

Also adds a 1d public-cache duration for the /static/ dir (was 10m).

### DIFF
--- a/webapp/service-worker-installer.js
+++ b/webapp/service-worker-installer.js
@@ -1,0 +1,18 @@
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker
+    .register('/service-worker.js', {
+      scope: '/'
+    })
+    .then(function (reg) {
+      if (reg.installing) {
+        console.log('Service worker installing');
+      } else if (reg.waiting) {
+        console.log('Service worker installed');
+      } else if (reg.active) {
+        console.log('Service worker active');
+      }
+    }).catch(function (error) {
+      // registration failed
+      console.log(`Registration failed with ${error}`);
+    });
+}
--- a/webapp/service_worker_handler.go
+++ b/webapp/service_worker_handler.go
@@ -6,12 +6,8 @@ package webapp
 
 import (
 	"fmt"
-	"io/ioutil"
 	"net/http"
-	"path"
-	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 	"text/template"
 
@@ -41,35 +37,10 @@ func serviceWorkerHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	var files []string
-	_, filename, _, _ := runtime.Caller(0)
-	dir := filepath.Dir(filename)
-	for _, folder := range []string{"components", "static"} {
-		appendFiles(path.Join(dir, folder), &files)
-	}
-
 	data := struct {
 		Version string
-		Files   []string
 	}{
 		Version: version,
-		Files:   files,
 	}
 	swTemplate.Execute(w, data)
-}
-
-func appendFiles(dirPath string, files *[]string) error {
-	dir, err := ioutil.ReadDir(dirPath)
-	if err != nil {
-		return err
-	}
-	for _, f := range dir {
-		name := path.Join(dirPath, f.Name())
-		if f.IsDir() {
-			appendFiles(name, files)
-		} else {
-			*files = append(*files, name)
-		}
-	}
-	return nil
 }
--- a/webapp/service_worker_handler.go
+++ b/webapp/service_worker_handler.go
@@ -6,8 +6,8 @@ package webapp
 
 import (
 	"fmt"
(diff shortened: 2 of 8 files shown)


Hexcles:
https://aaaaaaa-dot-wptdashboard-staging.appspot.com/service-worker.js returns 404?

webapp/service-worker-installer.js
```diff
@@ -0,0 +1,18 @@
+if ('serviceWorker' in navigator) {
```
Hexcles: Looks like this file should be in `static/`?

webapp/templates/_head_common.html
```diff
@@ -1,4 +1,5 @@
 <title>web-platform-tests dashboard</title>
 <link rel="stylesheet" href="/static/common.css">
+<script src="/service-worker-installer.js"></script>
```
Hexcles: And correspondingly, this needs to be `/static/service-worker-installer.js`.

Hexcles:
NVM, there's already a route config for `service-worker-installer.js` and the installer script works fine. Chrome is complaining about `service-worker.js`:
> Registration failed with TypeError: Failed to register a ServiceWorker: A bad HTTP response code (404) was received when fetching the script.

webapp/service_worker_handler.go
```diff
@@ -0,0 +1,67 @@
...
+	swTemplate = template.Must(template.ParseFiles("templates/service-worker.js"))
+	// NOTE(lukebjerring): If tweaking service worker locally, change to
+	// sevenCharSHA, _ = regexp.Compile("^[0-9a-f]{7}|None$")
+	sevenCharSHA, _ = regexp.Compile("^[0-9a-f]{7}$")
+)
+
+func serviceWorkerHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	aeAPI := shared.NewAppEngineAPI(ctx)
+	if !aeAPI.IsFeatureEnabled("serviceWorker") {
```
Hexcles: Ahh I see; it's behind a flag. I went ahead to create this flag in Datastore (on staging) and set it to true.

webapp/service_worker_handler.go
```diff
@@ -0,0 +1,67 @@
...
+	if !sevenCharSHA.MatchString(version) {
+		http.Error(w, fmt.Sprintf("Service worker not implemented for version '%s'", version), http.StatusNotImplemented)
+		return
+	}
+
+	var files []string
+	for _, folder := range []string{"components", "static"} {
+		_, filename, _, _ := runtime.Caller(0)
+		dir := filepath.Dir(filename)
+		err := filepath.Walk(path.Join(dir, folder), func(path string, info os.FileInfo, err error) error {
```
Hexcles: I'm now getting am empty 200 response from https://aaaaaaa-dot-wptdashboard-staging.appspot.com/service-worker.js, so I'm suspecting a non-nil error here. Can we actually do `filepath.Walk` on AppEngine standard at all? Given that direct network access isn't available, I won't be surprised if there's no real filesystem on AppEngine standard.

webapp/templates/service-worker.js
```diff
@@ -0,0 +1,25 @@
...
+  'fetch',
+  e => {
+    return e.respondWith(
+      caches.match(e.request)
+        .then(r => {
+          if (r) {
+            return r;
+          }
+          return fetch(e.request)
+            .then(r => {
```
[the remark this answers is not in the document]
lukebjerring: Done.

webapp/templates/service-worker.js
```diff
@@ -0,0 +1,25 @@
...
+              const components = new RegExp('^/((bower_)?components|static)/');
+              if (components.test(e.request.url.toString())) {
+                let clone = r.clone();
+                caches.open('{{ .Version }}').then(cache => {
```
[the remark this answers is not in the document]
lukebjerring: Done.

webapp/templates/service-worker.js
```diff
@@ -0,0 +1,43 @@
...
+    return e.respondWith(
+      caches.match(e.request)
+        .then(r => {
+          if (r) {
+            return r;
+          }
+          return fetch(e.request)
+            .then(r => {
+              if (r.ok) {
+                const components = new RegExp('^/((bower_)?components|static)/');
```
Hexcles: I suppose with #982 we'll need to cache `node_modules` as well. Shall we do it in this PR?

lukebjerring: No, I'll do it in #982.

PR #980, opened 2019-01-04
Consolidate history query params

lukebjerring:
## Description
Changes the behaviour of the history components to correctly re-use the query parameters.
In turn, this ensures that we're fetching runs that match the query for the main results above the history section.

### DIFF
--- a/webapp/components/test-results-chart.html
+++ b/webapp/components/test-results-chart.html
@@ -150,7 +150,7 @@ found in the LICENSE file.
       }
 
       // eslint-disable-next-line no-unused-vars
-      loadResults(tests, query) {
+      async loadResults(tests, query) {
         if (!this.show || !this.tests || !this.tests.length) {
           return;
         }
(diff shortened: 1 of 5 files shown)


lukebjerring:
Note that https://github.com/web-platform-tests/wpt.fyi/pull/980/commits/844b632ecdd6d2247e0f703630ae47815c4d6996 won't work for any query with labels, until #981 lands, but that's an independent bug in either case.

webapp/components/test-results-chart.html
```diff
@@ -157,41 +144,33 @@
...
-      // from prior to `to` iff `show`.
-      async loadNext(tests, chunkSize, maxRuns, labels, to, show) {
-        if (!show || tests.length === 0 || this.values.length >= maxRuns) {
+      computeTestRunQueryParams(shas, aligned, master, labels, productSpecs, to, from, maxCount) {
+        maxCount = this.chunkSize;
+        return super.computeTestRunQueryParams(shas, aligned, master, labels, productSpecs, to, from, maxCount);
+      }
+
+      // eslint-disable-next-line no-unused-vars
+      loadResults(tests, query) {
```
Hexcles: Any benefits to stop using async/await here and`loadNext`? `await` looks clearer to me.

lukebjerring: More obvious that it's returning a chainable promise, but could be swayed to keep await and leave understanding as an exercise for the reader.

Hexcles: I feel the opposite: without type annotation (of the return value), `async` makes it more obvious that the function returns a promise.

lukebjerring: Reverted.

webapp/components/test-results-history-grid.html
```diff
@@ -97,35 +96,34 @@
         };
       }
 
-      async onPathChanged(path) {
+      // eslint-disable-next-line no-unused-vars
+      loadResults(path, query) {
```
Hexcles: Perhaps this needs to be `async` too for consistency? Or at least the early return when `!path` should be a thenable?

webapp/components/test-runs.html
```diff
@@ -75,30 +75,31 @@
...
-          const fetches = await Promise.all(runs);
-
-          // Filter unresolved fetches and flatten any array-fetches into the array.
-          const nonEmpty = fetches.filter(e => e);
-          const flattened = nonEmpty.reduce((sum, item) => {
-            return sum.concat(Array.isArray(item) ? item : [item]);
-          }, []);
-          this.testRuns = flattened;
-          return flattened;
+          return Promise.all(runs).then(fetches => {
```
Hexcles: Same here. This change seems to be just converting an async function to a non-async one. If so, let's revert the change in this function as well to keep this PR simple (and for my preference of `async`).

PR #981, opened 2019-01-04
Add TestRunFilter custom marshalling

lukebjerring:
## Description
Fixes pagination when a product has a label (or, a label is used in general).

### DIFF
--- a/shared/product_spec.go
+++ b/shared/product_spec.go
@@ -117,12 +117,6 @@ func (p ProductSpecs) Len() int           { return len(p) }
 func (p ProductSpecs) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
 func (p ProductSpecs) Less(i, j int) bool { return p[i].String() < p[j].String() }
 
-type productSpecNoCustomMarshalling ProductSpec
-type marshallableProductSpec struct {
-	productSpecNoCustomMarshalling
-	Labels []string `json:"labels,omitempty"`
-}
-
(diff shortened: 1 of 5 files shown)


shared/product_spec.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return s
+}
+
+func (p ProductSpecs) Len() int           { return len(p) }
+func (p ProductSpecs) Swap(i, j int)      { p[i], p[j] = p[j], p[i] }
+func (p ProductSpecs) Less(i, j int) bool { return p[i].String() < p[j].String() }
+
+type productSpecNoCustomMarshalling ProductSpec
+type marshallableProductSpec struct {
```
Hexcles: What are these two types for? They don't seem to be referenced anywhere.

shared/util.go
```diff
@@ -90,6 +90,18 @@ func ToStringSlice(set mapset.Set) []string {
 	return result
 }
 
+// ToStringSet converts a typed slice to a set.
+func ToStringSet(slice []string) mapset.Set {
```
Hexcles: This seems the same as `NewSetFromStringSlice`.

shared/product_spec.go
lukebjerring: Remnants of making the same change to product spec, only to decide that stringifying and parsing was a better call. Removed.

shared/util.go
lukebjerring: Whoops. Looks like my naming choices aren't deterministic.

PR #982, opened 2019-01-04
Migrate to Polymer 3

lukebjerring:
## Description
Migrating the project from the deprecated HTML imports (Polymer 2) to ES6 imports (Polymer 3).

This PR is intended to document everything done to achieve a working migration.
So far:
- Follow basic steps at https://polymer-library.polymer-project.org/3.0/docs/upgrade
  - `cd webapp`
  - `npm install --user polymer-cli`
  - `npm install --user polymer-modulizer`
  - `modulizer --out .`
- Replace incorrect `../../@polymer` paths with `../node_modules/@polymer`
- Delete the generated FIXMEs
- Tweak the copyright comments
- Switch the app.yaml static dir for bower_components to node_modules


lukebjerring:
We're now in a state where our components imports seem to be working, but the @polymer imports are relative, so we need to run `polymer build` and change our static dirs, or run `polymer serve` as part of the `web_server.sh` script.

lukebjerring:
I added a Makefile rule for making all relative imports absolute, but this will clash with genuine use of node (e.g. `wct`), so we'll need to be more clever.

lukebjerring:
In https://github.com/web-platform-tests/wpt.fyi/pull/982/commits/aef040bdb5d3ed8b795f62578f18d03ac0e7860c, I've change the path resolution to occur when we serve the file, with a regex, in a custom handler for /node_modules/.

The cost of doing this will be drastically mitigated when we land https://github.com/web-platform-tests/wpt.fyi/pull/979 (service-worker for caching the components).

lukebjerring:
@Hexcles this is now ready for review :) tests are green... so everything we test should be working at least!

webapp/app.yaml
```diff
@@ -11,15 +11,12 @@ builtins:
 
 handlers:
   # Couple of special-case dynamic components.
-- url: /components/wpt-env-flags.html
+- url: /components/wpt-env-flags.js
```
Hexcles: We can simply remove this entry and fall back to the default.

webapp/app.yaml
```diff
@@ -33,18 +30,7 @@ handlers:
...
-- bower_components/chai/
-- bower_components/lodash/
-- bower_components/mocha/
-- bower_components/sinon-chai/
-- bower_components/sinonjs/
-- bower_components/stacky/
-- bower_components/test-fixture/
-- bower_components/web-component-tester/
 - components/test/
-- node_modules/
```
Hexcles: I'm a bit concerned about this. I understand we now need `node_modules` to include all the dependent components, but `node_modules` can contain a lot of stale dependencies & dev-only dependencies on our local machines.

I feel like we might want a new make rule to build release assets after all. And if so, we can use `polymer-build` instead of using the server-side rewriter. WDYT?

Hexcles: Alternatively, we can use [`npm prune --production`](https://docs.npmjs.com/cli/prune.html) before deploying.

webapp/components_handler.go
```diff
@@ -0,0 +1,38 @@
...
+	"io/ioutil"
+	"mime"
+	"net/http"
+	"path/filepath"
+	"regexp"
+
+	"github.com/gorilla/mux"
+)
+
+var packageRegex = regexp.MustCompile(`(import|from) (['"])(@[^/]*/)`)
```
Hexcles: Since this is a dumb string replacer (instead of an AST walker), I'd rather having a stricter regex, perhaps only matching `import ... from @...` isnstead of accepting either `import` or `from` to minimize the risk of overwriting something else.

webapp/bower.json
```diff
@@ -1,29 +1,6 @@
 {
```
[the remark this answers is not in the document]
lukebjerring: No, unfortunately `wct` relies on it.

webapp/app.yaml
lukebjerring: Not quite; it's a template-generated compontent (dynamic). Needs to be handled server-side.

webapp/app.yaml
lukebjerring: So, polymer build is very specifically designed for an app that has a shell/entrypoint, and expects a range of things we don't do. `npm prune --production` on deploy sounds good; I'll use that for web_server.sh too.

webapp/app.yaml
Hexcles: Oh sorry I didn't see the next entry (for `/components`). nvm then.

webapp/app.yaml
```diff
@@ -38,18 +35,7 @@ handlers:
 
 # Don't upload mock-data static files to AppEngine
 skip_files:
```
Hexcles:
```suggestion
skip_files:
- bower_components/
```

If the only thing using bower now is `wct`, we can and should skip the whole `bower_components` directory when deploying.

util/docker-dev/web_server.sh
```diff
@@ -27,7 +27,10 @@ while getopts ':rh' flag; do
   esac
 done
 
-info "Installing web server code dependencies"
+info "Pruning node_modules so dev_appserver can handle watching file updates..."
```
Hexcles: Is it because there are too many files and dev_appserver would bail otherwise?

webapp/templates/service-worker.js
```diff
@@ -26,8 +26,10 @@ self.addEventListener(
           return fetch(e.request)
             .then(r => {
               if (r.ok) {
-                const components = new RegExp('^/((bower_)?components|static)/');
-                if (components.test(e.request.url.toString())) {
+                const components = new RegExp('^/((bower_)?components|static|node_modules)/');
+                const wptEnvFlags = new RegExp('^/components/wpt-env-flags.js$');
```
Hexcles: Can this be a simple string comparison instead of a regex?

lukebjerring: Done.

PR #984, opened 2019-01-07
Distinguish release version and promote

lukebjerring:
## Description
Changes the staging pushes to deploy and promote a versioned release, rather than clobbering the `master` version (which happens to serve all the traffic).

No tangible change to our release flow, except that staging.wpt.fyi's version will we a 7-char SHA, same as prod, which allows us to leverage a service-worker without risk of stale cache (see #979)

### DIFF
--- a/util/cleanup-versions.sh
+++ b/util/cleanup-versions.sh
@@ -20,9 +20,10 @@ function cleanup() {
   info "Cleaning stale versions of $1..."
 
   local SERVICE_ARG="-s $1"
+  local FILTER_ARG="traffic_split=0.0 last_deployed_time.datetime<$CUTOFF"
   local versions_to_delete=()
 
-  for version in $( gcloud app versions list $PROJECT_ARG $SERVICE_ARG --filter="last_deployed_time.datetime<$CUTOFF" --format="value(id)" ); do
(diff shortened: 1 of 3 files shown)


util/cleanup-versions.sh
```diff
@@ -20,9 +20,10 @@ function cleanup() {
   info "Cleaning stale versions of $1..."
 
   local SERVICE_ARG="-s $1"
+  local FILTER_ARG="traffic_split=0.0 last_deployed_time.datetime<$CUTOFF"
   local versions_to_delete=()
 
-  for version in $( gcloud app versions list $PROJECT_ARG $SERVICE_ARG --filter="last_deployed_time.datetime<$CUTOFF" --format="value(id)" ); do
+  for version in $( gcloud app versions list $PROJECT_ARG $SERVICE_ARG --filter="$FILTER_ARG" --format="value(id)" ); do
```
Hexcles: @lukebjerring FYI

Hexcles:
I've merged this PR. It remains to be seen if we'll run into severe race conditions by trying to promote multiple versions together. If anything goes wrong, I'll revert it.

lukebjerring:
Theory: I think it'd just fail in the same way that 2 requests attempting to update `master` currently fail (with error / operation already in progress).

PR #988, opened 2019-01-07
fixup! [processor] Only upload results when they exist

Hexcles:
Damn my butter fingers... Urgent fix. TBR


lukebjerring:
This LGTM but I feel like Travis should pack a sad for a syntax error; no tests for that file?

Hexcles:
It's sadly valid syntax... The trailing comma makes this a tuple with only one element.

This module does warrant some tests. It started out simple, and with so much network I/O and syscalls I didn't bother, but now at this point it does deserve some mocks.

PR #989, opened 2019-01-08
Preserve subtest order better

### LINKED ISSUE #965: Preserve order better when merging subtests across runs
In https://staging.wpt.fyi/results/css/css-typed-om/interfaces.html?product=chrome%5Bexperimental%2Ctaskcluster%5D%401bdc13d850f73a419ae2a4266710fdbbf2002f29&product=chrome%407c6c6afac5e1f9ed9cae1159c3f8a58ad27ff9b8&diff, some subtests have been removed and some added, at the same place in the overall order of subtests.

The added tests, which are only in the second run, appear at the very bottom. This is because the current algorithm is to take the union of tests simply by appending the tests that aren't already in the list when processing the 2nd, 3rd, etc. run.

Inserting tests in the same place as they were in the original run would lead to renamed tests appearing next to each other, and would make it easier to understand what has changed.

The algorithm for this probably has some similarities with a diff algorithm, but something pretty simple would probably do the trick.

lukebjerring:
## Description
Changes subtest appending to append after the last match. Fixes #965


foolip:
In https://subtest-merge-dot-wptdashboard-staging.appspot.com/results/css/css-typed-om/interfaces.html?product=chrome%5Bexperimental%2Ctaskcluster%5D%401bdc13d850f73a419ae2a4266710fdbbf2002f29&product=chrome%407c6c6afac5e1f9ed9cae1159c3f8a58ad27ff9b8&diff=true there are two tests at the very end which ought not have been at the end of either reports. Can you verify that this case is working as intended?

lukebjerring:
Fixed in b7f76a2, PTAL

Issue #329, opened 2018-06-28
Add an API to list all labels


Hexcles:
@kereliuk asked if there's a way to list all labels. Sounds like a useful feature to me for discovering data.

lukebjerring:
@kereliuk is going to tackle this as a good introductory PR.

PR #990, opened 2019-01-08
Add /api/labels endpoint

lukebjerring:
## Description
Resolves https://github.com/web-platform-tests/wpt.fyi/issues/329

Adds a endpoint /api/labels that lists all of the labels used on any runs.

Loads `TestRun` entities with all distinct `Labels`, aggregates them into a set, spits out a sorted array, caches the whole request for 5 min.

### DIFF
--- a/api/labels.go
+++ b/api/labels.go
@@ -20,7 +20,7 @@ type LabelsHandler struct {
 	ctx context.Context
 }
 
-// apiLabelsHandler is responsible for emitting just the revision SHAs for test runs.
+// apiLabelsHandler is responsible for emitting just all labels used for test runs.
 func apiLabelsHandler(w http.ResponseWriter, r *http.Request) {
 	// Serve cached with 5 minute expiry. Delegate to LabelsHandler on cache miss.
 	ctx := shared.NewAppEngineContext(r)
--- a/api/labels.go
+++ b/api/labels.go
@@ -1,0 +1,53 @@
+// Copyright 2019 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"sort"
+	"time"
+
+	mapset "github.com/deckarep/golang-set"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// LabelsHandler is an http.Handler for the /api/labels endpoint.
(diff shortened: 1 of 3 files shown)


api/labels.go
```diff
@@ -0,0 +1,53 @@
...
+	mapset "github.com/deckarep/golang-set"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// LabelsHandler is an http.Handler for the /api/labels endpoint.
+type LabelsHandler struct {
+	ctx context.Context
+}
+
+// apiLabelsHandler is responsible for emitting just the revision SHAs for test runs.
```
foolip: Is this comment copypasta? It emits labels, I hope :)

api/labels.go
```diff
@@ -0,0 +1,53 @@
...
+	store := shared.NewAppEngineDatastore(h.ctx)
+	var runs shared.TestRuns
+	_, err := store.GetAll(store.NewQuery("TestRun").Project("Labels").Distinct(), &runs)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	all := mapset.NewSet()
+	for _, run := range runs {
```
foolip: Is this really able to iterate all runs, won't this run into some limit that requires pagination just like /api/runs?

api/labels_medium_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"google.golang.org/appengine/datastore"
+)
+
+func TestLabelsHandler(t *testing.T) {
+	ctx, done, err := sharedtest.NewAEContext(true)
+	assert.Nil(t, err)
+	defer done()
+
+	key := datastore.NewIncompleteKey(ctx, "TestRun", nil)
+	_, err = datastore.Put(ctx, key, &shared.TestRun{Labels: []string{"a"}})
```
foolip: Can you add the labels out of order to also test the sorting, and perhaps also a duplicate?

api/labels.go
lukebjerring: Sure was.

api/labels.go
lukebjerring: My theory (seems to work in staging, which has more labels than prod) is that the `.Project("Labels").Distinct()` part means that we get less than 500 distinct results, since only distinct sets of labels are included in the result.

api/labels_medium_test.go
lukebjerring: Done.

PR #994, opened 2019-01-09
Add master runs in dev_data

lukebjerring:
## Description
Tweaks the `make dev_data` script to grab `master` runs, and de-duplicate (ignore) the runs that it's already seen for other label queries.

### DIFF
--- a/util/populate_dev_data.go
+++ b/util/populate_dev_data.go
@@ -21,9 +21,10 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


lukebjerring:
TBR'ing this trivial dev change.

Hexcles:
What's up with `package-lock.json`? Do you have a different (perhaps older) version of `npm` locally? All checksums were switched from SHA256 to SHA1 in this PR. I'd like to avoid switching back and forth every time we touch `webapp`.

PR #996, opened 2019-01-09
Fix google-chart package

lukebjerring:
## Description
Changes to use the (correct + working) `@google-web-components` package.
https://www.webcomponents.org/element/@google-web-components/google-chart


lukebjerring:
TBR'ing this bugfix

PR #998, opened 2019-01-09
Create prod wpt.fyi checks app ID

lukebjerring:
## Description
Adds the ID for the production wpt.fyi check to the known app IDs.


lukebjerring:
My intention is to create a fork of wpt that follows the repo's updates, and install staging there.
Production will be applied to wpt itself.

PR #999, opened 2019-01-09
Restore wct-legacy-browser dependency for devDependencies

lukebjerring:
Also, change the installer to a more modern FF.


Hexcles:
`TestLabelParam_Results` still fails.

lukebjerring:
Eh, merging anyway. Can't repro the flake and this isn't making it worse.

PR #1000, opened 2019-01-09
Search cache: Force monitor to run every N IngestRun() invocations

mdittmer:
This change uses a channel to synchronize running a memory monitor check every N `IngestRun()` invocation on the index. The monitor trusts the index to synchronize on every invocation (which is not ideal, but works in practice).


lukebjerring:
#1000 :tada:

api/query/cache/monitor/monitor.go
```diff
@@ -114,19 +116,36 @@ func (m *indexMonitor) Start() error {
...
-		if heapBytes > m.maxHeapBytes {
-			m.logger.Errorf("Out of memory: %d > %d", heapBytes, m.maxHeapBytes)
-			m.idx.EvictRuns(m.percent)
-		} else {
-			m.logger.Debugf("Monitor: %d heap-allocated bytes OK", heapBytes)
+		timer := make(chan bool, 1)
+		go func() {
+			time.Sleep(m.interval)
+			timer <- true
+		}()
```
Hexcles: I think you'd need to extract this channel and goroutine out of the infinite loop and add an infinite loop within the function; i.e., have a single, long-living goroutine that writes to the same `timer` channel periodically.

Imagine if you have runs coming in such a fast rate that eviction is never triggered by the timer, you'd end up with a whole bunch of unnecessary goroutines sleeping. Besides, `timer` is created in each iteration but never closed, and I'm not sure if this would confuse the GC.

api/query/cache/monitor/monitor.go
```diff
@@ -114,19 +116,36 @@ func (m *indexMonitor) Start() error {
...
 			return errStopped
 		}
 
-		heapBytes := m.rt.GetHeapBytes()
-		if heapBytes > m.maxHeapBytes {
-			m.logger.Errorf("Out of memory: %d > %d", heapBytes, m.maxHeapBytes)
-			m.idx.EvictRuns(m.percent)
-		} else {
-			m.logger.Debugf("Monitor: %d heap-allocated bytes OK", heapBytes)
+		timer := make(chan bool, 1)
```
Hexcles: No need to specify the capacity 1.

api/query/cache/monitor/monitor.go
```diff
@@ -114,19 +116,36 @@ func (m *indexMonitor) Start() error {
...
+		go func() {
+			time.Sleep(m.interval)
+			timer <- true
+		}()
+
+		for {
+			select {
+			case <-timer:
+				m.check()
```
Hexcles: Move `m.check()` (and the other one in the next `case`) out of this `for` loop.

api/query/cache/monitor/monitor.go
```diff
@@ -175,13 +195,23 @@ func (m *indexMonitor) start() error {
 	return nil
 }
 
+func (m *indexMonitor) check() {
+	heapBytes := m.rt.GetHeapBytes()
+	if heapBytes > m.maxHeapBytes {
+		m.logger.Errorf("Out of memory: %d > %d", heapBytes, m.maxHeapBytes)
```
Hexcles: This is not part of the PR, but I think we shouldn't use `Errorf` here as this is expected to happen eventually; and perhaps we should change the wording of the message to something like "%d memory allocated, exceeding the threshold %d" to avoid confusion with actual, fatal OOM.

api/query/cache/monitor/monitor.go
mdittmer: I don't think there's a GC issue. There are no references to `timer` after this goroutine exits.

The trouble with a separate goroutine is that it doesn't start its timeout after the last `check()`, it starts it independently. That means the monitor may run more frequently than intended.

api/query/cache/monitor/monitor.go
mdittmer: The capacity of 1 ensures that a goroutine that writes to `timer` can exit (and be cleaned up) immediately, rather than waiting for the `select` to choose `timer`.

api/query/cache/monitor/monitor.go
mdittmer: The `m.check()` in the other case is conditional. That's why I don't have it outside the `for` loop.

api/query/cache/monitor/monitor.go
mdittmer: Sounds good. We can make this a `Warningf` and soften the wording.

Hexcles:
Please wait until #1005 is merged and rebase.

PR #1002, opened 2019-01-09
Move WCT out of Makefile and ensure it fails reliably

Hexcles:
Based on my observation, L112 in the original `Makefile` behaves differently on Travis. Namely, the `exit 1` statement seems to kill the whole `make` process. For example, this [Travis log](https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/169019618#L3330) did not have the following critical lines as on my local workstation:

```
Makefile:103: recipe for target 'web_components_test' failed
make: *** [web_components_test] Error 1   
```

L112 has too much magic after all (subshells connected with boolean operators). The reason was to ensure we could kill Xvfb regardless of test results. This can be better achieved by using a `trap` in a separate Bash script, which is what this PR does.

## Review

We should see web components test **red** on Travis with this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,13 +29,6 @@ VERBOSE := -v
 GO_FILES := $(shell find $(WPTD_PATH) -type f -name '*.go')
 GO_TEST_FILES := $(shell find $(WPTD_PATH) -type f -name '*_test.go')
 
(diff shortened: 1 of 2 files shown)


Hexcles:
Seems to be working: https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/169370861#L3306

PR #1003, opened 2019-01-10
Fix go_firefox_test

lukebjerring:
## Description
#999 updated FF v62 to v64, which we didn't notice fully breaking go_firefox_test because of other ongoing issues.
It's broken because they launched shadow dom in v63 (yay!) but don't support sending keys to shadow dom inputs (boo!)

### DIFF
--- a/webdriver/search_test.go
+++ b/webdriver/search_test.go
@@ -49,16 +49,20 @@ func testSearch(t *testing.T, path, elementName string) {
 	err = wd.WaitWithTimeout(resultsLoadedCondition, time.Second*10)
 	assert.Nil(t, err)
 
-	// Run the search.
-	searchBox, err := getSearchElement(wd, elementName)
-	if err != nil {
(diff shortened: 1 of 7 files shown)


webdriver/search_test.go
```diff
@@ -49,16 +49,20 @@ func testSearch(t *testing.T, path, elementName string) {
...
 
-	// Run the search.
-	searchBox, err := getSearchElement(wd, elementName)
-	if err != nil {
-		assert.FailNow(t, fmt.Sprintf("Error getting search element: %s", err.Error()))
-	}
 	folder := "2dcontext"
-	if err := searchBox.SendKeys(folder + selenium.EnterKey); err != nil {
-		assert.FailNow(t, fmt.Sprintf("Error sending keys: %s", err.Error()))
+	// NOTE(lukebjerring): firefox can't take sendKeys for shadow elements.
```
Hexcles: Nit: change this to TODO or FIXME

lukebjerring:
Only WCT still broken :tada:

PR #1004, opened 2019-01-10
Distinguish hostname and versioned hostname

lukebjerring:
## Description
We were checking for `master` to return staging.wpt.fyi, but this is convoluting two different cases (are we staging, or prod? vs which version am I).

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"os/exec"
 	"strings"
 	"time"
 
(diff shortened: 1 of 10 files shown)


shared/appengine.go
```diff
@@ -121,17 +128,37 @@ func (a AppEngineAPIImpl) GetUploader(uploader string) (Uploader, error) {
...
+	return hostname
+}
+
+// GetVersion returns the version name for the current environment.
+func (a AppEngineAPIImpl) GetVersion() string {
 	version := strings.Split(appengine.VersionID(a.ctx), ".")[0]
-	if version == "master" && hostname == "wptdashboard-staging.appspot.com" {
-		return "staging.wpt.fyi"
+	if appengine.IsDevAppServer() {
+		out, err := exec.Command("/usr/bin/git", "rev-parse", "--abbrev-ref", "HEAD").Output()
```
Hexcles: Maybe don't do this... It'd shell out to `git` in a lot of requests and I don't see much benefit. Besides, even if we want to keep it, we should perhaps use `"git"` instead of the explicit absolute path as it'd be different on different platforms.

lukebjerring: 1) only happens when running dev_appserver
1a) some requests use a version, but most are checks related so won't get called locally. The other is this use-case, which is nicer to see the git branch name as the version.
2) There's nothing on the path inside appengine, so can't use non-absolute
2b) failing to find git will just fall back to `dev_appserver`

PR #1006, opened 2019-01-11
Include PR number for PRs across forks

lukebjerring:
## Description
Fixes PR number storage for checks when the check is force-created from a PR that originates on a fork.


lukebjerring:
These changes were included in another PR.

PR #1007, opened 2019-01-11
Create suite for different app depending on environment

lukebjerring:
## Description
Uses the production wpt.fyi installation if the events are arriving in production.

### DIFF
--- a/api/checks/api.go
+++ b/api/checks/api.go
@@ -8,11 +8,13 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/google/go-github/github"
 	"github.com/web-platform-tests/wpt.fyi/api/checks/summaries"
 	"github.com/web-platform-tests/wpt.fyi/shared"
+	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
 	"google.golang.org/appengine/taskqueue"
 )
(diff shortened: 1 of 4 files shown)


lukebjerring:
This is also the last piece of the checks, so resolves https://github.com/web-platform-tests/wpt.fyi/issues/712

api/checks/webhook.go
```diff
@@ -16,11 +16,14 @@ import (
...
 const (
-	wptfyiCheckAppID         = int64(23318) // https://github.com/apps/wpt-fyi-status-check
-	wptfyiStagingCheckAppID  = int64(19965) // https://github.com/apps/staging-wpt-fyi-status-check
-	checksStagingAppID       = int64(21580) // https://github.com/apps/checks-staging-instance
+	wptfyiCheckAppID        = int64(23318) // https://github.com/apps/wpt-fyi-status-check
+	wptfyiStagingCheckAppID = int64(19965) // https://github.com/apps/staging-wpt-fyi-status-check
+	checksStagingAppID      = int64(21580) // https://github.com/apps/checks-staging-instance
+
+	wptRepoInstallationID        = int64(577173)
+	wptRepoStagingInstallationID = int64(449270)
```
Hexcles: Nit: move these constants to `api/checks/api.go` where they are used.

api/checks/api.go
```diff
@@ -176,3 +179,17 @@ func (s checksAPIImpl) CreateWPTCheckSuite(appID, installationID int64, sha stri
 	}
 	return suite != nil, err
 }
+
+func (s checksAPIImpl) GetWPTRepoAppInstallationIDs() (appID, installationID int64) {
+	aeID := appengine.AppID(s.ctx)
+	domainAndID := strings.Split(aeID, ":")
```
Hexcles: Nit: maybe a comment to give an example of `aeID` (I tend to do this whenever I'm doing non-trivial string manipulation).

Hexcles: Besides, in the method below, you split on "." instead of ":". Can you confirm?

https://github.com/web-platform-tests/wpt.fyi/blob/bff764ad5ee1df69402e10253bd50a13316167c9/shared/appengine.go#L145

lukebjerring: Yah, not sure why the inconsistency, but the ID format is documented on the `AppID` method itself.

PR #1009, opened 2019-01-11
Search cache: Write on read

mdittmer:
- Set limit on number of run IDs per request
- Fix access to `someShardedIndex.runs[]` that was not using mutex
- Perform write-on-read for run IDs not found in cache

These changes required `Run()` rather than `Runs()` on the `Index` interface.

### DIFF
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -51,6 +51,9 @@ func ErrRunLoading() error {
 type Index interface {
 	query.Binder
 
+	// Run loads the metadata associated with the given RunID value. It returns
+	// an error if the Index does not understand the given RunID value.
+	Run(RunID) (shared.TestRun, error)
 	// Runs loads the metadata associated with the given RunID values. It returns
 	// an error if the Index does not understand one or more of the given RunID
 	// values.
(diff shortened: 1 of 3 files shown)


mdittmer:
N.B. This won't work live until #1015 #1016 and #1017 land.

api/query/cache/index/index.go
```diff
@@ -330,6 +334,34 @@ func NewReportLoader() ReportLoader {
...
+
+func (i *shardedWPTIndex) syncGetRuns(ids []RunID) ([]shared.TestRun, error) {
+	i.m.RLock()
+	defer i.m.RUnlock()
+
+	runs := make([]shared.TestRun, len(ids))
+	for j := range ids {
+		run, ok := i.runs[ids[j]]
+		if !ok {
+			return nil, fmt.Errorf("Unknown run ID: %v", ids[j])
```
Hexcles: It's a bit counter-intuitive that `Run` and `Runs` will return different types of errors when a run is not loaded. WDYT?

api/query/cache/service/main.go
```diff
@@ -36,6 +40,9 @@ var (
 	evictRunsPercent       = flag.Float64("evict_runs_percent", 0.1, "Decimal percentage indicating what fraction of runs to evict when soft memory limit is reached")
 	updateInterval         = flag.Duration("update_interval", time.Second*10, "Update interval for polling for new runs")
 	updateMaxRuns          = flag.Int("update_max_runs", 10, "The maximum number of latest runs to lookup in attempts to update indexes via polling")
+	maxRunsPerRequest      = flag.Int("max_runs_per_request", 16, "Maximum number of runs that may be queried per request")
+
+	maxRunsPerRequestMsg string
```
Hexcles: Nit: maybe add a comment to say it will be assigned in `init()`.

PR #1010, opened 2019-01-11
Bake in master-only default query

lukebjerring:
## Description
Feature is launched in prod, so let's remove the flag.

### DIFF
--- a/webapp/components/test-runs-query-builder.js
+++ b/webapp/components/test-runs-query-builder.js
@@ -88,11 +88,9 @@ class TestRunsQueryBuilder extends WPTFlags(TestRunsUIQuery(PolymerElement)) {
     <paper-item>
       <paper-checkbox checked="{{diff}}" disabled="{{!canShowDiff}}">Show diff</paper-checkbox>
     </paper-item>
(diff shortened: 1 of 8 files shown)


Hexcles:
Many tests are broken.

Hexcles:
Integration tests are still broken.

webapp/components/test-runs-query-builder.js
```diff
@@ -88,11 +88,9 @@ class TestRunsQueryBuilder extends WPTFlags(TestRunsUIQuery(PolymerElement)) {
     <paper-item>
       <paper-checkbox checked="{{diff}}" disabled="{{!canShowDiff}}">Show diff</paper-checkbox>
     </paper-item>
-    <template is="dom-if" if="[[masterRunsOnly]]">
-      <paper-item>
-        <paper-checkbox id="master-checkbox" checked="{{master}}">Only master branch</paper-checkbox>
-      </paper-item>
-    </template>
+    <paper-item>
+      <paper-checkbox id="master-checkbox" checked="{{master}}">Only master branch</paper-checkbox>
```
Hexcles: We might want to tick this by default. See #1023 I just filed.

PR #1018, opened 2019-01-11
Optimize the front-end performance

Hexcles:
1. Use `webcomponents-loader.js` instead of `webcomponents-bundle.js` to
   load only the necessary polyfills on demand.
2. Fix the path matching in servieworker to make it actually functional.

## Review

To be verified in staging deployment.

### DIFF
--- a/webapp/templates/service-worker.js
+++ b/webapp/templates/service-worker.js
@@ -13,6 +13,8 @@ self.addEventListener(
   }
 );
 
+const cacheablePath = new RegExp('^/(components|static|node_modules)/');
+
 // Locally cache eligible components/files.
 self.addEventListener(
   'fetch',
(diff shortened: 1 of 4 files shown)


webapp/templates/service-worker.js
```diff
@@ -23,16 +25,20 @@ self.addEventListener(
           if (r) {
             return r;
           }
-          return fetch(e.request)
+
+          // IMPORTANT: Clone the request to reuse in fetch.
```
Hexcles: @lukebjerring according to https://developers.google.com/web/fundamentals/primers/service-workers/ , we also need to clone the request.

webapp/templates/service-worker.js
```diff
@@ -23,16 +25,20 @@ self.addEventListener(
...
+          const request = e.request.clone();
+          const url = new URL(e.request.url);
+          const path = url.pathname;
```
Hexcles: This is the part that was broken. `request.url` is just a string and does not have a method called `path`.

Hexcles:
Serviceworker is caching properly now. Cached load time is now down from 2s to 1s on my machine.

Some more digging in the network panel shows that the major bottleneck now is `https://www.gstatic.com/charts/loader.js`, whose response header has `no-cache` set and it takes up to 500ms. This JS is loaded via `wpt-results` -> `test-results-chart` -> `google-chart`. Let me try to make it non-blocking (lazy loading).

Hexcles:
Oh well, it's not as simple as I thought: the easiest approach (without restructuring the components) is to use [dynamic import()](https://developers.google.com/web/updates/2017/11/dynamic-import) which isn't supported by Firefox yet and there's no easy polyfill for it. Maybe we'll just wait (Firefox nightly has dynamic import behind a flag now).

Hexcles:
Re hosting Google charts loader ourselves: no, we probably can't do that.

> Can I use charts offline?
> Your users' computers must have access to https://www.gstatic.com/charts/loader.js in order to use the interactive features of Google Charts. This is because the visualization libraries that your page requires are loaded dynamically before you use them. The code for loading the appropriate library is part of the included script, and is called when you invoke the google.charts.load() method. Our terms of service do not allow you to download the google.charts.load or google.visualization code to use offline.

> Can I download and host the chart code locally, or on an intranet?
> Sorry; our terms of service do not allow you to download and save or host the google.charts.load or google.visualization code. However, if you don't need the interactivity of Google Charts, you can screenshot the charts and use them as you wish.

PR #1019, opened 2019-01-11
Restore package-lock.json

Hexcles:
Primary doc: https://docs.npmjs.com/files/package-locks

> It is highly recommended you commit the generated package lock to source control: this will allow anyone else on your team, your deployments, your CI/continuous integration, and anyone else who runs npm install in your package source to get the exact same dependency tree that you were developing on. Additionally, the diffs from these changes are human-readable and will inform you of any changes npm has made to your node_modules, so you can notice if any transitive dependencies were updated, hoisted, etc.

To answer my own question earlier, the difference between `package-lock.json` and `npm-shrinkwrap.json` is that the latter can be published to npm.

> Unlike package-lock.json, npm-shrinkwrap.json may be included when publishing a package.

Since we don't publish any package and the project is a top-level package, we don't need `npm-shrinkwrap.json`.

Also, `npm install [--production]` (without any other arguments) will install exactly the versions described in `package-lock.json`. Only installing a new package, or explicitly updating packages (e.g. `npm update`) will get us new versions.

Lastly, please make sure you have a recent-ish `npm` so that SHA512 is used as checksum instead of SHA1; otherwise, we'd run into endless back-and-forth updates.

### DIFF
--- a/webapp/package-lock.json
+++ b/webapp/package-lock.json
@@ -1,0 +1,9873 @@
+{
+  "name": "wptdashboard",
+  "requires": true,
+  "lockfileVersion": 1,
+  "dependencies": {
+    "@babel/code-frame": {
+      "version": "7.0.0",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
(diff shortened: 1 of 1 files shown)


lukebjerring:
@mdittmer @LukeZielinski, see above:

> Lastly, please make sure you have a recent-ish npm so that SHA512 is used as checksum instead of SHA1; otherwise, we'd run into endless back-and-forth updates.

Issue #1020, opened 2019-01-14
Fetch full runs from Azure Pipelines daily runs


foolip:
https://github.com/web-platform-tests/wpt/issues/14836 tracks performing daily runs of Safari on Azure Pipelines. There's an unresolved problem with actually triggering the runs in the main wpt repo, but https://dev.azure.com/foolip/wpt/_build/results?buildId=343 is an example of what it will look like. Using that repository to develop the code to ingest these results should work.

@lukebjerring @Hexcles is this in either of your ballparks?

lukebjerring:
Currently the Azure handler expects exactly one report in the artifact. We'll need to tweak this code to pull out and attach all of the artifact's report shards.

https://github.com/web-platform-tests/wpt.fyi/blob/master/api/azure/webhook.go#L132

lukebjerring:
This is now fixed and exposed for notification via #1054

PR #1022, opened 2019-01-14
Handle multiple report file (shards) in azure check webhook

lukebjerring:
## Description
Part of #1020 

Iterates, and wildcards, over the contents of an artifact, instead of assuming a single file.


api/azure/api.go
```diff
@@ -48,7 +51,7 @@ type ArtifactResource struct {
 type API interface {
 	HandleCheckRunEvent(*github.CheckRunEvent) (bool, error)
 	GetAzureArtifactsURL(owner, repo string, buildID int64) string
-	FetchAzureArtifact(BuildArtifact, io.Writer) error
+	FetchAzureArtifact(BuildArtifact, *multipart.Writer) error
```
Hexcles: I would much prefer using `io.Writer` instead of any concrete writer type, because:

1. It is immediately obvious that this is a writer that doesn't need to close, as the standard `io` interfaces are well known, unlike the concrete writer types that can vary (e.g. a HTTP response writer needs to be closed).
2. It is easier to test by using mocks or other writers.

And it's also the convention that I've observed in the standard library.

api/azure/api.go
```diff
@@ -96,24 +99,37 @@ func (a apiImpl) FetchAzureArtifact(artifact BuildArtifact, w io.Writer) error {
...
+		if reportPath.MatchString(f.Name) {
+			fileName := f.Name[len(artifactName)+1:]
+			fileField, err := writer.CreateFormFile("result_file", fileName)
 			var fileData io.ReadCloser
 			if fileData, err = f.Open(); err != nil {
-				log.Errorf("Failed to extract %s", reportPath)
+				log.Errorf("Failed to extract %s", f.Name)
 				return err
 			}
 			defer fileData.Close()
```
Hexcles: This is problematic. `defer` pushes the execution to the end of the **function**, but to the end of the current scope. Hence, we would hold onto a lot of readers.

I'd suggest wrapping everything inside this `if` block in a function, because the alternative (explicitly calling `Close()` on every exit path is more error-prone).

api/azure/api.go
```diff
@@ -122,10 +138,13 @@ func (a apiImpl) FetchAzureArtifact(artifact BuildArtifact, w io.Writer) error {
 				log.Errorf("Failed to close gzip writer")
 				return err
 			}
-			return nil
+			extracted++
 		}
 	}
-	return fmt.Errorf("File %s not found in zip", reportPath)
+	if extracted < 1 {
+		return errors.New("No wpt_report.json files found in zip")
```
Hexcles:
```suggestion
		return errors.New(`No "wpt_report.*\.json" files found in zip`)
```

to make it clear that we allow arbitrary suffixes after `wpt_report`.

api/azure/api_test.go
```diff
@@ -0,0 +1,38 @@
...
+	"path"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestExtractFiles(t *testing.T) {
+	_, filename, _, _ := runtime.Caller(0)
+	data, err := ioutil.ReadFile(path.Join(path.Dir(filename), "artifact_test.zip"))
```
Hexcles: Perhaps add a comment to explain where is `artifact_test.zip` and what is inside.

api/azure/api.go
lukebjerring: Reverted back to writer. It's not obvious to me whether creating multiple multipart writers into the same buffer will behave as expected.

api/azure/api.go
lukebjerring: Scoped.

api/azure/api_test.go
lukebjerring: Done.

api/azure/api.go
Hexcles: In fact, I didn't notice in your original version you called `CreateFormFile` inside the function, which means you did need a concrete type. I mistakenly thought you only needed `Write` and `Close`.

PR #1024, opened 2019-01-14
[receiver] Allow multiple result_file fields

Hexcles:
This PR allows uploaders to send multiple (chunked) files in a multipart
/api/results/upload request.

Works alongside with #1022 .

### DIFF
--- a/api/receiver/receive_results.go
+++ b/api/receiver/receive_results.go
@@ -7,7 +7,6 @@ package receiver
 import (
 	"fmt"
 	"io"
-	"mime/multipart"
 	"net/http"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


api/receiver/receive_results_test.go
```diff
@@ -144,120 +129,203 @@ func TestHandleResultsUpload_extra_params(t *testing.T) {
...
-func TestHandleFilePayload(t *testing.T) {
+func TestHandleResultsUpload_single_file(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	f := &os.File{}
-	extraParams := map[string]string{
-		"browser_name": "firefox",
-	}
+	buffer := &bytes.Buffer{}
```
lukebjerring:
```suggestion
	buffer := new(bytes.Buffer)
```

api/receiver/receive_results_test.go
```diff
@@ -144,120 +129,203 @@ func TestHandleResultsUpload_extra_params(t *testing.T) {
 	assert.Equal(t, resp.Code, http.StatusOK)
 }
 
-func TestHandleFilePayload(t *testing.T) {
+func TestHandleResultsUpload_single_file(t *testing.T) {
```
lukebjerring: There's a lot of overlap between this and multiple.
Have a look at `testing.T.Run` for a clean way of collapsing the cases (and we can get rid of  `_single_file` suffixes, yay).

Hexcles:
@lukebjerring PTAL again

api/receiver/receive_results.go
```diff
@@ -60,58 +59,58 @@ func HandleResultsUpload(a AppEngineAPI, w http.ResponseWriter, r *http.Request)
 		}
 	}
 
-	f, _, err := r.FormFile("result_file")
```
lukebjerring: @Hexcles - FYI - we were ignoring the error here, and `FormFile` actually calls `ParseMultipartForm` in effect.

Hexcles: Now I'm really confused. `FormValue` also calls `ParseMultipartForm`. And we were not ignoring the error; it was checked later on L77. I don't see anything wrong yet.

PR #1030, opened 2019-01-16
wpt-results component: Retry /api/search on HTTP 422

mdittmer:
This change makes client code aware of write-on-read attempts that are triggered when `HTTP 422` is returned from `/api/search`. The client will retry searches that return `HTTP 422` up to five times, retrying every five seconds.


webapp/components/wpt-results.js
```diff
@@ -944,6 +955,23 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
     this.refreshDisplayedNodes();
     this.loadData();
   }
+
+  retry(f, shouldRetry, num, wait) {
```
lukebjerring: Let's migrate this to loading-state.js and test it :)

PR #1031, opened 2019-01-16
searchcache: When some runs missing, HTTP 422 + partial results

mdittmer:
This change allows `searchcache` to return `HTTP 422` and partial results when at least one, but not all, runs needed to service a query are found in the cache.

This allows clients to render partial results and decide whether or not to retry the query later to get more complete results.

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -118,8 +118,15 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Return to client `http.StatusUnprocessableEntity`immediately if any runs
 	// are missing.
-	if len(missing) != 0 {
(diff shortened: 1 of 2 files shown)


api/query/cache/service/main.go
```diff
@@ -118,8 +118,15 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 
 	// Return to client `http.StatusUnprocessableEntity`immediately if any runs
 	// are missing.
-	if len(missing) != 0 {
-		http.Error(w, fmt.Sprintf("Test run(s) with ID(s) %v not loaded into cache. Try again later.", missing), http.StatusUnprocessableEntity)
+	if len(runs) == 0 {
```
lukebjerring: Might be redundant based on earlier code, but check that `len(missing) > 0` too.

PR #1032, opened 2019-01-16
Rejig the indentation + fix retry

lukebjerring:
Suggestion For #1030

### DIFF
--- a/webapp/components/wpt-results.js
+++ b/webapp/components/wpt-results.js
@@ -593,11 +593,12 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
         this.searchResults = json.results;
         this.refreshDisplayedNodes();
       }),
-    (e) => {
-      // eslint-disable-next-line no-console
-      console.log(`Failed to load: ${e}`);
-      this.resultsLoadFailed = true;
-    });
+      (e) => {
+        // eslint-disable-next-line no-console
+        console.log(`Failed to load: ${e}`);
+        this.resultsLoadFailed = true;
+      }
+    );
   }
 
   fetchDiff() {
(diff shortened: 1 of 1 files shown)


webapp/components/wpt-results.js
```diff
@@ -575,19 +575,24 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
...
+          return r.json();
+        },
+        err => err === 422,
+        5,
+        5000
+      ).then(json => {
+        this.searchResults = json.results;
+        this.refreshDisplayedNodes();
+      }),
     (e) => {
```
mdittmer: nit: Shouldn't this (and subsequent lines) be indented one more step? `(e) ...` is second arg of `this.load`, which is indented one step in.

webapp/components/wpt-results.js
```diff
@@ -958,7 +963,11 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
         if (count >= num || !shouldRetry(err)) {
           throw err;
         }
-        window.setTimeout(retry, wait);
+        return new Promise((resolve, reject) => {
+          window.setTimeout(() => {
+            return retry().then(resolve, reject)
```
mdittmer: nit: `return` is misleading. Nothing cares about the return of f'n passed to `window.setTimeout`. Should drop return, and could consider simply `() => retry().then(resolve, reject)`.

mdittmer:
:trophy: Achievement unlocked :trophy:

PR Inception :clock1: :clock10: BWWWWAAAA! :clock1030: :clock11: 

Buildings

:clock1030: :clock11: BWWWWAAAA! :clock1: :clock10:

Moar buildings

PR #1033, opened 2019-01-16
Show a toast when /api/search returns a 422

lukebjerring:
## Description
Lets the user know that they can retry.


lukebjerring:
They can't both appear on the screen simultaneously, but besides that they're fine.

Issue #912, opened 2018-12-13
Detect regressions on master


foolip:
Related to https://github.com/web-platform-tests/wpt.fyi/issues/866

@beaufortfrancois has reported this recent regression in picture-in-picture:
good: https://wpt.fyi/results/picture-in-picture?product=chrome[experimental,taskcluster]&sha=5dd605cdfa
bad: https://wpt.fyi/results/picture-in-picture?product=chrome[experimental,taskcluster]&sha=c522b884f7

The full diff between those runs:
https://wpt.fyi/results/?label=experimental&product=chrome%5Btaskcluster%5D%405dd605cdfa&product=chrome%5Btaskcluster%5D%40c522b884f7&diff

There are lots of regressions, introduced by https://github.com/web-platform-tests/wpt/pull/13966. We should have detected that before landing the changes, but *if* it had been contained to picture-in-picture we should still have detected the regression when it happened. In addition to changes in the wpt repo, infrastructure changes and browser releases can also cause regresssions.

@jgraham @gsnedders FYI

foolip:
Constructing the diff URLs in https://github.com/web-platform-tests/wpt/issues/14495 was quite time-consuming, so automatically comparing all pairs of master runs for a configuration would help immensely.

foolip:
Doing https://github.com/web-platform-tests/wpt/issues/13263 would mitigate the need for this somewhat, but I'd still love to have links to diffs for each master run, even if there aren't regressions.

lukebjerring:
@foolip - the wpt.fyi custom check does compare each full master run to the most recent master run for the same product, and surfaces the regressions. It's noisy due to flakiness, but is that what this issue was asking for?

foolip:
Flakiness isn't what I wanted to discover, what I have in mind is a change to testharness.js or something in tools/ breaking a lot of tests. In other words, we'd need to compare two runs, somehow account for flakiness, and see if there are regressions outside of the affected tests. If there are, some human should be notified.

lukebjerring:
OK - Then this isn't a wpt.fyi issue. wpt.fyi does detect regressions, but it gets ignored because of the flaky tests being noisy. The remaining work lies in fixing the flaky tests.

foolip:
Well, dealing with flakiness might be a blocker to make this useful, but there's still wpt.fyi work:
- see if there are regressions outside of the affected tests (or just if there are very many regressions)
- human should be notified

[...]

PR #1035, opened 2019-01-16
Allow multiple results per product when more than 1 sha

lukebjerring:
## Description
Changes the behaviour of querying for multiple shas, to not implicitly set a max-count of one result for each product. (For other queries, such as no params, the default is one result per product).

If a user wants exactly one result per product (across 2 shas) they can add the param explicitly.

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -103,8 +103,8 @@ func LoadTestRunKeysForFilters(store shared.Datastore, filters shared.TestRunFil
 	limit := filters.MaxCount
 	offset := filters.Offset
 	from := filters.From
-	if limit == nil && from == nil {
-		// Default to a single, latest run when from & max-count both empty.
+	// Default to a single, latest run when not using any "more than one results" filters.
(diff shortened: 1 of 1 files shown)


lukebjerring:
Example from https://github.com/web-platform-tests/wpt.fyi/issues/912 :
https://shas-dot-wptdashboard-staging.appspot.com/results/?label=master&shas=5dd605c,c522b88&product=chrome&diff

lukebjerring:
Sorry, I typed many word do the explain. Revised to be interpretable as English. 

We added support for multiple SHA params as part of the ?pr=123 support, but it behaves strangely. this pr makes it behaves more as you might expect (if you ask for more than one Sha, you get all results across those shas)

Hexcles:
If there's existing tests related to the default value of `limit`, I'd like you to make them cover the new condition. Otherwise, it's perhaps not that important.

lukebjerring:
Currently untested - tests would need a refactor. Separate PR.

Issue #1025, opened 2019-01-15
PR runs showing up on https://wpt.fyi/results/?label=experimental


foolip:
https://wpt.fyi/results/?label=experimental and links like it have been passed around for a long time, and today https://wpt.fyi/results/preload?label=experimental&q=srcset was linked from [Intent to Ship: imagesrcset and imagesizes attributes on &lt;link>](https://groups.google.com/a/chromium.org/d/msg/blink-dev/07CbItWfA4M/CG1qF8IoEAAJ).

https://wpt.fyi/results/preload?label=experimental is very broken for Chrome and Firefox:
![image](https://user-images.githubusercontent.com/498917/51173608-0f38b580-18b6-11e9-9747-fc7107672be7.png)

That's because https://github.com/web-platform-tests/wpt/commit/7ca362d756 was a commit for https://github.com/web-platform-tests/wpt/pull/14847, and the run was an affected tests run on Taskcluster. The run doesn't seem to have been labeled correctly, however.

foolip:
@lukebjerring can you take a look?

lukebjerring:
This is an unfortunate consequence of enabling non-master runs in production.
I've enabled the query-builder yesterday, which should mitigate the need for manual construction of URLs, which exacerbate this problem, (because the default query + builder behaviour include `&label=master` by default).

lukebjerring:
Closing this is WAI - include the `master` label if you want master-only runs.

foolip:
I guess that different ways of making master implied based on other labels were considered and have other not-nice properties?

foolip:
@fbeaufort was also bitten by this, but found this issue and the workaround.

beaufortfrancois:
Could we redirect all requests with the `label=master` in the mean time if `label` is only `experimental` in the mean time?

foolip:
I noticed/remembered https://wpt.fyi/results/?label=experimental is linked in ["Intent to {Implement,Experiment,Ship}" email templates for blink-dev](https://docs.google.com/document/d/1vlTlsQKThwaX0-lj_iZbVTzyqY7LioqERU8DK3u3XjI/edit#bookmark=id.dog2hyrj4r72). I've updated the link, but that's probably where some of the links have come from.

A redirect or other fix would be nice, there will be links in places that can't be updated too.

foolip:
[...]

PR #1038, opened 2019-01-17
Warn about ?label=experimental

lukebjerring:
## Description
Resolves https://github.com/web-platform-tests/wpt.fyi/issues/1025

Shows a warning toast when you query for label=experimental, but not for label=master, with a button for adding the master label.


lukebjerring:
It should dismiss straight away, but 

>  this.handleDismissToast is not a function

Issue #1037, opened 2019-01-16
Split the staging deployment job on Travis


Hexcles:
Due to the increasing number of services, the whole deployment of all services on master sometimes takes close to 1 hour. We'd need to split it (e.g. deploying one service per task), which will also help alleviate #581 .

PR #1042, opened 2019-01-18
Add feature for copying a permalink

### LINKED ISSUE #645: Expose a permalink URL
When pasting link to wpt.fyi into bug trackers as evidence of some problem, it would be very nice to paste links that will remain stable. When visiting that link, the ability to see the same results but for the latest runs would be great too.

I have manually been adding `sha=12345` when I wanted this myself.

lukebjerring:
## Description
Fixes #645 

Adds a small "Link" button next to "Edit" in the query banner, which opens a dialog allowing the user to tweak the URL a bit and copy to the clipboard with a click (in Chrome/FF, at least).

## Review Information
- Head to /flags, enable the permalink flag
- Visit the homepage, nav to a directory somewhere
- Hit edit, toggle the tickboxes, copy to clipboard

### DIFF
--- a/webapp/components/wpt-permalinks.js
+++ b/webapp/components/wpt-permalinks.js
@@ -4,13 +4,13 @@
  * found in the LICENSE file.
  */
 
-import '../node_modules/@polymer/paper-dialog/paper-dialog.js';
-import '../node_modules/@polymer/paper-item/paper-item.js';
-import '../node_modules/@polymer/paper-toast/paper-toast.js';
-import '../node_modules/@polymer/paper-input/paper-input.js';
 import '../node_modules/@polymer/paper-checkbox/paper-checkbox.js';
+import '../node_modules/@polymer/paper-dialog/paper-dialog.js';
+import '../node_modules/@polymer/paper-input/paper-input.js';
+import '../node_modules/@polymer/paper-item/paper-item.js';
+import '../node_modules/@polymer/paper-tab/paper-tab.js';
 import '../node_modules/@polymer/paper-tabs/paper-tabs.js';
-import '../node_modules/@polymer/polymer/lib/elements/dom-repeat.js';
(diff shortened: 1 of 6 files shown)


foolip:
@lukebjerring no staging instance was deployed for this, a bug?

webapp/components/wpt-permalinks.js
```diff
@@ -0,0 +1,150 @@
...
+        paper-tabs {
+          --paper-tabs-selection-bar-color: var(--paper-blue-500);
+        }
+        paper-tab {
+          --paper-tab-ink: var(--paper-blue-300);
+        }
+      </style>
+      <paper-dialog>
+        <paper-tabs selected="{{selectedTab}}">
+          <paper-tab title="Link to these specific runs, via their IDs">These runs</paper-tab>
```
foolip: Nice!

webapp/components/wpt-permalinks.js
```diff
@@ -0,0 +1,150 @@
...
+    return url;
+  }
+
+  async handleCopyToClipboard() {
+    try {
+      const input = this
+        .shadowRoot.querySelector('paper-input')
+        .shadowRoot.querySelector('input');
+      input.select();
+      document.execCommand('copy');
```
foolip: This returns false when it doesn't work, it doesn't throw. That means the success toast will always show and the failure never. I haven't tested in Safari, but are you sure it works?

lukebjerring:
Yeah we temporarily broke staging deployment in #1037 - should be fixed with the master-merge push.

lukebjerring:
cc @Hexcles

lukebjerring:
https://permalink-dot-wptdashboard-staging.appspot.com

lukebjerring:
Works on safari 12 (tested on browser stack)

PR #1046, opened 2019-01-19
Fix the result collection in the Azure webhook

Hexcles:
Bug was introduced in
https://github.com/web-platform-tests/wpt.fyi/pull/1022 because of an
incorrect suggestion from me. This PR fixes the bug by reusing the
same multipart.Writer throughout a request.

This PR also improves error logging in the results receiver to help
future debugging.


lukebjerring:
Needs to fix the package name for the mockgen

PR #1057, opened 2019-01-23
Add exists root node in structured queries

lukebjerring:
## Description
Modifies the behaviour of structured queries to do disjuction of runs at the root node, via an `exists` root.

The exists root has an array of query objects, each of which must be true for some (single) run. This allows for conjunction of atoms over a single, specific run, allowing the combination of requirements.


lukebjerring:
Example query:

    curl -v -X POST  https://exists-bind-dot-wptdashboard-staging.appspot.com/api/search --data '{"run_ids":[6339218469027840,6246586443104256,5849786024460288,5105477922848768,6635882278813696,4529664676593664,5146098582683648,5068494160789504,4823266896117760,5127544793726976],"query":{"exists":[{"browser_name":"chrome","status":"FAIL"},{"browser_name":"chrome","status":"PASS"}]}}'

lukebjerring:
And a query that uses `and`, which wouldn't work before:

> (chrome:pass or chrome:ok) (chrome:!pass and chrome:!ok)

    curl -v -X POST  \
      https://exists-bind-dot-searchcache-dot-wptdashboard-staging.appspot.com/api/search/cache \
      --data '{"run_ids":[6339218469027840,6246586443104256,5849786024460288,5105477922848768,6635882278813696,4529664676593664,5146098582683648,5068494160789504,4823266896117760,5127544793726976],"query":{"and":[{"or":[{"browser_name":"chrome","status":"PASS"},{"browser_name":"chrome","status":"OK"}]},{"and":[{"browser_name":"chrome","status":{"not":"PASS"}},{"browser_name":"chrome","status":{"not":"OK"}}]}]}}

lukebjerring:
@mdittmer - note that as of #1058, PRs that deploy the webapp and searchcache will now point at each other for the same deployed version names (meaning that we can run a search on the webapp and see its results visually).

api/query/atoms.go
```diff
@@ -36,10 +36,36 @@ type TestNamePattern struct {
...
+			byRun[j] = e.Args[i].BindToRuns(run)
+		}
+		// Each separate Exists query is true if there exists a run that makes its concrete query true.
+		queries[i] = Or{
+			Args: byRun,
+		}
+	}
+	// And the overall node is true if all its exists queries are true.
+	return And{
+		Args: queries, // T
```
mdittmer: I don't understand this comment.

api/query/atoms_test.go
```diff
@@ -368,7 +383,56 @@ func TestStructuredQuery_bindStatusSomeRuns(t *testing.T) {
 			},
 		},
 	}
-	assert.Equal(t, expected, q.BindToRuns(runs))
+	assert.Equal(t, expected, q.BindToRuns(runs...))
+}
+
+func TestStructuredQuery_bindExists(t *testing.T) {
```
mdittmer: I know it makes the test longer, but could include a test that's more realistic? Multiple branches to the exists and multiple runs (with different browser names). That scenario is what the exists structure was created for.

webapp/components/test-search.js
```diff
@@ -8,27 +8,7 @@ import '../node_modules/@polymer/paper-tooltip/paper-tooltip.js';
...
-skipSpacesIfInSyntacticContext:function(){return this.inSyntacticContext()?this.skipSpa
```
mdittmer: +1 to de-polluting test-search!

webapp/components/test-search.js
```diff
@@ -39,7 +19,11 @@ const QUERY_GRAMMAR = ohm.grammar(`
 
     Exp = NonemptyListOf<OrPart, or>
 
-    OrPart = NonemptyListOf<AndPart, and>
+    OrPart
```
mdittmer: Do we need this change? It looks like the non-default alternatives for `OrPart` and `AndPart` are now the same, which is redundant.

api/query/atoms.go
lukebjerring:
```suggestion
		Args: queries,
```

api/query/atoms_test.go
Hexcles: I also agree that we should have multiple branches in the top level. The most recent change added multiple runs, but there's still only one `And` in the top-level `Exists`.

lukebjerring: Done (added a separate BrowserStatusNeq).

api/query/cache/index/filter.go
```diff
@@ -144,7 +171,7 @@ func newFilter(idx index, q query.ConcreteQuery) (filter, error) {
 		}
 		return Not{idx, f}, nil
 	default:
-		return nil, errUnknownConcreteQuery
+		return nil, fmt.Errorf("Unknown ConcreteQuery type %s", reflect.TypeOf(q))
```
Hexcles: Hmm... I'd like to avoid using reflection if possible, especially for an informative error message.

api/query/atoms.go
```diff
@@ -18,7 +18,7 @@ var browsers = shared.GetDefaultBrowserNames()
 // AbstractQuery is an intermetidate representation of a test results query that
 //  has not been bound to specific shared.TestRun specs for processing.
 type AbstractQuery interface {
-	BindToRuns(runs []shared.TestRun) ConcreteQuery
+	BindToRuns(runs ...shared.TestRun) ConcreteQuery
```
Hexcles: After reviewing the PR, I agree this is a much better interface. Good job!

api/query/cache/index/filter.go
lukebjerring: This line should basically never get hit, and when it does, debugging the type is quite important. This is how I noticed that we weren't translating True/False across filters.
We're an open source project; revealing the concrete type of a json.Unmarshal and bind isn't a security problem.

PR #1058, opened 2019-01-23
Add api method for getting versioned service hostname

lukebjerring:
## Description
Ensures we get a versioned external service hostname, so that branches like https://github.com/web-platform-tests/wpt.fyi/pull/1057 keep the webapp and searchcache in sync, so the changes can be tested in the UI.

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -163,6 +163,7 @@ func (a AppEngineAPIImpl) GetVersionedHostname() string {
 }
 
 // GetServiceHostname returns a versioned canonical hostname for the given service (module).
+// If the service does not have the current version, AppEngine routing will fall back to its default version.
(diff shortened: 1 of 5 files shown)


shared/appengine.go
```diff
@@ -161,6 +162,20 @@ func (a AppEngineAPIImpl) GetVersionedHostname() string {
 	return fmt.Sprintf("%s-dot-%s", a.GetVersion(), hostname)
 }
 
+// GetServiceHostname returns a versioned canonical hostname for the given service (module).
```
Hexcles:
```suggestion
// GetServiceHostname returns a versioned canonical hostname for the given service (module).
// If the service does not have the current version, AppEngine routing will fall back to its default version.
```

PR #1060, opened 2019-01-24
Cleanup searchcache automatically too

lukebjerring:
## Description
Include `searchcache` in the set of things we clean up.


lukebjerring:
TBR'ing this simple change.

PR #1074, opened 2019-01-25
Always pass non-regressing checks, and use action_required instead of failure

lukebjerring:
## Description
Regardless of whether we use neutral for newly-failing / deleted results, always set a pass status when we don't see anything amiss.

Change the non-neutral status to "action_required" instead of failing - fail is too aggressive and might discourage people adding failing tests.

## Review
Docs for valid "conclusion" field values:
https://developer.github.com/v3/checks/runs/#update-a-check-run

cc @jgraham

### DIFF
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -199,7 +199,7 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, suite sha
 	regressions := diff.Differences.Regressions()
 	neutral := "neutral"
 	checkState.Conclusion = &neutral
-	checksCanFailAndPass := aeAPI.IsFeatureEnabled(failChecksOnRegressionFeature)
+	checksCanBeNonNeutral := aeAPI.IsFeatureEnabled(failChecksOnRegressionFeature)
 
 	var summary summaries.Summary
 	host := aeAPI.GetHostname()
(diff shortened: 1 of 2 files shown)


foolip:
Alright, so from the docs I guess neutral and action_required are the only conclusions that we have to work with for the uncertain cases.

Not in this PR, but would it make sense to use the text that says "wpt.fyi - Chrome results" to highlight the most important thing to note about the results? This would effectively give us more states. I might look into "regressed existing tests" but not "new failing tests" assuming they both use the same conclusion.

api/checks/update.go
```diff
@@ -263,9 +261,9 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, suite sha
 				data.More++
 			}
 		}
-		if checksCanFailAndPass {
-			failure := "failure"
-			data.CheckState.Conclusion = &failure
+		if checksCanBeNonNeutral {
+			actionRequired := "action_required"
```
foolip: I'm happy to try this and see how it works in practice. I'd expect it to occasionally be a nuisance that the same action is required on as many checks as we have browsers. This could be changed back to "neutral" if a quickfix for that or other problem isn't possible.

jgraham:
I think I would prefer to leave it all as neutral until we have the better data to work with so that we don't train people to interpret the statuses in one way and then change the correct interpretation later on.

lukebjerring:
So, leaving it as neutral is independent of this change. The change is, when the feature for non-neutral is enabled, do we show "failure" outcome or "action_required" outcome.

api/checks/update.go
lukebjerring: As many checks as we have _failing_ browsers, to be precise.

foolip:
@lukebjerring, there is still the optioin of using neutral as the conclusion for some checks where we're unsure if action_required is the right trade-off. I believe that's what @jgraham was getting at, not to leave it neutral for every case.

foolip:
Well actually I don't know what @jgraham intended

lukebjerring:
Yes, but that'd be a revision of the behaviour separate to this - which was a change of considering failure to (always) be too extreme of a conclusion.

Issue #81, opened 2018-04-23
Search doesn't do anything for individual test-result files (subtest pages)


lukebjerring:
Behaviour should probably be filtering on the subtest name, at a minimum. We'll eventually want to filter on particular result types too (TIMEOUT, FAIL, etc). See https://github.com/web-platform-tests/wpt.fyi/issues/73.

foolip:
Being able to search on both file contents and the failure messages would be useful too.

foolip:
(Mostly failure messages, file contents is readily searchable with grep already.)

mdittmer:
Search box no longer shows up in these contexts. Should we close this, or leave it open as a feature request to implement single-top-level-test search?

lukebjerring:
Search now supports filtering by results.

PR #1075, opened 2019-01-25
Add includeSubtests flag for emitting subtest names

lukebjerring:
## Description
Adds a (yet unused) flag to the aggregator, to emit an array of subtest names.
This will be followed by a change to set the flag to true when searching for a specific test-file path.

### DIFF
--- a/api/query/cache/index/aggregator.go
+++ b/api/query/cache/index/aggregator.go
@@ -17,8 +17,9 @@ type aggregator interface {
 type indexAggregator struct {
 	index
 
-	rus []RunID
-	agg map[uint64]query.SearchResult
+	rus             []RunID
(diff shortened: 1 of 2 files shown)


Hexcles:
I know this PR is a no-op at the moment, but what is the space implication of this when we actually start to load subtests to memory? Doesn't it bloat the memory usage by a few magnitudes (and hence reduce the cached time span by several magnitudes)?

lukebjerring:
I'm not sure that it affects the contents of the cache at all (unless you mean full-request result caching)?

Hexcles:
Ooops, sorry this is the aggregator not the indexer. Never mind.

lukebjerring:
This was working towards https://github.com/web-platform-tests/wpt.fyi/issues/81

Issue #149, opened 2018-05-15
Also link to reference on w3c-test.org


foolip:
Until https://github.com/web-platform-tests/wpt.fyi/issues/57 is resolved, which is probably very hard, it would be handy to be able to see the reference of a test more easily. Example:
https://wpt.fyi/svg/extensibility/foreignObject/stacking-context.html

Here, a link to https://w3c-test.org/svg/extensibility/foreignObject/stacking-context-expected.html and possible also to its source on GitHub would be handy.

foolip:
Note that the link will break after https://github.com/w3c/web-platform-tests/pull/11004 is merged.

foolip:
Merged: https://w3c-test.org/svg/extensibility/foreignObject/stacking-context-ref.html

PR #1080, opened 2019-01-28
Include a link to the ref for reftests

lukebjerring:
## Description
Resolves https://github.com/web-platform-tests/wpt.fyi/issues/149

Adds a feature for including a link to the ref, in addition to the test itself. This is implemented by leveraging the manifest.

Naively assumes a single ref, with an `==`. Other cases (rare) will be added in the future.


webapp/components/wpt-results.js
```diff
@@ -441,8 +450,22 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
...
+    }
+    return new URL(`${this.scheme}://w3c-test.org${path}`);
+  }
+
+  computeTestRefURL(testType, path, manifest) {
+    if (!this.showTestRefURL || testType !== 'reftest') {
+      return;
+    }
+    const item = Object.values(manifest.items['reftest']).find(v => v.find(i => i[0] === path));
+    // In item[0], the 2nd item is the refs array, and we take the first ref (0).
```
foolip: If you know where it's generated, can you point to the Python code generating the manifest that this needs to be in sync with? Maybe @Hexcles can find it quickly.

PR #1082, opened 2019-01-28
Change browser_name to product in searchcache

lukebjerring:
## Description
Consolidates the product-spec parsing behaviour of `product` query params and "browser_name" in structured queries to use the same parsing logic server-side.

In turn, this facilitates searching for version-specific statuses, which we can use to find potential regressions in major releases, with `chrome-64:fail chrome-63:pass` type of thing.

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -120,6 +120,12 @@ func ParseProductSpec(spec string) (productSpec ProductSpec, err error) {
 	return productSpec, nil
 }
 
+// ParseProductSpecUnsafe ignores any potential error parsing the given product spec.
+func ParseProductSpecUnsafe(s string) ProductSpec {
+	parsed, _ := ParseProductSpec(s)
+	return parsed
+}
+
 // ParseProduct parses the `browser-version-os-version` input as a Product struct.
 func ParseProduct(product string) (result Product, err error) {
 	pieces := strings.Split(product, "-")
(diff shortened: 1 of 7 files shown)


shared/params.go
```diff
@@ -120,14 +120,20 @@ func ParseProductSpec(spec string) (productSpec ProductSpec, err error) {
...
+
 // ParseProduct parses the `browser-version-os-version` input as a Product struct.
 func ParseProduct(product string) (result Product, err error) {
 	pieces := strings.Split(product, "-")
 	if len(pieces) > 4 {
 		return result, fmt.Errorf("invalid product: %s", product)
 	}
 	result = Product{
-		BrowserName: pieces[0],
+		BrowserName: strings.ToLower(pieces[0]),
```
Hexcles: Why is `strings.ToLower` needed? Doesn't the frontend (more specifically ohm) already convert the keywords to lower-case?

lukebjerring:
Yes, because there'll be URLs out in the wild with the old query.

shared/params.go
lukebjerring: Because the tests had capitalized names and there's no reason that we shouldn't parse `product=Chrome` as the chrome browser.

Hexcles:
But the URL would only contain the `q` parameter with human-readable queries, not the structured form, so neither `browser_name` nor `product` would show up in the URL. Or did I miss something?

lukebjerring:
True; only covers direct searchcache queries, but still better safe than sorry.

lukebjerring:
Whoops, that comment got misinterpreted.

lukebjerring:
Once searchcache deploys, an example query:
https://specs-dot-wptdashboard-staging.appspot.com/results/?sha=fc57d7ad7a&product=chrome-72&product=chrome-73&q=chrome-72%3Apass+chrome-73%3Afail

lukebjerring:
Concern noted. Won't make a habit of polluting the json objects.

Issue #1076, opened 2019-01-28
Newly added tests show up as flakiness


foolip:
Using the insight tab and tweaking a bit to get Safari TP runs for Azure Pipelines, I finally ended up at this view:
https://wpt.fyi/results/xhr/sync-xhr-supported-by-feature-policy.html?run_id=5701098400645120&run_id=5648138584981504&run_id=5164405260025856&run_id=6496107718770688&run_id=5458240213090304

Here, the test was added on Jan 24, so reporting it as flakiness is not helpful. As flakiness is improved, the view ought to become dominated by recently added tests.

lukebjerring:
This is no longer the case - flakes require a test going from pass to fail _and_ fail to pass.

PR #1084, opened 2019-01-29
Simplify the flakes query

lukebjerring:
## Description
After #1057 we can now make "and" combinations for atoms, since the disjunction of the runs is done at the root instead of at the status equality (leaf). And after #1073 we don't need to be product-specific (since the flakes queries load a single product).

So, `(chrome:pass|chrome:ok) (chrome:fail|chrome:error|chrome:timeout|...)` simplifies to `(status:pass|status:ok) (status:!pass&status:!ok)`

### DIFF
--- a/webapp/components/wpt-insights.js
+++ b/webapp/components/wpt-insights.js
@@ -82,7 +82,7 @@ class Flakes extends ProductInfo(PolymerElement) {
       },
       query: {
         type: String,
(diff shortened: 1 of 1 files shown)


lukebjerring:
Also, ignoring `unknown` fixes https://github.com/web-platform-tests/wpt.fyi/issues/1076

Issue #1070, opened 2019-01-25
Expand results in "Visual comparison of the results" for checks


foolip:
### Check
A link to the GitHub Check run containing my issue:
https://github.com/web-platform-tests/wpt/pull/15067/checks?check_run_id=54652355

### Description
Following the link to https://wpt.fyi/results/?diff=&filter=ADC&run_id=6600560987865088&run_id=4704131331653632 it's still necessary to drill down a few steps before seeing anything interesting.

At the very least it would be good to link into the most shallow path where there's more than one thing, i.e. link into a directory if all affected tests are in that directory.

Somewhat more fancy would be to expand all results, but that wouldn't be usable if the number of affected tests is high.

PR #1085, opened 2019-01-29
Set diff results path to deepest shared dir

lukebjerring:
## Description
Resolves https://github.com/web-platform-tests/wpt.fyi/issues/1070

Finds the deepest shared path, and set that as the URL's path.

## Review
To test this change, we'll need an [example](https://github.com/web-platform-tests/wpt/pull/15067/checks?check_run_id=54652355) that does a poor job.
Once the enviroment is deployed, we can POST to /api/checks/[SHA]?product=chrome (example is b842c84)
That will recompute the links, with this PR's changes.

### DIFF
--- a/shared/util.go
+++ b/shared/util.go
@@ -307,6 +307,8 @@ func GetSharedPath(paths ...string) string {
 				if part == otherParts[i] {
 					continue
 				}
+				// Crop to the matching parts, append empty last-part
+				// so that we have a trailing slash.
 				parts = append(parts[:i], "")
 				break
 			}
--- a/shared/util.go
+++ b/shared/util.go
@@ -294,3 +294,23 @@ func CropString(s string, i int) string {
 	}
 	return s[:i]
 }
+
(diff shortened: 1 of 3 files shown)


shared/util.go
```diff
@@ -294,3 +294,23 @@ func CropString(s string, i int) string {
...
+	for _, path := range paths {
+		if parts == nil {
+			parts = strings.Split(path, "/")
+		} else {
+			otherParts := strings.Split(path, "/")
+			for i, part := range parts {
+				if part == otherParts[i] {
+					continue
+				}
+				parts = append(parts[:i], "")
```
foolip: Can you commend that this truncates parts to the common bits? Seeing append here makes it first look like something quite different, and interpreting the arguments is a bit of work. Is there not a slice method that would be more obvious?

lukebjerring:
Verified: https://github.com/web-platform-tests/wpt/pull/15067/checks?check_run_id=54651201
(vs its merge base now includes /shape-detection/).

PR #1086, opened 2019-01-29
Add Metadata structs to wpt.fyi/shared

lukebjerring:
## Description
Early stab at a strongly-typed interpretation of the wpt-metadata yml format.

### DIFF
--- a/shared/metadata_test.go
+++ b/shared/metadata_test.go
@@ -1,0 +1,28 @@
+// +build small
+
+// Copyright 2019 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import (
+	"testing"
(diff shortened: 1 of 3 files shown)


shared/metadata_test.go
```diff
@@ -0,0 +1,28 @@
...
+// Copyright 2019 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import (
+	"testing"
+
+	"github.com/go-yaml/yaml"
```
Hexcles: Which version does this give us? The project README instructs us to use `gopkg.in/yaml.v2`, so I wonder if there's any difference.

lukebjerring: Yah it's v2; The v2 git branch is set as the default for the package I think.

PR #1087, opened 2019-01-29
Always log the event sha for check_suite/run events

lukebjerring:
## Description
Ensures that we can find logs for a SHA[:7] by text search in appengine, even for failed requests.


lukebjerring:
Yes, it will.

PR #1088, opened 2019-01-29
Fix simple (empty) queries

lukebjerring:
## Description
#1057 broke the check for a "simple" query, causing simple queries to bypass the summaries, causing a regression in performance (particularly for uncommon data), which in turn is timing out some integration tests.

### DIFF
--- a/api/query/search.go
+++ b/api/query/search.go
@@ -126,8 +126,17 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	simpleQ, ok := rq.AbstractQuery.(TestNamePattern)
-	if !ok {
+	// Check if the query is a simple (empty, or test name only) query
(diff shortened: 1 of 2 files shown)


api/query/search.go
```diff
@@ -126,8 +126,17 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	simpleQ, ok := rq.AbstractQuery.(TestNamePattern)
-	if !ok {
+	// Check if the query is a simple (empty, or test name only) query
+	var simpleQ TestNamePattern
+	var isSimpleQ bool
+	{
```
Hexcles: Why the scoping?

lukebjerring: No major reason; avoids variable pollution.

Hexcles: Doesn't seem necessary to me, but no strong opinion.

PR #1094, opened 2019-01-30
Refactor Datastore interface

Hexcles:
This PR refactors the Datastore abstraction layer (again).

The main purpose is to extract caching and querying specific to TestRun out so that the `Datastore` interface is more generic and can be used to read/write other entities.

* Caching: some entities, e.g. the upcoming screenshot hash, are mutable and shouldn't be cache. We should not bake caching into the Datastore interface. The PR splits `AppEngineDatastore` into `AppEngineDatastore` (uncached) and `AppEngineCachedDatastore`, and the call sites get to choose which one to use.
* Querying: currently we have some querying methods for TestRun in the `Datastore` interface and some not, which bloats the interface unnecessarily and makes it hard to mock TestRun queries in tests. This PR moves all TestRun-related queries into a simple interface so that higher-level code can choose to mock the query methods instead of the low-level Datastore queries. Later the screenshot PR will introduce a similar query interface for screenshot hashes.

Some drive-by fixes:
* The `LoadTestRunsByKeys` doesn't need to be different when caching is used. The method effectively always uses `GetMulti` and the caching should happen in `GetMulti` (similar to the caching in `Get`).
* Abstract the iterator sentinels (`Done`) as they are different depending on the libraries.


api/diff.go
```diff
@@ -52,7 +52,7 @@ func handleAPIDiffGet(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	store := shared.NewAppEngineDatastore(ctx)
+	store := shared.NewAppEngineCachedDatastore(ctx)
```
lukebjerring: I'd prefer `NewAppEngineDatastore` take a `cached` boolean param than 2 separate functions. Copy pasta is a little harder that way (all call sites have the param, rather than not knowing about an overload method).

lukebjerring:
LGTM, but can you collapse the `NewAppEngine{,Cached}Datastore` signatures into a `cached` param on the method?

api/diff.go
Hexcles: Good point! will do.

Hexcles:
Merged. If anything goes wrong, please feel free to revert.

PR #1101, opened 2019-01-31
Plumb param for including subtests in searchcache response

lukebjerring:
## Description
Checks for `?subtests` param and plumbs the option to include subtests in the response.

### DIFF
--- a/api/query/cache/index/aggregator.go
+++ b/api/query/cache/index/aggregator.go
@@ -5,6 +5,7 @@
 package index
 
 import (
+	mapset "github.com/deckarep/golang-set"
 	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
@@ -32,7 +33,13 @@ func (a *indexAggregator) Add(t TestID) error {
 			return err
 		}
 
-		r = query.SearchResult{Test: name, LegacyStatus: nil}
+		r = query.SearchResult{
+			Test:         name,
+			LegacyStatus: nil,
+		}
+		if a.includeSubtests {
+			r.Subtests = mapset.NewSet()
+		}
 	}
 
 	rus := r.LegacyStatus
--- a/api/query/cache/index/aggregator.go
+++ b/api/query/cache/index/aggregator.go
@@ -54,7 +54,7 @@ func (a *indexAggregator) Add(t TestID) error {
 
 		if a.includeSubtests {
 			if _, name, err := ts.GetName(t); err == nil && name != nil {
-				r.Subtests = append(r.Subtests, *name)
+				r.Subtests.Add(*name)
 			}
 		}
 	}
--- a/api/query/cache/index/aggregator.go
+++ b/api/query/cache/index/aggregator.go
@@ -72,10 +72,11 @@ func (a *indexAggregator) Done() []query.SearchResult {
 	return res
 }
 
-func newIndexAggregator(idx index, rus []RunID) aggregator {
+func newIndexAggregator(idx index, rus []RunID, opts query.AggregationOpts) aggregator {
 	return &indexAggregator{
-		index: idx,
-		rus:   rus,
-		agg:   make(map[uint64]query.SearchResult),
+		index:           idx,
+		rus:             rus,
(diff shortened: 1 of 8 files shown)


lukebjerring:
Example query: 
curl -X POST --data '{"run_ids":[5844727341514752,6163721139257344,6582181900058624,4697720035999744],"query":{"exists":[{"and":[{"status":"TIMEOUT"},{"pattern":"/2dcontext/imagebitmap/canvas-createImageBitmap-video-resize.html"}]}]}}' https://subtests-dot-searchcache-dot-wptdashboard-staging.appspot.com/api/search/cache?subtests

api/query/cache/index/aggregator.go
```diff
@@ -72,10 +72,11 @@ func (a *indexAggregator) Done() []query.SearchResult {
 	return res
 }
 
-func newIndexAggregator(idx index, rus []RunID) aggregator {
+func newIndexAggregator(idx index, rus []RunID, opts query.AggregationOpts) aggregator {
 	return &indexAggregator{
```
KyleJu: Syntax question - looks like &indexAggregator returns the reference to this object, which is a pointer. But how come the return type of the function is aggregator instead of *aggregator?

api/query/cache/service/main.go
```diff
@@ -141,13 +141,19 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	// resident in `idx`. In the unlikely event that a run in `ids`/`runs` is no
 	// longer in `idx`, `idx.Bind()` below will return an error.
 	q := cq.PrepareUserQuery(ids, rq.AbstractQuery.BindToRuns(runs...))
+
+	// Configure format, from request params.
+	_, subtests := r.URL.Query()["subtests"]
+	opts := query.AggregationOpts{
+		IncludeSubtests: subtests,
+	}
```
KyleJu: this is a boolean value?

api/query/cache/index/aggregator.go
```diff
@@ -54,7 +61,7 @@ func (a *indexAggregator) Add(t TestID) error {
 
 		if a.includeSubtests {
 			if _, name, err := ts.GetName(t); err == nil && name != nil {
-				r.Subtests = append(r.Subtests, *name)
+				r.Subtests.Add(*name)
```
KyleJu: why do we need a set to store pointers? to make sure no pointers that point to the same memory address?

api/query/search.go
```diff
@@ -46,9 +47,24 @@ type SearchResult struct {
 	LegacyStatus []LegacySearchRunResult `json:"legacy_status"`
 
 	// Subtests (names) which are included in the LegacyStatus summary.
+	Subtests mapset.Set `json:"-"`
+}
+
+type searchResultNoCustomMarshalling SearchResult
+type marshallableSearchResult struct {
+	searchResultNoCustomMarshalling
 	Subtests []string `json:"subtests,omitempty"`
```
KyleJu: Subtests []string `json:"subtests,omitempty" - just a quick question about this syntax - is Subtests a variable being declared here? if so what is `json:"subtests,omitempty" in the end?

api/query/cache/service/main.go
lukebjerring: https://blog.golang.org/go-maps-in-action

> A two-value assignment tests for the existence of a key

`_` is the (ignored) value, `subtests` is whether the value was in the map.

api/query/cache/index/aggregator.go
lukebjerring: `*name` is dereferencing the pointer`s value (string).
In most cases, the same subtests exist for all of the runs. We're taking the actual string values in the set, so that we don't have duplication.

api/query/search.go
lukebjerring: https://gobyexample.com/json

On the struct, `Subtests` is the field name. the `json:"field_name"` tag renames the field's name when serializing (marshalling) as JSON. `,omitempty` is a helper that drops the field when its value is empty.

lukebjerring:
This approach (using a set, and having custom marshalling) isn't as clean as the follow-up PR which only appends to the array once per test, avoiding the problem.

Issue #1048, opened 2019-01-21
Filter not applied on individual test pages


jgraham:
On a browser comparison page like:

https://staging.wpt.fyi/results/dom?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned&q=%21firefox%3Apass+%21firefox%3Aok+%28chrome%3Apass%7Cchrome%3Aok%29+%28safari%3Apass%7Csafari%3Aok%29+dom

one gets directory-level results for tests that pass in one browser but fail in others e.g. currently on that link you get 0/3 for firefox on `historical.html`, reflecting the fact that 3 subtests match the filter and by construction none of them pass in Firefox. But if you click on the individual test link you suddenly see all the subtests in that file, not just the ones that matched the applied filter, which is confusing and kind of defeats the point of the feature in cases where files have a large number of subtests.

lukebjerring:
The `/api/search` request, handled by `searchcache`, is designed to return aggregated numbers.
We'll need to expand the functionality of the search to return included subtests, either via a query param, or whenever the path is a test-file (the latter is simpler, so probably that).

lukebjerring:
Duplicate of and old issue https://github.com/web-platform-tests/wpt.fyi/issues/81
Back then, we worked around it by just not having a search box, but we can now actually search subtests via the searchcache, so revisiting it to provide an actual fix.

PR #1103, opened 2019-01-31
Filter the matching subtests for individual tests pages

lukebjerring:
## Description
Checks for ?subtests param and plumbs the option to include subtests in the response.
Then, leverages the functionality from the client in order to filter the results on an individual test file page.

Fix for #81 (and duplicate #1048)

### DIFF
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -93,7 +93,7 @@ class TestFileResults extends WPTFlags(LoadingState(TestRunsUIQuery(
       this.fetchTestFile(path, testRuns),
     ]);
 
-    if (searchResults) {
+    if (resultsTable && searchResults) {
       const test = searchResults.results.find(r => r.test === path);
       if (test) {
         let subtests = new Set(test.subtests);
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -111,17 +111,12 @@ class TestFileResults extends WPTFlags(LoadingState(TestRunsUIQuery(
     }
 
     // Combine the query with " and [path]".
-    const q = Object.assign({}, structuredSearch);
-    if (q.exists) {
-      q.exists = q.exists.map(node => {
-        return {
-          and: [
-            node,
-            {pattern: path}
-          ]
-        };
-      });
-    }
+    let q = {
+      and: [
+        {pattern: path},
+        structuredSearch,
+      ]
+    };
 
     let url = new URL('/api/search', window.location);
     url.searchParams.set('subtests', '');
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -96,8 +96,8 @@ class TestFileResults extends WPTFlags(LoadingState(TestRunsUIQuery(
     if (resultsTable && searchResults) {
       const test = searchResults.results.find(r => r.test === path);
       if (test) {
-        let subtests = new Set(test.subtests);
-        let [first, ...others] = resultsTable;
+        const subtests = new Set(test.subtests);
+        const [first, ...others] = resultsTable;
         const matches = others.filter(t => subtests.has(t.name));
         resultsTable = [first, ...matches];
       }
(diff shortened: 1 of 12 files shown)


foolip:
Starting at https://subtests-2-dot-wptdashboard-staging.appspot.com/results/dom/historical.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned&q=chrome%3Afail, I wonder if there should be any UI to reveal all of the tests? Removing the search and hitting enter works, but is something else planned?

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+    // Run a search query, including subtests, as well as fetching the results file.
+    let [searchResults, resultsTable] = await Promise.all([
+      this.fetchSearchResults(path, testRuns, structuredSearch),
+      this.fetchTestFile(path, testRuns),
+    ]);
+
+    if (resultsTable && searchResults) {
+      const test = searchResults.results.find(r => r.test === path);
+      if (test) {
+        let subtests = new Set(test.subtests);
```
foolip:
```suggestion
        const subtests = new Set(test.subtests);
```

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+        let [first, ...others] = resultsTable;
```
foolip:
```suggestion
        const [first, ...others] = resultsTable;
```

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+        const matches = others.filter(t => subtests.has(t.name));
```
foolip: Just to confirm my understanding of this, it's the server that does the filtering, then we fetch the full results and filter that to match?

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+    this.resultsTable = resultsTable;
+  }
+
+  async fetchSearchResults(path, testRuns, structuredSearch) {
+    if (!testRuns || !testRuns.length || !this.structuredQueries || !structuredSearch) {
+      return;
+    }
+
+    // Combine the query with " and [path]".
+    let q = {
```
foolip:
```suggestion
    const q = {
```

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+      and: [
+        {pattern: path},
+        structuredSearch,
+      ]
+    };
+
+    let url = new URL('/api/search', window.location);
```
foolip:
```suggestion
    const url = new URL('/api/search', window.location);
```

lukebjerring:
No plans. The issue of "the view is only showing results that have been searched for" is quite separate to this PR - all views that have a search applied are privy to the same "remove this search" need.

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+
+  async fetchSearchResults(path, testRuns, structuredSearch) {
+    if (!testRuns || !testRuns.length || !this.structuredQueries || !structuredSearch) {
+      return;
+    }
+
+    // Combine the query with " and [path]".
+    let q = {
+      and: [
+        {pattern: path},
```
foolip: Will something terrible happen if this matches more than one test? It could happen with variants, where both `/some/test.html?a` and `/some/test.html?aa` could exist.

webapp/components/test-file-results.js
```diff
@@ -80,7 +83,65 @@ class TestFileResults extends TestRunsUIQuery(
...
+      body: JSON.stringify({
+        run_ids: testRuns.map(r => r.id),
+        query: q,
+      }),
+    };
+    return await this.retry(
+      async() => {
+        const r = await window.fetch(url, fetchOpts);
+        if (!r.ok) {
+          if (fetchOpts.method === 'POST' && r.status === 422) {
```
foolip: Why the `fetchOpts.method === 'POST'` part of the condition? Nothing can have changed it at this point. Should it have been `r.method === 'POST'`?

lukebjerring: Copy-paste from a code branch that actually has both GET/POST.
For interop, we can't run a server-side search with GET.

PR #1105, opened 2019-01-31
Add permissive CORS for /api/search

lukebjerring:
## Description
Mozilla would like to use it externally.

### DIFF
--- a/api/query/routes.go
+++ b/api/query/routes.go
@@ -9,7 +9,10 @@ import "github.com/web-platform-tests/wpt.fyi/shared"
 // RegisterRoutes binds query API handlers to URL routes.
 func RegisterRoutes() {
 	// API endpoint for searching results over given runs.
-	shared.AddRoute("/api/search", "api-search", apiSearchHandler)
+	shared.AddRoute(
+		"/api/search",
+		"api-search",
+		shared.WrapPermissiveCORS(apiSearchHandler))
 	// API endpoint for search autocomplete.
 	shared.AddRoute("/api/autocomplete", "api-autocomplete", apiAutocompleteHandler)
 }
--- a/shared/routing.go
+++ b/shared/routing.go
@@ -41,11 +41,15 @@ func WrapHSTS(h http.Handler) http.HandlerFunc {
 
 // WrapPermissiveCORS wraps the given handler func in one that sets an
 // all-permissive CORS header on the response.
-func WrapPermissiveCORS(h http.HandlerFunc) http.HandlerFunc {
-	cors := handlers.CORS(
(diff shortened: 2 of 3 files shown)


jgraham:
I tried updating my app to point at the staging server here, and I"m still getting a 400 for the options request; you should be able to see the failure at https://jgraham.github.io/wptdash/

lukebjerring:
> Request Method: OPTIONS

Looks like we don't implement the OPTIONS http methods (just POST and GET). I'll open a separate issue for that.

If we did, it'd return.. the information I am giving you manually, I guess.

jgraham:
The `OPTIONS` request has to return the right `Access-Control-*` headers so that the browser knows it's permitted to make the real `POST` request, c.f. https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request

jgraham:
Seems like the most straightforward option here is just to add a route matching all options requests when configuring the router somewhere like https://github.com/web-platform-tests/wpt.fyi/blob/e604b98f451b6c7c01f2ccfb2a2d7dde029c34a2/shared/routing.go#L19

```
router.Methods("OPTIONS").handlerFunc(
  func(w http.ResponseWriter, r *http.Request) {
     w.Header().Add("Access-Control-Allow-Origin", "*")
     w.Header().Add("Access-Control-Allow-Methods", "POST,GET,OPTIONS")
     // whatever's needed here to actually send the response
  })
```

Or since you are already using mux, there's existing gorilla middleware to handle this: https://github.com/gorilla/handlers/blob/master/cors.go

lukebjerring:
Driveby fixes https://github.com/web-platform-tests/wpt.fyi/issues/1106 too

lukebjerring:
OK, allowing content-type seemed to stop the 403. I don't actually do anything with it yet.

Note that your query shows much less results than [this equivalent](https://cors-dot-wptdashboard-staging.appspot.com/results/?sha=e1d9e660a5b1eb9ee26193f2c71d26673f764e19&label=master&label=experimental&product=chrome&product=firefox&product=safari&q=%28firefox%3A%21pass+%26+firefox%3A%21ok%29+%28chrome%3Aok%7Cchrome%3Apass%29+%28safari%3Aok%7Csafari%3Apass%29) - I think the search should be `(firefox:!pass & firefox:!ok) (chrome:ok|chrome:pass) (safari:ok|safari:pass)`

(which expands as 

    {"query":{"exists":[{"and":[{"product":"firefox","status":{"not":"PASS"}},{"product":"firefox","status":{"not":"OK"}}]},{"or":[{"product":"chrome","status":"OK"},{"product":"chrome","status":"PASS"}]},{"or":[{"product":"safari","status":"OK"},{"product":"safari","status":"PASS"}]}]}

PR #1107, opened 2019-01-31
Update the Azure Pipelines artifact name matching

foolip:
## Description
Follows https://github.com/web-platform-tests/wpt/pull/15110, and also allows for the old names.

## Review Information
Can this be tested in webhook_test.go?

### DIFF
--- a/api/azure/webhook.go
+++ b/api/azure/webhook.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"io/ioutil"
 	"net/url"
+	"regexp"
 	"strconv"
 
 	mapset "github.com/deckarep/golang-set"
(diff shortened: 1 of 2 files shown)


api/azure/webhook.go
```diff
@@ -17,6 +18,15 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+// Labels for runs from Azure Pipelines are determined from the artifact names.
+// For master runs, artifact name may be either just "results" or something
+// like "safari-results".
+var (
+	masterRegex = regexp.MustCompile(`[^-]results$`)
```
lukebjerring: `[^-]` is "not a dash", right? so this would match aresults but not safari-preview-results ?

In either case, add a unit test to webhook_test.go that covers the positive cases :)

foolip: D'oh! Can `processAzureBuild` as a whole be tested, or should I write tests for the regexes?

lukebjerring:
Just the regexps is fine/easier.

foolip:
Tested!

PR #1096, opened 2019-01-31
Increase the number of processors

Hexcles:
TBR

PR #1109, opened 2019-02-01
Increase the maximum rate of the results processor

foolip:
Follow up to https://github.com/web-platform-tests/wpt.fyi/pull/1096.


foolip:
Actually, I was looking at prod and Maximum Concurrent is still 2.

Still, in staging, there are now 10 tasks in the queue and only 2 running, presumably for this reason.

PR #1111, opened 2019-02-01
Fix loop variable pointer reuse

lukebjerring:
## Description
Fixes a bug where only one (repeated) subtest name is ever returned for any test.


lukebjerring:
TBR'ing.

Issue #655, opened 2018-10-15
Interop query against a custom set of browsers.


mikewest:
https://staging.wpt.fyi/results/content-security-policy?aligned&product=chrome%5Bexperimental%5D&product=edge%5Bstable%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D shows experimental versions of Chrome, Firefox, and Safari, and a stable version of Edge. Clicking "Interoperability" gives a "failed to fetch interop data" error (https://staging.wpt.fyi/interop/content-security-policy?aligned&product=chrome%5Bexperimental%5D&product=edge%5Bstable%5D&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D).

foolip:
@lukebjerring @mdittmer we can do this after the Spanner work, right? (Which is the best tracking issue for that?)

lukebjerring:
This is blocked on #931

foolip:
@lukebjerring do you have a URL for what @mikewest was looking to achieve?

lukebjerring:
The link that @mikewest provided (as a failure case) is now functional.

For anyone who stumbles here looking how to check interop for a custom set of browsers:
- Visit the /interop page (a.k.a. the Interoperability tab)
- Hit the 'EDIT' button on the query builder banner
- Select the products you want to include
- Hit submit

PR #1112, opened 2019-02-01
Add a query param for including interop summary from /api/search

lukebjerring:
## Description
This will allow us to deprecate the results-analysis pipeline, and compute interop on the fly, including filtering to particular searches, e.g. what's the interoperability of things that Chrome is passing?
And, including custom product sets (see #655)

This PR builds off of #1103, so if you're reviewing, do so with 5599a4e...7a36891

## Review
- Grab a specific view that we have precomputed interop for
- Extract the run_id, and POST to /api/search?interop
  - `{"run_ids":[6347092704362496,4879930668089344,5070059928027136,6495293554032640],"query":{"exists":[{"pattern":"/2dcontext/"}]}}`

Hopefully, see some interop scores.

### DIFF
--- a/api/query/cache/index/aggregator.go
+++ b/api/query/cache/index/aggregator.go
@@ -17,7 +17,7 @@ type aggregator interface {
 type indexAggregator struct {
 	index
 
-	rus             []RunID
+	runIDs          []RunID
 	agg             map[uint64]query.SearchResult
 	includeSubtests bool
 	interopFormat   bool
@@ -42,12 +42,11 @@ func (a *indexAggregator) Add(t TestID) error {
 
 	if a.interopFormat {
 		if r.Interop == nil {
-			r.Interop = make([]int, len(a.rus)+1)
+			r.Interop = make([]int, len(a.runIDs)+1)
 		}
 		passing := 0
-		for _, ru := range a.rus {
-			res := shared.TestStatus(a.runResults[ru].GetResult(t))
-			// Only include tests with non-UNKNOWN status for this run's total.
+		for _, id := range a.runIDs {
+			res := shared.TestStatus(a.runResults[id].GetResult(t))
 			if res.IsPassOrOK() {
 				passing++
 			}
@@ -57,10 +56,10 @@ func (a *indexAggregator) Add(t TestID) error {
 
 	rus := r.LegacyStatus
 	if rus == nil {
-		rus = make([]query.LegacySearchRunResult, len(a.rus))
+		rus = make([]query.LegacySearchRunResult, len(a.runIDs))
 	}
 
-	for i, ru := range a.rus {
+	for i, ru := range a.runIDs {
 		res := shared.TestStatus(a.runResults[ru].GetResult(t))
 		// TODO: Switch to a consistent value for Total across all runs.
 		//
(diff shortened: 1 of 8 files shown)


api/query/cache/index/aggregator.go
```diff
@@ -32,7 +33,26 @@ func (a *indexAggregator) Add(t TestID) error {
...
+	}
+
+	if a.interopFormat {
+		if r.Interop == nil {
+			r.Interop = make([]int, len(a.rus)+1)
+		}
+		passing := 0
+		for _, ru := range a.rus {
+			res := shared.TestStatus(a.runResults[ru].GetResult(t))
+			// Only include tests with non-UNKNOWN status for this run's total.
```
foolip: The commend doesn't match what the code seems to be doing, as there are many statuses other than PASS or OK. Which is correct?

api/query/cache/index/aggregator.go
```diff
@@ -32,7 +33,26 @@ func (a *indexAggregator) Add(t TestID) error {
...
+			Interop:      nil,
+		}
+	}
+
+	if a.interopFormat {
+		if r.Interop == nil {
+			r.Interop = make([]int, len(a.rus)+1)
+		}
+		passing := 0
+		for _, ru := range a.rus {
```
foolip: I see that the naming is not new, but in this context even knowing that the type is `[]RunID` I can't guess what `rus` or `ru` means. Run... unit? If it can be renamed to something more transparent, doing that here on or followup would be nice.

api/query/cache/monitor/monitor_test.go
```diff
@@ -64,6 +64,7 @@ func TestDoubleStart(t *testing.T) {
 		defer wg.Done()
 		err1 = mon.Start()
 	}()
+	time.Sleep(time.Microsecond * 100)
```
foolip: Debugging left in? If not, can you add a comment?

api/query/concrete_query.go
```diff
@@ -8,6 +8,13 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+// AggregationOpts are options for the aggregation format used when collecting
+// the results.
+type AggregationOpts struct {
+	IncludeSubtests bool
```
foolip: Can you say something about whether these are mutually exclusive or what they do when combined? Is there some documentation that should be updated too?

api/query/cache/index/aggregator.go
lukebjerring: The code :)

api/query/cache/index/aggregator.go
lukebjerring: Totally agree - no idea what ru was.

api/query/cache/monitor/monitor_test.go
lukebjerring: No; fixes a flaky test.

api/query/concrete_query.go
lukebjerring: I'm not going to document until we have all of the pieces in place, end-to-end. Don't want to have to change the documentation when we change the impl.

PR #1113, opened 2019-02-04
Add client-side feature for fetching interop from searchcache

### LINKED ISSUE #655: Interop query against a custom set of browsers.

lukebjerring:
## Description
- Adds code to map (refactor) the current interop format to the format returned by the server
- Adds a feature flag for fetching from the server, using a structured search POST request

Fixes #655 

## Review
Note: Feature for `searchCacheInterop` is already globally enabled in staging.
Change to a custom set of browser (e.g. remove Edge) and visit the interop tab.

### DIFF
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -345,8 +345,8 @@ class WPTInterop extends WPTColors(WPTFlags(SelfNavigation(LoadingState(
         5,
         5000
       ).then(
-        json => {
-          this.searchResults = json.results;
+        results => {
+          this.searchResults = results;
           this.refreshDisplayedNodes();
         },
         (e) => {
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -161,15 +161,15 @@ class WPTInterop extends WPTColors(WPTFlags(SelfNavigation(LoadingState(
     <section class="runs">
       <table>
         <thead>
-        <tr>
-          <template is="dom-repeat" items="{{ passRateMetadata.test_runs }}" as="testRun">
-            <th>
-              <test-run test-run="[[testRun]]"></test-run>
-            </th>
-          </template>
-        </tr>
-
-      </thead></table>
+          <tr>
+            <template is="dom-repeat" items="[[testRuns]]" as="testRun">
+              <th>
+                <test-run test-run="[[testRun]]"></test-run>
+              </th>
+            </template>
+          </tr>
+        </thead>
+      </table>
     </section>
 
     <table>
@@ -205,8 +205,7 @@ class WPTInterop extends WPTColors(WPTFlags(SelfNavigation(LoadingState(
   </template>
 
   <template is="dom-if" if="[[ pathIsATestFile ]]">
-    <test-file-results test-runs="[[passRateMetadata.test_runs]]" path="[[path]]">
-    </test-file-results>
+    <test-file-results test-runs="[[testRuns]]" path="[[path]]"></test-file-results>
   </template>
 
   <paper-toast id="runsNotInCache" duration="5000" text="One or more of the runs requested is currently being loaded into the cache. Trying again..."></paper-toast>
@@ -221,10 +220,8 @@ class WPTInterop extends WPTColors(WPTFlags(SelfNavigation(LoadingState(
     return {
       passRateMetadata: Object,
       testRuns: Array,
-      allMetrics: Object,
-      searchResults: {
-        type: Object,
-      },
+      precomputedInterop: Object,
+      searchResults: Object,
       displayedTests: {
         type: Array,
         computed: 'computeDisplayedTests(path, searchResults)'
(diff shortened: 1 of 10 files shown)


webapp/components/interop.js
```diff
@@ -258,39 +260,108 @@ class WPTInterop extends QueryBuilder(
       this.handleSearchAutocomplete(e.detail.path);
     };
     this.onLoadingComplete = () => {
-      // passRateMetadata contains the url for the JSON blob of allMetrics;
+      // passRateMetadata contains the url for the JSON blob of precomputedInterop;
       // both fetches need to succeed + parse.
```
foolip: This comment seems out of date now, the following code only considers one thing.

webapp/components/interop.js
```diff
@@ -258,39 +260,108 @@ class WPTInterop extends QueryBuilder(
...
           }
-          this.passRateMetadata = await r.json();
-          this.allMetrics = await fetch(this.passRateMetadata.url).then(r => r.json());
+          const metadata = await r.json();
+          this.passRateMetadata = metadata;
+          this.testRuns = metadata && metadata.test_runs;
+          this.precomputedInterop = await fetch(this.passRateMetadata.url).then(r => r.json());
+        })
+        .then(() => {
+          if (this.search) {
```
foolip: Is it significant that this is not in the above block. Given that async/await is used there, just folding this into the above block should work and would be easier to read.

webapp/components/interop.js
```diff
@@ -258,39 +260,108 @@ class WPTInterop extends QueryBuilder(
...
+          }
+        })
+    );
+  }
+
+  fetchSearchCacheInterop() {
+    this.load(
+      Promise.resolve(this.testRuns || this.loadRuns())
+        .then(runs => {
+          if (!runs) {
```
foolip: Looks like this is the only thing `runs` is used for? Can you either use `runs` below as well or `this.testRuns` in both places so that it's clear they're the same?

webapp/components/interop.js
```diff
@@ -258,39 +260,108 @@ class WPTInterop extends QueryBuilder(
...
+
+          // Fetch search results and refresh display nodes. If fetch error is HTTP'
+          // 422, expect backend to attempt write-on-read of missing data. In such
+          // cases, retry fetch up to 5 times with 5000ms waits in between.
+          const toast = this.shadowRoot.querySelector('#runsNotInCache');
+          return this.retry(
+            async() => {
+              const r = await window.fetch(url, fetchOpts);
+              if (!r.ok) {
+                if (fetchOpts.method === 'POST' && r.status === 422) {
```
foolip: Same question about `fetchOpts.method` here as in https://github.com/web-platform-tests/wpt.fyi/pull/1103#discussion_r253500441. This part of the condition will always be true.

webapp/components/interop.js
```diff
@@ -319,32 +390,38 @@ class WPTInterop extends QueryBuilder(
...
-    let fileMetrics = [];
-    for (const metric of allMetrics.data) {
-      if (this.computePathIsATestFile(metric.dir)) {
-        fileMetrics.push(metric);
+  precomputedInteropLoaded(precomputedInterop) {
+    const searchResults = {
+      runs: this.testRuns,
+      results: [],
+    };
+    if (precomputedInterop) {
```
foolip: Can this be false, and if so is an else-branch not needed?

webapp/components/interop.js
```diff
@@ -319,32 +390,38 @@ class WPTInterop extends QueryBuilder(
...
+      for (const metric of precomputedInterop.data) {
+        if (this.computePathIsATestFile(metric.dir)) {
```
foolip: This line shows that calling it `dir` in the format is a bit odd. Is it to match the old format, or can it be called `path` instead?

webapp/components/interop.js
```diff
@@ -319,32 +390,38 @@ class WPTInterop extends QueryBuilder(
...
+          searchResults.results.push({
+            test: metric.dir,
+            interop: metric.pass_rates,
+          });
+        }
+      }
+      const q = this.search;
+      if (q  && q.length) {
+        searchResults.results = searchResults.results.filter(t => t.test.toLowerCase().includes(q));
```
foolip: Looks like client-side filtering. Isn't there an existing single place where this is handled, if we change the rules for how it's matched, or move it back to the server?

webapp/components/interop.js
lukebjerring: Done.

webapp/components/interop.js
lukebjerring: Done.

webapp/components/interop.js
lukebjerring: Done.

webapp/components/interop.js
lukebjerring: Done.

webapp/components/interop.js
lukebjerring: Tweaked.

webapp/components/interop.js
lukebjerring: dir is the existing precomputed interop's field name, because we (redundantly) group by each directory as well as test-file. searchcache's results calls it `path`.

webapp/components/interop.js
lukebjerring: Interop wasn't moved to a server-side search when Results was, so this is legacy code. Note that I haven't authored it in this PR, simply moved it.

lukebjerring:
The code is (partially) integration tested, as well as some basic web components tests.
I'm hoping to refactor the view(s) to be more consistent / consolidated, since changing tab really only means changing the display table and search params now.

PR #1116, opened 2019-02-05
Add a minimal configuration file for Azure Pipelines

foolip:
@lukebjerring as suggested, here's a skeleton of how we could use Azure Pipelines to bunch of steps to run every week.

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,0 +1,13 @@
+# This is the configuration file for Azure Pipelines.
+# Documentation to help understand this setup:
+# https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema
+# https://docs.microsoft.com/en-us/azure/devops/pipelines/process/conditions
+
+jobs:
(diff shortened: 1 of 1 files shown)


foolip:
In order to actually register the pipeline in Azure DevOps I had to let it create a PR: https://github.com/web-platform-tests/wpt.fyi/pull/1117

(There's a way around that, but it's elaborate and I forgot.)

Hexcles:
I'm out of loop: any reason we'd like to do the release on Azure instead of Travis?

foolip:
Because it's possible to schedule builds at specific times. With Travis it's the time you set it up the first time.

foolip:
One can of course make sure to be around at the right time once and then it's fine, but @lukebjerring liked this idea. I guess it'll duplicate some setup of gcloud SDK, etc.?

Hexcles:
Oh, that's sad... Hopefully our Dockerized setup would just work without a hitch on Azure.

PR #1118, opened 2019-02-05
Start running mypy on results-processor

gsnedders:
This adds mypy to results-processor. This doesn't actually quite work yet, though. `wptreport.py` almost passes; we could maybe just enable it for that first.

### DIFF
--- a/results-processor/.gitignore
+++ b/results-processor/.gitignore
@@ -1,3 +1,4 @@
 env/
 .tox/
 __pycache__/
+.mypy_cache/
--- a/results-processor/config.py
+++ b/results-processor/config.py
@@ -1,25 +1,25 @@
 import os
 
 
-def _is_prod():
+def _is_prod() -> bool:
     return os.getenv('GOOGLE_CLOUD_PROJECT') == 'wptdashboard'
 
 
-def raw_results_bucket():
+def raw_results_bucket() -> str:
     """Returns the bucket name for storing raw, full results."""
     if _is_prod():
         return 'wptd-results'
     return 'wptd-results-staging'
 
 
-def results_bucket():
+def results_bucket() -> str:
     """Returns the bucket name for storing split results."""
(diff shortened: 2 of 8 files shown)


Hexcles:
This is great! `wptreport.py` is meant to be self-contained, well-documented and thoroughly tested, so I also think we should start checking types there. Hard-to-test glue code is kept out of this module and can be dealt with later (I was planning to test the other modules with mocks, and I can tackle some typing issues then).

The only concern is that, due to various environmental complications, we aim at Python>=3.5, which means we can't use [PEP526](https://www.python.org/dev/peps/pep-0526/) quite yet.

Hexcles:
To elaborate, the complications are mostly because Debian's py3 still defaults to 3.5 (https://wiki.debian.org/Python). I could be convinced to explicitly require Python 3.6 for this repo, but PEP526 alone perhaps isn't enough.

Hexcles:
Good news: Python 3.6 is coming. Once #1235 lands, I'll resurrect this PR. (I myself was bitten by type issues a few times yesterday, so it's nice to have some checks...)

gsnedders:
Can you get it running with `--strict`? How much effort would that be? (I'd hardcode the various current options `--strict` enables in the config file.)

Hexcles:
mypy 0.670 currently has:

```
Strict mode; enables the following flags:
--warn-unused-configs,
--disallow-subclassing-any,
--disallow-any-generics,
--disallow-untyped-calls,
--disallow-untyped-defs,
--disallow-incomplete-defs,
--check-untyped-defs,
--disallow-untyped-decorators,
--no-implicit-optional,
--warn-redundant-casts,
--warn-unused-ignores,
--warn-return-any
```

I'll enable a few more of them that don't require much effort, but probably not all of them right now.

gsnedders:
That looks great!

PR #1119, opened 2019-02-05
Normalize product names to handle underscore delimited variants

gsnedders:
See https://github.com/web-platform-tests/wpt/issues/15174#issuecomment-459961243
for the origins of this feature.

This also changes labels to be stored as a set while being processed and only normalised into a sorted list at the last moment.

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -460,7 +460,7 @@ def prepare_labels(report, labels_str, uploader):
         uploader: The name of the uploader.
 
     Returns:
-        A sorted list of unique strings.
+        A set of strings.
     """
     labels = set()
     labels.add(report.run_info['product'])
@@ -481,7 +481,27 @@ def prepare_labels(report, labels_str, uploader):
     # Remove any empty labels.
     if '' in labels:
         labels.remove('')
-    return sorted(labels)
+    return labels
+
+
+def normalize_product(report):
+    """Normalizes the product identifier.
+
+    Computes what labels need to be added while normalizing the product.
+
+    Args:
+        report: A WPTReport
+
+    Returns:
+       A set of strings.
+    """
+    product = report.run_info['product']
+    if "_" in product:
+        tokens = product.split("_")
+        report.run_info['product'] = tokens[0]
+        return set(tokens[1:])
+    else:
+        return set()
 
 
 def create_test_run(report, labels_str, uploader, secret,
@@ -509,10 +529,12 @@ def create_test_run(report, labels_str, uploader, secret,
     labels = prepare_labels(report, labels_str, uploader)
     assert len(labels) > 0
 
+    labels |= normalize_product(report)
+
     payload = report.test_run_metadata
     payload['results_url'] = results_url
     payload['raw_results_url'] = raw_results_url
-    payload['labels'] = labels
+    payload['labels'] = sorted(labels)
 
     response = requests.post(
         callback_url,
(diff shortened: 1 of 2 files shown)


results-processor/wptreport.py
```diff
@@ -481,7 +481,27 @@ def prepare_labels(report, labels_str, uploader):
...
+        report: A WPTReport
+
+    Returns:
+       A set of strings.
+    """
+    product = report.run_info['product']
+    if "_" in product:
+        tokens = product.split("_")
+        report.run_info['product'] = tokens[0]
+        return set(tokens[1:])
```
lukebjerring: I'm wondering whether labelling as the full original product string is a better idea here (e.g. `edge_webdriver`) to avoid reusing labels that might have other meaning. A little paranoid, maybe.

@foolip - thoughts?

foolip: Given that a URL like https://wpt.fyi/results/?label=edge&label=master doesn't redirect and people might assume it to be the same as https://wpt.fyi/results/?label=master&product=edge, the most conservative and also non-lossy option seems to be labeling with both "edge" and "edge_webdriver" if the product was "edge_webdriver".

gsnedders: @lukebjerring note that because this is called after `prepare_labels` we end up with both `edge` and `edge_webdriver` in this case.

foolip:
What's still blocking getting this landed? I think it's the only thing still blocking getting Edge runs from Azure automatic published.

results-processor/wptreport.py
```diff
@@ -509,10 +529,12 @@ def create_test_run(report, labels_str, uploader, secret,
     labels = prepare_labels(report, labels_str, uploader)
     assert len(labels) > 0
 
+    labels |= normalize_product(report)
```
Hexcles: I'd prefer calling this inside `prepare_labels` (or perhaps even inline it there).

gsnedders: This means `prepare_labels` can change the name of the product within the report, which seems an even more surprising outcome.

Hexcles: Ahh, I didn't see `normalize_product` would also modify `report.run_info`. What about not doing that?

In other words, instead of "normalizing the product name" in the report, think of it as adding a special extra label. That way, it's more similar to `prepare_labels` than `normalize_version` (which is a hack that I don't like much).

gsnedders: Doing what exactly? We need these runs to show up as "Edge" runs in the UI (that it's using a different runner should probably be considered an implementation detail), and likely want Chrome/CDP runs to show up as "Chrome" in the UI.

If we add a magic label, we'll need to encode some way of expressing "this is a product name override", and ensuring everywhere which reads the product name reads from there.

It seems very much easier to change the actual product name in the report, because then we don't need to deal with some override. That said, given we end up with the report labelled with both "edge" and "edge_webdriver", we might want to clarify which label is the original product name (though it is fairly obvious to a human!).

Hexcles: Right, I was thinking not modifying `report.run_info`, in which case we'd need to override the product name in the request param to `/api/resuls/create`, which would be convoluted...

I was a bit unhappy about the fact that two `normalize_*` functions ended up in different places (in and out of the class) with slightly different behaviours, but I guess it's acceptable.

gsnedders: I do agree it's not ideal, but I'm not sure what the best solution here really is. All the options seem not great, and at least this isn't a large change now…

PR #1127, opened 2019-02-07
Improve discoverability of structured search

### LINKED ISSUE #1126: Document query syntax in UI
It's currently not possible to discover that the results can be filtered in the UI.

lukebjerring:
## Description
Adds some autocomplete behaviour for the search for basic atoms.
Adds a small link to some docs about the search syntax. Adds the docs, too.

Fixes #1126

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -1,0 +1,141 @@
+# wpt.fyi Search queries
+
+wpt.fyi supports a structured search syntax, allowing the user to filter specific results.
+
+## wpt.fyi search syntax
+
+As outlined below, the `/api/search` endpoint takes a structured query object. wpt.fyi's
+UI contains a search-box that converts a search syntax into the required structured query.
+Listed below are the "atoms" that can be used in a search.
+
+### Status
+
+Filters to results with a specific status (or, _not_ a specific status).
+
+Valid statuses are:
+
+ - `unknown`
+ - `pass`
+ - `ok`
+ - `error`
+ - `timeout`
+ - `notrun`
+ - `fail`
+ - `crash`
+ - `skip`
+ - `assert`
+
+> NOTE: `ok` is the status of the test harness setup. Individual subtests will have
+> a status of `pass` - it may be necessary to search for both.
+
+There are a couple of different ways to filter by status.
+
+#### Any product
+
+    status:[status]
+
+or, negation,
+
+    status:![status]
+
+
+#### Specific product
+
+    [product]:[status]
+
+Where `[product]` is a product specification (e.g. `safari`, `chrome-69`).
+
+### And-conjuction
+
+    [query1] and [query2] [and ...]
+
+Combines filters, such that they must all apply, e.g.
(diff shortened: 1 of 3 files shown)


lukebjerring:
Note: the link is broken until this lands, but if you replace `.../blob/master/...` with `.../blob/2641ea94ee4afc6947d05911bc7ae96ab098c428/...` it will show what it'll look at.

api/query/README.md
```diff
@@ -0,0 +1,141 @@
+# wpt.fyi Search queries
```
foolip: Great docs!

api/query/README.md
```diff
@@ -0,0 +1,141 @@
...
+Combines filters, such that any must apply, e.g.
+
+    chrome:pass or chrome:ok
+
+> NOTE: Or-conjuction takes less precedence than `and`. Precedence can be modified
+> using parens, e.g. `chrome:pass and (firefox:!pass or safari:!pass)`
+
+### Sequential
+
+    seq([query1] [query2] [...])
```
foolip: I already saw this live, but neat!

api/query/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+> NOTE: If, rather than a specific set of runs, the user wishes to query for the latest
+> results for a set of products, the `/api/search` endpoint supports the same query
+> parameters as /api/runs, outlined [in the API docs](../README.md)
+
+### Structured query objects
+
+Structured query objects are produced by the syntax parser on wpt.fyi.
+
+The easiest way to build the query you need is to use the syntax above, and inspect
```
foolip: Would it be hard to take the human-readable query in the API as well?

webapp/components/test-search.js
```diff
@@ -9,8 +9,28 @@ import { html } from '../node_modules/@polymer/polymer/lib/utils/html-tag.js';
...
+  'crash',
+  'skip',
+  'assert',
+  'unknown',
+];
+
+const atoms = {
+  status: statuses,
+};
+for (const b of DefaultBrowserNames) {
```
foolip: Another blank line here?
```suggestion

for (const b of DefaultBrowserNames) {
```

webapp/components/test-search.js
```diff
@@ -184,12 +192,19 @@ class TestSearch extends WPTFlags(PolymerElement) {
         padding: 0.5em 0;
         width: 100%;
       }
+      .help {
+        font-size: x-small;
```
foolip: The computed size for me is 10px. I suspect this is smaller than would get past an accessibility review, if we ever do that.

api/query/README.md
lukebjerring: Yes, since there's no ohm equivalent for Golang.

webapp/components/test-search.js
lukebjerring: I intend to deprecate this with a search builder in the future.

PR #1128, opened 2019-02-07
Add feature for showing iframes for reftests

lukebjerring:
## Description
Inlines a couple of iframes when we load the manifest, and are looking at a `reftest` test type.

## Review Information
- Visit /flags, enable the feature for reftest iframes
- Visit /results, find a reftest, e.g. /2dcontext/drawing-images-to-the-canvas/drawimage_canvas_self.html
- See the iframes when the manifest loads.

### DIFF
--- a/webapp/components/product-info.js
+++ b/webapp/components/product-info.js
@@ -94,6 +94,10 @@ function parseProduct(name) {
 const ProductInfo = (superClass) => class extends superClass {
   static get properties() {
     return {
+      activeBrowser: {
+        type: Object,
+        value: getActiveBrowser(),
+      }
     };
   }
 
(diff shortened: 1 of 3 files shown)


webapp/components/product-info.js
```diff
@@ -159,6 +163,30 @@ const ProductInfo = (superClass) => class extends superClass {
...
+  const ua = navigator.userAgent;
+  let tem, M = ua.match(/(opera|chrome|safari|firefox|msie|trident(?=\/))\/?\s*(\d+)/i) || [];
+  if(/trident/i.test(M[1])){
+    tem = /\brv[ :]+(\d+)/g.exec(ua) || [];
+    return { name:'IE', version: tem[1] || '' };
+  }
+  if (M[1] === 'Chrome') {
+    tem = ua.match(/\bOPR|Edge\/(\d+)/);
+    if (tem !== null) {
+      return { name: 'Opera', version: tem[1] };
```
foolip: Looks like this will say that Edge is Opera?

Rather than trying to parse out the browser and misclassify some, can the help text just say "your browser"?

lukebjerring: Done.

Issue #1071, opened 2019-01-25
Removing a failing test is reported as a regression


foolip:
### Check
A link to the GitHub Check run containing my issue:
https://github.com/web-platform-tests/wpt/pull/15066/checks?check_run_id=54650589

### Description
This check links to https://wpt.fyi/results/?diff&run_id=5792420276994048&run_id=6578543039348736 and the change is that "PaymentAddress interface: attribute regionCode" was removed, which matches the IDL change.

This shouldn't be reported as a regression.

lukebjerring:
I disagree that it shouldn't be flagged. Imagine that the change is a user typo from "PaymentAddress" to "PaymnetAddress", and a subtest gets dropped. That isn't distinguishable from an intentional subtest delete.

lukebjerring:
OK, another example. You have 10 failing tests. You botch the setup of the test, and now, the first failing test causes none of the other tests to be run. Your count goes from 10 failing to 1 failing.

My point is, a user confirming that they intended to reduce the number of subtests is the only foolproof way to avoid accidental regressions. The argument that "accidents are rare" could be applied generally to the overall regression concept, and we could choose to not bother reporting anything except PASS becomes FAIL. That seems risky and counter to the idea of the project as a whole.

foolip:
Yes, the sort of thing you describe can happen with idlharness.js tests or other cases where the tests are generated from input in some convoluted way.

In many other cases, I would expect the change in number of tests to be accompanied by something else which is itself worth flagging:
- For subtests that create other subtests in a chain, the last subtest before the ones that disappeared probably went from passing to failing, since the next test is commonly created at the very end of a test.
- A harness error can cause tests to never be created, or tests to get the NOTRUN status. Both a new harness error and NOTRUN are strong signals that something is wrong.

In addition to the frequency of accidental regressions ("accidents are rare") we have to look at how frequent intentional changes that look like regressions are. Among my hunches are that *just* removing tests (passing or failing) is very common and then usually intentional. Would it be straightforward to get pairs of reports for all the PRs we've gotten into wpt.fyi so far and run them through the regression heuristic we have and just look at a random sample of things that would be flagged? It would be interesting to see if anything other than "convoluted and indirect setup" has resulted in accidental subtest removal so far.

PR #1129, opened 2019-02-08
Add support for diffs using searchcache

lukebjerring:
## Description
Adds diff param support to /api/search, computing differences for the subtests directly (instead of via summary numbers, which can cause things to cancel each other out).

This is a step towards https://github.com/web-platform-tests/wpt.fyi/issues/1071
We'll need to compute the regression diffs via the same API

### DIFF
--- a/shared/run_diff.go
+++ b/shared/run_diff.go
@@ -112,26 +112,62 @@ func (s ResultsSummary) Add(k string, other TestSummary) {
 // TestDiff is an array of differences between 2 tests.
 type TestDiff []int
 
+const (
+	newlyPassingIndex = 0
+	newlyFailingIndex = 1
+	totalDeltaIndex   = 2
+)
+
 // NewlyPassing is the delta/increase in the number of passing tests when comparing before/after.
 func (d TestDiff) NewlyPassing() int {
-	return d[0]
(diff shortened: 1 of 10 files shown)


lukebjerring:
@foolip - do we have any examples of regressions obfuscated by the inability to compare subtest results directly?

foolip:
@lukebjerring no, spotting regressions manually is pretty hard, and I haven't caught any. If you compare the two methods across pairs of runs going back in time I suspect you'll find some though.

shared/run_diff.go
```diff
@@ -112,26 +112,73 @@ func (s ResultsSummary) Add(k string, other TestSummary) {
...
 func (d TestDiff) Add(other TestDiff) {
-	d[0] += other[0]
-	d[1] += other[1]
-	d[2] += other[2]
+	d[newlyPassingIndex] += other[newlyPassingIndex]
+	d[newlyFailingIndex] += other[newlyFailingIndex]
+	d[totalDeltaIndex] += other[totalDeltaIndex]
+}
+
+// Append appends the difference between the two given statuses, if any.
```
foolip:
```suggestion
// Appends the difference between the two given statuses, if any.
```

PR #1135, opened 2019-02-11
Add "path" atom to structured queries

### LINKED ISSUE #1132: Allow prefix-only searches
If I search for a path like `/dom/` I get both `/dom/*` and `/html/dom/*` when I really only want the former. If later someone adds `/css/css-foo/subdir/dom` I will start to get that too. It would be great if paths could be searched with something like `prefix:/dom/` (or just with regexp, although that doesn't really match the rest of the API) to allow more predictable selections.

lukebjerring:
## Description
Add a `path:[test path]` atom, distinct from the `pattern` atom in that it's a prefix-match, not a case-insensitive search (and whatever else we end up doing to make it even more fuzzy).

Fixes #1132

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -55,6 +55,17 @@ func (tnp TestNamePattern) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return tnp
 }
 
+// TestPath is a query atom that matches exact test path prefixes.
+// It is an inflexible equivalent of TestNamePattern.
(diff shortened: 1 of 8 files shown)


lukebjerring:
@jgraham - the "integration test" is to run a search for `/dom/`:
https://path-dot-wptdashboard-staging.appspot.com/results/?q=path%3A%2Fdom%2F

Seems to include /dom/ and ignore /html/dom/ as expected.

PR #1139, opened 2019-02-13
Make the page wider and expand reftest previews

Hexcles:
## Description

800px is too crowded when we have more than four runs (which is getting more common now that search is heavily used). And reftest previews can be expanded a bit (also to match the default screenshot size 600x600).

## Review Information

See the staging deployment.

### DIFF
--- a/webapp/static/common.css
+++ b/webapp/static/common.css
@@ -6,6 +6,7 @@ body {
   font-size: 16px;
   font-family: Arial;
   color: #333;
+  padding: 0 16px;
 }
 #content {
   margin: 0 auto;
--- a/webapp/static/common.css
+++ b/webapp/static/common.css
@@ -9,7 +9,7 @@ body {
 }
 #content {
   margin: 0 auto;
-  max-width: 800px;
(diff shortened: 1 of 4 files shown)


Hexcles:
Example: https://wider-page-dot-wptdashboard-staging.appspot.com/results/css/css-ui/box-sizing-001.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned

Hexcles:
Please go to https://wider-page-dot-wptdashboard-staging.appspot.com/flags to enable reftest preview first.

webapp/static/common.css
```diff
@@ -9,7 +9,7 @@ body {
 }
```
lukebjerring:
```suggestion
  padding: 16px;
}
```

Hexcles: This wouldn't make any difference normally, unless the viewport is narrower than 1200px, in which case this will apply some padding on the sides. I assume this is your intention.

PR #1145, opened 2019-02-14
Automatically generate mocks when building packages

Hexcles:
## Description

Background: I ran into problems when trying to modify `api/receiver/appengine.go`. I couldn't get `mockgen` to generate a new mock for it. After some digging, I found out I was violating some conventions and/or best practices (e.g. having unexported functions in exported interfaces) and it happened to work with an older version of `mockgen` and only in source mode. I then realized the problem was widespread: we didn't really document how we generated mocks; some used source mode while the others used reflect mode, and other parameters (like the package name) also varied. This is bad for maintainability.

See the commit message of the first commit for more rationales of the change.

## Review Information

Changes are mostly mechanical. Tests should continue to pass.


lukebjerring:
What's the need for the separate packages? The mockgen command can take a package param?

Issue #1154, opened 2019-02-19
Make browser headers in results view be position:sticky


stephenmcgruer:
If you look at a results page like https://wpt.fyi/results/fetch/data-urls/processing.any.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned, it is easy to scroll down and then lose the context of which column is which browser.

If we apply the following to the <th>* above the <test-run> entry:

element.style {
    background: white;
    position: sticky;
    top: 0;
    z-index: 1;
}

Then you get a view like the attached which I feel is more usable.

![itglqtnykqe](https://user-images.githubusercontent.com/153264/53024188-8967d700-342c-11e9-82e2-14e393ddf495.png)

* In the best world, you would apply is to the <thead> so that it works on all the rows without duplicating style. But sticky is poorly supported for tables in Chrome and so that wouldn't work :(.

PR #1156, opened 2019-02-19
Add CSS for sticky table headers

lukebjerring:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1154

### DIFF
--- a/webapp/components/abstract-test-file-results-table.js
+++ b/webapp/components/abstract-test-file-results-table.js
@@ -6,68 +6,71 @@
 import { TestRunsBase } from './test-runs.js';
 import { WPTColors } from './wpt-colors.js';
 import './test-file-results.js';
-const $_documentContainer = document.createElement('template');
+import { html } from '../node_modules/@polymer/polymer/lib/utils/html-tag.js';
 
-$_documentContainer.innerHTML = `<dom-module id="abstract-test-file-results-table">
-  <template>
-    <style include="wpt-colors">
-      table {
-        width: 100%;
-        border-collapse: collapse;
-      }
-      td {
-        padding: 0;
-        height: 1.5em;
-      }
-      td code {
-        padding: 0.25em;
-      }
-      td.sub-test-name {
-        font-family: monospace;
-      }
-      td.result {
-        background-color: #eee;
-      }
-      tbody tr:first-child {
-        border-bottom: 8px solid white;
-      }
-    </style>
-
-    <table>
-      <thead>
-        <tr>
-          <th width="[[computeSubtestThWidth(testRuns)]]">Subtest</th>
-          <template is="dom-repeat" items="[[testRuns]]" as="testRun">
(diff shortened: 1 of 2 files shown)


lukebjerring:
@stephenmcgruer - the Travis CI build will automatically deploy a staging enviroment including the fix, which you can use to verify it's WAI.

stephenmcgruer:
Wow, thanks for the quick PR!

This appears to work for the 'overall' results view (e.g. https://sticky-dot-wptdashboard-staging.appspot.com/results/?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned) but not for results for a specific test (e.g. https://sticky-dot-wptdashboard-staging.appspot.com/results/fetch/data-urls/processing.any.worker.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned). Maybe there's somewhere else the style needs applied?

webapp/components/abstract-test-file-results-table.js
```diff
@@ -6,70 +6,73 @@
...
   }
 
+  static get template() {
+    return html`
+<style include="wpt-colors">
+  table {
+    width: 100%;
+    border-collapse: collapse;
+  }
+  th {
```
lukebjerring: This `th` style is the only actual change; the remainder is just cleanup of the auto-upgrade tool for Polymer 2 => 3 from #982 a while back.

webapp/components/abstract-test-file-results-table.js
```diff
@@ -6,70 +6,73 @@
 import { TestRunsBase } from './test-runs.js';
 import { WPTColors } from './wpt-colors.js';
 import './test-file-results.js';
-const $_documentContainer = document.createElement('template');
+import { html } from '../node_modules/@polymer/polymer/lib/utils/html-tag.js';
```
Hexcles:
```suggestion
import { html } from '@polymer/polymer/polymer-element.js';
```
Same problem as last time.

lukebjerring: So, this link is copied from another file. It's strange to me that it's broken in some files and that goes unnoticed - I'll send a separate PR for changing all of them.

Hexcles: Yeah that's indeed strange. Looks like it's only broken when you directly import `html-tag.js` in *tests*. Might have something to do with the dynamic import path overwrite of `wct-server`.

lukebjerring: https://github.com/web-platform-tests/wpt.fyi/pull/1157

stephenmcgruer:
The latest version deployed to the staging server does appear to have sticky headers in both places, thanks!

PR #1157, opened 2019-02-19
Update invalid legacy import

lukebjerring:
This import clashes with `wct`, so get rid of it to ensure we don't keep causing test failures.

I search+replaced, then auto-sorted the imports in all affected files.


lukebjerring:
Ignoring the Chrome failures, will investigate those separately.

PR #1163, opened 2019-02-20
Add feature for whether to process azure check_run events

### LINKED ISSUE #1130: Duplicate (partial) runs from Azure Pipelines
https://wpt.fyi/runs?label=master&max-count=100&from=2019-02-04T00%3A00 now looks like this:
image

At least for Azure Pipelines, there shouldn't have been this many actual runs on master.

@lukebjerring @Hexcles is it possible to find the source of each of these? I see that the raw report URLs and timestamps aren't the same, so maybe there's something wrong on the Azure Pipelines side?

lukebjerring:
## Description
Should fix #1130 

Disables the azure pipelines `CheckRunEvent` processing for GitHub webhooks, unless feature `processAzureCheckRunEvents` is enabled.

This PR should be coupled with a change to the non-sharded (affected-tests) Azure jobs, to fire the completion webhook.

### DIFF
--- a/api/checks/webhook.go
+++ b/api/checks/webhook.go
@@ -175,7 +175,11 @@ func handleCheckRunEvent(aeAPI shared.AppEngineAPI, checksAPI API, azureAPI azur
 	status := checkRun.GetCheckRun().GetStatus()
 	shouldSchedule := false
 	if appID == azure.PipelinesAppID {
-		return azureAPI.HandleCheckRunEvent(checkRun)
+		if aeAPI.IsFeatureEnabled("processAzureCheckRunEvents") {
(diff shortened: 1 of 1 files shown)


Hexcles:
@foolip already removed the webhook in https://github.com/web-platform-tests/wpt/pull/15477

We'd need to revert that removal.

foolip:
Yeah, I sent https://github.com/web-platform-tests/wpt/pull/15477 before @lukebjerring discussed and came up with a new plan.

lukebjerring:
TBR'ing this - we want this fix in prod ASAP, so let's get it landed and tested.

Issue #1138, opened 2019-02-11
In /runs, link to /results/run_id=...


foolip:
Trying to understand an issue like https://github.com/web-platform-tests/wpt.fyi/issues/1130, it would be most helpful if one could get to each individual run by clicking the browser icons. This should now be possible with the `run_id` parameter.

lukebjerring:
Fixed via https://github.com/web-platform-tests/wpt.fyi/pull/1164

PR #1164, opened 2019-02-20
Support selecting runs in the /runs UI

### LINKED ISSUE #114: Support easily creating a diff query from the /test-runs UI
When https://github.com/web-platform-tests/wpt.fyi/issues/112 and https://github.com/web-platform-tests/wpt.fyi/issues/113 and potentially a few other issues are fixed, it'd be great to make it easy to compare two arbitrary runs from a view like https://wpt.fyi/test-runs.

lukebjerring:
## Description
Fixes #114 and https://github.com/web-platform-tests/wpt.fyi/issues/1138

Clicking a run in the /runs UI selects that run, and triggers a toast, which has a "view runs" button.
Selecting exactly two runs also shows a "view diff" button (which includes the ?diff param).

## Review Information
- Visit https://select-dot-wptdashboard-staging.appspot.com/runs
- Click on runs
- Be amazed

### DIFF
--- a/webapp/components/wpt-runs.js
+++ b/webapp/components/wpt-runs.js
@@ -4,8 +4,10 @@
  * found in the LICENSE file.
  */
 
+import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
 import '../node_modules/@polymer/iron-collapse/iron-collapse.js';
 import '../node_modules/@polymer/paper-button/paper-button.js';
+import '../node_modules/@polymer/paper-toast/paper-toast.js';
 import '../node_modules/@polymer/paper-spinner/paper-spinner-lite.js';
 import '../node_modules/@polymer/paper-styles/color.js';
 import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-runs.js
```diff
@@ -374,6 +433,42 @@ class WPTRuns extends WPTFlags(SelfNavigation(LoadingState(TestRunsUIBase))) {
...
+    }
+    if (diff) {
+      url.searchParams.set('diff', true);
+    }
+    window.location = url;
+  }
+
+  handleSelectRun(e) {
+    const id = e.target.getAttribute('data-run-id');
+    if (this.selectedRuns.find(r => r === id)) {
```
foolip: Can this be just `e.target.hasAttribute('selected')`? Is it possible for that to be out of sync with the condition here?

lukebjerring: It's safer to use the set as the source of truth, but shouldn't be possible to be out of sync.

lukebjerring:
The feature is difficult to test with WCT because it triggers a window.location change; we'd need to write an integration test for it.

Issue #1151, opened 2019-02-15
Implement a TestRun lifecycle


lukebjerring:
There are several times that we know/think a TestRun is going to be created. Being able to identify TestRun entities that have not yet completed is important in a few use-cases:

 - The non-deterministic processing of PR results (with/without the patch)
 - Supporting "in progress" statuses for the custom checks
 - Displaying pending TestRuns in the wpt.fyi /runs view

The `PendingTestRun` entity will simplify the separation of in-progress runs and completed runs, avoiding the need for a big refactor (and any requirement of all queries to consider what statuses they care about.)

Example of race condition for non-distinct runs: https://github.com/web-platform-tests/wpt.fyi/issues/1147

While we are using a separate entity, it is possible to [preserve a `TestRun` ID](
https://cloud.google.com/datastore/docs/concepts/entities#assigning_identifiers) and re-use it as the identifier for the `PendingTestRun` too. This would ensure that we can reuse the TestRun ID across different mediums (the PendingTestRun, TaskQueue ID, etc), making for a smoother consolidation of the different services involved in the lifecycle.

lukebjerring:
Closing in favor of tracking more issues in https://github.com/web-platform-tests/wpt.fyi/projects/11

PR #1168, opened 2019-02-21
Pre-allocate TestRun IDs

lukebjerring:
## Description
Adds a datastore method for pre-allocating an ID for a `TestRun` and plumbs it through to the processor and back. We surface the (unique) ID as the Task's name (currently the names are meaningless).

Partial implementation of https://github.com/web-platform-tests/wpt.fyi/issues/1151


shared/datastore.go
```diff
@@ -42,6 +42,7 @@ type Datastore interface {
 	Done() interface{}
 	NewQuery(typeName string) Query
 	NewKey(typeName string, id int64) Key
+	ReserveKey(typeName string) (Key, error)
```
Hexcles: Nit: name this `ReserveID` as it can only reserve a numeric ID automatically.

results-processor/processor.py
```diff
@@ -115,6 +115,7 @@ def process_report(params):
     gcs_paths = params.getlist('gcs')
     result_type = params['type']
     # Optional fields:
+    run_id = params.get('run_id') or '0'
```
Hexcles:
```suggestion
    run_id = params.get('run_id', '0')
```

api/receiver/appengine.go
```diff
@@ -57,12 +58,24 @@ func NewAppEngineAPI(ctx context.Context) AppEngineAPI {
...
+		store:            shared.NewAppEngineDatastore(ctx, false),
 	}
 }
 
 func (a *appEngineAPIImpl) addTestRun(testRun *shared.TestRun) (*DatastoreKey, error) {
-	key := datastore.NewIncompleteKey(a.Context(), "TestRun", nil)
-	key, err := datastore.Put(a.Context(), key, testRun)
+	ctx := a.Context()
+	var key *datastore.Key
+	if testRun.ID > 0 {
```
Hexcles:
```suggestion
	if testRun.ID != 0 {
```

I think the ID *could* be negative although I've never seen one. And all the documentation only says 0 is the special case, so let's be a bit more cautious here.

api/receiver/appengine.go
```diff
@@ -57,12 +58,24 @@ func NewAppEngineAPI(ctx context.Context) AppEngineAPI {
...
+		key = datastore.NewKey(ctx, "TestRun", "", testRun.ID, nil)
+		err := datastore.Get(ctx, key, new(shared.TestRun))
```
Hexcles: This needs to go into a transaction: https://cloud.google.com/datastore/docs/concepts/entities#creating_an_entity (see `insert` an entity)

While doing so, let's make `Insert` a new method on the `shared.Datastore` interface.

lukebjerring: I disagree that this needs to be in a transaction; we're not doing multiple actions. Either the entity doesn't exist (so, nothing happens), or it does and we return. We don't ever do both a real get/put.

shared/datastore.go
lukebjerring: Done.

api/receiver/appengine.go
Hexcles: There's a `Put` right after the whole `if` block. The race condition would be:

Two parallel requests trying to create two test runs with the same ID (bugs in the code or retries, etc.). Two requests may both think the ID doesn't exist, and they'll simultaneously try to write to that key. One test run will be lost that way, and in case the runs are different, we have data loss.

lukebjerring: Oh, I see. Your contrived error relies on some incorrect code somewhere. Done.

api/receiver/appengine.go
```diff
@@ -57,12 +58,21 @@ func NewAppEngineAPI(ctx context.Context) AppEngineAPI {
...
 	}
 }
 
 func (a *appEngineAPIImpl) addTestRun(testRun *shared.TestRun) (*DatastoreKey, error) {
-	key := datastore.NewIncompleteKey(a.Context(), "TestRun", nil)
-	key, err := datastore.Put(a.Context(), key, testRun)
+	ctx := a.Context()
+	var key shared.Key
+	var err error
+	store := shared.NewAppEngineDatastore(ctx, false)
```
Hexcles: Use `a.store` instead of creating another one.

api/receiver/appengine.go
```diff
@@ -57,12 +58,21 @@ func NewAppEngineAPI(ctx context.Context) AppEngineAPI {
...
+	key = store.NewIDKey("TestRun", testRun.ID)
```
Hexcles: `tesRun.ID` needs to be manually set here:
https://github.com/web-platform-tests/wpt.fyi/blob/03bc0e68e21cd92b57fad345acd5475d873968d5/api/receiver/create_run.go#L63

api/receiver/appengine.go
lukebjerring: Done.

api/receiver/appengine.go
lukebjerring: No it doesn't - the `id` field in the JSON either is or is not set by the POSTer.

https://github.com/web-platform-tests/wpt.fyi/pull/1168/files#diff-696fea2c57fe75ba995a921eb5585dddR535

Hexcles: Ahh I didn't see that `TestRun.ID` also had a `json` tag, so this would just work automatically. Great.

Hexcles:
@lukebjerring feel free to merge after you resolve the merge conflicts.

PR #1172, opened 2019-02-22
Screenshot storage backend

Hexcles:
## Description

This implements API endpoint [1](https://docs.google.com/document/d/1IhZa4mrjK1msUMhtamKwKJ_HhXD-nqh_4-BcPWM6soQ/edit?ts=5c4b4334#heading=h.ratj488usbcf) and [2](https://docs.google.com/document/d/1IhZa4mrjK1msUMhtamKwKJ_HhXD-nqh_4-BcPWM6soQ/edit?ts=5c4b4334#heading=h.xczylqsapugz) in the design.

### DIFF
--- a/api/screenshot/cache.go
+++ b/api/screenshot/cache.go
@@ -17,15 +17,19 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+func parseParams(r *http.Request) (browser, browserVersion, os, osVersion string) {
+	browser = r.FormValue("browser")
+	browserVersion = r.FormValue("browser_version")
+	os = r.FormValue("os")
+	osVersion = r.FormValue("os_version")
+	return browser, browserVersion, os, osVersion
+}
+
 func getHashesHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := shared.NewAppEngineContext(r)
 	ds := shared.NewAppEngineDatastore(ctx, false)
 
-	browser := r.FormValue("browser")
-	browserVersion := r.FormValue("browser_version")
-	os := r.FormValue("os")
-	osVersion := r.FormValue("os_version")
-
+	browser, browserVersion, os, osVersion := parseParams(r)
 	hashes, err := RecentScreenshotHashes(ds, browser, browserVersion, os, osVersion, nil)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
@@ -59,10 +63,7 @@ func uploadScreenshotHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	bucket := gcs.Bucket(bucketName)
 
-	browser := r.FormValue("browser")
-	browserVersion := r.FormValue("browser_version")
-	os := r.FormValue("os")
-	osVersion := r.FormValue("os_version")
+	browser, browserVersion, os, osVersion := parseParams(r)
 	hashMethod := r.FormValue("hash_method")
 	if r.MultipartForm == nil || r.MultipartForm.File == nil || len(r.MultipartForm.File["screenshot"]) == 0 {
 		http.Error(w, "no screenshot file found", http.StatusBadRequest)
--- a/api/screenshot/cache.go
+++ b/api/screenshot/cache.go
@@ -106,7 +106,7 @@ func storeScreenshot(ctx context.Context, bucket *storage.BucketHandle, hashMeth
 	if hashMethod == "" {
 		hashMethod = "sha1"
 	}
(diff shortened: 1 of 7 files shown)


lukebjerring:
Let me know when this is ready for a thorough review.

api/screenshot/cache.go
```diff
@@ -0,0 +1,103 @@
...
+)
+
+func getHashesHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	ds := shared.NewAppEngineDatastore(ctx, false)
+
+	browser := r.FormValue("browser")
+	browserVersion := r.FormValue("browser_version")
+	os := r.FormValue("os")
+	osVersion := r.FormValue("os_version")
```
lukebjerring: Why the separate params? This could just be a product(s) param, and leverage the existing syntax + parsing for, say, `chrome-72-linux` type thing?

Hexcles: This is somewhat complicated: the request will come from `wptrunner`, where browser names and versions are not normalized (the normalization happens in the processor), so we'd see invalid versions (e.g. Safari TP) if we use `ParseProduct`.

Hexcles:
@lukebjerring this is now ready for review.

api/screenshot/cache.go
```diff
@@ -0,0 +1,136 @@
...
+	bucketName := "wptd-screenshots-staging"
+	if aeAPI.GetHostname() == "wpt.fyi" {
+		bucketName = "wptd-screenshots"
+	}
+	bucket := gcs.Bucket(bucketName)
+
+	browser := r.FormValue("browser")
+	browserVersion := r.FormValue("browser_version")
+	os := r.FormValue("os")
+	osVersion := r.FormValue("os_version")
```
lukebjerring: nit: probably nicer to have a helper function that aggregates these into a product struct. (here and above)

api/screenshot/model.go
```diff
@@ -0,0 +1,154 @@
...
+	HashMethod string
+	Labels     []string
+	// These two fields can be left empty and will be filled by Store().
+	Counter  int
+	LastUsed time.Time
+}
+
+// NewScreenshot creates a new Screenshot with the given labels (empty labels
+// are omitted).
+func NewScreenshot(labels []string) *Screenshot {
```
lukebjerring: IMO, callers shouldn't have to wrap labels in an array when there's only one parameter type.

```suggestion
func NewScreenshot(labels ...string) *Screenshot {
```

api/screenshot/model.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+		query.Order("-LastUsed")
+		query.Limit(totalLimit)
+
+		var hits []Screenshot
+		if _, err := ds.GetAll(query, &hits); err != nil {
+			return shared.ToStringSlice(all), err
+		}
+		for _, s := range hits {
+			all.Add(s.Hash())
```
lukebjerring: If we're only pulling the hash, you might consider doing a `.Project("HashMethod","HashDigest")` on the `Screenshot` query.

api/screenshot/model.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+
+	for all.Cardinality() < totalLimit {
+		query := ds.NewQuery("Screenshot")
+		for _, l := range labels {
+			query.Filter("Labels =", l)
+		}
+		query.Order("-LastUsed")
+		query.Limit(totalLimit)
```
lukebjerring: The GetAll maximum Limit value is 1000.

> Note that a keys-only query can return more than 1000 results, but GetAll can only retrieve 1000 keys at a time, and will fail with an error if called on a larger result. Therefore, we recommend adding a limit of 1000 keys to the query.

https://cloud.google.com/appengine/docs/standard/go/datastore/queries

Hexcles: This is a valid concern. However, contrary to the documentation, it seems to work in staging (`GetAll` is returning me 7000+ items with no complaints). We can leave it as is and fix it later if it indeed becomes a problem as it'd be an easy change.

api/screenshot/model.go
Hexcles: Yeah it used to have more parameters. Fixed.

api/screenshot/model.go
Hexcles: I'll send a follow-up to PR to fix this as well as changing the Datastore abstraction to take variable-length parameters.

For now, I've added a TODO.

Hexcles:
@lukebjerring a query with 7k hits seems to work and I've addressed the other comments. Can you give another pass of review?

Issue #1173, opened 2019-02-23
SHA links on /runs don't show the runs listed


foolip:
https://wpt.fyi/runs now looks like this:
<img width="813" alt="screen shot 2019-02-23 at 11 42 49 am" src="https://user-images.githubusercontent.com/498917/53285399-33629e80-3760-11e9-9f8a-66196e99e46d.png">

The links for the rows with Safari runs are:
https://wpt.fyi/?sha=70df598b89 ([permalink](https://wpt.fyi/results/?run_id=5106130036457472&run_id=5484645806243840&run_id=5582530191294464))
https://wpt.fyi/?sha=5803f5e36a ([permalink](https://wpt.fyi/results/?run_id=4703540874313728&run_id=6004742656360448&run_id=4975408411836416))
https://wpt.fyi/?sha=8d7111a760 ([permalink](https://wpt.fyi/results/?run_id=4531892908982272&run_id=6566352210886656&run_id=4736764862267392))

These aren't the same runs as are shown in the /runs view. In the first two, the Safari runs are pr_base runs. This had me thinking that our Safari runs are totally broken before finally understanding what I was seeing.

PR #1177, opened 2019-02-25
Propagate /runs query params to /results?sha= links

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1173

Propagates any query params used to filter the `/runs` view across to the `/results` view for specific SHAs.

### DIFF
--- a/webapp/components/wpt-runs.js
+++ b/webapp/components/wpt-runs.js
@@ -178,7 +178,7 @@ class WPTRuns extends Pluralizer(WPTFlags(SelfNavigation(LoadingState(TestRunsUI
         <template is="dom-repeat" items="{{ testRunsBySHA }}" as="results">
           <tr>
             <td>
-              <a class="github" href="/?sha={{ results.sha }}">
+              <a class="github" href="{{ revisionLink(results) }}">
                 <template is="dom-if" if="[[results.commitType]]">
                   <img src="/static/[[results.commitType]].svg">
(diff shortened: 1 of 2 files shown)


webapp/components/wpt-runs.js
```diff
@@ -389,6 +389,15 @@ class WPTRuns extends Pluralizer(WPTFlags(SelfNavigation(LoadingState(TestRunsUI
     return link.toString();
   }
 
+  revisionLink(results) {
+    const url = new URL('/results', window.location);
+    url.search = this.query;
+    url.searchParams.set('sha', results.sha);
+    url.searchParams.set('max-count', 1);
```
foolip: Isn't this the default for /results, so that just omitting does the right thing?

lukebjerring: Only when no constraints are supplied. Providing a SHA prevents an implicit max-count of 1 (you may want all of the results).

foolip: That doesn't sound like what I've observed. I've always seen at most one run per product when following links from /runs, has this changed?

lukebjerring: If there are multiple shas, or a from param, etc, then we don't set max count 1 implicitly.

foolip: Oh, so some other conditions than what I usually look at, I see.

PR #1180, opened 2019-02-26
Detect master branch for azure builds

lukebjerring:
## Description
Artifact name alone isn't sufficient; they can be a manual full-run.

### DIFF
--- a/api/azure/api.go
+++ b/api/azure/api.go
@@ -11,6 +11,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"regexp"
 
 	"github.com/google/go-github/github"
 	"github.com/web-platform-tests/wpt.fyi/shared"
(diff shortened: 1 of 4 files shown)


lukebjerring:
Tested locally using https://dev.azure.com/web-platform-tests/wpt/_apis/build/builds/8120

api/azure/api.go
```diff
@@ -62,10 +74,40 @@ func (a apiImpl) HandleCheckRunEvent(checkRun *github.CheckRunEvent) (bool, erro
...
+		log.Errorf("Failed to read request response: %s", err.Error())
+		return false
+	}
+	var build AzureBuild
+	if err := json.Unmarshal(data, &build); err != nil {
+		log.Errorf("Failed to unmarshal request response: %s", err.Error())
+		return false
+	}
+	log.Debugf("Source branch: %s", build.TriggerInfo.SourceBranch)
+	return build.TriggerInfo.SourceBranch == "master"
```
foolip: This will fail to add the master label for builds triggered from epochs/* branches. It's a bit ugly to know about branch topology here, but the only other option is to have a checkout.

lukebjerring: Can you give a specific build ID for this case?

PR #1183, opened 2019-02-27
React to taskcluster checkrun events

lukebjerring:
## Description
This is untested; we'll need to deploy it as staging.wpt.fyi to test that it works as intended with another repo (e.g. lukebjerring/wpt) that has the wpt.fyi checks installed.


api/taskcluster/webhook.go
```diff
@@ -293,7 +328,6 @@ func getAuth(ctx context.Context) (username string, password string, err error)
 func createAllRuns(
 	log shared.Logger,
 	aeAPI shared.AppEngineAPI,
-	checksAPI checks.API,
```
lukebjerring: This was a circular dep, and was only used for pending checks (which we have disabled for now).
Pending checks will be re-implemented (if at all) with lifecycle - see #1151

api/taskcluster/webhook.go
```diff
@@ -194,12 +204,40 @@ func shouldProcessStatus(log shared.Logger, processAllBranches bool, status *sta
 	return true
 }
 
-func extractTaskGroupID(targetURL string) string {
-	lastSlash := strings.LastIndex(targetURL, "/")
-	if lastSlash == -1 {
-		return ""
+// handleCheckRunEvent processes an Azure Pipelines check run "completed" event.
+func handleCheckRunEvent(aeAPI shared.AppEngineAPI, event *github.CheckRunEvent) (bool, error) {
```
Hexcles: It might be less confusing to move this function into `api/taskcluster/api.go` (and perhaps even inline it into the impl), since nothing else is using this function and we can make this file specific for legacy Taskcluster webhooks.

api/taskcluster/webhook.go
```diff
@@ -115,27 +155,8 @@ func (b branchInfos) GetNames() []string {
...
-
-	if status.TargetURL == nil {
-		return false, errors.New("No target_url on taskcluster status event")
-	}
-	taskGroupID := extractTaskGroupID(*status.TargetURL)
-	if taskGroupID == "" {
-		return false, fmt.Errorf("unrecognized target_url: %s", *status.TargetURL)
-	}
-
 	log.Debugf("Taskcluster task group %s", taskGroupID)
```
Hexcles: Log `taskID` as well if it's not empty.

api/taskcluster/webhook.go
```diff
@@ -115,27 +155,8 @@ func (b branchInfos) GetNames() []string {
 	return names
 }
 
-func handleStatusEvent(ctx context.Context, payload []byte) (bool, error) {
+func processTaskclusterBuild(ctx context.Context, taskGroupID, taskID string, sha string, labels ...string) (bool, error) {
```
Hexcles: Take an `aeAPI` here instead of `ctx`. All the callers have already created an `aeAPI` before calling this function.

api/taskcluster/webhook.go
```diff
@@ -194,12 +204,40 @@ func shouldProcessStatus(log shared.Logger, processAllBranches bool, status *sta
...
+	taskGroupID, taskID := extractTaskGroupID(detailsURL)
+	if taskGroupID == "" {
+		return false, fmt.Errorf("unrecognized target_url: %s", detailsURL)
+	}
+
+	return processTaskclusterBuild(aeAPI.Context(), taskGroupID, taskID, sha, shared.ToStringSlice(labels)...)
+}
+
+func extractTaskGroupID(targetURL string) (string, string) {
+	regex := regexp.MustCompile("(?:/task-group-inspector/#/|/groups/)([^/]*)(?:/tasks/([^/]*))?")
```
Hexcles: I'd prefer two simpler, stricter regexes instead of one with two clauses with two options each (i.e. four total combinations).

lukebjerring: Done.

lukebjerring:
OK, it's now working
https://github.com/web-platform-tests/wpt/pull/15583/checks?check_run_id=70460037

Issue #57, opened 2018-04-17
Show expected and actual renderings for reftests


foolip:
This is the frontend part of https://github.com/web-platform-tests/results-collection/issues/217. If we record the screenshots in the runners and store them somewhere, then the frontend would also need to show them, and possibly diffs between them.

foolip:
Renaming to also track what the receiver (@Hexcles) would have to do.

(This issue is quite aspirational at this point, it'd be a lot of work, but also very valuable if we could pull it off.)

foolip:
@mrego, you asked in https://chromium-review.googlesource.com/c/chromium/src/+/1054770#message-b3087aea32526011c2a83a3a094013263e2aedfb: "Where I can see the screenshots from wpt.fyi?"

The answer is that you can't and this is the issue for it. Could you describe what it is you were trying to do and how much time (if any) you ended up spending to get to the same answer without having the screenshots? Is it only for failed tests that you wanted to see screenshots, or passing ones too?

mrego:
Ok, good to know they're nowhere, so it's expected I couldn't find them. :smile: 

I needed to see the output of a WPT test on Mac, I ended up using an actual Mac to check it.
In this case it was a test passing in Mac and failing on the rest of platforms, but I admit that looking for screenshots of passing tests don't seem the most common thing.
Just my two cents.

foolip:
Thanks @mrego. It's interesting that you wanted to see why a test was passing, as it seems like it might be easier/faster to only record failure cases. But I think that many times when you have some browser passing and some failing, looking at the passing ones will tell you something interesting.

mrego:
I agree it's the strange case probably, but the test was failing in all browsers and passing only in Safari on Mac. So I wanted to see a screenshot of what was going on there.

foolip:
I'm not sure it is a strange case, will keep asking people who ask about screenshots to see if they're normally interested in the passing or the failing cases.

foolip:
IRC between me and @jgraham starts at https://w3.logbot.info/testing/20180808#c430986

There I suggest that we'd want to get the screenshots into wpt.fyi so that old links keep working, although it isn't strictly speaking necessary to keep all screenshots forever. However, I suspect that if we deduplicate, perhaps by hashing the decoded size+RGBA data, that the size of the screenshot bucket wouldn't grow very fast. (Even flaky tests won't have very many different renderings, except perhaps animation tests.)

foolip:
[...]

PR #1188, opened 2019-02-28
Show both before/after pixel colors in the reftest-analyzer

lukebjerring:
## Description
Ensures that we can see any differences.


webapp/components/reftest-analyzer.js
```diff
@@ -157,75 +167,93 @@ class ReftestAnalyzer extends PolymerElement {
...
+    for (const before of [true, false]) {
+      const paths = before ? pathsBefore : pathsAfter;
+      for (let x = 0; x < 5; x++) {
+        paths.push([]);
+        for (let y = 0; y < 5; y++) {
+          const path = document.createElementNS(nsSVG, 'path');
+          const offsetX = x * 50 + 1;
+          const offsetY = y * 50 + 1;
+          if (before) {
+            path.setAttribute('d', `M${offsetX},${offsetY} H${offsetX + 48} L${offsetX},${offsetY + 48} V${offsetY}`);
```
lukebjerring: This is an upper-left half of the square. Below is the lower-right half (cut diagonally).

webapp/components/reftest-analyzer.js
```diff
@@ -157,75 +167,93 @@ class ReftestAnalyzer extends PolymerElement {
...
-      const beforePixels = beforeCtx.getImageData(0, y, before.width, 1).data;
-      const afterPixels = afterCtx.getImageData(0, y, after.width, 1).data;
-      for (let x = 0; x < Math.min(before.width, after.width); x++) {
-        for (let i = 0; i < 4; i++) {
-          const pxlBefore = beforePixels[(x * 4) + i];
-          const pxlAfter = afterPixels[(x * 4) + i];
-          if (pxlBefore !== pxlAfter) {
-            outCtx.fillRect(x, y, 1, 1);
-            break;
+    return this.load(new Promise(resolve => {
```
lukebjerring: This code is unchanged except for the `this.load` so that we show a spinner while it's computing. On my machine, this only shows for like, 100ms. On slower machine it may take longer.

lukebjerring:
Part of https://github.com/web-platform-tests/wpt.fyi/issues/57

PR #1191, opened 2019-03-01
Implement screenshot uploader

Hexcles:
wptscreenshot is a multiprocessing batch uploader of screenshots (for
better performance), which warrants careful testing.

The test setup has a simple Flask server in the background that prints
out the number of screenshots it receives in each request. The tests are
more like end-to-end integration tests: it points WPTScreenshot at the
local test server to record all requests made by worker processes.

Not hooked up to the processor yet.


results-processor/test_server.py
```diff
@@ -0,0 +1,31 @@
...
+import logging
+import sys
+
+import flask
+
+logging.basicConfig(level=logging.ERROR, stream=sys.stdout)
+app = flask.Flask(__name__)
+
+
+@app.route('/api/screenshots/upload', methods=['POST'])
```
lukebjerring: I'm somewhat surprised to see a push model; are we going to be underneath the appengine restrictions on upload file sizes?

results-processor/wptscreenshot.py
```diff
@@ -0,0 +1,115 @@
...
+    global _run_info
+    _api = api
+    _run_info = run_info
+
+
+def _upload(images):
+    files = []
+    for i in range(len(images)):
+        files.append((
+            'screenshot', (('{}.png').format(i), images[i], 'image/png')))
```
lukebjerring: Is `.format` preferred to `'%s.png' % i`?

results-processor/wptscreenshot_test.py
```diff
@@ -0,0 +1,92 @@
...
+            s = WPTScreenshot(f.name, run_info={}, api=self.api)
+            s.start(processes=1)
+            s.process()
+            s.wait()
+        self.server.terminate()
+        _, err = self.server.communicate()
+        sizes = self._batch_sizes(err)
+        self.assertListEqual(sizes, [2])
+
+    def test_multiple_batches(self):
```
lukebjerring: Add a test for a gzipped file.

results-processor/test_server.py
Hexcles: If by push you mean the processor uploading screenshots to `/api/screenshots/upload` (which is in the webapp), then yes it's push. The `MAXIMUM_BATCH_SIZE` is small enough to make sure we are always far away from the request size limit (100 * 100KB = 10MB < 32MB, and rarely we see screenshots close to 100KB). I could implement a higher but flexible (based on sizes) batch size, but that'd be an overkill given batch size of 100 already gives us superb performance.

results-processor/wptscreenshot.py
Hexcles: Not really. They have subtle differences which don't matter in this case. Let me change it to `%` for consistency.

results-processor/wptscreenshot_test.py
Hexcles: Good advice. This in fact helped me to identify a defect in the implementation: when things go wrong (e.g. invalid gzip or encoding), file handlers won't be closed and we will have gradual resource leak...

Issue #28, opened 2018-04-10
Enable finding the set of tests that pass in 3 or greater browsers


lukebjerring:
_From @jdm on October 3, 2017 13:16_

Servo is a big consumer of web-platform-tests but doesn't really warrant inclusion as a data point (due to instability, lack of meaningful audience, etc.) However, it would be useful to be able to see the set of tests that are passing in most other browsers, and to do so in a way that is easy to consume in a scripted fashion (eg. JSON). Historical data for other browsers is not particularly interesting to us; we're only interested in a snapshot on any given day so we can compare our own test results against the latest data from other browsers.

_Copied from original issue: w3c/wptdashboard#135_

lukebjerring:
_From @foolip on October 3, 2017 16:52_

This kind of filtering is a topic of https://github.com/w3c/wptdashboard/issues/83, but having a JSON endpoint isn't part of the plan so far. Would https://github.com/w3c/wptdashboard/issues/136  solve the problem if it's possible to just download the latest results instead of everything?

Actually, if you run wpt.fyi it will download some data files that may be exactly what you need, I see https://storage.googleapis.com/wptd/85a171624e/chrome-62.0-linux-summary.json.gz and others in developer tools. Can you see if that has the data you need, and if a nicer endpoint for just that might suffice?

lukebjerring:
_From @jdm on November 1, 2017 19:55_

Yes, that data looks like it's thorough enough for what we're looking for.

lukebjerring:
_From @jgraham on November 1, 2017 20:5_

Note that that data is per-file rather than per-subtest so it's generally not possible to tell if each browser is passing the same subset of the tests except in trivial cases (e.g. all subtests have the same behaviour).

PR #1198, opened 2019-03-04
Add 'count' searchcache operator

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/28

Adds support for a `count` operator in searchcache queries, where the number of runs in any given row that satisfy the requirement must be a precise number (count).

e.g. `count:3(status:pass|status:ok)`

Also added `three(...)`, `two(...)`, and `one(...)` aliases for simplicity.


lukebjerring:
Example query:
[three(status:pass|status:ok)](https://count-dot-wptdashboard-staging.appspot.com/results/?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned&q=three%28status%3Apass%7Cstatus%3Aok%29)

api/query/atoms.go
```diff
@@ -78,8 +78,10 @@ func (e AbstractExists) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	queries := make([]ConcreteQuery, len(e.Args))
 	for i, arg := range e.Args {
 		var query ConcreteQuery
+		// For sequential + count, we pass all runs.
 		if _, isSeq := arg.(AbstractSequential); isSeq {
-			// For sequential, we pass all runs.
+			query = arg.BindToRuns(runs...)
+		} else if _, isCount := arg.(AbstractCount); isCount {
```
KyleJu: Looks like count operator is only implemented at the root level. If I understand correctly, a case like
 exists: [{
          and: [
            {pattern: 'a'},
           {
          count: 5,
          where: {
            or: [{status: 'PASS'}, {status: 'OK'}],
          },
        }
      ]}]

api/query/atoms.go
```diff
@@ -529,6 +551,35 @@ func (e *AbstractSequential) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// UnmarshalJSON for AbstractCount attempts to interpret a query atom as
+// {"count": int, "where": query}.
+func (c *AbstractCount) UnmarshalJSON(b []byte) error {
```
KyleJu: what is the reasoning behind BindToRuns implements on a AbstractCount while UnmarshalJSON implements on *AbstractCount?

api/query/atoms_test.go
```diff
@@ -297,6 +297,34 @@ func TestStructuredQuery_sequential(t *testing.T) {
 			}}}, rq)
 }
 
+func TestStructuredQuery_count(t *testing.T) {
+	var rq RunQuery
+	err := json.Unmarshal([]byte(`{
+		"run_ids": [0, 1, 2],
+		"query": {
```
KyleJu: Question outside of this PR - if AbstractExists is the root, should every single query in this test file should start with "exists"?

webapp/components/test-search.js
```diff
@@ -39,9 +39,20 @@ const QUERY_GRAMMAR = ohm.grammar(`
     Q = ListOf<RootExp, space*>
 
     RootExp
-      = "seq(" ListOf<Exp, space*> ")" -- sequential
+      = Sequential
+      | Count
```
KyleJu: count operator exists at the root level?

api/query/atoms.go
```diff
@@ -575,5 +626,10 @@ func unmarshalQ(b []byte) (AbstractQuery, error) {
 	if err == nil {
 		return s, nil
 	}
+	var c AbstractCount
+	err = json.Unmarshal(b, &c)
+	if err == nil {
+		return c, nil
+	}
 	return nil, errors.New(`Failed to parse query fragment as test name pattern, test status constraint, negation, disjunction, or conjunction`)
```
KyleJu: nit: `Failed to parse query fragment as test name pattern, test status constraint, negation, disjunction, conjunction, sequential or count`

api/query/atoms.go
lukebjerring: You pass a ref if A) you're going to mutate the ref or B) it's costly to copy (a large struct).

api/query/atoms_test.go
lukebjerring: No; Server-side, AbstractExists is only the root when `{exists: ...}` is passed in from the client (which, it always is).

webapp/components/test-search.js
lukebjerring: We shouldn't be combining it with other queries; you'd get unexpected behaviour

api/query/atoms.go
lukebjerring: FYI the little +/- icon 
image

Lets you make a suggestion I can apply with the click of a button.

```suggestion
	return nil, errors.New(`Failed to parse query fragment as test name pattern, test status constraint, negation, disjunction, conjunction, sequential or count`)
```

PR #1204, opened 2019-03-06
Add a flag to enable web-platform-tests.live instead of w3c-test.org

### LINKED ISSUE #380: Link to web-platform-tests.live instead of w3c-test.org
https://web-platform-tests.live/ is now live. @jugglinmike will update this issue when it's a state where we can depend on it.

KyleJu:
## Description
Fix #380. Add a flag to enable web-platform-tests.live.

## Review Information
- Visit /flags, enable the feature for web-platform-tests.live
- Visit /results, find a test
- Click on Run in your browser on web-platform-tests.live
- See if it directs to a web-platform-tests.live page and the result is shown properly

### DIFF
--- a/webapp/components/wpt-results.js
+++ b/webapp/components/wpt-results.js
@@ -205,20 +205,10 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
           <ul>
             <li><a href\$="https://github.com/web-platform-tests/wpt/blob/master[[sourcePath]]" target="_blank">View source on GitHub</a></li>
             <template is="dom-if" if="[[ showTestURL ]]">
-              <template is="dom-if" if="[[ !webPlatformTestsLive ]]">
-                <li><a href="[[showTestURL]]" target="_blank">Run in your browser on w3c-test.org</a></li>
-              </template>
-              <template is="dom-if" if="[[ webPlatformTestsLive ]]">
-               <li><a href="[[showTestURL]]" target="_blank">Run in your browser on web-platform-tests.live</a></li>
-              </template>
+              <li><a href="[[showTestURL]]" target="_blank">Run in your browser on [[ liveTestDomain ]]</a></li>
             </template>
             <template is="dom-if" if="[[ showTestRefURL ]]">
-              <template is="dom-if" if="[[ !webPlatformTestsLive ]]">
-                <li><a href="[[showTestRefURL]]" target="_blank">View ref in your browser on w3c-test.org</a></li>
-              </template>
-              <template is="dom-if" if="[[ webPlatformTestsLive ]]">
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-results.js
```diff
@@ -204,11 +204,21 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
...
-              <li><a href="[[testW3CURL]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+            <template is="dom-if" if="{{ showTestURL }}">
+              <template is="dom-if" if="{{ !webPlatformTestsLive }}">
+                <li><a href="[[showTestURL]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+              </template>
+              <template is="dom-if" if="{{ webPlatformTestsLive }}">
+               <li><a href="[[showTestURL]]" target="_blank">Run in your browser on web-platform-tests.live</a></li>
+              </template>
+            </template>
+            <template is="dom-if" if="[[ showTestRefURL ]]">
```
lukebjerring: For the record, it's strange that we mix `[[ property ]]` and `{{ property ]}` syntax - the difference is whether changes to the property (on the destination element) are propagated back to the parent. In this case, it's read-only, so the `[[ property ]]` syntax is appropriate.

webapp/components/wpt-results.js
```diff
@@ -204,11 +204,21 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
         <div class="links">
           <ul>
             <li><a href\$="https://github.com/web-platform-tests/wpt/blob/master[[sourcePath]]" target="_blank">View source on GitHub</a></li>
-            <template is="dom-if" if="{{ testW3CURL }}">
-              <li><a href="[[testW3CURL]]" target="_blank">Run in your browser on w3c-test.org</a></li>
+            <template is="dom-if" if="{{ showTestURL }}">
+              <template is="dom-if" if="{{ !webPlatformTestsLive }}">
```
lukebjerring: You shouldn't need to duplicate the `dom-if` here; you've already forked what the URL computes to (so just revert lines 207 to 212.

KyleJu: I thought I have to dom-if because
1) showTestURL could return empty string as line 530
2) the text when the flag is on/off is different - Run in your browser on w3c-test.org vs Run in your browser on web-platform-tests.live

lukebjerring: Oh, the anchor text. In that case, extract the host out of the anchor's sentence into a variable, i.e.

    Run in your browser on [[ liveTestDomain ]]

(and add a computed property for liveTestDomain).

And then we don't need the dom-if's.

webapp/components/wpt-results.js
KyleJu: Ack

PR #1209, opened 2019-03-07
Rearrange manifest to be test-path centric

### LINKED ISSUE #1208: When manifest is enabled, navigating from a file view to a directory view is very slow
To reproduce:

1. Make sure manifest is enabled in https://staging.wpt.fyi/flags
2. Go to https://staging.wpt.fyi/results/css/css-flexbox
3. Navigate into any file
4. Navigate back

There's no network request between 3 & 4 as the directory view is already in memory, but the navigation is still extremely slow. Profiling shows JS is busy doing `computeTestType`.

lukebjerring:
## Description
Otherwise, it's computationally expensive to find a test path (we have to iterate the whole manifest and look up the `manifest.items[type][file][0]` entry).

Fixes #1208

### DIFF
--- a/webapp/components/wpt-results.js
+++ b/webapp/components/wpt-results.js
@@ -35,6 +35,8 @@ import { WPTFlags } from './wpt-flags.js';
 import './wpt-permalinks.js';
 import './wpt-prs.js';
 
+const TEST_TYPES = ['manual', 'reftest', 'testharness', 'visual', 'wdspec'];
+
 class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRunsUIBase)))) {
   static get template() {
     return html`
@@ -503,10 +505,9 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
     if (!this.computePathIsATestFile(path) || !manifest) {
       return path;
     }
-    // Filter in case any types are fully missing.
     const metadata = manifest.get(path);
     if (metadata) {
-      return metadata.file.startsWith('/') ? metadata.file : `/${metadata.file}`;
+      return metadata.file;
     }
   }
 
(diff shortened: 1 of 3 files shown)


lukebjerring:
Still too slow async; let's rearrange the manifest instead.

Hexcles:
@lukebjerring is this ready for review?

lukebjerring:
Yep

webapp/components/wpt-results.js
```diff
@@ -505,15 +504,10 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
       return path;
     }
     // Filter in case any types are fully missing.
```
Hexcles:
```suggestion
    // Ensure leading slash.
```

webapp/components/wpt-results.js
```diff
@@ -534,12 +528,10 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
...
     }
-    const item = Object.values(manifest.items['reftest']).find(v => v.find(i => i[0] === path));
-    // In item[0], the 2nd item is the refs array, and we take the first ref (0).
-    // Then, the ref's 1st item is the url (0). (2nd is the condition, e.g. "==".)
-    // See https://github.com/web-platform-tests/wpt/blob/master/tools/manifest/item.py#L141
-    const refPath = item && item[0][1][0][0];
-    return this.computeTestURL(testType, refPath);
+    const metadata = manifest.get(path);
+    if (metadata && metadata.refPath) {
+      return this.computeTestW3CURL(testType, metadata.refPath);
```
Hexcles: I don't see `computeTestW3CURL` anywhere in master or this patch? Where's it defined?

webapp/components/wpt-results.js
Hexcles: In fact, do we still need to do it? Looks like the updated `fetchManifestForSHA` already ensures the leading slash.

webapp/components/wpt-results.js
```diff
@@ -851,14 +857,10 @@ class WPTResults extends WPTColors(WPTFlags(SelfNavigation(LoadingState(TestRuns
     // Add an empty row for all the tests known from the manifest.
     const knownNodes = {};
     if (this.manifest && !this.search) {
-      for (const type of Object.keys(this.manifest.items)) {
+      for (const [path, {type}] of Object.entries(this.manifest)) {
         if (['manual', 'reftest', 'testharness', 'wdspec'].includes(type)) {
```
Hexcles: Might be worthwhile to keep the `TEST_TYPES` constant and use it here.

PR #1212, opened 2019-03-08
Set ID before ingesting

lukebjerring:
## Description
Ingestion relies on the ID being set.


lukebjerring:
TBR'ing this fairly crucial bugfix.

PR #1214, opened 2019-03-08
Fix the custom error type MultiError

Hexcles:
The New method has to return an error interface instead of a concrete
type. This is not only in accordance with the best practice (errors are
checked by type assertions[1]), but also avoids a weird issue when
assigning the return value to a predeclared variable of type error:
(error)(nil) != (*MultiError)(nil).

Add a test to prevent regressing this subtle intentional behaviour.

Drive-by fix: remove unnecessary tags in browsers_test.go.

[1]: https://blog.golang.org/error-handling-and-go

### DIFF
--- a/shared/browsers_test.go
+++ b/shared/browsers_test.go
@@ -1,4 +1,4 @@
-// +build small medium large
+// +build small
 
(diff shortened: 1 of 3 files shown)


Hexcles:
TBR'ing

PR #1215, opened 2019-03-08
Handle screenshots in Azure artifacts

Hexcles:
along with some new tests.

I'll test it manually end to end after the staging instance is up.


api/receiver/azure.go
```diff
@@ -35,65 +46,80 @@ func newAzureArtifact(artifactName string, zip io.Reader) (azureArtifact, error)
...
 	}
 	files := make([]*zip.File, 0, len(a.z.File))
 	for _, f := range a.z.File {
-		if reportPathRegex.MatchString(f.Name) {
+		if p.MatchString(f.Name) {
 			files = append(files, f)
 		}
 	}
-	return files, nil
+	return len(files), files, nil
```
lukebjerring: Why is the caller not able to len the array themselves?

api/receiver/azure.go
```diff
@@ -35,65 +46,80 @@ func newAzureArtifact(artifactName string, zip io.Reader) (azureArtifact, error)
...
+// the given slice of *zip.File (the return value is intended to be used by
+// sendResultsToProcessor).
+func gzipReaderFromZip(files []*zip.File) func(int) (io.ReadCloser, error) {
+	return func(i int) (io.ReadCloser, error) {
+		zipR, err := files[i].Open()
+		if err != nil {
+			return nil, err
+		}
+		buf := new(bytes.Buffer)
+		gzipW := gzip.NewWriter(buf)
```
lukebjerring: Seems strange to force the content back into gzip just to be read out again?

api/receiver/azure.go
Hexcles: For convenience, so that its only caller can directly assign the return value to `results, resultFiles, err` (or screenshots) in one line. I wouldn't have written it this way if the function was more generic.

api/receiver/azure.go
Hexcles: Yeah this is not ideal. It's the same as what the code currently does, which is because in a later part of the receiver, the file is unconditionally treated as gzipped (there's a TODO to detect whether the file is gzipped). Let me also add a TODO here.

Hexcles:
Yes, it works! https://staging.wpt.fyi/results/css/css-flexbox?run_id=5429338749009920&run_id=4762797283999744

PR #1216, opened 2019-03-11
Add atom for filtering by test-file contents

lukebjerring:
## Description
Adds support for `contains:foo` for searching file contents.

Leverages a gcloud search index, populated with all the web-platform-tests/wpt files as documents.

Note: Does not currently take into account generated tests (.any.js) - only the original source file(s).

## Review
- Visit https://contents-dot-wptdashboard-staging.appspot.com/
- Search for `contains:something`
e.g. [contains:shadowroot](https://contents-dot-wptdashboard-staging.appspot.com/results/?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned&q=contains%3Ashadowroot)


shared/util.go
```diff
@@ -7,18 +7,58 @@ package shared
...
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/appengine"
 	gaelog "google.golang.org/appengine/log"
 )
 
+// ProjectID is a flag for the Google Cloud Platform project ID, if different from ID detected from metadata service.
+var ProjectID = flag.String("project_id", "", "Google Cloud Platform project ID, if different from ID detected from metadata service")
+
+// InitProjectID changes the ProjectID flag to fall back to the ID fetched from metadata.
+func InitProjectID() {
```
Hexcles: I don't think it's a good idea to move this into `shared`. The `metadata` package is only available on Flex, not on Standard.

api/query/cache/service/main.go
```diff
@@ -235,30 +232,19 @@ func init() {
...
-	}
+	flag.Parse()
+	shared.InitProjectID()
 
 	log.Infof("Serving index with %d shards", *numShards)
 	// TODO: Use different field configurations for index, backfiller, monitor?
 	logger := log.StandardLogger()
 
-	idx, err = index.NewShardedWPTIndex(index.HTTPReportLoader{}, *numShards)
+	idx, err := index.NewShardedWPTIndex(index.HTTPReportLoader{}, *numShards)
```
Hexcles: This is the problem. You are not assigning to the global `idx`, but creating a new local `idx` instead, which caused the readiness check to fail.

shared/util.go
lukebjerring: The package needs to be available in both api/search/cache/service and api/search/cache/index, and it's not very "cache" specific. Just because code doesn't work in both standard/flex doesn't disqualify it from /shared (e.g. datastore_ae & datastore_cloud)

api/query/cache/service/main.go
lukebjerring: Thanks again for spotting this. Digging through logs would never have found it.

lukebjerring:
@Hexcles this is ready for a review now :)

shared/util.go
lukebjerring: Reverted :)

lukebjerring:
@Hexcles - I've published the script I wrote to pre-populate the index @ https://github.com/web-platform-tests/wpt/pull/19640

I ran it with `go run populate_index.go -remote_api_host contents-dot-wptdashboard-staging.appspot.com` (note that we hadn't set up the remote_api correctly before that PR).

lukebjerring:
Note: There seem to be issues with the staging instances of the searchcache; it has trouble producing results, and ends up serving a 502 to the wpt.fyi side. I _was_ temporarily able to see that the code is working after https://github.com/web-platform-tests/wpt.fyi/pull/1216/commits/15aeab9bb11072decc33ba542ecee9470d374315

PR #1219, opened 2019-03-11
Show screenshots in expanded mode

### LINKED ISSUE #1217: Show screenshot or diff thumbnail when "expand" is toggled
Reftest screenshots are coming!

A great addition to what's in staging would be seeing screenshots or their diffs inclined as thumbnails in the "expand" view instead of the "this URL didn't match that URL" message from the report.

lukebjerring:
## Description
Fixes #1217 

Collapses terse/verbose into one file, with a dom-if (much simpler).


Hexcles:
What is the benefit of having the redirect handler? Besides, I'd prefer to have the handler in `api/screenshot`.

lukebjerring:
We need a redirect to send it to the correct bucket. The alternative is to have environment logic on the client somewhere; I prefer server side

lukebjerring:
Alrighty, all dusted off and ready for some scrutiny!

Hexcles:
@lukebjerring looks like a bad rebase/merge? Or did you intentionally refactor `webapp/components/test-file-results-table.js`?

lukebjerring:
Intentional refactor - collapsed the two classes into a single class that has conditional CSS based on the difference (whether it's `verbose`)

webapp/components/test-file-results-table.js
```diff
@@ -0,0 +1,281 @@
...
+    width: 100%;
+    border-collapse: collapse;
+  }
+  th {
+    background: white;
+    position: sticky;
+    top: 0;
+    z-index: 1;
+  }
+  td, .ref-button {
```
Hexcles:
```suggestion
  td {
```

webapp/components/test-file-results-table.js
```diff
@@ -0,0 +1,281 @@
...
+    padding: 0;
+    height: 1.5em;
+  }
+  td code, .ref-button {
```
Hexcles:
```suggestion
  td code {
```

webapp/components/test-file-results-table.js
```diff
@@ -0,0 +1,281 @@
...
+              </a>
+            </template>
+          </td>
+        </template>
+      </tr>
+
+      <template is="dom-if" if="[[verbose]]">
+        <template is="dom-if" if="[[anyScreenshots(row)]]">
+          <tr>
+            <td>Screenshot</td>
```
Hexcles:
```suggestion
            <td class="sub-test-name"><code>Screenshot</code></td>
```

PR #1221, opened 2019-03-12
Handle screenshots in Taskcluster webhook

Hexcles:
The webhook will unconditionally add `screenshot_url` constructed from a hard-coded pattern (see https://github.com/web-platform-tests/wpt/pull/15771) to the request payload to `/api/results/upload` regardless of whether the artifact exists. The results receiver will then try to download the screenshot artifact and will ignore any failures.

There are also some drive-by changes/cleanup.

(Do not merge yet; wait for https://github.com/web-platform-tests/wpt/pull/15771 to land first in case we need to change the filename.)


api/azure/webhook.go
```diff
@@ -62,9 +61,8 @@ func processBuild(aeAPI shared.AppEngineAPI, azureAPI API, owner, repo, sender,
 	log := shared.GetLogger(aeAPI.Context())
 	log.Infof("Fetching %s", artifactsURL)
 
-	slowClient, cancel := aeAPI.GetSlowHTTPClient(time.Minute)
-	defer cancel()
-	resp, err := slowClient.Get(artifactsURL)
+	client := aeAPI.GetHTTPClient()
```
KyleJu: Could you briefly explain the difference between GetHTTPClient and GetSlowHTTPClient, as well as what does defer cancel() do?

Hexcles: The difference is the timeout. The default timeout is 5s, while `GetSlowHTTPClient` allows you to specify a timeout (and if you do, you'd need to call `cancel()` to destroy the timeout context when you're done). Here we are only fetching a small JSON from an API so we don't really need 1min timeout.

PR #1222, opened 2019-03-12
More no-op code cleanup

Hexcles:
It'd be easier to review each individual commit as the changes are explained in the commit messages.

No behaviours should be affected. No non-mechanical/non-trivial changes were mixed in. Tests should pass.

### DIFF
--- a/api/azure/webhook_test.go
+++ b/api/azure/webhook_test.go
@@ -4,115 +4,14 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-package azure_test
+package azure
 
 import (
-	"context"
 	"encoding/json"
-	"net/http"
-	"net/http/httptest"
-	"net/url"
-	"strings"
 	"testing"
 
-	"github.com/golang/mock/gomock"
-	"github.com/google/go-github/github"
-	logrustest "github.com/sirupsen/logrus/hooks/test"
 	"github.com/stretchr/testify/assert"
-
-	"github.com/web-platform-tests/wpt.fyi/api/azure"
-	"github.com/web-platform-tests/wpt.fyi/api/azure/mock_azure"
-	"github.com/web-platform-tests/wpt.fyi/shared"
-	"github.com/web-platform-tests/wpt.fyi/shared/sharedtest"
 )
-
-func TestHandleCheckRunEvent(t *testing.T) {
-	mockCtrl := gomock.NewController(t)
-	defer mockCtrl.Finish()
-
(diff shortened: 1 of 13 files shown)


shared/appengine.go
```diff
@@ -27,70 +27,79 @@ import (
...
-// context.
-func (a AppEngineAPIImpl) GetSlowHTTPClient(timeout time.Duration) (*http.Client, context.CancelFunc) {
+func (a appEngineAPIImpl) GetSlowHTTPClient(timeout time.Duration) (*http.Client, context.CancelFunc) {
 	slowCtx, cancel := context.WithTimeout(a.ctx, timeout)
 	return urlfetch.Client(slowCtx), cancel
 }
 
-// GetGitHubClient returns a github client using the stored API token.
-func (a AppEngineAPIImpl) GetGitHubClient() (*github.Client, error) {
+func (a *appEngineAPIImpl) GetGitHubClient() (*github.Client, error) {
```
KyleJu: what is the reasoning behind using *appEngineAPIImpl here? the only case I see is to help update githubClient but its a githubClient *github.Client

api/azure/webhook_test.go
```diff
@@ -4,116 +4,15 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-package azure_test
```
KyleJu: Briefly read about the implication of using azure_test vs azure, and the difference is being that using azure_test will restrict the test from using non-exported identifiers. In the past experience with Java testing, its almost preferable to only use exported variables/functions in the tests and test from the perspective of a package user, instead of opening "back door" to allow private variables to be tested.

Hexcles: > In the past experience with Java testing, its almost preferable to only use exported variables/functions in the tests and test from the perspective of a package user

That is true, but the convention in Go is to prefer having tests in the same package as the source, as seen in standard libraries and examples. In fact, Go generally requires all files in the same directory to be in the same package (it'll complain otherwise). Since Go differentiates exported/private names by capitalization, it should be pretty clear if you are accessing/testing a private method in a test. In other words, it won't happen "by accident" as in Java, but rather intentionally (authors can make the judgement call to depend on private stuff in tests if necessary).

`pacakge_test` is really a very special case in Go -- the only time Go does not complain when there are multiple packages in the same directory. It has legitimate uses, to enforce black-box testing, and more commonly to work around cyclic dependency (e.g. `util` depends on `shared`, and you're writing a test for `shared` that depends on `util`).

PR #1223, opened 2019-03-13
Protect the screenshot upload API

Hexcles:
Export an authentication function from the receiver package and use it in the screenshot package to guard the private API.

The branch is based on #1222 .

### DIFF
--- a/api/receiver/create_run.go
+++ b/api/receiver/create_run.go
@@ -23,12 +23,10 @@ const InternalUsername = "_processor"
 
 // HandleResultsCreate handles the POST requests for creating test runs.
 func HandleResultsCreate(a API, s checks.API, w http.ResponseWriter, r *http.Request) {
-	username, password, ok := r.BasicAuth()
-	if !ok || username != InternalUsername || !a.AuthenticateUploader(username, password) {
-		http.Error(w, "Authentication error", http.StatusUnauthorized)
+	if AuthenticateUploader(a, r) != InternalUsername {
+		http.Error(w, "This is a private API.", http.StatusUnauthorized)
 		return
 	}
-
 	body, err := ioutil.ReadAll(r.Body)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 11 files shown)


foolip:
@KyleJu can you review this?

Hexcles:
@foolip Kyle might still need some time to ramp up. This PR is rather important and is blocking the prod launch. Perhaps you should review it?

KyleJu:
@foolip @Hexcles I can have a look in terms of the golang syntax but it gonna take me a long time to ramp up on what its actually doing

api/receiver/create_run.go
```diff
@@ -23,12 +23,10 @@ const InternalUsername = "_processor"
 
 // HandleResultsCreate handles the POST requests for creating test runs.
 func HandleResultsCreate(a API, s checks.API, w http.ResponseWriter, r *http.Request) {
-	username, password, ok := r.BasicAuth()
-	if !ok || username != InternalUsername || !a.AuthenticateUploader(username, password) {
-		http.Error(w, "Authentication error", http.StatusUnauthorized)
+	if !AuthenticateProcessor(a, r) {
+		http.Error(w, "This is a private API.", http.StatusUnauthorized)
```
foolip: s/private/authenticated/ and point to documentation?

api/receiver/receive_results.go
```diff
@@ -34,29 +34,31 @@ const DownloadTimeout = time.Second * 10
...
-	if uploader == "" {
+	var uploader string
+	if a.IsAdmin() {
 		uploader = r.FormValue("user")
 		if uploader == "" {
 			http.Error(w, "Cannot identify uploader", http.StatusBadRequest)
 			return
 		}
+	} else {
+		username, password, ok := r.BasicAuth()
```
foolip: This looks a lot like the code in api.go. Can that code return the authenticated user on success and otherwise error to make it possible to share this code? Or just have the bool-vending version be a type cast on top of this?

foolip:
@Hexcles done, sorry I didn't review sooner.

api/receiver/create_run.go
Hexcles: No this is indeed private in the sense that no one should ever use it except the processor service.

Hexcles:
@foolip thanks for reviewing. Another big one is on the way to fix #1227 ; expect to see it on Monday :)

foolip:
> Another big one is on the way to fix #1227 ; expect to see it on Monday :)

Looking forward to it, or at least the outcome :)

api/receiver/receive_results.go
Hexcles: Done.

PR #1224, opened 2019-03-13
Fix mockgen: remove an unused & broken mock

Hexcles:
With the new 1.12 release of `mockgen`, it's now failing because of `Loading input failed: loading package failed` when generating mock for `query.go`.

I'm not exactly sure what exactly is wrong, but since we don't need `query_mock.go` anyway, the easiest & safest fix is to remove it.


Hexcles:
TBR @KyleJu to fix master

PR #1226, opened 2019-03-15
Give a longer timeout to download artifacts

Hexcles:
Azure is now producing much larger artifacts because of the screenshots, so we need to increase the download timeout a bit to accommodate that.


Hexcles:
TBR @foolip

foolip:
LGTM

PR #1229, opened 2019-03-18
Move fetching of result_url payload into processor

### LINKED ISSUE #1227: Azure artifacts exceed urlfetch limit and can't be ingested
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=9985 is the latest scheduled run.

It failed with "API error: HTTP 400: Failed to fetch Azure artifact".

This may have been fixed by https://github.com/web-platform-tests/wpt.fyi/pull/1226, but it would also be good to ingest the missed runs.

Hexcles:
## Background

The results receiver supports the [`result_url`](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md#url-payload) payload. Currently, the Go API fetches the URL(s) and stores the artifact(s) into the GCS buffer bucket, same as the `result_file` payload. Then the GCS paths are passed to the processor for both types of payloads.

## Changes

This is now becoming problematic as the screenshots push the sizes of artifacts over 32MB, the limit of `urlfetch` response size (#1227 ). The PR solves the problem by moving the fetching of `result_url`, as well as the special handling of Azure artifacts, into the processor. The Go API will now directly pass `result_url` to the processor for it to download (Go API will continue to store the directly uploaded `result_file` in GCS buffer as before).

This PR will fix #1227 without causing externally visible changes in APIs.

## Review

There'll be a few new tests in the processor, but the PR is otherwise ready for review.


api/receiver/receive_results.go
```diff
@@ -71,138 +61,92 @@ func HandleResultsUpload(a API, w http.ResponseWriter, r *http.Request) {
...
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusBadRequest)
+			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
+	} else if artifactName := getAzureArtifactName(r.PostForm.Get("result_url")); artifactName != "" {
+		// Special Azure case for result_url payload
+		azureURL := r.PostForm.Get("result_url")
+		log.Debugf("Found Azure URL: %s", azureURL)
+		extraParams["azure_url"] = azureURL
```
foolip: This is used in processor.py. The change there is so large it's hidden by default.

api/receiver/azure_test.go
```diff
@@ -7,67 +7,11 @@
...
-	"path"
-	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 )
 
-// artifact_test.zip has a single dir, artifact_test/, containing 4 files: wpt_report_{1,2}.json, wpt_screenshot_{1,2}.txt.
-
-func readZip(t *testing.T) *os.File {
```
foolip: I guess these are the tests that need to be rewritten in Python for processor.py?

results-processor/processor.py
```diff
@@ -18,200 +21,282 @@
...
+        only a best effort to avoid duplicate runs.
+        """
+        q = self._datastore.query(kind='TestRun')
+        q.add_filter('RawResultsURL', '=', self.raw_results_url)
+        q.keys_only()
+        run = list(q.fetch(limit=1))
+        return len(run) > 0
+
+    @staticmethod
+    def known_extension(path):
```
foolip: Looks like this is used to give temporary files the right extension. But does that matter, can't they be called anything as long as they have the right content?

(This is in the first refactoring commit, but seems like it's new?)

results-processor/processor.py
```diff
@@ -18,200 +21,282 @@
...
+        os.close(fd)
+        return path
+
+    def _download_single(self, uri):
+        if uri.startswith('gs://'):
+            return self._download_gcs(uri)
+        return self._download_http(uri)
+
+    def _download_azure(self, azure_url):
+        artifact = self._download_http(azure_url)
```
foolip: If this fails I guess the task will be retried later?

results-processor/processor.py
```diff
@@ -18,200 +21,282 @@
...
+            self.report.load_file(r)
+
+    def upload_raw(self):
+        """Uploads the merged raw JSON report to GCS."""
+        with tempfile.NamedTemporaryFile(
+                suffix='.json.gz', dir=self._temp_dir) as temp:
+            self.report.serialize_gzip(temp.name)
+            gsutil.copy(temp.name, self.raw_results_gs_url, gzipped=True)
+
+    def upload_split(self):
```
foolip: Not reviewing this since it's in the refactoring commit and is similar to old code, even if this name is new.

api/receiver/azure_test.go
Hexcles: Correct.

results-processor/processor.py
Hexcles: Yes. And it matters because the loaders (both for screenshots and for reports) rely on file extensions instead of magic bytes to detect gzip. This is borderline "refactor": since the API requires uploaded files to be gzipped, the non-gzipped code path won't be exercised in reality. And the exact refactoring would be always returning ".gz" here. Sorry for making the already big change more confusing.

results-processor/processor.py
Hexcles: Yes. It will throw and the server will return 500, which will cause TaskQueue to retry.

results-processor/processor.py
Hexcles: Yes this (as well as `upload_raw`) are indeed pure refactoring.

results-processor/processor.py
foolip: Nah, it was nicely split, just this bit caught my eye :)

api/receiver/azure_test.go
Hexcles: Done. And even more tests for various downloading functions.

PR #1230, opened 2019-03-18
[processor] Upgrade pip packages

Hexcles:
Trivial mechanical upgrade to get rid of some deprecation warnings from grpc. Tests should pass.


gsnedders:
I wonder if it's worthwhile running pyup.io on this?

PR #1232, opened 2019-03-19
Fix #1225 - set SVG rect width through attribute rather than CSS

### LINKED ISSUE #1225: Diffing feature of reftest analyzer does not work on Firefox
Example: https://staging.wpt.fyi/analyzer?screenshot=sha1%3A339d253814be71ad2052253c38020082d003598c&screenshot=sha1%3A26802d00e57a78123a7a789c81502f0cafee27ab

Reported by @gsnedders

gsnedders:
## Description
Fixes #1225 

## Review Information
SVG2 requires CSS to be supported for this, but it hasn't historically been supported, and isn't in Firefox, so let's not rely on that.

### DIFF
--- a/webapp/components/reftest-analyzer.js
+++ b/webapp/components/reftest-analyzer.js
@@ -44,11 +44,6 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
         #source.after #before {
           display: none;
         }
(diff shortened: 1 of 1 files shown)


Hexcles:
Thank you!

PR #1235, opened 2019-03-19
[processor] Upgrade to Python 3.6

Hexcles:
Nothing is changed except our test setup.

Unfortunately, the dev Docker image we use does not have Python 3.6 (it's running Debian stretch). Instead of using debports etc. to get Python 3.6 into stretch, I changed the Travis configuration for the `python_test` target to run tests outside of Docker using Travis' native Python support.

### DIFF
--- a/results-processor/README.md
+++ b/results-processor/README.md
@@ -1,0 +1,28 @@
+## Basics
+
+The results processor runs on Python 3.6. The entry point is a Flask web server
+(`main.py`). In production, gunicorn is used as the WSGI (see `Dockerfile`) and
+the container runs as a custom AppEnging Flex instance (see `app.yaml`).
+
+## Getting started
+
+We can create a virtualenv to recreate a setup close to production for daily
+development.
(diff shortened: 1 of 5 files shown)


results-processor/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+```bash
+virtualenv env -p python3.6
+. env/bin/activate
+pip install -r requirements.txt
+```
+
+## Running tests
+
+We strongly recommend you to run tests **outside of** the virtualenv created
```
foolip: By just reading these docs I don't understand where the nested virtualenvs are. There's just one activate step described above.

Hexcles: `tox` will automatically create one, so if you're still in the virtualenv created above, running tox gives you a nested virtualenv, which is usually fine but occasionally you run into weird PATH issues.

foolip: Oh right, tox.

Issue #1228, opened 2019-03-15
Missing features in reftest analyzer


Hexcles:
From @jgraham on IRC:

> 6:22:51 AM	jgraham  Hexcles: Even in Chrome you're missing the feature that tells you the actual pixel values, which is important for making fuzzy annotations
> 6:23:00 AM	jgraham  Other features I'd like:
> 6:23:16 AM	jgraham  * Click to lock the zoom at the current point
> 6:23:46 AM	jgraham  * Summary of total pixels different / max pixels different (you could even write a fuzzy annotation for people to copy)

jgraham:
To be clear, only the first feature is required for parity, but the latter are features I often miss in the gecko analyzer.

gsnedders:
Having some sort of zoom would be very nice for looking at single-pixel differences, right now it's very easy to miss it trying to move to the cursor to the appropriate place.

PR #1239, opened 2019-03-22
Improve reftest analyzer

Hexcles:
## Description

1. Show the pixel position & RGB values of the current position (#1228).
2. Mask same pixels even more to highlight the border of the different areas.

## Review Information

Demo: https://reftest-analyzer2-dot-wptdashboard-staging.appspot.com/analyzer?screenshot=sha1%3A8cfce4c0c8e873c12e613a94e53c8c0a8e44eb3e&screenshot=sha1%3Ac8fbf3067a700b1916632ec1eed6c379a7113efd

### DIFF
--- a/webapp/components/reftest-analyzer.js
+++ b/webapp/components/reftest-analyzer.js
@@ -29,6 +29,9 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
           height: 250px;
           width: 250px;
         }
+        #zoom #info {
+          width: 280px;
+        }
         #display {
           position: relative;
           height: 800px;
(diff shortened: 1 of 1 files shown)


Hexcles:
cc @jgraham

foolip:
Still doesn't work in Firefox, I guess this change wasn't intended to fix that?

webapp/components/reftest-analyzer.js
```diff
@@ -52,12 +55,18 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
...
+      <div id="zoom">
         <svg xmlns="http://www.w3.org/2000/svg" shape-rendering="optimizeSpeed">
           <g id="zoomed">
             <rect width="250" height="250" fill="white"/>
           </g>
         </svg>
+
+        <div id="info">
+          <strong>Pixel at:</strong> [[curX]], [[curY]] <br>
+          <strong>Image before:</strong> [[getRGB(canvasBefore, curX, curY)]] <br>
```
foolip: Can we change this to "Test" and "Reference" or something, and eventually the test URLs?

Hexcles:
@foolip no, it should totally work on Firefox; the original issue was fixed, but I introduced another one in this PR... Firefox would hard fail if the coordinations passed to `getImageData` are invalid.

webapp/components/reftest-analyzer.js
Hexcles: We should change the display text along with the DOM identifiers together to avoid confusion. Besides, I'll need to double check whether "before" is always guaranteed to be the test image. I'll do it in another PR.

PR #1244, opened 2019-03-26
[processor] Handle HTTP errors better

Hexcles:
1. Enforce a timeout when issuing HTTP requests (the default is no
   timeout, which is not good).
2. When a HTTP error occurs (timeout or unsuccessful status), we wait 1
   second and retry again.
3. If no results are downloaded successfully, we abort early.

### DIFF
--- a/results-processor/processor.py
+++ b/results-processor/processor.py
@@ -8,6 +8,7 @@ import re
 import shutil
 import sys
 import tempfile
+import time
 import traceback
 import zipfile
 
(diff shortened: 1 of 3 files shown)


results-processor/processor.py
```diff
@@ -19,12 +20,13 @@
...
 from wptscreenshot import WPTScreenshot
 
-
 _log = logging.getLogger(__name__)
 
 
 class Processor(object):
     USERNAME = '_processor'
+    # Timeout waiting for remote HTTP servers to respond
+    TIMEOUT_WAIT = 10
```
foolip: Isn't this too short to fetch a zipfile full of screenshots over a slow network? Do you know what the actual times are in the processor before this change?

Or is the timeout only for the server to send *headers* and the response body can be arbitrarily slow after that?

Hexcles: This is the timeout waiting for the server to start sending any byte. Unfortunately, I don't see an easy way to control the overall timeout provided by `requests`.

foolip: Alright, good in this context. I guess if a server begins responding, then not continuing to send data will eventually cause the same type of error as if a network cable is cut, some TCP-level timeout.

Issue #1245, opened 2019-03-27
Reftest "compare" results seem to be showing broken images


tabatkins:
In a test like <https://staging.wpt.fyi/results/css/css-content/element-replacement.html?label=master&product=chrome%5Bexperimental%5D&product=edge&product=firefox%5Bexperimental%5D&product=safari%5Bexperimental%5D&aligned>, the reftest "compare" button loads up a page that just shows broken images <https://staging.wpt.fyi/analyzer?screenshot=sha1%3A3e5e80df178687ec83e24ecfe57582a99992ad17&screenshot=sha1%3A232ebf37746e2fbd2b27f003c2cd9196632ba4bc>.

tabatkins:
ping @Hexcles

Hexcles:
This run is from BuildBot. I forgot to change results-collection to enable the screenshot logging.

foolip:
@Hexcles I guess additional work needed for https://github.com/web-platform-tests/wpt/pull/16107 and https://github.com/web-platform-tests/wpt/pull/16108 to get screenshots as well?

Hexcles:
#16107 is fine; the necessary command line arg was already added. I commented on #16108 .

Hexcles:
I'm closing this as "fixed", because 1. new runs should have screenshots; 2. when historical runs are missing screenshots, we now show an error message with a link to file an issue.

PR #1246, opened 2019-03-27
Show an error message when screenshots can't load

Hexcles:
Hope to alleviate #1245 .

### DIFF
--- a/webapp/components/reftest-analyzer.js
+++ b/webapp/components/reftest-analyzer.js
@@ -88,8 +88,8 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
         upload screenshots despite having screenshot checksums in the reports.</p>
 
         <div id="display">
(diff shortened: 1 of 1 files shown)


webapp/components/reftest-analyzer.js
```diff
@@ -80,9 +85,12 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
         </div>
 
 
+        <p id="error-message">Failed to load images. Some historical runs and some runners did not
+        upload screenshots despite having screenshot checksums in the reports.</p>
```
foolip: The bit about checksums is a bit much information for someone who doesn't know our infrastructure. Can you just say that screenshots should be available for all runs from 2019-04-01 and may be missing for older runs? Then maybe a link to file a bug?

Hexcles: Done. I created an issue template.

PR #1247, opened 2019-03-27
Backend structure for displaying metadata info on wpt-result page

KyleJu:
## Description
Collect all Metadata YML files from Wpt-metadata repository and return to the frontend if the showMetadataInfo flag is set.

## Review Information
Visit /flags
Enable the feature - click on "Show metadata Information on wpt.fyi result page"
Execute any api/search query request, metadata data should be return along with test run results.

Notes:
- The return format will matched to the results fields - test name maps to a list bug url list, whose index corresponds to browser type requested.

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -202,7 +202,6 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if showMetadata, _ := shared.ParseBooleanParam(urlQuery, shared.ShowMetadataParam); showMetadata != nil && *showMetadata {
-		log.Warningf("*********************GOing into metadata part***********")
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -200,6 +200,15 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	if len(missing) != 0 {
 		resp.IgnoredRuns = missing
 	}
+
+	if _, showMetadata := urlQuery["metadataInfo"]; showMetadata {
+		// Five seconds default timeout.
+		var netClient = &http.Client{
+			Timeout: time.Second * 5,
+		}
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient)
+	}
+
 	data, err = json.Marshal(resp)
 	if err != nil {
 		http.Error(w, "Failed to marshal results to JSON", http.StatusInternalServerError)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -205,7 +205,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
+		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -201,12 +201,11 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		resp.IgnoredRuns = missing
 	}
 
-	if showMetadata, _ := shared.ParseBooleanParam(urlQuery, shared.MetadataKey); showMetadata != nil && *showMetadata {
+	if showMetadata, _ := shared.ParseBooleanParam(urlQuery, shared.ShowMetadataParam); showMetadata != nil && *showMetadata {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		ctx := context.Background()
-		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, shared.GetLogger(ctx))
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -26,7 +26,7 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/datastore"
-	log "github.com/sirupsen/logrus"
+	log "github.com/Hexcles/logrus"
 )
 
 var (
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -201,11 +201,12 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		resp.IgnoredRuns = missing
 	}
 
-	if _, showMetadata := urlQuery[shared.MetadataFlag]; showMetadata {
+	if _, showMetadata := urlQuery[shared.MetadataKey]; showMetadata {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
+		ctx := context.Background()
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, shared.GetLogger(ctx))
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -201,8 +201,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		resp.IgnoredRuns = missing
 	}
 
-	if _, showMetadata := urlQuery["metadataInfo"]; showMetadata {
-		// Five seconds default timeout.
+	if _, showMetadata := urlQuery[shared.MetadataFlag]; showMetadata {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -202,6 +202,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if showMetadata, _ := shared.ParseBooleanParam(urlQuery, shared.ShowMetadataParam); showMetadata != nil && *showMetadata {
+		log.Warningf("*********************GOing into metadata part***********")
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -205,7 +205,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient)
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log)
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -201,7 +201,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		resp.IgnoredRuns = missing
 	}
 
-	if _, showMetadata := urlQuery[shared.MetadataKey]; showMetadata {
+	if showMetadata, _ := shared.ParseBooleanParam(urlQuery, shared.MetadataKey); showMetadata != nil && *showMetadata {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -205,7 +205,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log)
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/query.go
+++ b/api/query/query.go
@@ -144,11 +144,6 @@ func getMemcacheKey(testRun shared.TestRun) string {
 }
 
 func isRequestCacheable(r *http.Request) bool {
-	q := r.URL.Query()
-	if _, showMetadata := q[shared.MetadataFlag]; showMetadata {
-		return false
-	}
-
 	if r.Method == "GET" {
 		ids, err := shared.ParseRunIDsParam(r.URL.Query())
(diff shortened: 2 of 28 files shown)


shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
...
+
+			var fullTestName = ""
+			if strings.HasPrefix(link.TestPath, folderPath) {
+				fullTestName = link.TestPath
+			} else {
+				fullTestName = path.Join(folderPath, link.TestPath)
+			}
+
+			if _, ok := testMap[fullTestName]; !ok {
+				testMap[fullTestName] = make([]string, len(testRuns))
```
Hexcles: This is both incorrect and unnecessary.

1. The second param of `make` is the length, not the capacity. Yet we don't really know the length at this point, so we might end up with extra empty strings at the end of the slice.
2. We should make use of the zero value of slice and `append`. If a key does not exist in the map, the map will give you a zero value, which is a nil slice in this case. You can actually `append` to a nil slice without running into nil pointer errors.

shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
...
+			isMatches := false
+
+			for _, url := range urlsVal {
+				if url != "" {
+					isMatches = true
+				}
+			}
+
+			// No matching testRuns.
+			if !isMatches {
```
Hexcles: Once you've done the above, you can simply check whether `urlsVal` is nil here.

shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
...
+				fullTestName = path.Join(folderPath, link.TestPath)
+			}
+
+			if _, ok := testMap[fullTestName]; !ok {
+				testMap[fullTestName] = make([]string, len(testRuns))
+			}
+			urls = testMap[fullTestName]
+
+			for i, run := range testRuns {
+				// Only matches browser type.
```
Hexcles: Why?

api/query/cache/service/main.go
```diff
@@ -200,6 +200,15 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	if len(missing) != 0 {
 		resp.IgnoredRuns = missing
 	}
+
+	if _, showMetadata := urlQuery["metadataInfo"]; showMetadata {
```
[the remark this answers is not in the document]
KyleJu: sounds good. I created a const MetadataFlag in metadata.go

shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
 	TestPath string `yaml:"test"`
 	URL      string
 }
+
+func (m MetadataResults) Len() int           { return len(m) }
+func (m MetadataResults) Swap(i, j int)      { m[i], m[j] = m[j], m[i] }
+func (m MetadataResults) Less(i, j int) bool { return m[i].String() < m[j].String() }
```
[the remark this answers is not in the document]
KyleJu: Right!

shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
...
+
+func (m MetadataResult) String() string {
```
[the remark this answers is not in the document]
KyleJu: I created this String() to break ties for Less(i, j, int) bool function but don't think I need it anymore because of your suggestion.

shared/metadata.go
```diff
@@ -20,3 +50,94 @@ type MetadataLink struct {
...
+func constructMetadataResponse(testRuns []TestRun, metadata map[string]Metadata) MetadataResponse {
+	res := MetadataResults{}
+	for folderPath, data := range metadata {
+		testMap := make(map[string][]string)
+
+		for _, link := range data.Links {
+			var urls []string
+
+			var fullTestName = ""
+			if strings.HasPrefix(link.TestPath, folderPath) {
```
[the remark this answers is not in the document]
KyleJu: Yea I agree on a relative path. Its also defined at https://docs.google.com/document/d/1oWYVkc2ztANCGUxwNVTQHlWV32zq6Ifq9jkkbYNbSAg.

shared/metadata.go
KyleJu: The reasoning here is because I want to return Metadata in a similar fashion to rest results, e.g. when querying for 4 browsers, the metadata response will look like 

test: "IndexedDB/bindings-inject-key.html"
urls: ["bugs.chromium.org/p/chromium/issues/detail?id=934844", "", "", ""]

while a test result looks like

test: "/2dcontext/building-paths/canvas_complexshapes_arcto_001.htm"
legacy_status: [{passes: 0, total: 1}, {passes: 0, total: 1}, {passes: 0, total: 1}, {passes: 0, total: 1}

So if there are empty strings, it just means the corresponding browser for this test does not have any bug associated with it. Len(testRuns) gives me both the length and the capacity in this case.

shared/metadata.go
KyleJu: If I understand what we discussed in the meeting correctly, there is no versioning in the Metadata YML files - Browser version or WPT Test Version (everything against the tip of the tree). When users query metadata against browsers, we will simply return the bug that fulfills the browser type. I guess there are two implications

1) I would assume its user's responsibility to check the browser version this bug files against;
2) User won't file the same bug against two different versions of a same browser.

Well obviously it is a big flaw, I think I do need to flash out more details on this issue.

shared/metadata.go
KyleJu: relate to the comment above

shared/metadata.go
Hexcles: Ah I see. You're right!

shared/metadata.go
Hexcles: Instead of matching only the browser part of the product spec, I'd suggest encouraging users not to put versions in the product specs when they are annotating. Essentially, we leave the option up to the users. If they have legitimate reasons to include browser versions in META.yml, we should let them do so.

shared/metadata.go
```diff
@@ -20,3 +54,81 @@ type MetadataLink struct {
...
+
+// parseMetadata collects and parses all META.yml files from
+// wpt-metadata reposiroty.
+func parseMetadata(metadataByteMap map[string][]byte) map[string]Metadata {
+	var metadataMap = make(map[string]Metadata)
+	for path, data := range metadataByteMap {
+		var metadata Metadata
+		err := yaml.Unmarshal(data, &metadata)
+		if err != nil {
+			log.Warningf("Failed to unmarshal %s.yml.", path)
```
Hexcles: This file is in the shared package. We usually try to make sure `shared` can run in both Standard and Flex environments. Different loggers are used in the two environments, so instead of using `logrus` directly, we should take a `shared.Logger` here and let the caller create the logger.

Hexcles:
@KyleJu I don't see you've addressed my open comments?

shared/metadata.go
```diff
@@ -4,6 +4,40 @@
...
+)
+
+// MetadataFlag determines whether Metadata Information returns along
+// with a test result query request.
+// TODO(kyleju): Remove this flag once feauture is launched.
+const MetadataFlag = "metadataInfo"
+
+// MetadataResponse is a response to a wpt-metadata query.
+type MetadataResponse struct {
+	Response MetadataResults
```
Hexcles:
```suggestion
	Response MetadataResults `json:"response"`
```

KyleJu:
> @KyleJu I don't see you've addressed my open comments?
Yesterday when we cleaned up the history, we lost a few commits

Hexcles:
Also, sorry I might have asked this before, but do we have a doc for the format of `META.yml` files? For example, whether the `test` field is a relative path?

Hexcles:
> Yesterday when we cleaned up the history, we lost a few commits

Oops, sorry about that...

shared/metadata.go
Hexcles: Actually, do we really need this extra level?

shared/metadata.go
KyleJu: > Also, sorry I might have asked this before, but do we have a doc for the format of `META.yml` files? For example, whether the `test` field is a relative path?

Right its right here (search for metadata format header) https://docs.google.com/document/d/1oWYVkc2ztANCGUxwNVTQHlWV32zq6Ifq9jkkbYNbSAg/edit#heading=h.7nki9mck5t64

KyleJu: I modified the format of META.yml slightly and paste to the https://github.com/web-platform-tests/wpt-metadata/blob/master/README.

- I set test name field as mandatory 
- When a browser name is not specified, it will match to all browsers

shared/metadata.go
KyleJu: yea we don't need it

api/query/cache/service/main.go
```diff
@@ -200,6 +200,14 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 	if len(missing) != 0 {
 		resp.IgnoredRuns = missing
 	}
+
+	if _, showMetadata := urlQuery[shared.MetadataFlag]; showMetadata {
+		var netClient = &http.Client{
```
lukebjerring: We have a method in `shared` for getting an http client (on the AppEngineAPI).

api/query/cache/service/main.go
```diff
@@ -200,6 +200,14 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
...
+			Timeout: time.Second * 5,
+		}
+		resp.MetadataResult = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
```
lukebjerring: Generally, we get the logger from the context. This lets tests drop logs without issue.

api/query/query.go
```diff
@@ -142,6 +144,11 @@ func getMemcacheKey(testRun shared.TestRun) string {
 }
 
 func isRequestCacheable(r *http.Request) bool {
+	q := r.URL.Query()
+	if _, showMetadata := q[shared.MetadataFlag]; showMetadata {
```
lukebjerring: I would probably still cache the request, even if it's showing metadata; we don't cache forever or anything.
In either case, this would fall under the should-cache, not can-cache (they're separate).

shared/metadata.go
```diff
@@ -4,6 +4,39 @@
...
+	"sort"
+
+	"github.com/go-yaml/yaml"
+	"github.com/web-platform-tests/wpt-metadata/util"
+)
+
+// MetadataFlag determines whether Metadata Information returns along
+// with a test result query request.
+// TODO(kyleju): Remove this flag once feauture is launched.
+const MetadataFlag = "metadataInfo"
```
lukebjerring: I don't think you need to hide this behind a flag server-side - it's controlled by a flag on the client and is not a dangerous feature.

api/query/cache/service/main.go
Hexcles: This is part of the searchcache that runs on AppEngine Flex, so no urlfetch.

shared/metadata.go
KyleJu: This is the key value of Metadata param in the url. I have renamed it to MetadataKey to avoid confusions.

api/query/search.go
```diff
@@ -211,6 +216,11 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 	}
 	runIDsStr := strings.Join(runIDStrs, ",")
 	r2.URL.RawQuery = fmt.Sprintf("run_ids=%s&q=%s", url.QueryEscape(runIDsStr), url.QueryEscape(simpleQ.Pattern))
+
+	if _, showMetadata := q[shared.MetadataKey]; showMetadata {
```
lukebjerring: use the utility for this (I think it's shared.ParseBooleanParam or similar). ?foo=false would still be "present", but should not be considered true.

api/query/search.go
```diff
@@ -223,10 +233,16 @@ func (sh unstructuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
 
 	resp := prepareSearchResponse(filters, testRuns, summaries)
 
+	q := r.URL.Query()
+	if _, showMetadata := q[shared.MetadataKey]; showMetadata {
```
lukebjerring: Same here (shared.ParseBooleanParam)

shared/metadata.go
```diff
@@ -4,6 +4,39 @@
...
+	"path"
+	"sort"
+
+	"github.com/go-yaml/yaml"
+	"github.com/web-platform-tests/wpt-metadata/util"
+)
+
+// MetadataKey determines whether Metadata Information returns along
+// with a test result query request.
+// TODO(kyleju): Remove this flag once feauture is launched.
```
lukebjerring:
```suggestion
```

lukebjerring: (It's not a flag, we'll always have the query param)

api/query/cache/service/main.go
Hexcles: Again, this is on AppEngine Flex. The background context doesn't actually have a logger associated with it, so all you'll get is "Context without logger; logs will be dropped". We do need to use `log.StandardLogger()` here. And it is true that tests will print out a lot of stuff into stdout (which is already happening in search cache & revision announcer), but we should do a separate refactor to address that.

Sorry for the back and forth, Kyle.

shared/metadata.go
```diff
@@ -4,6 +4,38 @@
...
+	"path"
+	"sort"
+
+	"github.com/go-yaml/yaml"
+	"github.com/web-platform-tests/wpt-metadata/util"
+)
+
+// MetadataKey determines whether Metadata Information returns along
+// with a test result query request.
+const MetadataKey = "metadataInfo"
```
Hexcles: Nit: would `ShowMetadataParam` be more readable?

shared/metadata.go
```diff
@@ -4,6 +4,38 @@
...
+
+// MetadataResponse is a response to a wpt-metadata query.
+type MetadataResponse struct {
```
Hexcles: Again, what is the purpose of this struct type? Can't `MetadataResult` simply be of type `shared.MetadataResults`?

api/query/cache/service/main.go
KyleJu: yea No worries!

shared/metadata.go
```diff
@@ -20,3 +47,84 @@ type MetadataLink struct {
...
+				continue
+			}
+
+			linkResult := MetadataResult{Test: nameKey, URLs: urlsVal}
+			res = append(res, linkResult)
+		}
+	}
+	sort.Sort(res)
+	return res
+
```
Hexcles:
```suggestion
```

api/query/search.go
```diff
@@ -223,10 +233,16 @@ func (sh unstructuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Req
...
+	q := r.URL.Query()
+	if showMetadata, _ := shared.ParseBooleanParam(q, shared.ShowMetadataParam); showMetadata != nil && *showMetadata {
+		resp.MetadataResponse = shared.GetMetadataResponse(testRuns, sh.queryHandler.client, sh.queryHandler.logger)
+	}
+
 	data, err := json.Marshal(resp)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 	}
+
```
Hexcles:
```suggestion
```

PR #1253, opened 2019-04-04
Tweak wptreport to accept multiple reports

Hexcles:
This only affects the main entry point of wptreport for more convenience
(i.e. when using wptreport directly from command line).

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -565,7 +565,7 @@ def create_test_run(report, run_id, labels_str, uploader, auth,
 def main():
     parser = argparse.ArgumentParser(
         description='Parse and transform JSON wptreport.')
-    parser.add_argument('report', metavar='REPORT', type=str,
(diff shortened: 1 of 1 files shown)


results-processor/wptreport.py
```diff
@@ -565,7 +565,7 @@ def create_test_run(report, run_id, labels_str, uploader, auth,
 def main():
     parser = argparse.ArgumentParser(
         description='Parse and transform JSON wptreport.')
-    parser.add_argument('report', metavar='REPORT', type=str,
+    parser.add_argument('report', metavar='REPORT', type=str, nargs='+',
```
KyleJu: Is this guaranteed to have at least one report? maybe '*' is a safer choice?

Hexcles: Yes at least one report is required. The CLI interface is used by humans, so this is not a "guarantee" but a "requirement".

Hexcles:
Travis is broken due to #1254. Force-merging it.

PR #1257, opened 2019-04-09
Remove some redundant abstractions

lukebjerring:
## Description
Not sure the exact problem, but mockgen is choking on these files.
Obviated the problem by just deleting the interfaces.

### DIFF
--- a/api/query/cache/backfill/backfill.go
+++ b/api/query/cache/backfill/backfill.go
@@ -2,8 +2,6 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
 
-//go:generate mockgen -destination mock_backfill/backfill_mock.go github.com/web-platform-tests/wpt.fyi/api/query/cache/backfill RunFetcher
-
 package backfill
 
 import (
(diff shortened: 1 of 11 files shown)


Hexcles:
Could you paste the mockgen errors? I don't see mockgen errors on Travis, but only the build error caused by `unsafe`.

lukebjerring:
Can't repro anymore, but as it was, there's nothing to post other than 

> Failed to load package

(mockgen doesn't print anything other than that), so not very informative. 

While not blocked on this change, it does simplify things somewhat, so I'll leave the PR as-is.

Hexcles:
Yeah mockgen was broken at one point and I fixed it. I was wondering if you
got caught in some intermediate state.

api/query/cache/backfill/backfill.go
```diff
@@ -52,31 +38,21 @@ const bytesPerRun = uint64(6.5e+7)
 
 var errNilIndex = errors.New("Index to backfill is nil")
 
-// NewDatastoreRunFetcher constructs a RunFetcher that loads runs from Datastore
+// GetDatastore constructs a RunFetcher that loads runs from Datastore
```
Hexcles:
```suggestion
// GetDatastore constructs a shared.Datastore interface that loads runs from Datastore
```

Hexcles:
Sorry for the delay. I pushed a few commits to fix some nits (including a typo that broke the build) myself.

PR #1259, opened 2019-04-10
[WIP] Use iron-list for the test-file-results component

lukebjerring:
## Description
iron-list is designed to be an "infinite, virtual" list. More performant than dom-repeat (which genuinely creates N DOM elements).

Should help with pages like https://wpt.fyi/results/encoding/api-invalid-label.any.html?label=master&label=stable&aligned
(> 1000 results, takes 15s to render and be responsive)


Hexcles:
Comparing:
* https://iron-list-dot-wptdashboard-staging.appspot.com/results/encoding/api-invalid-label.any.html?label=master&label=stable&aligned
* https://staging.wpt.fyi/results/encoding/api-invalid-label.any.html?label=master&label=stable&aligned

Bug:
* Some CSS rules no longer apply, e.g. line height and ellipses. This is probably an easy fix.
* Search in page cannot find items far out of the page. I guess this is the intrinsic disadvantage of a virtual list?

The performance is indeed *a lot* better.

WDYT about find-in-page? I think it's a deal breaker. When the page is this long, people definitely want to be able to search...

Hexcles:
One idea: use `iron-list` only in terse mode and show a banner to tell users to expand into the verbose mode if they want to search reliably.

PR #1260, opened 2019-04-10
Load more runs on scroll-to-bottom

lukebjerring:
## Description
Removes the 'load more' button in favour of hitting the bottom of the document when scrolling.


Hexcles:
https://infinite-scroll-dot-wptdashboard-staging.appspot.com/runs?label=master doesn't work

```
element-mixin.js:224 Cannot redefine computed property 'query'.
element-mixin.js:224 Cannot redefine computed property 'queryParams'.
element-mixin.js:224 Cannot redefine computed property 'isDefaultProducts'.
element-mixin.js:224 Cannot redefine computed property 'isLatest'.
element-mixin.js:224 Cannot redefine computed property 'resultsRangeMessage'.
element-mixin.js:224 Cannot redefine computed property 'encodedPath'.
```

lukebjerring:
See #1269 - the version "deployed" is actually master (see /about page for proof).

Those are just warnings, not breaking anything (polymer started emitting them recently).

PR #1266, opened 2019-04-15
Use 1+2 keys for toggling before+after on /analyzer

### LINKED ISSUE #1264: Reftest analyzer key bindings
Mozilla's reftest analyzer has keys "1" and "2" bound to the before and after images, respectively. That makes it very easy to toggle back and forth between the two images, and also makes that possible while keeping your pointer over a part of the image you're interested in. (I.e. the zoomed view should update on toggling.) It would be helpful if the WPT reftest analyzer could have the same behavior.

lukebjerring:
## Description
Fixes #1264 

## Review Information
Also drive-by fixes the width when the error message is shown, and handling the spinner stopping onerror.

### DIFF
--- a/webapp/components/reftest-analyzer.js
+++ b/webapp/components/reftest-analyzer.js
@@ -173,17 +173,17 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
     this._createMethodObserver('computeDiff(canvasBefore, canvasAfter)');
 
     // Set the img srcs manually so that we can promisify them being loaded.
-    const beforeImg = this.shadowRoot.querySelector('#before');
-    const afterImg = this.shadowRoot.querySelector('#after');
-    const beforeLoaded = new Promise((resolve) => beforeImg.onload = resolve);
-    const afterloaded = new Promise((resolve) => afterImg.onload = resolve);
-    beforeImg.src = this.before;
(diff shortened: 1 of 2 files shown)


webapp/templates/analyzer.html
```diff
@@ -2,11 +2,23 @@
 <html>
 <head>
 {{ template "_head_common.html" }}
-<script type="module" src="/components/reftest-analyzer.js"></script>
+<script type="module">
+import '/components/reftest-analyzer.js';
+
+document.body.addEventListener('keydown', e => {
+  if (e.which == 49) {
```
Hexcles: Let's add a comment to say they are numkey '1' and '2'.

webapp/components/reftest-analyzer.js
```diff
@@ -101,9 +104,11 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
         </div>
```
Hexcles: Optional: add tooltips to the two radio buttons to show their keyboard bindings.

lukebjerring: I'll avoid this for now as the binding behaviour is added in the page (body keydown), not the component.

lukebjerring:
Merging since comments didn't break anything.

PR #1267, opened 2019-04-16
Migrate webapp codebase to use 1.11 on appengine

### LINKED ISSUE #983: Consider migrating to the second generation AppEngine standard enviroment
https://cloud.google.com/appengine/docs/standard/appengine-generation

This will provide us with several benefits, mostly to eliminate discrepencies between flex & standard environments in our Go codebase.

1. Go 1.11
2. Direct network access without having to use URLFetch
3. Same client libraries for various GCP products (e.g. Datastore, GCS, etc.)
4. Next-gen TaskQueue: Cloud Tasks

Currently the second generation is still in beta and some libraries (e.g. Cloud Tasks) are in alpha. We might want to come back to this when it's a bit more mature.

lukebjerring:
## Description
Following instructions here:
https://cloud.google.com/appengine/docs/standard/go111/go-differences

I suspect this won't work when deployed, but we'll see.

Would fix #983.


Hexcles:
typos in `travis.yml`

Hexcles:
BTW, we might as well move to 1.12 directly. I think the difference between 1.11 and 1.12 is small, and it'd buy us more support life from AppEngine.

lukebjerring:
> If your application uses the App Engine Go SDK APIs, please upgrade to Go 1.11.

Think we will need to step.

lukebjerring:
Note: appengine uses the go.mod file as an indicator for the application's root, which breaks some assumptions about paths for files being read by the app, so I've got a go.mod for wpt.fyi/webapp too.

lukebjerring:
So it's worth noting there's supposed to be a cool meta command that can help show you what your `.gcloudignore` file is doing. It doesn't work for this case:

    gcloud meta list-files-for-upload webapp/web

(It just lists stuff in that dir). When `gcloud app deploy`ing, the go.mod file is treated as the root, but the .gcloudignore file that's siblings with the app.yaml file is what's used for ignoring.

Workaround: symlink `wpt.fyi/.gcloudignore` to `wpt.fyi/webapp/web/.gcloudignore`, and run `gcloud meta list-files-for-upload` from wpt.fyi's root.

lukebjerring:
Huzzah! https://github.com/web-platform-tests/wpt.fyi/pull/1267/commits/166e8377aa329c81da90912ff342169fa12294f6 produced a working deployment (see deployed webapp environment)

Makefile
```diff
@@ -42,12 +42,13 @@ prepush: go_build go_test lint
...
-go_build: git mockgen $(WPTD_GO_PATH)
-	cd $(WPTD_GO_PATH); go get -v ./...
-	cd $(WPTD_GO_PATH); go generate ./...
+go_build: git mockgen packr
+	cd $(WPTD_PATH); go get -v ./...
+	cd $(WPTD_PATH); go generate ./...
+	cd $(WPTD_PATH); go build ./...
 
 go_build_test: go_build gcc
-	cd $(WPTD_GO_PATH); go get -v -t -tags="small medium large" ./...
```
Hexcles: I tried this locally, outside of Docker, with Go 1.13 and the `-t` flag was needed to get the test packages.

I dug a bit deeper and found this annoying back-and-forth:
* Modules were formally introduced in Go 1.11 along with a module-aware mode of `go get` in which `-t` was a no-op: https://github.com/golang/go/blob/83bd88129ea3b5a1272c06db3e865ee75740cee8/src/cmd/go/alldocs.go#L2445
* Go devs reverted the change in 1.13 to make `go get` behave more consistently with and without modules and `-t` is required again: https://golang.org/doc/go1.13#modules:
> By default, go get -u in module mode upgrades only non-test dependencies, as in GOPATH mode. It now also accepts the -t flag, which (as in GOPATH mode) causes go get to include the packages imported by tests of the packages named on the command line.

Makefile
```diff
@@ -42,12 +42,13 @@ prepush: go_build go_test lint
...
+	cd $(WPTD_PATH); go get -v -tags="small medium large" ./...
```
lukebjerring:
```suggestion
	cd $(WPTD_PATH); go get -v -t -tags="small medium large" ./...
```

Makefile
```diff
@@ -234,8 +240,8 @@ eslint: webapp_node_modules_all
 
 dev_data: FLAGS := -remote_host=staging.wpt.fyi
 dev_data: git
-	cd $(WPTD_GO_PATH)/util; go get -t ./...
-	go run $(WPTD_GO_PATH)/util/populate_dev_data.go $(FLAGS)
+	cd $(WPTD_PATH)/util; go get ./...
```
lukebjerring:
```suggestion
	cd $(WPTD_PATH)/util; go get -t ./...
```

Makefile
lukebjerring: Welp. Good to know, I guess.

Hexcles: Actually, this doesn't matter because we don't want to upgrade the packages in these targets.

We can have a new make target for updating everything including test packages.

lukebjerring:
Looks like the only failures now are the aetest failures, because aetest is broken by the latest version of appengines golang SDK.

Hexcles:
> Looks like the only failures now are the aetest failures, because aetest is broken by the latest version of appengines golang SDK.

Is there a migration path for aetest? If not, we could temporarily freeze gcloud SDK.

lukebjerring:
https://issuetracker.google.com/issues/142004500 is totally unanswered thusfar. Freezing feels short-term, but happy for you to make that call.

lukebjerring:
I'm going to merge this now, since:
- The medium tests are already broken with the same cause
- Local development without this update hinges on dev_appserver.py, which fails (same cause as medium tests).

TBR.

PR #1268, opened 2019-04-16
Modularize query/cache/index for the go-farm dep

lukebjerring:
Fixes the build (maybe).


lukebjerring:
@Hexcles - force merged to fix build, but I think, in hindsight, `killall` might permanently be missing (see wct failure)

PR #1269, opened 2019-04-16
Fetch before cleanup

lukebjerring:
## Description
We're false-negativing on the existence of branches, which is killing our staging environments.


Hexcles:
I'm going to force-merge this.

PR #1271, opened 2019-04-17
Install up-to-date deps for Firefox

Hexcles:
Fixes broken Firefox tests


Hexcles:
TBR to fix build. (The failure is a transient deployment failure for revision announcer, which is totally unrelated.)

PR #1275, opened 2019-04-18
Implement logging via a gcloud client in flex contexts

lukebjerring:
We already have an nginx "parent" request log spanning the request's duration.

This PR creates a gcloud logging client, for which all entries are aggregated by the request that they belong to (by replicating its traceId).

To verify:
- Visit appengine logs for `GAE application > searchcache > logging`
- Run a request that will be logged, e.g. `curl -X POST -d '{"run_ids":[5418908555149312,6317650003099648,5797013731934208,6197500226568192],"query":{"exists":[{"contains":"shadowRoot"}]}}' https://logging-dot-searchcache-dot-wptdashboard-staging.appspot.com/api/search/cache`
- See the log, expand, see child logs including the error.

Caveat: If we use the existing "parent" log from nginx, we are not able to set an overarching status (of the most severe log level from the "child" logs).
The alternative, of producing a parent log too, was tried but causes a double-entry (nginx vs request_log), which isn't more desirable.

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -219,7 +219,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
+		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, logrus.StandardLogger())
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -219,7 +219,7 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, logrus.StandardLogger())
+		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, log)
 	}
 
 	data, err = json.Marshal(resp)
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -69,18 +69,23 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	ctx := r.Context()
+	log := shared.GetLogger(ctx)
 	data, err := ioutil.ReadAll(r.Body)
 	if err != nil {
+		log.Errorf("Failed to read request body: %s", err.Error())
 		http.Error(w, "Failed to read request body", http.StatusInternalServerError)
 	}
 	err = r.Body.Close()
 	if err != nil {
+		log.Errorf("Failed to close request body: %s", err.Error())
 		http.Error(w, "Failed to finish reading request body", http.StatusInternalServerError)
 	}
 
 	var rq query.RunQuery
 	err = json.Unmarshal(data, &rq)
 	if err != nil {
+		log.Errorf("Failed to unmarshal RunQuery: %s", err.Error())
 		http.Error(w, err.Error(), http.StatusBadRequest)
 		return
 	}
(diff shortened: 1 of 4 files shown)


Hexcles:
`api/query/cache/service/main.go:222:74: log.StandardLogger undefined (type shared.Logger has no field or method StandardLogger)` ?

shared/logger.go
```diff
@@ -0,0 +1,219 @@
...
+
+// NewAppEngineFlexContext creates a new Google App Engine Flex-based
+// context, with a Google Cloud logger client bound to an http.Request.
+func NewAppEngineFlexContext(r *http.Request, project string, commonResource *mrpb.MonitoredResource) (ctx context.Context, err error) {
+	ctx = r.Context()
+	client, err := gclog.NewClient(ctx, project)
+	if err != nil {
+		return nil, err
+	}
+	ctx = context.WithValue(ctx, cloudLoggingClientKey, client)
```
Hexcles: Do we really need to store the client in the context? Doesn't `gcLogger` contain all we need?

shared/logger.go
```diff
@@ -0,0 +1,219 @@
...
+
+	traceID := strings.Split(r.Header.Get("X-Cloud-Trace-Context"), "/")[0]
+	if traceID != "" {
+		traceID = fmt.Sprintf("projects/%s/traces/%s", project, traceID)
+	}
+	ctx = context.WithValue(ctx, traceKey, traceID)
```
Hexcles: ditto

shared/logger.go
```diff
@@ -0,0 +1,219 @@
...
+
+	childLogger := client.Logger("request_log_entries", gclog.CommonResource(commonResource))
+	ctx = WithLogger(ctx, &gcLogger{
```
Hexcles: You might want to move `NewAppEngineContext` & `NewRequestContext` to this file and change them to use `WithLogger`, too, for consistency.

shared/logger.go
```diff
@@ -0,0 +1,219 @@
...
+// context, with a Google Cloud logger client bound to an http.Request.
+func NewAppEngineFlexContext(r *http.Request, project string, commonResource *mrpb.MonitoredResource) (ctx context.Context, err error) {
+	ctx = r.Context()
+	client, err := gclog.NewClient(ctx, project)
+	if err != nil {
+		return nil, err
+	}
+	ctx = context.WithValue(ctx, cloudLoggingClientKey, client)
+
+	traceID := strings.Split(r.Header.Get("X-Cloud-Trace-Context"), "/")[0]
```
Hexcles: Nit: perhaps add a comment and link to https://cloud.google.com/appengine/docs/flexible/go/writing-application-logs

shared/logger.go
```diff
@@ -0,0 +1,219 @@
...
+
+type gcLogger struct {
+	childLogger *gclog.Logger
+	maxSeverity gclog.Severity
+	traceID     string
+}
+
+func (gcl *gcLogger) log(severity gclog.Severity, format string, params ...interface{}) {
+	if severity > gcl.maxSeverity {
+		gcl.maxSeverity = severity
```
Hexcles: What is this for?

api/query/cache/service/main.go
```diff
@@ -205,11 +219,12 @@ func searchHandler(w http.ResponseWriter, r *http.Request) {
 		var netClient = &http.Client{
 			Timeout: time.Second * 5,
 		}
-		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
+		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, logrus.StandardLogger())
```
Hexcles: We are inside a request context here, right? Shall we pass in `log` instead?

lukebjerring:
```suggestion
		resp.MetadataResponse = shared.GetMetadataResponse(runs, netClient, log)
```

shared/logger.go
lukebjerring: https://github.com/web-platform-tests/wpt.fyi/pull/1275/commits/18e6c76d67a7895b26c4cb13db911dc459fa18cf#diff-b4ef6754bf56e8bface731e15bad4cb9L157

I was, earlier, creating a parent request manually (which reflected the max severity). These are redundant given that I am no longer doing that.

shared/logger.go
lukebjerring: Done.

shared/logger.go
lukebjerring: As above, was used for parent log entry.

lukebjerring:
PTAL

shared/logger.go
```diff
@@ -0,0 +1,228 @@
...
+	return logger
+}
+
+type cloudLoggingClientKeyType struct{}
+
+var cloudLoggingClientKey = cloudLoggingClientKeyType{}
+
+type traceKeyType struct{}
+
+var traceKey = traceKeyType{}
```
Hexcles: This type and var are no longer used.

PR #1278, opened 2019-04-23
Tweak retry params for results-arrival

Hexcles:


Hexcles:
I'm merging this directly. There's nothing on Travis that would process `queue.yaml`.

Note that this file needs to be deployed to staging & production manually.

PR #1281, opened 2019-04-25
Deploy smaller staging searchcache instances

lukebjerring:


api/query/cache/service/app.staging.yaml
```diff
@@ -0,0 +1,23 @@
...
+env: flex
+service: searchcache
+
+# AppEngine Flex supports custom machine types with up to 32 vCPUs.
+# Custom machine type specifications:
+# https://cloud.google.com/compute/docs/instances/creating-instance-with-custom-machine-type#specifications
+# Relevant AppEngine Flex documentation:
+# https://cloud.google.com/appengine/docs/flexible/nodejs/reference/app-yaml#resource-settings
+resources:
+  # TODO: Would prefer up to 96 CPUs for performance and memory capacity if/when
```
Hexcles: Replace this comment with explanation that fewer resources are used for staging instances.

lukebjerring:
PTAL.

Note that this change probably needs to be preceded with a memory monitor that isn't a fixed value.

util/travis-deploy-staging.sh
```diff
@@ -25,6 +25,7 @@ if [[ "${APP_PATH}" == "webapp" ]]; then APP_DEPS="${APP_DEPS}|api|shared"; fi
 if [[ "${APP_PATH}" == "revisions/service" ]]; then APP_DEPS="${APP_DEPS}|revisions|shared"; fi
 # Be more conservative: only deploy searchcache when it's directly modified.
 # if [[ "${APP_PATH}" == "api/query/cache/service" ]]; then APP_DEPS="shared|api/query"; fi
+# if [[ "${APP_PATH}" == "api/query/cache/service/app.staging.yaml" ]]; then APP_DEPS="shared|api/query"; fi
```
Hexcles:
```suggestion
if [[ "${APP_PATH}" == "api/query/cache/service/app.staging.yaml" ]]; then APP_DEPS="api/query"; fi
```

We need to have this otherwise searchcache won't be deployed even if we modify `api/query/cache/...`.

Makefile
```diff
@@ -162,7 +162,7 @@ mockgen: git
 	fi
 
 package_service: var-APP_PATH
-	if [[ "$(APP_PATH)" == "revisions/service" || "$(APP_PATH)" == "api/query/cache/service" ]]; then \
+	if [[ "$(APP_PATH)" == "revisions/service" || "$(APP_PATH)" == "api/query/cache/service"* ]]; then \
```
Hexcles: @lukebjerring FYI

Hexcles: It didn't work. I tweaked the workaround a bit (and it's now uglier...)

Hexcles: Duh apparently I got confused by Makefile & shell variables again...

lukebjerring:
Yeah LGTM - they're fairly sane.

PR #1283, opened 2019-04-26
Allow empty reports in index

### LINKED ISSUE #1276: PR runs fail to load after a number of retries
### Check
A link to the GitHub Check run containing my issue:
https://github.com/web-platform-tests/wpt/runs/106964177

### Description

https://wpt.fyi/results/?diff&filter=ADC&run_id=5683815137673216&run_id=5644081153703936 never loads correctly, after the "trying again" toast shows a number of times I see this:
Screenshot_20190421-172407

lukebjerring:
## Description
Fixes #1276 

Gently error when a report is empty, and serve a response that has no results (since the contents are never going to be non-empty)

### DIFF
--- a/api/query/cache/index/index.go
+++ b/api/query/cache/index/index.go
@@ -17,10 +17,10 @@ import (
 	"sync"
 
 	mapset "github.com/deckarep/golang-set"
-	"github.com/web-platform-tests/wpt.fyi/shared/metrics"
 	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/lru"
 	"github.com/web-platform-tests/wpt.fyi/shared"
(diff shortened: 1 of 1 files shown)


api/query/cache/index/index.go
```diff
@@ -34,6 +34,7 @@ var (
 	errSomeShardsRequired = errors.New("Index must have at least one shard")
 	errUnexpectedRuns     = errors.New("Unexpected number of runs")
 	errZeroRun            = errors.New("Cannot ingest run with ID of 0")
+	errEmptyReport        = errors.New("Report contains no results")
```
Hexcles: Drive-by optional nit: the convention is to not capitalize the error strings (they are usually logged in formats like "Failed to ...: [error string]").

lukebjerring: Maybe in another PR.

PR #1284, opened 2019-04-26
Add api/metadata endpoints and support Metadata search

KyleJu:
## Description
Add api/metadata endpoint/{products} and handle all metadata related queries.

Endpoint supports:
Get api/metadata?product=...&product=...
Post api/metadata?product=... , when search link:[url]

Search supports:
Search links by a substring of url  - link: [substring]
Search links by urls that do not have a [substring] - link: ! [substring]
Search untriaged issues - (...) and !link:[substring]
Search for issues that have been triaged - (...) and link:[substring]

## Review Information
Verify Get api/metadata:
Visit [api/metadata?product=chrome&product=safari](https://metadata-endpoint-dot-wptdashboard-staging.appspot.com/api/metadata?product=chrome)

Verify Post api/metadata:
http POST localhost:8080/api/metadata?product=chrome\\&product=firefox \exists:='[{"link":"bugs.chromium.org"}]'

Verify link on searchcache, e.g.
[html/browsers/history/the-history-interface & link:bugs.chromium.org](https://metadata-endpoint-dot-wptdashboard-staging.appspot.com/results/?q=+html%2Fbrowsers%2Fhistory%2Fthe-history-interface+%26+link%3Abugs.chromium.org)

View Untriaged metadata
1. remove all browsers besides Chrome
2. search the-history-interface & !link:chromium
3. 007.html is traiged in the wpt-metadata repository so its not shown


shared/metadata.go
```diff
@@ -54,12 +54,29 @@ func (m MetadataResults) Less(i, j int) bool { return m[i].Test < m[j].Test }
 
 // GetMetadataResponse retrieves the response to a WPT Metadata query.
 func GetMetadataResponse(testRuns []TestRun, client *http.Client, log Logger) MetadataResults {
+	var productAtRevisions = make([]ProductAtRevision, len(testRuns))
```
lukebjerring: I wonder about the `revision` part of this. We don't serve different metadata for different revisions, so as long as its ignored somewhere then it's not too problematic.

But, it's more effort here to try and pipe the `revision` part; why not just take the TestRun's `Product` struct?

webapp/components/test/test-search.html
```diff
@@ -275,6 +275,14 @@
         });
       }
     });
+
+    test('simple link search', () => {
```
lukebjerring: Leave the `searchcache` work / ohm stuff for another PR

api/query/atoms.go
```diff
@@ -580,6 +599,27 @@ func (c *AbstractCount) UnmarshalJSON(b []byte) error {
...
+	if err != nil {
+		return err
+	}
+	patternMsg, ok := data["link"]
+	if !ok {
+		return errors.New(`Missing Link pattern property: "link"`)
+	}
+	var pattern string
+	if err := json.Unmarshal(*patternMsg, &pattern); err != nil {
+		return errors.New(`Missing test name pattern property "pattern" is not a string`)
```
KyleJu:
```suggestion
		return errors.New(`Missing link pattern property "pattern" is not a string`)
```

api/query/concrete_query.go
```diff
@@ -97,6 +103,10 @@ func (RunTestStatusEq) Size() int { return 1 }
 // lookup in a test run result mapping per test.
 func (RunTestStatusNeq) Size() int { return 1 }
 
+// Size of AbstractLink has a size of 1: servicing such a query requires a
```
KyleJu:
```suggestion
// Size of Link has a size of 1: servicing such a query requires a
```

api/query/atoms.go
```diff
@@ -146,6 +149,22 @@ func (c AbstractCount) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
...
+type AbstractLink struct {
+	Pattern string
+}
+
+// BindToRuns for AbstractLink is a no-op; it is independent of test runs
+func (l AbstractLink) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
+	var netClient = &http.Client{
+		Timeout: time.Second * 5,
+	}
+	metadata := shared.GetMetadataResponse(runs, netClient, log.StandardLogger())
```
Hexcles: We only do query binding in a request, right? Is it possible to get a logger from the request context instead of using the standard logger?

api/metadata_handler.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	} else if len(productSpecs) == 0 {
+		http.Error(w, fmt.Sprintf("Missing required 'product' param"), http.StatusBadRequest)
+		return
+	}
+
+	Metadata := shared.GetMetadataResponseOnProducts(productSpecs, h.httpClient, h.logger)
+	MetadataResponse := filterMetadata(abstractLink, Metadata)
```
Hexcles: Nit: we generally don't capitalize local variables.

webapp/components/test-search.js
```diff
@@ -53,6 +53,9 @@ const QUERY_GRAMMAR = ohm.grammar(`
       | "two"           -- count2
       | "one"           -- count1
 
+    Link
```
Hexcles: Nit: move this to around `statusExp` and rename it to `linkExp`.

api/query/search.go
```diff
@@ -141,6 +137,54 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
+	// Check if the query is a simple link query.
```
lukebjerring: As discussed offline: This was intended for the case that no test runs are passed, only a metadata search.

I don't think we should be handling this case server-side; a client that is fetching only metadata, without respect to any particular runs, will be in a different context than one that is calling the `/api/search` endpoint (and can simply call `/api/metadata` directly).

api/routes.go
```diff
@@ -60,4 +60,7 @@ func RegisterRoutes() {
 
 	// API endpoint for redirecting to a run's summary JSON blob.
 	shared.AddRoute("/api/results", "api-results", shared.WrapPermissiveCORS(apiResultsRedirectHandler))
+
+	// API endpoint for searching Metadata for the products.
+	shared.AddRoute("/api/metadata", "api-medata", shared.WrapPermissiveCORS(apiMetadataHandler))
```
lukebjerring:
```suggestion
	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
```

shared/metadata.go
KyleJu: Because in the metadata endpoint, I use productSpecs, err = shared.ParseProductOrBrowserParams(q), thought it would be clean to provider an API using ProductAtRevision

api/query/atoms.go
```diff
@@ -146,6 +149,22 @@ func (c AbstractCount) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
...
+
+	return Link{l.Pattern, metadata}
```
Hexcles: Nit: I'd strongly suggest you to explicitly initialize the struct (naming all the fields). This is a non-trivial struct (with type embedding) and defined quite far away from here so it'd be brittle / less readable to initialize like this.

api/metadata_handler.go
```diff
@@ -0,0 +1,171 @@
...
+			if strings.Contains(url, linkQuery.Pattern) {
+				res = append(res, data)
+				break
+			}
+		}
+	}
+	return res
+}
+
+var cacheKey = func(r *http.Request) interface{} {
```
KyleJu: Refactor this - same method in search

api/README.md
```diff
@@ -753,3 +754,74 @@ empty string, which will yield all test results for the selected runs.
...
+
+### /api/metadata
+
+This endpoint accepts POST and GET requests.
+
+- GET request returns Metadata Link Information by product, and requires product parameters;
+
+- POST request searches Metadata Link by link url, and requires product parameters and payload.
+
+__Parameters__
```
Hexcles: Clarify these are *URL* params.

api/README.md
```diff
@@ -753,3 +754,74 @@ empty string, which will yield all test results for the selected runs.
...
+
+__`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
+
+#### Link Query
+  link: [pattern]
```
Hexcles: For consistency of style:
```suggestion
__`link`__: [pattern]
```

Also remove the spaces at the beginning of next line.

api/README.md
```diff
@@ -753,3 +754,74 @@ empty string, which will yield all test results for the selected runs.
...
+
+- GET request returns Metadata Link Information by product, and requires product parameters;
+
+- POST request searches Metadata Link by link url, and requires product parameters and payload.
+
+__Parameters__
+
+__`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
+
+#### Link Query
```
Hexcles: So IIUC this goes to the body as a JSON, right? Make that clear.

api/metadata_handler_medium_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestFilterMetadataHanlder_Success(t *testing.T) {
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "../shared/metadata_testdata/gzip_testfile.tar.gz")
+	}
+	server := httptest.NewServer(http.HandlerFunc(handler))
+	defer server.Close()
+
+	r := httptest.NewRequest("GET", "/abd/api/metadata?product=chrome&product=safari", nil)
+	w := httptest.NewRecorder()
+	client := &http.Client{}
```
Hexcles:
```suggestion
	client := server.Client()
```

The benefit of this is that this test client will suppress any TLS errors with the local test server.

shared/metadata.go
```diff
@@ -52,14 +57,40 @@ func (m MetadataResults) Len() int           { return len(m) }
 func (m MetadataResults) Swap(i, j int)      { m[i], m[j] = m[j], m[i] }
 func (m MetadataResults) Less(i, j int) bool { return m[i].Test < m[j].Test }
 
+// GetMetadatURL returns the URL that fetches Metadata information.
+// For testing, it returns a MetadataTestingURL set by users.
+func GetMetadatURL(isTesting bool) string {
```
Hexcles: What is this for? I don't see any `GetMetadatURL(true)` anywhere? The empty "testing" URL never seems to be used.

Also, it's `Metadata` not `Metadat`.

api/metadata_handler_medium_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	metadataHandler := MetadataHandler{nil, client, server.URL}
+	metadataHandler.ServeHTTP(w, r)
+
+	assert.Equal(t, http.StatusOK, w.Code)
+	res := w.Body.String()
+	assert.Equal(t, "[{\"test\":\"/IndexedDB/bindings-inject-key.html\",\"urls\":[\"bugs.chromium.org/p/chromium/issues/detail?id=934844\",\"\"]},{\"test\":\"/html/browsers/history/the-history-interface/007.html\",\"urls\":[\"bugs.chromium.org/p/chromium/issues/detail?id=592874\",\"\"]}]", res)
```
Hexcles: Use backticks (``) instead of double quotes to avoid having to escape all the double quotes in the JSON string.

api/metadata_handler_medium_test.go
```diff
@@ -0,0 +1,133 @@
...
+				"link": "bugs.chromium.org"
+			}]
+		}
+	}`
+	bodyReader := strings.NewReader(body)
+	r := httptest.NewRequest("GET", "/abd/api/metadata?", bodyReader)
+	w := httptest.NewRecorder()
+	client := &http.Client{}
+
+	metadataHandler := MetadataSearchHandler{nil, client,""}
```
Hexcles: Nit:
```suggestion
	metadataHandler := MetadataSearchHandler{nil, client, ""}
```

Make sure you run `gofmt` automatically in your editor. (Unfortunately, we can't run `gofmt` on Travis because of annoying version mismatches.)

api/metadata_handler.go
```diff
@@ -0,0 +1,184 @@
...
+
+	var abstractLink query.AbstractLink
+	var isLinkQuery = false
+	if exists, isExists := rq.AbstractQuery.(query.AbstractExists); isExists && len(exists.Args) == 1 {
+		abstractLink, isLinkQuery = exists.Args[0].(query.AbstractLink)
+	}
+
+	if !isLinkQuery {
+		h.logger.Errorf("Error from request: non Link search query %s for api/metadata", rq.AbstractQuery)
+		http.Error(w, "Error from request: non Link search query for api/metadata", http.StatusInternalServerError)
```
Hexcles:
```suggestion
		http.Error(w, "Error from request: non Link search query for api/metadata", http.StatusBadRequest)
```
This is a user error.

api/metadata_handler_medium_test.go
```diff
@@ -0,0 +1,133 @@
...
+	client := &http.Client{}
+
+	metadataHandler := MetadataSearchHandler{shared.NewNilLogger(), client, ""}
+	metadataHandler.ServeHTTP(w, r)
+
+	assert.Equal(t, http.StatusInternalServerError, w.Code)
+}
+
+func TestFilterMetadata(t *testing.T) {
+	metadata := shared.MetadataResults(shared.MetadataResults{shared.MetadataResult{Test: "/foo/bar/b.html", URLs: []string{"", "https://aa.com/item", "https://bug.com/item"}}, shared.MetadataResult{Test: "bar", URLs: []string{"", "https://external.com/item", ""}}})
```
Hexcles: Nit: can we wrap this line a bit?

api/metadata_handler_medium_test.go
```diff
@@ -0,0 +1,133 @@
...
+	abstractLink := query.AbstractLink{Pattern: "bug.com"}
+
+	res := filterMetadata(abstractLink, metadata)
+
+	assert.Equal(t, 1, len(res))
+	assert.Equal(t, "/foo/bar/b.html", res[0].Test)
+	assert.Equal(t, "", res[0].URLs[0])
+	assert.Equal(t, "https://aa.com/item", res[0].URLs[1])
+	assert.Equal(t, "https://bug.com/item", res[0].URLs[2])
+}
```
Hexcles: Nit: This file is missing a newline at EOF. Again, try to make sure `gofmt` is working in your VSCode.

api/query/atoms_test.go
```diff
@@ -616,6 +673,43 @@ func TestStructuredQuery_bindCount(t *testing.T) {
...
+		},
+	}
+
+	handler := func(w http.ResponseWriter, r *http.Request) {
+		http.ServeFile(w, r, "../../shared/metadata_testdata/gzip_testfile.tar.gz")
+	}
+	server := httptest.NewServer(http.HandlerFunc(handler))
+	defer server.Close()
+
+	shared.MetadataTestingURL = server.URL
```
Hexcles: Does this really work? I don't see `MetadataTestingURL` ever used in any code paths.

I feel like this test *coincidentally* passed because your dummy test data happens to be the same as the current tip of tree of `wpt-metadata`.

A great way to make sure test data actually works is to put non-existent stuff in it (e.g. `foobarbaz`).

api/query/cache/index/index_filter_test.go
```diff
@@ -13,9 +13,9 @@ import (
 	mapset "github.com/deckarep/golang-set"
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
-	metrics "github.com/web-platform-tests/wpt.fyi/shared/metrics"
 	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/shared"
+	metrics "github.com/web-platform-tests/wpt.fyi/shared/metrics"
```
Hexcles: Nit:
```suggestion
	"github.com/web-platform-tests/wpt.fyi/shared/metrics"
```
Again `gofmt` would catch this.

webapp/components/test-search.js
```diff
@@ -79,6 +79,7 @@ const QUERY_GRAMMAR = ohm.grammar(`
 
     Fragment
       = not Fragment -- not
+      | LinkExp
```
Hexcles: nit: can we make this more consistent?
```suggestion
      | linkExp
```

Hexcles:
To reiterate a high-level but optional suggestion I mentioned to Kyle offline yesterday:

I'm a bit concerned about making `Bind` no longer purely computational ("functional"). There are now side effects and external network requests in the `Bind` phase. The downside of this change can partly be demonstrated by the complexity of the test setup.

Looking at the requirements, especially that we only ever care about the latest metadata, I believe fetching metadata might be better suited to happen in the background polling thread of `searchcache`. This will not only solve a lot of the headache in terms of testing (`Bind` really wasn't designed to easily take options like the metadata URL), but also improve error handling and saving time & space (we can keep just one global metadata map instead of having to fetch and create one for each `link` search).

lukebjerring:
Your concerns mirror the suggestion that I made during design; it was considered an optional optimization to build an index from the metadata, rather than just fetching it when needed.

It's a balance of engineering time and performance problems; we're making the same trade-off by fetching from the search index via the remote API in #1216 
Let's wait and let usage dictate whether a polling update and in-memory index is justified.

api/query/cache/index/index_filter_test.go
KyleJu: umm I tried to go fmt on this file and didn't see the suggested change, plus `metrics` was there originally

api/metadata_handler_medium_test.go
KyleJu: Aha thanks!

api/README.md
```diff
@@ -753,3 +754,74 @@ empty string, which will yield all test results for the selected runs.
...
+         ""
+      ]
+   }
+]
+```
+</details>
+
+#### Post Examples
+- POST /api/metadata?product=chrome\&product=firefox \
+    run_ids:="[1, 2, 3]" query:='{"exists":[{"link":"bugs.chromium.org"}]}'
```
KyleJu: update this

KyleJu:
> As discussed last week, can we change the dummy test data (`shared/metadata_testdata/gzip_testfile.tar.gz`) first to make it _very different_ from a real archive so that we can be sure the test setup isn't depending on GitHub API and the current head of the `wpt-metadata` repo?

yea that makes sense. Will do

api/query/atoms_test.go
```diff
@@ -616,6 +674,54 @@ func TestStructuredQuery_bindCount(t *testing.T) {
 	assert.Equal(t, expected, q.BindToRuns(runs...))
 }
 
+func tearDownLinkSubTest() func(t *testing.T, url string) {
+	return func(t *testing.T, url string) {
+		shared.MetadataArchiveURL = url
+		t.Log(fmt.Sprintf("Assign origin URL %s", shared.MetadataArchiveURL))
```
Hexcles:
```suggestion
```

api/query/atoms_test.go
```diff
@@ -616,6 +674,54 @@ func TestStructuredQuery_bindCount(t *testing.T) {
 	assert.Equal(t, expected, q.BindToRuns(runs...))
 }
 
+func tearDownLinkSubTest() func(t *testing.T, url string) {
+	return func(t *testing.T, url string) {
```
Hexcles: This is a bit convoluted. You don't need to return a functor to capture `url`. Parameters are captured at the `defer` statement. e.g. https://play.golang.org/p/3znsrNkGFB_F

And then you can simply do `defer tearDownLinkSubTest(t, shared.MetadataArchiveURL)`

api/README.md
```diff
@@ -753,3 +754,74 @@ empty string, which will yield all test results for the selected runs.
...
+
+- GET request returns Metadata Link Information by product, and requires product parameters;
+
+- POST request searches Metadata Link by link url, and requires product parameters and payload.
+
+__URL Parameters__
+
+__`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
+
+#### JSON Payload
```
lukebjerring: This (`link`) is a query parameter, not part of the JSON payload?

api/query/README.md
```diff
@@ -139,3 +139,20 @@ Same as satuts, but with a specific product-spec.
       "product": "chrome-69",
       "status": "ok",
     }
+
+#### link
+
+`link` query objecsts performs a search on a test that has a Metadata Link
```
lukebjerring:
```suggestion
`link` query atoms perform a search for tests that have some matching link metadata
```

lukebjerring: (And drop L146)

api/README.md
KyleJu: Nope, link is a part of payload. I think it could be either in the query parameter or a part of the JSON payload, but I did so because that is what searchcache does.

webapp/components/wpt-flags.js
```diff
@@ -227,11 +226,6 @@ class WPTFlagsEditor extends FlagsEditorClass(/*environmentFlags*/ false) {
         Use web-platform-tests.live.
       </paper-checkbox>
     </paper-item>
-    <paper-item>
-      <paper-checkbox checked="{{showMetadataInfo}}">
-        Show metadata Information on wpt.fyi result page.
-      </paper-checkbox>
-    </paper-item>
```
lukebjerring: What's going on here?

webapp/components/test-search.js
```diff
@@ -165,6 +169,10 @@ const QUERY_SEMANTICS = QUERY_GRAMMAR.createSemantics().addOperation('eval', {
   CountSpecifier_count3: (_) => 3,
   CountSpecifier_count2: (_) => 2,
   CountSpecifier_count1: (_) => 1,
+  linkExp: (l, colonBang, r) => {
```
lukebjerring:
```suggestion
  linkExp: (l, colon, r) => {
```

shared/metadata.go
```diff
@@ -70,35 +89,34 @@ func parseMetadata(metadataByteMap map[string][]byte, log Logger) map[string]Met
...
 	}
 	return metadataMap
 }
 
-// ConstructMetadataResponse constructs the response to a WPT Metadata query.
-// When parsing 'link' nodes, assume there is no mising information nor duplicates;
-// assume each test for each browser type is only associated with one bug.
-func constructMetadataResponse(testRuns []TestRun, metadata map[string]Metadata) MetadataResults {
+// constructMetadataResponse constructs the response to a WPT Metadata query, given ProductSpecs.
+func constructMetadataResponse(productAtRevisions []ProductAtRevision, metadata map[string]Metadata) MetadataResults {
```
lukebjerring: Use `ProductSpec` not `ProductAtRevision` (the latter drops labels, which can be important).

Add a function for TestRun to get the `ProductSpec` (ProductSpec uses a set of labels, TestRun uses an array. Annoying, I know).

api/query/cache/index/index_filter_test.go
```diff
@@ -332,3 +332,140 @@ func TestBindExecute_TestStatus(t *testing.T) {
...
+}
+
+func TestBindExecute_LinkNoMatchingPattern(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	loader := NewMockReportLoader(ctrl)
+	idx, err := NewShardedWPTIndex(loader, testNumShards)
+	assert.Nil(t, err)
+
+	matchingTestName := "/a/b/c"
```
lukebjerring: Probably not the greatest variable name (copy/pasta) when the test is for no matches.

api/metadata_handler.go
lukebjerring: ?

api/README.md
lukebjerring: searchcache has a complicated JSON blob; this example doesn't imply a JSON blob. Is `link` a field on a JSON object?

webapp/components/wpt-flags.js
KyleJu: This is a flag to indicate that the metadata information will be requested along with TestRuns, but we don't need it anymore

api/README.md
```diff
@@ -753,3 +754,80 @@ empty string, which will yield all test results for the selected runs.
...
+
+__URL Parameters__
+
+__`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
+
+#### JSON Request Payload
+```json
+[
+  {
+    "link": [pattern]
```
lukebjerring:
```suggestion
    "link": "[pattern]"
```

api/README.md
```diff
@@ -753,3 +754,80 @@ empty string, which will yield all test results for the selected runs.
...
+  }
+]
+```
+
+  Where `[pattern]` is a susbtring of the url field of a Metadata Link Node.
```
lukebjerring:
```suggestion
  Where `[pattern]` is any substring of the url field of a wpt-metadata `link` node.
```

api/README.md
```diff
@@ -753,3 +754,80 @@ empty string, which will yield all test results for the selected runs.
...
+
+<details><summary><b>Example JSON</b></summary>
+
+```json
+[
+   {
+      "test":"/IndexedDB/bindings-inject-key.html",
+      "urls":[
+         "bugs.chromium.org/p/chromium/issues/detail?id=934844",
+         ""
```
lukebjerring: what's with the empty strings? (here and L801)

api/metadata_handler.go
```diff
@@ -0,0 +1,185 @@
...
+		delegate,
+		shared.NewGZReadWritable(shared.NewMemcacheReadWritable(ctx, 5*time.Minute)),
+		shared.AlwaysCachable,
+		cacheKey,
+		shared.CacheStatusOK).ServeHTTP(w, r)
+}
+
+func (h MetadataHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	q := r.URL.Query()
+	var productSpecs shared.ProductSpecs
```
lukebjerring:
```suggestion
```

lukebjerring: (You re-declare the variable with `:=` on the next line)

api/metadata_handler.go
```diff
@@ -0,0 +1,185 @@
...
+	var isLinkQuery = false
+	if len(ae.Args) == 1 {
+		abstractLink, isLinkQuery = ae.Args[0].(query.AbstractLink)
+	}
+
+	if !isLinkQuery {
+		h.logger.Errorf("Error from request: non Link search query %s for api/metadata", ae)
+		http.Error(w, "Error from request: non Link search query for api/metadata", http.StatusBadRequest)
+		return
+	}
```
lukebjerring: It looks to me like the remainder of this function mainly matches the `MetadataHandler` above. Why not collapse the handlers into an option if-statement (whether there's a query or not) ?

api/README.md
KyleJu: I added a url link for safari. It was empty because this test didn't have a corresponding bug link in safari. But I guess if it is just for demonstration, the link doesn't have to be real?

KyleJu:
@Hexcles since Luke is gone for the week, would you mind taking a second look?

api/query/cache/index/filter.go
```diff
@@ -144,6 +151,26 @@ func (c Count) Filter(t TestID) bool {
...
 }
 
+// Filter interprets a Link as a filter function over TestIDs.
+func (l Link) Filter(t TestID) bool {
+	name, _, err := l.tests.GetName(t)
+	if err != nil {
+		return false
+	}
+
+	urls, isExist := l.metadata[name]
```
lukebjerring: nit: Remane `isExist` to `ok` (standard name for the second param in a map lookup)

api/query/cache/index/index_filter_test.go
```diff
@@ -332,3 +332,140 @@ func TestBindExecute_TestStatus(t *testing.T) {
...
+					},
+					&metrics.TestResults{
+						Test:   "/d/e/f",
+						Status: "FAIL",
+					},
+				},
+			},
+		},
+	})
+	metadata := map[string][]string{"/foo/bar/b.html": []string{"https://bug.com/item", "https://bug.com/item", "https://bug.com/item"}, matchingTestName: []string{"", "https://external.com/item", ""}}
```
lukebjerring:
```suggestion
	metadata := map[string][]string{
		"/foo/bar/b.html": []string{"https://bug.com/item", "https://bug.com/item", "https://bug.com/item"},
		matchingTestName: []string{"", "https://external.com/item", ""},
	}
```

PR #1289, opened 2019-04-30
[searchcache] Dynamically detect total memory

Hexcles:
Instead of hard-coding it to ~200GB.

Needs to land before #1281 .


Hexcles:
`time="2019-04-30T19:43:01Z" level=info msg="Detected total system memory: 219764219904; setting max heap size to 217764219904"`

(unit is bytes) looks correct to me

PR #1290, opened 2019-04-30
Fix cleanup script by forcing an unshallow fetch

Hexcles:
Travis by default uses `git clone --depth=50` which will affect all
future `git fetch` in this checkout. We need to explicitly add
`--unshallow` to make sure branches older than 50 commits ago can also
be fetched.

e.g. https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/195992237#L417
This job will later go on deleting `smaller` incorrectly.


Hexcles:
No point waiting for Travis for this PR. I'm merging it now.

PR #1293, opened 2019-05-02
Make sure preallocated run IDs are used

### LINKED ISSUE #1237: Preallocation of run IDs doesn't work
#1168 doesn't seem to be working as intended.

We can see IDs successfully reserved and used as task names, but the IDs of the created runs don't match their task names (pre-allocated IDs).

Hexcles:
There was a mix of `run_id` and `id` between services.

This PR unifies the param to `id` and adds tests.

Fixes #1237 .


Hexcles:
Welp, apparently Travis CI does not respond to "draft" PRs, so I'm marking this as "ready for review" although it isn't yet.

PR #1304, opened 2019-05-10
Add infrastructure for puppeteer

lukebjerring:
## Description
Paves the way for migrating our `go_chrome_test` suite to puppeteer (and we can follow with firefox at a later date).

Note: we also verified that running multiple webdriver tests in the same suite (dev_appserver) works fine, bypassing whatever bug happens in Golang; this leads me to the theory that some unconsumed input buffer is getting exhausted.


webdriver/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "wptdashboard-webdriver",
+  "description": "Node packages leveraged by build rules for the wpt.fyi webdriver tests.",
+  "license": "BSD-3-Clause",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/web-platform-tests/wpt.fyi.git"
+  },
+  "devDependencies": {
```
Hexcles: Did you forget to save the dependencies? There's no prod or dev dependencies at all in `package.json`.

webdriver/dev_appserver.js
```diff
@@ -0,0 +1,105 @@
+const { spawn } = require('child_process');
```
Hexcles: Do we need a copyright header?

webdriver/webdriver.js
```diff
@@ -0,0 +1,38 @@
+const dev_appserver = require('./dev_appserver');
```
Hexcles: ditto

webdriver/webdriver.js
```diff
@@ -0,0 +1,38 @@
...
+  test('/results/', async () => {
+    browser = await puppeteer.launch();
+    const page = await browser.newPage();
+    await page.goto(`${server.url}/results/`);
+    await browser.close();
+  }).timeout(10000);
+
+  suiteTeardown(async () => {
+    log('closing dev_appserver...');
+    server.process.kill();
```
Hexcles:
```suggestion
    server.close();
```

Just better encapsulation. Furthermore, I think we might as well make `process` a private member.

webdriver/dev_appserver.js
lukebjerring: Done

webdriver/webdriver.js
lukebjerring: Done

PR #1306, opened 2019-05-13
Remove instead of hiding empty logo

### LINKED ISSUE #1305: Firefox attempts to load /static/.svg a bunch of times on staging
This is only happening on Firefox and staging (i.e. it's a Firefox-only regression) and can be easily reproduced by opening https://staging.wpt.fyi .

There are a bunch of requests to https://staging.wpt.fyi/static/.svg , which of course all got 404.

I suspect this is related to the run info builder: these might be the CI icons, but are empty when the runs aren't loaded. I'm not sure why this only happens on Firefox though.

Hexcles:
Fixes #1305 

https://polymer-library.polymer-project.org/3.0/docs/devguide/templates#dom-if


Hexcles:
Verified manually on Firefox. Not sure how to test it. (Perhaps a Puppeteer test to listen on 404 events?)

PR #1309, opened 2019-05-15
Upload test coverage to codecov.io

foolip:
Just experimenting, would be cool if we could get coverage information for both python and golang, or even the frontend JS code.


foolip:
I've added a `CODECOV_TOKEN` secret to Travis to test this.

foolip:
Abandoning this, leaving it to @Hexcles to decide if code coverage is useful or not.

Issue #1308, opened 2019-05-15
PR results fail to load


foolip:
### Check
A link to the GitHub Check run containing my issue:
https://github.com/web-platform-tests/wpt/pull/16788/checks?check_run_id=124826939
https://github.com/web-platform-tests/wpt/pull/16788/checks?check_run_id=124827047
https://github.com/web-platform-tests/wpt/pull/16788/checks?check_run_id=124825192

### Description
Following the "Visual comparison of the results" links doesn't work, they say "Failed to fetch test runs" and in the devtools network tab I see a POST to https://wpt.fyi/api/search?diff=true&filter=ADC which failed with 500, response being "Error connecting to search API cache".

foolip:
I suspected something related to the service worker, but loading https://wpt.fyi/results/?diff&filter=ADC&run_id=5707609709477888&run_id=6213144875630592 in an incognito window still doesn't work.

Hexcles:
Closing this as fixed. Feel free to reopen if this happens again.

foolip:
Confirmed my link works now.

PR #1310, opened 2019-05-15
Give 15s to return searchcache response

lukebjerring:
## Description
Attempts to avoid #1308

### DIFF
--- a/api/query/search.go
+++ b/api/query/search.go
@@ -167,7 +167,7 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 		logger := shared.GetLogger(ctx)
 		logger.Infof("Forwarding structured search request to %s: %s", hostname, string(data))
 
-		client := sh.api.GetHTTPClient()
+		client, _ := sh.api.GetSlowHTTPClient(time.Second * 15)
 		req, err := http.NewRequest("POST", fwdURL.String(), bytes.NewBuffer(data))
 		if err != nil {
 			logger.Errorf("Failed to create request to POST %s: %v", fwdURL.String(), err)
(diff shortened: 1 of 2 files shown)


api/query/search.go
```diff
@@ -167,7 +167,7 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 		logger := shared.GetLogger(ctx)
 		logger.Infof("Forwarding structured search request to %s: %s", hostname, string(data))
 
-		client := sh.api.GetHTTPClient()
+		client, _ := sh.api.GetSlowHTTPClient(time.Second * 15)
```
Hexcles:
```suggestion
		client, cancel := sh.api.GetSlowHTTPClient(time.Second * 15)
		defer cancel()
```

I think this would be a resource leak otherwise. (Tests might need to be fixed accordingly.)

api/query/search_test.go
```diff
@@ -133,7 +133,7 @@ func TestStructuredSearchHandler_success(t *testing.T) {
 
 	api.EXPECT().Context().Return(sharedtest.NewTestContext())
 	api.EXPECT().GetServiceHostname("searchcache").Return(hostname)
-	api.EXPECT().GetHTTPClient().Return(server.Client())
+	api.EXPECT().GetSlowHTTPClient(gomock.Any()).Return(server.Client(), nil)
```
Hexcles:
```suggestion
	api.EXPECT().GetSlowHTTPClient(gomock.Any()).Return(server.Client(), func() {})
```

api/query/search_test.go
```diff
@@ -163,9 +163,7 @@ func TestStructuredSearchHandler_failure(t *testing.T) {
 
 	api.EXPECT().Context().Return(sharedtest.NewTestContext())
 	api.EXPECT().GetServiceHostname("searchcache").Return(hostname)
-	api.EXPECT().GetHTTPClient().DoAndReturn(func() *http.Client {
-		return server.Client()
-	})
+	api.EXPECT().GetSlowHTTPClient(gomock.Any()).Return(server.Client(), nil)
```
Hexcles:
```suggestion
	api.EXPECT().GetSlowHTTPClient(gomock.Any()).Return(server.Client(), func() {})
```

PR #1312, opened 2019-05-16
Fix tests + run on Travis

lukebjerring:
## Description
Actually runs the (one) puppeteer tests on Travis.

### DIFF
--- a/webdriver/path-test.js
+++ b/webdriver/path-test.js
@@ -25,7 +25,7 @@ exports.tests = function(ctx) {
           return Array.from(results.shadowRoot.querySelectorAll('path-part'))
               .map(p => p.shadowRoot.querySelector('a').innerText.trim());
         })
-      expect(linkNames).to.eql([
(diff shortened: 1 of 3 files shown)


webdriver/path-test.js
```diff
@@ -15,14 +15,17 @@ exports.tests = function(ctx) {
...
+        return !!results && !results.isLoading;
+      }, {}, view);
       const linkNames = await page.$eval(
         `wpt-${view}`,
         results => {
           return Array.from(results.shadowRoot.querySelectorAll('path-part'))
               .map(p => p.shadowRoot.querySelector('a').innerText.trim());
         })
-      expect(linkNames).to.equal([
+      expect(linkNames).to.eql([
```
Hexcles:
```suggestion
      expect(linkNames).to.deep.equal([
```

`eql` is a bit too magical / not very readable (I would think it's a shorthand of `equal` if I didn't see this change).

PR #1318, opened 2019-05-24
Install Firefox 67.0

Hexcles:
Upgrade to the latest stable as an attempt to fix Firefox WebDriver test
failures on staging.


Hexcles:
This broke web component tests.

PR #1324, opened 2019-05-28
Improve caching for npm packages and hide SW installer behind a flag

Hexcles:
## Description

The main purpose of this PR is to improve the page load speed without (re)enabling the service worker (since we still have some cache invalidation bugs in our SW).

This is done via two approaches:
* Enable caching for `node_modules`.
* Do not load `service-worker-installer.js` if service worker isn't enabled (this saves both two unnecessary script loads and a console error).

## Review Information

Review the two commits separately. The second commit contains a refactor (renaming) of a private struct (as well as the function that populates it) to better reflect its meaning.

Open the staging instance. You shouldn't see any network errors in the console.

### DIFF
--- a/webapp/service-worker-uninstaller.js
+++ b/webapp/service-worker-uninstaller.js
@@ -1,0 +1,8 @@
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.getRegistrations().then(registrations => {
+    for (const reg of registrations) {
+      reg.unregister();
+      console.log('Service worker uninstalled');
+    }
+  });
+}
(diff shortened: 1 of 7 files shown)


webapp/service-worker-uninstaller.js
```diff
@@ -0,0 +1,8 @@
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.getRegistrations().then(registrations => {
```
KyleJu: Since we are not accessing anything outside of this scope, function(registrations) { could be better?

Hexcles: I'm not sure what you mean. In terms of scopes, arrow functions don't have `this`; both normal functions and arrow functions would capture lexical scopes and create closures.

I like it here because it's shorter ;)

Issue #1327, opened 2019-05-29
Race conditions make WebDriver flaky on staging


Hexcles:
I spent a lot of time debugging the flaky failures of the webapp deployment task on master, in particular the second part of it: running `go_firefox_test` against the newly deployed default staging version. Finally, I came to the realization that any WebDriver test that needs to load test results via searchcache is inherently flaky in this context, because:

We deploy all services at the same time in parallel (in different Travis tasks). When webapp is deployed and we are about to run WebDriver tests against it, the searchcache deployment may or may not have finished; and if the new searchcache has been deployed, the initial warmup (loading the most recent runs) may or may not have finished. Hence, we are likely to get 422 from the searchcache API which makes the WebDriver tests fail.

I can't think of an easy way to fix this race condition. Warmup requests in the webapp won't help because again when webapp is warming up, the searchcache service can be either the previous version or the new version. We need some more sophisticated wait mechanism and be cautious about potential deadlocks.

The immediate solution would be disabling the WebDriver tests that rely on test results when running against staging.

Hexcles:
This hasn't happened for a while.

PR #1328, opened 2019-05-29
Skip some WebDriver tests on staging and disable puppeteer_chrome_test

Hexcles:
Work around #1327 & #1330 to get the tree green.


Hexcles:
> Does it mean these tests will only run when a PR is being merged into the master?

No, these tests will only run against local dev_appserver, regardless of whether it's in a PR or on master.

Issue #464, opened 2018-08-21
Give subtests a testfile-unique hash to link to


tabatkins:
It was pointed out in the ecosystem-infra call that there shouldn't be any semantic difference between subtests in a file and individual tests in a folder; which organization method you use is purely an implementation detail.

However, individual tests have a reasonable name and you can link to them; subtests have neither. (They do have a name, but it's generally a human-readable sentence, not particularly appropriate for copy-pasting into a `<wpt>` element.)

Discussion during the call suggested that we derive a short, typeable name for each subtest, probably by taking the first 10 characters of the SHA hash of the subtest name.  This hash-name is then presented on wpt.fyi as an ID on the `<tr>` or test name, and shows up as a clickable link on the row.  It's also communicated in the individual-test-results API as another field in the subtest JSON.

In Bikeshed you'd then just link to the subtest like `foo.html#ba4`; same as with individual files, you can just right click -> copy link, paste into your spec, then remove the wpt.fyi link prefix.

foolip:
Implementation wise, this would be a matter of adding to the results JSON files that we serve, which we could do on the fly in `/api/results`.

PR #1337, opened 2019-05-31
Create UI Component for WPT Metadata

KyleJu:
## Description
Add the UI component of Metadata at the bottom of the page.

- The metadata will only shown in a sub-directory.
- All triaged metadata in a sub-directory, including the children directory ones are shown.

## Review Information
Open a staging instance and go to /IndexedDB.

## Follow Up
Show the specific status when there is one - plumbing test statue from backend


KyleJu:
Ready for a preliminary review

lukebjerring:
I'm not seeing anything on https://ui-wpt-metadata-dot-wptdashboard-staging.appspot.com/results/IndexedDB

Are there some instructions missing?

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,142 @@
...
+          margin-bottom: 4px;
+        }
+        .metadataNode img {
+          margin-right: 16px;
+          height: 24px;
+          width: 24px;
+        }
+      </style>
+      <div class="metadataNode">
+        <img src="/static/bug.svg" />
```
Hexcles: Use `<iron-icon icon="bug-report"></iron-icon>` instead.

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,154 @@
...
+        <iron-collapse id="metadata-collapsible">
+          <template is="dom-repeat" items="[[others]]" as="metadataNode">
+            <wpt-metadata-node
+              metadata-node="[[metadataNode]]"
+            ></wpt-metadata-node>
+          </template>
+        </iron-collapse>
+        <paper-button id="metadata-toggle" onclick="[[openCollapsible]]"
+          >Show more</paper-button
+        >
```
Hexcles: Nit: move the closing bracket to the previous line.

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,154 @@
...
+      </template>
+      <br />
```
Hexcles:
```suggestion
      <br>
```
XHTML requires tags to be closed. HTML does not.

webapp/views/wpt-app.js
```diff
@@ -237,6 +251,32 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
...
+          wptMetadata.push(wptMetadataNode);
+        }
+      }
+    }
+    return wptMetadata;
+  }
+
+  _loadAllMetadata() {
+    const url = new URL('/api/metadata', window.location);
+    url.searchParams.set('products', 'safari,chrome,edge,firefox')
```
Hexcles: The list of products should come from the query builder.

webapp/views/wpt-app.js
```diff
@@ -159,6 +165,11 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
         type: String,
         computed: '_computePath(subroute.path)',
       },
+      metadata: Array,
```
Hexcles: Perhaps this, along with `displayedMetadata` and related methods, should go to `wpt-metadata`. And then `<wpt-metadata>` would take only `path` as from `wpt-app`. @lukebjerring might have more experience regarding the tradeoff here.

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,198 @@
...
+      href: {
+        type: String,
+        computed: 'computeHref(metadataNode)'
+      }
+    };
+  }
+
+  computeHref(metadataNode) {
+    const prefix = 'https://';
+    if (metadataNode.url.substr(0, prefix.length) !== prefix) {
```
Hexcles:
```suggestion
    if (!metadataNode.url.startsWith(prefix)) {
```

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,198 @@
...
+      others: {
+        type: Array,
+        computed: 'computeOthers(displayedMetadata)'
+      }
+    };
+  }
+
+  constructor() {
+    super();
+    this.openCollapsible = this.handleOpenCollapsible.bind(this);
```
Hexcles: Do we need to declare `openCollapsible` as a property (with type function)?

webapp/components/wpt-metadata.js
```diff
@@ -0,0 +1,198 @@
...
+    const url = new URL('/api/metadata', window.location);
+    url.searchParams.set('products', productVal.join(','));
+    this.load(
+      window.fetch(url).then(r => r.json()).then(metadata => {
+        this.metadata = metadata;
+      })
+    );
+  }
+
+  computeDisplayedMetadata(path, metadata) {
```
Hexcles: We should have a test for this.

webapp/components/wpt-metadata.js
KyleJu: openCollapsible is used strictly within this wpt-metadata custom element. I think as long as openCollapsible is not invoked and changed outside of this class, it doesn't have to be declared as a property

KyleJu:
> The code looks good overall, but the demo doesn't seem to work -- I don't see any new stuff on the UI even after navigating to the indexeddb folder (which I believe has metadata). Could you provide a more detailed instruction of how to access the demo?

It wasnot deployed properly. I have deployed another instance and it should be good now!

lukebjerring:
Some feedback about the UI in the demo 
- For the given bug,
  - The test name has changed
  - It's hard to see which test it applies to (in the list)
  - There's nothing indicating that it's for _Chrome_ (and for FAIL status)

Perhaps action items are
- Make the test name a link (that navigates to that test)
- Show an icon of the product when there is one
- Show the specific status when there is one

This is what I'm imagining:
image

[Test path] > [Subtest] > [Product] > [Status]
The first two could be links (actually, we don't anchor to subtest names yet. See https://github.com/web-platform-tests/wpt.fyi/issues/464)

KyleJu:
> Some feedback about the UI in the demo
> 
> * For the given bug,
>   
>   * The test name has changed
>   * It's hard to see which test it applies to (in the list)
>   * There's nothing indicating that it's for _Chrome_ (and for FAIL status)
> 
> Perhaps action items are
> 
> * Make the test name a link (that navigates to that test)
> * Show an icon of the product when there is one
> * Show the specific status when there is one
> 
> This is what I'm imagining:
> image
> 
> [Test path] > [Subtest] > [Product] > [Status]
> The first two could be links (actually, we don't anchor to subtest names yet. See #464)

Implemented the first two suggestions in the follow-up commits - linking test name and showing icon. The third one requires some plumbing from the backend to retrieve metadata test status so I will implement it in a follow-up CL - https://github.com/web-platform-tests/wpt.fyi/projects/10#card-23632311

lukebjerring:
On https://ui-wpt-metadata-dot-wptdashboard-staging.appspot.com/results/entries-api?label=master&label=experimental&aligned

I see the metadata, but then when I navigate to the file itself, I don't. The metadata for that file should definitely show on the single file results page too.

lukebjerring:
"Triaged Metadata in /entries-api" doesn't read very well (is not user friendly).

How about "Relevant links for /entries-api results" ?

lukebjerring:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/1396 for detecting orphan cases (like the IndexedDB one, currently pointing to /IndexedDB/bindings-inject-key.html).

lukebjerring:
Can we open the external link in a new tab by default (`<a target="_blank">`, I think?)

webapp/components/path.js
```diff
@@ -57,6 +61,10 @@ const PathInfo = (superClass) => class extends superClass {
   computePathIsATestFile(path) {
     return /(\.(html|htm|py|svg|xhtml|xht|xml)(\?.*)?$)/.test(path);
   }
+
+  computePathIsRootDir(path) {
```
lukebjerring: Why isn't this just `path === '/'` ?

Issue #1240, opened 2019-03-22
Master runs from Taskcluster via Checks API cannot be ingested


foolip:
https://staging.wpt.fyi/runs?label=master&from=2019-03-15T00%3A00
https://wpt.fyi/runs?label=master&from=2019-03-15T00%3A00

After https://github.com/web-platform-tests/wpt/pull/15946 there have been no more runs, and no checks on PRs.

@Hexcles can you investigate and revert https://github.com/web-platform-tests/wpt/pull/15946 if we can't deploy a fix to prod today?

@jgraham FYI and sorry.

Hexcles:
Turns out we have a multitude of fatal issues here.

First, the Taskcluster runs did not run any tests at all (this is https://github.com/web-platform-tests/wpt/issues/16026).

On the wpt.fyi side, there are multiple fatal bugs that would prevent TC runs from being ingested even if the TC setup was correct. For now, I'm sure of the following issues:
1. The user whitelisting feature does not play well with `check_run` completion events: these events are sent by a `ghost` user. In fact, we should only enforce whitelisting for triggering events and always respond to all completion events.
2. A server-side flag wasn't properly set.

Hexcles:
So https://github.com/web-platform-tests/wpt.fyi/pull/1183#issuecomment-468309025 only verified that **PR** checks worked, but not master.

A fundamental difference between master and PRs is that master runs are sharded, so we should respond to `check_suite` events instead of individual runs so that we can merge all shards. (For a PR, we only care about the results of the affected tests, which are in a single run for each browser, so the approach in #1183 works.)

foolip:
@Hexcles I see results now, is this fully fixed? Thanks a bunch in either case. If there's concrete monitoring we need to avoid this again, can you file issues?

jgraham:
AIUI yhe checks API got disabled as that is still broken on the wpt.fyi side; what got fixed was not scheduling runs on master correctly.

foolip:
Oh, reverted again in https://github.com/web-platform-tests/wpt/pull/16027.

@Hexcles any ETA on when it can be enabled again? Every time we flip back and forth required checks have to be disabled for a period, so any way to dry run it in prod first would be great.

foolip:
@lukebjerring @Hexcles, has this been resolved or would things still break if we reland https://github.com/web-platform-tests/wpt/pull/15946?

[...]

PR #1343, opened 2019-06-06
Respond to completed CheckSuiteEvent from Taskcluster

lukebjerring:
## Description
Handles master-branch cases, where we have multiple shards that need to be grouped (by waiting until all shards are completed).
Fixes one of the many issues in https://github.com/web-platform-tests/wpt.fyi/issues/1240

### DIFF
--- a/api/azure/api_test.go
+++ b/api/azure/api_test.go
@@ -37,8 +37,8 @@ func TestHandleCheckRunEvent(t *testing.T) {
 	chrome := "chrome"
 	completed := "completed"
 	created := "created"
-	repoName := "wpt"
-	repoOwner := "web-platform-tests"
+	repoName := shared.WPTRepoName
+	repoOwner := shared.WPTRepoOwner
 	sender := "lukebjerring"
 	event := &github.CheckRunEvent{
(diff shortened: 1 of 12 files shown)


lukebjerring:
That's not my understanding. With this PR, we wait until the CheckSuiteEvent `completed`, meaning that all TaskGroups have completed or failed on TaskCluster. Thus, when asking for all the separate CheckRuns, we'll get multiple shard Task IDs (but they'll have the same TaskGroup). We group the DetailURLs by the TaskGroup (in a set), and then fetch all the metadata for that TaskGroup (same as the way we respond to StatusEvent already).

Hexcles:
Yes, the new code looks correct to me, but the problem is we didn't change the existing handler for CheckRunEvent, so when we turn on the flag to process Taskcluster check events again, we will react to both CheckRunEvent and CheckSuiteEvent.

lukebjerring:
Ah, now I get what you're saying. Removed the old code in https://github.com/web-platform-tests/wpt.fyi/pull/1343/commits/1452944414b6ec6aab511b82f081cbe881e56df6

Issue #1344, opened 2019-06-07
Edge Dev run incorrectly labeled as stable


foolip:
https://wpt.fyi/results/?run_id=230810012 is the first run from https://github.com/web-platform-tests/wpt/pull/16920.

It came from this build:
https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=18447

Maybe it would be fixed by https://github.com/web-platform-tests/wpt/pull/17224, but I wonder if the run_info shouldn't include this information anyway/already?

Hexcles:
Thanks, Philip! I'm fairly certain https://github.com/web-platform-tests/wpt/pull/17224 is the fix; I missed that during code review.

We do need to explicitly set that flag, unfortunately. wptrunner does not infer the browser channel itself.

Hexcles:
Also manually fixed the existing test runs on prod and staging.

foolip:
Confirmed fix in https://wpt.fyi/runs and looks like you relabeled the previous runs?

Hexcles:
Yes, I did.

PR #1345, opened 2019-06-07
Add further branching options for default homepage query

lukebjerring:
Relevant to https://github.com/web-platform-tests/wpt.fyi/issues/1344

## Description
Adds (separate) options for experimental, aligned runs, and ensuring `edge[edgechromium]`.

## Review
I've enabled `experimentalAligned` and `edgeChromiumByDefault` in staging.
Visit https://edgium-dot-wptdashboard-staging.appspot.com/


foolip:
From https://edgium-dot-wptdashboard-staging.appspot.com/ it looks like the default query is still "chrome[experimental], edge, firefox[experimental], safari[experimental]".

I'm not entirely sure what the effect of this PR will be, but is it necessary to look at the `edgechromium` label? Could we not instead just use edge[experimental], thus making just https://wpt.fyi/results/?label=master&label=experimental&aligned the default view / URL?

webapp/components/wpt-flags.js
```diff
@@ -27,7 +27,9 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
     return [
       'colorHomepage',
       'diffFromAPI',
+      'edgeChromiumByDefault',
```
Hexcles: I don't think this flag (and the corresponding branching) is necessary, because we will never have both EdgeHTML & EdgeChromium on the same channel, and this branching happens after the branching for channel preferences (`experimentalByDefault`). In other words, if `experimentalByDefault` is true, then we will only get EdgeChromium, so this flag becomes a no-op.

Even when we are looking at history runs, this flag is also a no-op. At no point in history have we ever received both EdgeHTML & EdgeChromium on the same channel, so the decision/timeframe decides which Edge you'll get.

PR #1346, opened 2019-06-07
Consolidate the path logic into a shared class

### LINKED ISSUE #1331: Live test link is broken on staging
Broken by #1295 : `scheme` is undefined on `wpt-app`.

lukebjerring:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1331

## Description
Pulls out some re-used properties relating to the path attribute (deduplication) and apply the shared mixin where needed to fix #1331


Hexcles:
The live link is still broken. The scheme doesn't get propagated correctly. Let me give it a try.

Hexcles:
Never mind. It seems to work after rebasing.

PR #1351, opened 2019-06-12
Attempt to fix WCT

Hexcles:
## Description

There is a multitude of problems that lead to flaky web components test. This PR addresses all of the issues I can find but is not guaranteed to solve the problem (only to alleviate it).

1. Some network fetches aren't always intercepted, which would cause flaky failures. Looking at the devtools, I've identified two cases:
    a) `test-runs-query-builder.html` stubs and restores `window.fetch` at the test level (`setup` and `teardown`). However, the query builder issues API requests after it becomes ready, which might happen after the test finishes, in which case the fetches would slip through. This is fixed by setting up the stub at the suite level (`suiteSetup` and `suiteTeardown`).
    b) `wpt-interop.html` has the same issue. In addition, its URL matching is incorrect: in one case, a pathname is matched against the full URL (https://storage.googleapis.com...); this is fixed by modifying the mock data. And to prevent future regression, an exception is thrown if a match can't be found.
2. Travis builders doesn't seem to have enough resources to run Chrome & Firefox in parallel reliably and the tests would hang. After a lot of trial and errors (including increasing the timeout), the only way to fix it is to test against one browser at a time (by taking away the `browsers` field from config and manually specifying the browsers from CLI).

## Review Information

Please review the final diff and ignore my trial and errors.

## Changes

In addition to fixing the issues above, there are some drive-by changes:

1. Upgrade npm packages.
2. Improve error handling in `reftest-analyzer.js` so that it doesn't attempt to fetch an empty URL when `before` or `after` isn't specified.

### DIFF
--- a/webapp/components/reftest-analyzer.js
+++ b/webapp/components/reftest-analyzer.js
@@ -146,8 +146,14 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
     return {
       curX: Number,
       curY: Number,
-      before: String,
-      after: String,
+      before: {
+        type: String,
+        value: '',
+      },
+      after: {
+        type: String,
+        value: '',
+      },
       selectedImage: {
         type: String,
         value: 'before',
@@ -174,6 +180,9 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
 
     // Set the img srcs manually so that we can promisify them being loaded.
     const imagePromises = ['before', 'after'].map(prop => {
+      if (!this[prop]) {
+        return Promise.reject(`${prop} is empty`);
+      }
       const img = this.shadowRoot.querySelector(`#${prop}`);
       const loaded = new Promise((resolve, reject) => {
         img.onload = resolve;
--- a/webapp/components/test/test-runs-query-builder.html
+++ b/webapp/components/test/test-runs-query-builder.html
@@ -22,27 +22,31 @@
 import '../test-runs-query-builder.js';
 import { PolymerElement } from '../../node_modules/@polymer/polymer/polymer-element.js';
 import { Channels } from '../product-info.js';
+
 suite('TestRunsQueryBuilder', () => {
   let queryBuilder, sandbox;
 
-  setup(() => {
+  suiteSetup(() => {
     sandbox = sinon.sandbox.create();
     // Spoof an empty result for /api/shas to speed up tests.
-    const realFetch = window.fetch;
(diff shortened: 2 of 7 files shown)


Hexcles:
TBR @foolip because master is broken.

webapp/components/reftest-analyzer.js
```diff
@@ -174,6 +180,9 @@ class ReftestAnalyzer extends LoadingState(PolymerElement) {
 
     // Set the img srcs manually so that we can promisify them being loaded.
     const imagePromises = ['before', 'after'].map(prop => {
+      if (!this[prop]) {
+        return Promise.reject(`${prop} is empty`);
```
foolip: I sent https://github.com/web-platform-tests/wpt.fyi/pull/1354 to simplify this a bit. When a function returns a promise in two places or does a lot of work outside the promified paths I always wonder if it's possible for it to throw an exception instead of rejecting, and just having a single promise with all the work inside makes it obvious.

webapp/components/test/test-runs-query-builder.html
```diff
@@ -22,29 +22,33 @@
...
+    // This is done in suiteSetup/suiteTeardown to avoid async fetches slipping
+    // through between tests.
+    const ignore = new RegExp('/api/(shas|versions)');
     sandbox.stub(window, 'fetch', url => {
-      const ignore = new RegExp('/api/(shas|versions)');
       if (ignore.test(url.pathname)) {
         return Promise.resolve(new Response('[]'));
       }
-      return realFetch(url);
+      throw url;
```
foolip: The `fetch` method which this stubs out never throws an exception, it rejects. Fix would be similar to https://github.com/web-platform-tests/wpt.fyi/pull/1354.

Hexcles: This wasn't intended to match the behaviour of `fetch`. Rather, an unusual exception is raised to show that there is a programming error. If we were to test the error handling of the component, we would then set up another set of expectations and reject the promise if the URL matches one of the failure expectations.

foolip: In other words, this error shouldn't be reached while executing the tests, that would be an error in the test code? Throwing an error with a more elaborate error message might make that clearer.

Hexcles: Yes. And you're right; I should write a message instead of just throwing the url.

PR #1353, opened 2019-06-13
Remove whitespaces between path components

Hexcles:
This is a trivial fix for a recent regression to make the top path copy-pastable again.


foolip:
Actually it doesn't entirely work, there's still a space after wpt: "wpt /ambient-light/AmbientLightSensor.https.html"

Hexcles:
@foolip yeah that was intentional, cuz `wpt/` isn't really part of the test path.

foolip:
Oh right, makes sense :)

Issue #1340, opened 2019-06-04
No link to get back to results from reftest comparison


gsnedders:
It seems less than ideal that one is reliant on the browser's back button to get back from the reftest comparison.

PR #1356, opened 2019-06-14
Improve buttons for reftest-analzyer

### LINKED ISSUE #1339: Unable to open reftest "compare" link behind current tab
The compare link doesn't provide "open link in [x]" options in its context menu, and cmd+click in Chrome/Firefox doesn't cause it to open in a new tab.

Hexcles:
See individual commits -- fixes #1339 and #1340 respectively.

Demo: https://analzyer-button-dot-wptdashboard-staging.appspot.com/results/css/css-fill-stroke/paint-order-001.tentative.html?label=master&label=experimental

Try the new COMPARE button, as well as the "go back" button in the reftest analyzer.

### DIFF
--- a/webapp/components/abstract-test-file-results-table.js
+++ b/webapp/components/abstract-test-file-results-table.js
@@ -31,7 +31,7 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
     top: 0;
     z-index: 1;
   }
-  td {
+  td, .ref-button {
     padding: 0;
     height: 1.5em;
   }
@@ -82,7 +82,7 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
             <template is="dom-if" if="[[result.screenshots]]">
               <a class="ref-button" href="[[ computeAnalyzerURL(result.screenshots) ]]">
                 <iron-icon icon="image:compare"></iron-icon>
-                <span>COMPARE</span>
+                COMPARE
               </a>
             </template>
           </td>
--- a/webapp/components/abstract-test-file-results-table.js
+++ b/webapp/components/abstract-test-file-results-table.js
@@ -138,10 +138,10 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
 
   computeAnalyzerURL(screenshots) {
     if (!screenshots) {
-      throw "empty screenshots";
+      throw 'empty screenshots';
     }
     const url = new URL('/analyzer', window.location);
-    for (const [_, sha] of screenshots) {
+    for (const sha of screenshots.values()) {
       url.searchParams.append('screenshot', sha);
     }
     return url.href;
--- a/webapp/components/abstract-test-file-results-table.js
+++ b/webapp/components/abstract-test-file-results-table.js
@@ -6,7 +6,6 @@
 
 import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
 import '../node_modules/@polymer/polymer/lib/elements/dom-repeat.js';
-import '../node_modules/@polymer/paper-button/paper-button.js';
 import '../node_modules/@polymer/iron-icon/iron-icon.js';
 import '../node_modules/@polymer/iron-icons/image-icons.js';
 import { html } from '../node_modules/@polymer/polymer/polymer-element.js';
@@ -39,11 +38,11 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
   td code {
     white-space: pre-wrap;
   }
-  td code, td paper-button {
+  td code, .ref-button {
     line-height: 1.6em;
     padding: 0 0.25em;
   }
-  td.sub-test-name {
+  td.sub-test-name, .ref-button {
     font-family: monospace;
   }
   td.result {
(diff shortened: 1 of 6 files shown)


Hexcles:
Note to myself: we should probably write a test for the `screenshots` shuffling logic.

webapp/components/abstract-test-file-results-table.js
```diff
@@ -78,10 +80,10 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
           <td class$="[[ colorClass(result.status) ]]">
             <code>[[ subtestMessage(result) ]]</code>
             <template is="dom-if" if="[[result.screenshots]]">
-              <paper-button onclick="[[compareReferences(result)]]">
+              <a class="ref-button" href="[[ computeAnalyzerURL(result.screenshots) ]]">
                 <iron-icon icon="image:compare"></iron-icon>
-                Compare
-              </paper-button>
+                <span>COMPARE</span>
```
foolip: Compare &rarr; COMPARE is because previously some style capitalized the text, right?

webapp/components/abstract-test-file-results-table.js
```diff
@@ -142,6 +135,17 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
     }
     return 'result';
   }
+
+  computeAnalyzerURL(screenshots) {
+    if (!screenshots) {
+      throw "empty screenshots";
+    }
+    const url = new URL('/analyzer', window.location);
+    for (const [_, sha] of screenshots) {
```
foolip:
```suggestion
    for (const sha of screenshots.values()) {
```

(`screenshots` is a Map)

webapp/components/test-file-results.js
```diff
@@ -164,8 +161,17 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
           status: data && data.status,
           message: data && data.message,
         };
-        if (this.reftestAnalyzer) {
-          result.screenshots = data && data.screenshots;
+        if (this.reftestAnalyzer && data && data.screenshots) {
+          // Clone the data because we might modify it.
+          const screenshots = JSON.parse(JSON.stringify(data.screenshots));
```
foolip: What is the type of `data.screenshots`? Is there a more straightforward way to clone it depending on type?

Hexcles:
@foolip question: I don't understand why the cursor keeps switching between pointer and default like crazy when I hover on "COMPARE". It's in an `<a>` tag with `display: block; float: right;`.

foolip:
@Hexcles I'm not sure either. It doesn't seem to be scripts toggling anything back and forth between states. Looking at the computed style of the `a` element it looks like it's always `cursor: pointer`. So... needs minifying to figure out what's up.

webapp/components/test-file-results.js
Hexcles: It is a plain object (path1: sha1, path2: sha2, ...) deserialized from JSON.

webapp/components/abstract-test-file-results-table.js
```diff
@@ -142,6 +135,17 @@ class AbstractTestFileResultsTable extends WPTColors(TestRunsBase) {
     }
     return 'result';
   }
+
+  computeAnalyzerURL(screenshots) {
+    if (!screenshots) {
+      throw "empty screenshots";
```
Hexcles:
```suggestion
      throw 'empty screenshots';
```

Hexcles:
Looks like something finicky with hit testing or I misunderstand CSS layout? There's a whitespace between the icon and the text "COMPARE", when I click on there, I end up selecting the `<code>` block on the back. I played with `z-index` which didn't seem to help. Is this supposed to happen? Both `<code>` and `<a>` are `block`s, and `<a>` floats to the right.

webapp/components/test-file-results.js
foolip: Could be copied with `Object.assign` then?

foolip:
Given the lack of scripts involved I would suspect a Chromium bug. Are you able to minimize it into a standalone test case?

Hexcles:
@foolip I've been trying to, but so far can't reproduce this in a minimized setup. My current suspicion is that this might have something to do with the shadow DOM of `iron-icon`.

Hexcles:
@foolip I've been trying to, but so far can't reproduce this in a minimized setup. My current suspicion is that this might have something to do with the shadow DOM of `iron-icon`.

foolip:
Hmm, I guess that capturing the state of the page and serializing it is even harder when Shadow DOM is involved, and I don't even know how to do it well without Shadow DOM because of style.

webapp/components/abstract-test-file-results-table.js
Hexcles: Yes.

webapp/components/test-file-results.js
Hexcles: TIL!

```suggestion
          const screenshots = Object.assign({}, data.screenshots);
```

Hexcles:
@lukebjerring helped me to figure out it was because of the neighbouring `<code>` block (that shows FAIL/PASS) has `position:absolute` which somehow confuses the layout.

Issue #1300, opened 2019-05-08
Claimed regressions when there appear to be done


foolip:
### Check
A link to the GitHub Check run containing my issue:
https://github.com/web-platform-tests/wpt/pull/16726/checks?check_run_id=121763196

### Description
This says "there are some newly-failing results"  but in https://wpt.fyi/results/html/dom/interfaces.https.html%3Finclude=HTML.%2A?diff&filter=ADC&run_id=5670774610329600&run_id=5178380483821568 it looks like the only difference is two tests fewer than before, and those tests were originally failing. Removing failing tests shouldn't count as a regression.

@lukebjerring

PR #1372, opened 2019-06-20
Add feature for changes-only regression detection

lukebjerring:
## Description
Should help with https://github.com/web-platform-tests/wpt.fyi/issues/1300

Changes the checks API to fetch the diff twice, with only `filter=C` the second time. The latter fetch is used when counting regressions, while the full diff is used for the summary text.

### DIFF
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -24,6 +24,7 @@ import (
 const CheckProcessingQueue = "check-processing"
 
 const failChecksOnRegressionFeature = "failChecksOnRegression"
+const onlyChangesAsRegressionsFeature = "onlyChangesAsRegressions"
 
 // updateCheckHandler handles /api/checks/[commit] POST requests.
 func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
--- a/api/checks/update.go
+++ b/api/checks/update.go
@@ -201,11 +201,14 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, suite sha
 		PRNumbers:  suite.PRNumbers,
 	}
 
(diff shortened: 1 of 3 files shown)


api/checks/update.go
```diff
@@ -201,7 +201,17 @@ func getDiffSummary(aeAPI shared.AppEngineAPI, diffAPI shared.DiffAPI, suite sha
 		PRNumbers:  suite.PRNumbers,
 	}
 
-	regressions := diff.Differences.Regressions()
+	var regressions mapset.Set
+	if aeAPI.IsFeatureEnabled("onlyChangesAsRegressions") {
```
foolip: Hmm, would this mean that changing a support file and breaking lots of tests wouldn't count as a regression, or what's the definition of changes again? Is it that total number of tests have changed?

Can this change, and other things that feed into what the check result and text will be, be tested?

lukebjerring: Changing a file does not affect the WPT report output - it only changes which tests are considered affected (the input).

`C` / changed means "existed before, and after, and the result is different". So for some `a.html` test, adding a new subtest B would not count as "changed"; i.e. if a new subtest is failing, that wouldn't be counted as a regression.

foolip: OK, so it's entirely based on the results, I should have been able to guess that.

Will a test changing its failure mode, going from FAIL to TIMEOUT for example, count as a regression?

lukebjerring: No; we just check for pass-or-not comparisons for the diff counts.

lukebjerring:
Actually, we still have a subset of queries that operate on the summaries (all the GET requests), so we should ensure `diff` always uses searchcache in those case before we tweak the docs.

PR #1374, opened 2019-06-20
Allow more flexible anomalies queries

lukebjerring:
## Description
Changes the anomalies query to have a multi-select for the other browsers (so that we can drop, say, one browser and compare against only 2 others).


lukebjerring:
A while! I think it was deleted in one PR and renamed (html => js) in another, and survived the merge.

PR #1380, opened 2019-06-26
Restore the test counts message when viewing results

lukebjerring:
## Description
It got dropped in the migration to `wpt-app`. The `N tests (M subtests) from` message is being computed, but on the nest `wpt-results` class, which is not what we're using to populate the banner.


lukebjerring:
Right you are, Hexcles!

PR #1391, opened 2019-07-17
Fetch any/all matching product keys in a single datastore lookup

lukebjerring:
## Description
Fixes the underlying cause of datastore timeouts, which in turn fixes https://github.com/web-platform-tests/wpt.fyi/issues/973

Example that is timing out: https://wpt.fyi/api/runs?pr=17560
(And its corresponding fix: https://runs-by-pr-dot-wptdashboard-staging.appspot.com/api/runs?pr=17560)

### DIFF
--- a/api/test_runs.go
+++ b/api/test_runs.go
@@ -7,6 +7,7 @@ package api
 import (
 	"encoding/json"
 	"net/http"
+	"strings"
 
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/appengine/datastore"
@@ -21,6 +22,7 @@ const paginationTokenFeatureFlagName = "paginationTokens"
 //     sha: SHA[0:10] of the repo when the tests were executed (or 'latest')
 func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := shared.NewAppEngineContext(r)
+	log := shared.GetLogger(ctx)
 	store := shared.NewAppEngineDatastore(ctx, true)
 	aeAPI := shared.NewAppEngineAPI(ctx)
 	q := r.URL.Query()
@@ -52,6 +54,11 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		if pr != nil && aeAPI.IsFeatureEnabled("runsByPRNumber") {
 			filters.SHAs = getPRCommits(aeAPI, *pr)
+			if len(filters.SHAs) < 1 {
+				log.Warningf("PR %s returned no commits from GitHub", *pr)
+			} else {
+				log.Infof("PR %s returned %v commits: %s", *pr, len(filters.SHAs), strings.Join(filters.SHAs.SevenCharSHAs(), ","))
+			}
 		}
 		var runsByProduct shared.TestRunsByProduct
 		runsByProduct, err = LoadTestRunsForFilters(store, filters)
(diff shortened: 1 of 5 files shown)


api/test_runs.go
```diff
@@ -52,6 +54,11 @@ func apiTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		if pr != nil && aeAPI.IsFeatureEnabled("runsByPRNumber") {
 			filters.SHAs = getPRCommits(aeAPI, *pr)
+			if len(filters.SHAs) < 1 {
+				log.Warningf("PR %v returned no commits from GitHub", *pr)
+			} else {
+				log.Infof("PR %v returned %v commits: %s", *pr, len(filters.SHAs), strings.Join(filters.SHAs.SevenCharSHAs(), ","))
```
Hexcles: Nit:
```suggestion
				log.Infof("PR %v returned %d commits: %s", *pr, len(filters.SHAs), strings.Join(filters.SHAs.SevenCharSHAs(), ","))
```

I'd like to avoid using `%v` so that `golint` can give me warnings if variables are not what I think they are...

shared/test_run_filter.go
```diff
@@ -32,6 +32,15 @@ func (s SHAs) FirstOrLatest() string {
 	return s[0]
 }
 
+// Returns an array of the given SHAs' first 7-chars.
+func (s SHAs) SevenCharSHAs() []string {
```
Hexcles: Maybe `ShortSHAs`?

shared/test_run_query.go
```diff
@@ -126,21 +128,23 @@ func (t testRunQueryImpl) LoadTestRunKeys(
...
 			var ids TestRunIDs
 			if ids, err = loadKeysForRevision(t.store, baseQuery, sha); err != nil {
 				return nil, err
 			}
 			for _, id := range ids {
-				globalKeyFilter.Add(id)
+				globalIDFilter.Add(id)
 			}
 		}
+		log.Debugf("Found %v keys across %v resivions", globalIDFilter.Cardinality(), len(revisions))
```
Hexcles:
```suggestion
		log.Debugf("Found %d keys across %d revisions", globalIDFilter.Cardinality(), len(revisions))
```

Two nits: 1. `%v`, 2. typo "resivions"

The same applies to the other logs in this file.

shared/test_run_query.go
```diff
@@ -296,14 +332,17 @@ func contains(s []string, x string) bool {
 
 // Loads any keys for a revision prefix or full string match
 func loadKeysForRevision(store Datastore, query Query, sha string) (result TestRunIDs, err error) {
```
Hexcles: Shall we rename this to `loadIDsForRevision`, and `loadKeysForBrowserVersion` to `loadIDsForBrowserVersion`?

By the way, `loadKeysForBrowserVersion` currently returns a mapset, which is inconsistent with `loadKeysForRevision`. Looking at the call sites, we should perhaps make both return mapsets.

shared/test_run_query.go
```diff
@@ -152,44 +156,76 @@ func (t testRunQueryImpl) LoadTestRunKeys(
...
-			query = query.Filter("TimeStart >=", *from)
-		}
-		if to != nil {
-			query = query.Filter("TimeStart <", *to)
+			log.Debugf("Found %v keys for %s", versionIDs.Cardinality(), product.BrowserVersion)
+			productIDFilter = merge(productIDFilter, versionIDs)
 		}
 
+		// If we have a specific set of possibilities, it's much cheaper to
+		// turn the query on its head (filter the entities).
```
Hexcles: One more nit: this function is quite big now, especially the part below which has too many levels of nesting. Let's pull the two branches of the if switch out into two functions.

shared/test_run_filter.go
lukebjerring: Done.

shared/test_run_query.go
lukebjerring: Done.

shared/test_run_query.go
lukebjerring: Done (for the client-side filtering only).

PR #1393, opened 2019-07-19
Surface platform information for test runs

lukebjerring:
## Description
Adds logos for the different OSes that we run, and surfaces them in the UI.

## Review
- Visit https://platform-dot-wptdashboard-staging.appspot.com/results/?label=master&label=experimental&aligned
- See windows/linux logos

And
- Visit https://platform-dot-wptdashboard-staging.appspot.com/runs
- See logos, but more compact.


LukeZielinski:
So looks like /runs places the icons in topLeft + bottomRight while /results uses bottomLeft + bottomRight. Should they be consistent (probably bottomLeft+bottomRight on both)?

lukebjerring:
Tweaked to keep bottom-left in both cases, PTAL.

PR #1397, opened 2019-07-24
Show diff + filtering on single file views

lukebjerring:
## Description
Includes the diff column in the single-file view, so that really large tests (e.g. /html/dom/interfaces.https.html) can be filtered when diffing.

### DIFF
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -35,6 +35,9 @@ class TestFileResultsTable extends WPTColors(PathInfo(TestRunsBase)) {
     padding: 0.25em;
     height: 1.5em;
   }
+  td.diff {
+    border-left: 8px solid white;
+  }
   td code {
     line-height: 1.6em;
     white-space: pre-wrap;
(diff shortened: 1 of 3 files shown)


webapp/components/test-file-results-table.js
```diff
@@ -276,6 +310,27 @@ class TestFileResultsTable extends WPTColors(PathInfo(TestRunsBase)) {
     }
     return `/api/screenshot/${shot}`;
   }
+
+  diffDisplay(results) {
+    if (results[0].status !== results[1].status) {
+      const passed = results.map(r => ['OK', 'PASS'].includes(r.status));
+      if (passed[0] && !passed[1]) {
+        return '↓ 1';
```
Hexcles: IMHO, `↓` and `↑` are actually more confusing to me than simple `-` and `+`.

lukebjerring:
```suggestion
        return '-1';
```

webapp/components/test-file-results-table.js
```diff
@@ -276,6 +310,27 @@ class TestFileResultsTable extends WPTColors(PathInfo(TestRunsBase)) {
...
+      } else if (passed[1] && !passed[0]) {
+        return '↑ 1';
```
lukebjerring:
```suggestion
        return '+1';
```

PR #1398, opened 2019-07-25
Show before and after in non-regressing check summaries

lukebjerring:
Currently we only show the "After", which is based on the premise that newly-failing tests would be considered regressing. Now, however, only existing passes that change to failure are considered regressions, so we should show before and after in all cases (in case new tests exist).

### DIFF
--- a/api/checks/summaries/completed.md
+++ b/api/checks/summaries/completed.md
@@ -6,10 +6,10 @@ There were no regressions detected in the results.
 
 ### Results
 
-Test | `{{ printf "%.7s" .HeadRun.FullRevisionHash }}`
+Test | `master` | `{{ printf "%.7s" .HeadRun.FullRevisionHash }}`
 --- | ---
 {{ range $test, $results := .Results -}}
-{{ escapeMD $test }} | {{ index $results 0 }} / {{ index $results 1 }}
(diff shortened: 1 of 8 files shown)


api/checks/summaries/completed.md
```diff
@@ -23,6 +23,6 @@ Other links that might be useful:
 {{- if .MasterDiffURL }}
 - [`{{ printf "%.7s" .HeadRun.FullRevisionHash }}` vs latest master]({{ .MasterDiffURL }})
 {{- end }}
-- [Latest results for `{{ printf "%.7s" .HeadRun.FullRevisionHash }}`]({{.HostURL}}?sha={{.HeadRun.Revision}})
+- [Latest results for `{{ printf "%.7s" .HeadRun.FullRevisionHash }}`]({{.HostURL}}?sha={{.HeadRun.Revision}}&label=pr_head)
```
Hexcles: I don't think the `pr_head` label is necessary -- the `HeadRun.Revision` is on a PR branch so it has to be a PR head run, but it doesn't matter either way.

lukebjerring: It is; this is an unrelated drive-by. Note that we "fake" the SHA for both runs as the SHA with changes, so there's a race condition.

PR #1399, opened 2019-07-26
Log the number of renames found

lukebjerring:
Gives more insight into why the UI might fail to consolidate renamed results


lukebjerring:
TBR-merging this trivial change.

PR #1400, opened 2019-07-26
Always show and detect renames when diffing

lukebjerring:
## Description
* Front-end: Always run a separate /api/diff fetch, so that we get any computed renames.
* Back-end: Include the renames fetch when we delegate to the searchcache for diffing.


shared/run_diff.go
```diff
@@ -424,12 +425,23 @@ func runDiffFromSearchResponse(before, after TestRun, scDiff SearchResponse) (Ru
 		}
 	}
 
+	var renames map[string]string
+	if aeAPI.IsFeatureEnabled("diffRenames") {
```
Hexcles: Ideally we should reuse this piece of code (it's exactly the same as in `GetRunsDiff`). Maybe next time.

PR #1407, opened 2019-07-30
npm update

Hexcles:
A simple `npm update --depth 9999` to address some security warnings. Tests should continue to pass.


KyleJu:
@Hexcles test failures?

Hexcles:
@KyleJu that's #1403 . Could you review #1404 ?

Issue #1401, opened 2019-07-29
Add support for displaying subtest in wpt-metadata 


KyleJu:
Per James Graham comments in the email thread, it is important to support triaging 
use cases where different subtests may be affected by entirely different 
bugs.

Documentation changes:
Add an optional subtest field in wpt-metadata repo README
Documentation changes

API changes:
api/metadata? will return one additional data field - subtest

Display Metadata UI changes:
Displaying subtest in the metadata
-  If path is a subfolder, only show the tests which have been triaged, do not display subtest information.
- if path is a test, show all subtests under this test which have been triaged

Amend Metdata UI changes:
support amending subtests in the UI

KyleJu:
@lukebjerring thoughts on the UI changes to display subtests?

lukebjerring:
If _any_ of the visible set of results is relevant to the metadata/link, it should be shown. For that reason, it might be best to return metadata from the searchcache, and only use a separate fetch when looking at an unfiltered view.

But, regarding subtests, if the subtests is being included in the summary numbers, include the metadata/link.

KyleJu:
> If _any_ of the visible set of results is relevant to the metadata/link, it should be shown. For that reason, it might be best to return metadata from the searchcache, and only use a separate fetch when looking at an unfiltered view.

Good point. Alternatively, we can filter metadata based on directories displayed ( from displayedNodes) on the page

> 
> But, regarding subtests, if the subtests is being included in the summary numbers, include the metadata/link.

I am afraid that it might be too many subtests at some points. What if we only display subtests in the subtest result page

KyleJu:
This has been fixed

PR #1409, opened 2019-08-01
Plumbing status and subtest Information

KyleJu:
## Description
Step towards fixing #1401. Plumbing status and subtest Information.

## Review Information
Go to /api/metadata?product=chrome

## Changes
Add subtest and status fields to the response of metadata


shared/metadata.go
```diff
@@ -28,9 +28,14 @@ type MetadataResult struct {
...
 	// the WPT source reposiory.
 	Test string `json:"test"`
+	// Subtest is the name of a subtest that is associated with this test.
+	Subtest string `json:"subtest,omitempty"`
 	// URLs represents a list of bug urls that are associated with
 	// this test.
 	URLs []string `json:"urls,omitempty"`
+	// URLs represents a list of test status that are associated with
+	// this test.
+	Status []TestStatus `json:"status,omitempty"`
```
Hexcles: Shall we return the string representation instead? Otherwise, we'd have to at least document the enum values somewhere.

api/README.md
```diff
@@ -816,17 +826,27 @@ __`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
...
+        "subtest":"subtestBar",
         "urls": [
             "bugs.chromium.org/p/chromium/issues/detail?id=592874",
             ""
-        ]
+        ],
+        "status": [
+            4,
+            0
+      ]
```
Hexcles:
```suggestion
        ]
```

shared/metadata.go
Hexcles: FWIW, the current result JSON uses strings (e.g. `status: "FAIL"`)

api/README.md
```diff
@@ -790,16 +790,26 @@ __`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
...
+      "subtest":"subtestFoo",
       "urls":[
          "bugs.chromium.org/p/chromium/issues/detail?id=934844",
          "bugs.webkit.org/show_bug.cgi?id=167052"
+      ],
+      "status":[
+         "FAIL",
+         "FAIL"
       ]
    },
```
lukebjerring: This format is now quite different than the format of the META.yml files (which have a single URL, and multiple results combinations to which it applies). Should we be revising the format of `/api/metadata` in a similar way to match?

KyleJu: I intended to format /api/metadata response similar to [/api/search,](https://github.com/web-platform-tests/wpt.fyi/blob/44d1b0905d9b194779ac4cca03bbe6c6ac256b54/api/README.md#apisearch). I think the current format (each combinations of test and subtest has a URL) is easy to use if users are searching for test names instead of URLs.

KyleJu:
Discussion offline. Will change the format of Metadata response similar to wpt-metadata YML format without su

Issue #1421, opened 2019-08-16
Triaging regressions is inefficient, especially for reftests


foolip:
Trying to go over all the results in https://github.com/web-platform-tests/wpt/pull/17219#issuecomment-521894965 I found myself wanting some power user tools for triage:
- Stepping from one test to the next without navigating up and down the directory structure. Now I have to be careful to never make a mistake about where I came from and what would be next, and it's time consuming. Maybe this could be implemented as flattening of all tests into a single list, or keyboard shortcuts.
- Seeing reftests failure screenshots inline or overlayed, so that one doesn't need to leave the main UI totally. In the reftest analyzer the test path is not shown, so to copy it one has to step out of it.

KyleJu:
Taken from offline chat between Stephen and Ian Kilpatrick
"Ultimately he would like to see a UI that looks like the layout tests result viewer, where it just shows a flattened list of tests and you can just spacebar through each test to see the diff, but at a very basic level it would also just be good to inline the reftest screenshots in a readable way (currently Show Details shows the tiniest screenshot ever)"

PR #1422, opened 2019-08-16
Move to next + prev individual test when n + p are pressed

lukebjerring:
### Description
When N is pressed on the document body, finds the next full test path that isn't the current path, or the first test path, and navigates to it.

Feature kinda requested in #1421. We can build off the keycodes and add "expand" shorthands in a follow-up PR.

### Review
Head to the deployed env and try it.
Please make sure to test the interop view's behaviour works too; hard to test locally.

### DIFF
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -68,10 +68,10 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
 
       <section class="search">
         <div class="path">
-          <a href="/[[page]]/?[[ query ]]" on-click="navigate">wpt</a>
+          <a href="/[[page]]/?[[ query ]]">wpt</a>
           <!-- The next line is intentionally formatted so to avoid whitespaces between elements. -->
           <template is="dom-repeat" items="[[ splitPathIntoLinkedParts(path) ]]" as="part"
-            ><span class="path-separator">/</span><a href="/[[page]][[ part.path ]]?[[ query ]]" on-click="navigate">[[ part.name ]]</a></template>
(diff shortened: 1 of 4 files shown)


foolip:
Sweet, it works! Improvements I'd like to see:
- If I've traversed into a directory, start from there rather from the individual test I was last look at.
- A way to go back one test, as one is likely to quickly step through tests and go one too far. Following [Gmail](https://support.google.com/mail/answer/6594?co=GENIE.Platform%3DDesktop&hl=en), maybe n+p?

webapp/views/wpt-app.js
```diff
@@ -273,6 +274,16 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
     return linkedParts;
   }
 
+  handleKeyDown(e) {
+    // Ignore when something other than body has focus.
```
foolip: Is that just the search box, or are there other things one has to avoid focusing?

webapp/views/wpt-results.js
```diff
@@ -946,6 +942,19 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
     this._fetchedQuery = query; // Debounce.
     this.reloadData();
   }
+
+  moveToNext() {
```
foolip: I wasn't able to test this because I got "Failed to fetch interop data"

lukebjerring:
Added 'p' for prev too, PTAL (once https://github.com/web-platform-tests/wpt.fyi/pull/1422/commits/471c1cb972e4848c455cdedc9efe5ca18b4d0f3d goes green it'll be deployed.)

webapp/views/wpt-results.js
foolip: Now it loads, but I can't get n/p to work. Does it work for you?

PR #1423, opened 2019-08-19
[lifecycle] Basic data models & WPTFYI_RECEIVED/VALID stages

Hexcles:
## Description

Design doc: https://docs.google.com/document/d/125dg-NST0xnEVqN7a-LrfGxPueSwXvDukAjOM8n5q50/

## Review Information

You can use https://lifecycle-dot-wptdashboard-staging.appspot.com/api/results/upload to upload a run, and check the `PendingTestRun` table in Datastore to see the new run going through WPTFYI_RECEIVED to VALID. I already did a manual integration test of the flow.

## Changes

See each commit title. Changes are split into logical commits, so it might be easier to review by commits.

### DIFF
--- a/shared/models.go
+++ b/shared/models.go
@@ -140,29 +140,51 @@ func (r TestRun) Channel() string {
 }
 
 // PendingTestRunStage represents the stage of a test run in its life cycle.
-type PendingTestRunStage string
+type PendingTestRunStage int
 
-// PendingTestRunLifeCycle is an ordered array of all valid stages in a life cycle.
-var PendingTestRunLifeCycle = [...]PendingTestRunStage{
-	"GITHUB_QUEUED",
-	"GITHUB_IN_PROGRESS",
-	"CI_RUNNING",
-	"CI_FINISHED",
-	"GITHUB_SUCCESS",
-	"GITHUB_FAILURE",
-	"WPTFYI_RECEIVED",
-	"WPTFYI_PROCESSING",
-	"VALID",
-	"INVALID",
+// Constant enums for PendingTestRunStage
+const (
+	StageGitHubQueued     PendingTestRunStage = 100
+	StageGitHubInProgress PendingTestRunStage = 200
+	StageCIRunning        PendingTestRunStage = 300
+	StageCIFinished       PendingTestRunStage = 400
+	StageGitHubSuccess    PendingTestRunStage = 500
+	StageGitHubFailure    PendingTestRunStage = 550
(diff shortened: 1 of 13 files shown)


webapp/templates/admin_upload.html
```diff
@@ -42,6 +42,7 @@
 <div class="column">
   <h2>File payload</h2>
   <form method="POST" action="/api/results/upload" enctype="multipart/form-data">
+    {{if .CallbackURL}}<input style="display:none;" name="callback_url" value="{{.CallbackURL}}">{{end}}
```
lukebjerring: Why not `type="hidden"` ?
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden

shared/models.go
```diff
@@ -139,25 +139,56 @@ func (r TestRun) Channel() string {
...
-const TestRunStatusCreated = TestRunStatus(0)
-
-// TestRunStatusRunning represents a PendingTestRun that has been announced as in-flight.
-const TestRunStatusRunning = TestRunStatus(1)
-
-// TestRunStatusProcessing represents a PendingTestRun that has completed the run,
-// and the results are being processed.
-const TestRunStatusProcessing = TestRunStatus(2)
+// PendingTestRunStage represents the stage of a test run in its life cycle.
+type PendingTestRunStage string
```
lukebjerring: IMO, we should have this as `type PendingTestRunStage int` with `String()` method (and a `PendingTestRunStageFromName(name)` helper). Less error prone when it's strongly typed consts in Golang.

Custom JSON marshalling can keep the human-readable JSON, and using 100, 200, ... values allows for inserting new values sequentially at a later date.

Hexcles: I converted this to int, and consequently we now have a series of constants; I'm a bit concerned about polluting the `shared` namespace. WDYT?

lukebjerring: I believe the style we use elsewhere is to prefix all the consts with a somewhat redundant "type", in this case, `STAGE_` or similar. This avoids a scattered pollution of the namespace, at the cost of slightly annoying verbosity.

Hexcles: golint does not like `STAGE_`, so `Stage*` I guess?

lukebjerring: Yes, sorry I always forget Golang is a snowflake that doesn't like all-caps.
https://github.com/web-platform-tests/wpt.fyi/blob/master/shared/statuses.go#L14 is the `TestStatus*` example I'm referring to.

PR #1424, opened 2019-08-19
Include master label for anomalies insight

lukebjerring:
Avoids showing empty PR results some of the time.


lukebjerring:
TBR'ing this trivial change (verified it myself in the staging env).

Issue #1375, opened 2019-06-20
Virtual navigation is causing searches to have + instead of spaces


lukebjerring:
Steps to repro:
- Load wpt.fyi
- Write a query with 2 atoms, separated by a space, e.g. `chrome:fail safari:pass`
- Submit said query
- Copy the URL in the address bar, paste it another tab
- Results fail to load.
- Search bar now contains a +

The URL copied from the LINK button in the info bar seems to work fine.

Hexcles:
Playing with the previous versions in prod, I can confirm that the regression happened in this push: https://github.com/web-platform-tests/wpt.fyi/issues/1342 (i.e. bbe393a...5c27b47), so most likely another breakage caused by `app-route`.

PR #1426, opened 2019-08-19
Fix bad decoding of space -> + -> %2B

lukebjerring:
## Description
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1375
Additionally fixes a bug with the search field not updating when navigating back/forward with the browser.

Our use of URLSearchParams causes spaces to become '+' (unnecessarily), which are further encoded by the `app-location` Polymer component. On the way back out, we don't undo the problem.

The fix in this PR ties into the hack for param cleanup and undo the + conversion, resulting in an unambiguous %20 encoding (which decodes back to a space correctly).

## Review Information
- Visit the deployed environment, run a search containing a space. (e.g. `firefox:pass chrome:fail`)
- Change the query (e.g. `firefox:pass chrome:fail safari:pass`)
- Hit the back button
- See previous query in search box, spaces as were (no unexpected +).
- Reload the current URL
- See query as-was (no unexpected +).

### DIFF
--- a/webapp/components/results-navigation.js
+++ b/webapp/components/results-navigation.js
@@ -72,7 +72,9 @@ const QueryBuilder = (superClass, opts_queryParamsComputer) => class extends sup
     const afterQ = url.search
       .replace(/=true/g, '')
       .replace(/:00.000Z/g, '')
-      .split('?',);
+      // Work around bug where space => + => %2B is not decoded correctly.
+      .replace(/\+/g, '%20')
+      .split('?');
     return afterQ.length && afterQ[1];
   }
 
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -233,6 +233,7 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
     if (this.activeView) {
       this.activeView.query = query;
     }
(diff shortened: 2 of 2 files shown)


Hexcles:
> The fix in this PR ties into the hack for param cleanup and undo the + conversion, resulting in an unambiguous %20 encoding (which decodes back to a space correctly).

So is this fix a "hack", too? I still don't quite understand the root cause of this issue. The symptom looks like a sign that we are doing double encoding somewhere?

lukebjerring:
The root cause is that we double-plumb the query params. Unfortunately there's no way to leverage the app-location without it encoding what we give it. Either we manually split and parse query params, or refactor our code to inherit the app-location and handle it's behaviour manually.

Both are non-trivial refractors, and will have their own bugs. Note that iron-location uses the same "hack" / workaround for a similar bug.

Hexcles:
I guess I'm most puzzled by why only spaces are problematic, while the other special chars are fine.

lukebjerring:
https://url.spec.whatwg.org/#urlencoded-serializing

0x20 is just _special_

Hexcles:
Ah I see! It's the difference in encoding spaces between the full percent encoding and the URL form encoding.

PR #1431, opened 2019-08-21
Fix treatment of slashes in testname query params

### LINKED ISSUE #1430: Tests with slashes in query string not shown correctly
https://github.com/web-platform-tests/wpt/pull/18414 introduced a test which uses variants like this:
```html
<meta name="variant" content="?q=cts:buffers/create_mapped:">
<meta name="variant" content="?q=cts:buffers/map:">
<meta name="variant" content="?q=cts:buffers/map_detach:">
<meta name="variant" content="?q=cts:buffers/map_oom:">
<meta name="variant" content="?q=cts:canvas/context_creation:">
<meta name="variant" content="?q=cts:command_buffer/basic:">
<meta name="variant" content="?q=cts:command_buffer/compute/basic:">
<meta name="variant" content="?q=cts:command_buffer/copies:">
<meta name="variant" content="?q=cts:command_buffer/render/basic:">
<meta name="variant" content="?q=cts:command_buffer/render/rendering:">
<meta name="variant" content="?q=cts:examples:">
<meta name="variant" content="?q=cts:fences:">
```

The results of these tests should be here:
https://wpt.fyi/results/webgpu?run_id=306280007&run_id=292860002&run_id=289070005&run_id=294680004

But:
image

Because of the slash in the query string, some of these are treated as directories, but drilling down into the directories doesn't work.

Note that this is new, when I wrote this no tests like this existed:
https://github.com/foolip/ad-hoc-wpt-results-analysis/blob/4c5210e1ab96e92fdd99841112bab3dc4e43763a/git-write.js#L111-L112

The problem might be just in the UI, but possibly also in results processing and in the search cache.

@lukebjerring @Hexcles

lukebjerring:
## Description
Fixes #1430 

We were naively splitting on slashes, and in the case that a slash appears (unencoded) in a URI's query params (for tests using `variant` metadata), we see the UI behave badly.

## Review Information
- Visit the auto-deployed environment
- Navigate to the webgpu test folder
- You should see some valid test names. cts.html has these variants:
  - ?q=cts:buffers/create_mapped:
  - ?q=cts:buffers/map:
  - ?q=cts:buffers/map_detach:
  - ?q=cts:buffers/map_oom:
  - ?q=cts:canvas/context_creation:
  - ?q=cts:command_buffer/basic:
  - ?q=cts:command_buffer/compute/basic:
  - ?q=cts:command_buffer/copies:
  - ?q=cts:command_buffer/render/basic:
  - ?q=cts:command_buffer/render/rendering:
  - ?q=cts:examples:
  - ?q=cts:fences:

### DIFF
--- a/webapp/components/path.js
+++ b/webapp/components/path.js
@@ -43,9 +43,14 @@ const PathInfo = (superClass) => class extends superClass {
   encodeTestPath(path) {
     path = path || '/';
     console.assert(path.startsWith('/'));
-    let parts = path.split('/').slice(1);
-    parts.push(encodeURIComponent(parts.pop()));
-    return '/' + parts.join('/');
+    const pathAndQuery = path.split('?');
+    let parts = pathAndQuery[0].split('/');
+    let lastPart = parts.pop();
+    if (pathAndQuery.length > 1) {
+      lastPart += '?' + pathAndQuery[1];
+    }
+    parts.push(encodeURIComponent(lastPart));
+    return parts.join('/');
   }
 
   computeTestScheme(path) {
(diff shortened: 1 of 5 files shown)


Hexcles:
According to the [spec](https://url.spec.whatwg.org/#concept-urlencoded-byte-serializer), slashes should be percent encoded in the query string, so it looks like the root issue is in wptrunner, which fails to encode test names properly?

lukebjerring:
Yes, but unless we're going to retroactively fix all the reports with invalid content, this workaround will still be the easier course of action.

webapp/components/path.js
```diff
@@ -57,20 +62,40 @@ const PathInfo = (superClass) => class extends superClass {
...
-      && path && path.split('/').filter(p => p).length > 0;
+    if (!path || this.computePathIsATestFile(path)) {
+      return false;
+    }
+    return new URL(path, window.location).pathname.split('/').filter(p => p).length > 0;
   }
 
   computePathIsATestFile(path) {
-    return /(\.(html|htm|py|svg|xhtml|xht|xml)(\?.*)?$)/.test(path);
+    return /(\.(html|htm|py|svg|xhtml|xht|xml)(\?.*)?$)/.test(new URL(path, window.location).pathname);
```
Hexcles: It took me a while (after looking up MDN) to understand what this does -- getting rid of the query string in `path` IIUC. This might warrant a comment.

PR #1432, opened 2019-08-21
Add endpoint for fetching pending test runs

lukebjerring:
Adds api endpoint for `/api/status` to fetch all pending runs.

### Review
See https://api-pending-runs-dot-wptdashboard-staging.appspot.com/api/status

### DIFF
--- a/api/pending_test_runs.go
+++ b/api/pending_test_runs.go
@@ -1,0 +1,33 @@
+// Copyright 2019 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// apiPendingTestRunsHandler is responsible for emitting JSON for
+// all the pending test runs.
+func apiPendingTestRunsHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	store := shared.NewAppEngineDatastore(ctx, true)
+
+	q := store.NewQuery("PendingTestRun")
+	var runs []shared.PendingTestRun
+	if _, err := store.GetAll(q, &runs); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	testRunsBytes, err := json.Marshal(runs)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	w.Write(testRunsBytes)
+}
--- a/api/pending_test_runs.go
+++ b/api/pending_test_runs.go
@@ -15,7 +15,7 @@ import (
 // all the pending test runs.
 func apiPendingTestRunsHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := shared.NewAppEngineContext(r)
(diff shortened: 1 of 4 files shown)


Hexcles:
As per offline discussions: the API for fetching pending runs will be `/api/status/pending` instead of `/api/runs` to distinguish `PendingTestRun` from `TestRun`.

api/pending_test_runs.go
```diff
@@ -0,0 +1,33 @@
...
+	"net/http"
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// apiPendingTestRunsHandler is responsible for emitting JSON for
+// all the pending test runs.
+func apiPendingTestRunsHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	store := shared.NewAppEngineDatastore(ctx, true)
```
Hexcles:
```suggestion
	store := shared.NewAppEngineDatastore(ctx, false)
```
We probably don't want to cache this because pending runs are mutable (and will change frequently in their lifecycle).

api/pending_test_runs.go
```diff
@@ -0,0 +1,33 @@
...
+
+	q := store.NewQuery("PendingTestRun")
+	var runs []shared.PendingTestRun
+	if _, err := store.GetAll(q, &runs); err != nil {
```
Hexcles: We need to hold onto the keys so that we can backfill the ID field of the pending runs (and this should be tested).

api/pending_test_runs.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// apiPendingTestRunsHandler is responsible for emitting JSON for
+// all the pending test runs.
+func apiPendingTestRunsHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	store := shared.NewAppEngineDatastore(ctx, true)
+
+	q := store.NewQuery("PendingTestRun")
```
Hexcles:
```suggestion
	q := store.NewQuery("PendingTestRun").Order("-Updated")
```

api/pending_test_runs.go
lukebjerring: I think the cache only affects `TestRun` entities in any case, but sure.

api/pending_test_runs.go
lukebjerring: Done in https://github.com/web-platform-tests/wpt.fyi/pull/1432/commits/99c9c375cf8fca11595e683264e81a9b2d809ab1

api/pending_test_runs.go
```diff
@@ -0,0 +1,37 @@
...
+	ctx := shared.NewAppEngineContext(r)
+	store := shared.NewAppEngineDatastore(ctx, false)
+
+	q := store.NewQuery("PendingTestRun").Order("-Updated")
+	var runs []shared.PendingTestRun
+	if keys, err := store.GetAll(q, &runs); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	} else {
+		for i, key := range keys {
```
Hexcles: Sigh, golint doesn't like it:
/home/user/go/src/github.com/web-platform-tests/wpt.fyi/api/pending_test_runs.go:25:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)

PR #1435, opened 2019-08-22
Bug 1574668 - use the Taskcluster client and taskcluster-lib-urls

djmitche:
## Description
As part of [bug 1574668](https://bugzilla.mozilla.org/show_bug.cgi?id=1574668) we will need the ability to easily switch from one TC deployment to another.  

## Review Information
I modified the existing tests to suit, but nothing tests `getTaskGroupInfo`.  In fact, in its existing incarnation it was incorrect and would under-report tasks in large task groups since it did not use continuation tokens.  I think "large" means over 1000 tasks, but depending on the details of the Azure tables backend, could be smaller than that.

## Changes
Deployments are identified by TASKCLUSTER_ROOT_URL.  So this PR modifies the TC webhook listener to use the TC client and taskcluster-lib-urls to query task groups and generate URLs, based on this env var.

## Requirements
Note that with this change, the API server must be run with `TASKCLUSTER_ROOT_URL` set to an appropriate value (`https://taskcluster.net` for now)

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -8,14 +8,16 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/http"
+	"os"
 	"regexp"
 	"strings"
 	"sync"
 
 	mapset "github.com/deckarep/golang-set"
 	"github.com/google/go-github/github"
+	tcurls "github.com/taskcluster/taskcluster-lib-urls"
+	"github.com/taskcluster/taskcluster/clients/client-go/tcqueue"
 	uc "github.com/web-platform-tests/wpt.fyi/api/receiver/client"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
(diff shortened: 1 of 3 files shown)


api/taskcluster/webhook_test.go
```diff
@@ -90,19 +91,19 @@ func TestIsOnMaster(t *testing.T) {
 
 func TestExtractTaskGroupID(t *testing.T) {
 	t.Run("Status", func(t *testing.T) {
-		group, task := extractTaskGroupID("https://tools.taskcluster.net/task-group-inspector/#/Y4rnZeqDRXGiRNiqxT5Qeg")
+		group, task := extractTaskGroupID("https://tc.example.com/task-group-inspector/#/Y4rnZeqDRXGiRNiqxT5Qeg")
```
Hexcles: Where's `tc.example.com` from? Does `tcurls.API` default to this URL when the root URL is empty?

api/taskcluster/webhook.go
```diff
@@ -167,18 +169,19 @@ func (b branchInfos) GetNames() []string {
 func processTaskclusterBuild(aeAPI shared.AppEngineAPI, taskGroupID, taskID string, sha string, labels ...string) (bool, error) {
 	ctx := aeAPI.Context()
 	log := shared.GetLogger(ctx)
+	rootURL := os.Getenv("TASKCLUSTER_ROOT_URL")
```
Hexcles: I'm debating whether we should handle the lack of this env var gracefully, e.g. default to `taskcluster.net`?

api/taskcluster/webhook_test.go
djmitche: In this case, it's just test input to the function, and a way of taking out the soon-to-be-defunct hostname.  Elsewhere in the tests, I use it as a testing rootURL.

api/taskcluster/webhook.go
djmitche: In general, we've tried to avoid defaulting, as a way of finding places where the value is hard-coded.  Otherwise, we'll all have a bad day when `taskcluster.net` stops resolving and 100 things where we *thought* we were passing a value explicitly break because they were actually using the default.

djmitche:
@Hexcles `app.yaml` updated.

djmitche:
@Hexcles is there another step to land this?

Hexcles:
@djmitche sorry just back from vacation. The PR failed the unit tests. The error messages are obscure; let me investigate a bit. (For a few tests, potentially we need to initialize the env var in tests as well, but it's not immediately clear to me why the other tests fail.)

Hexcles:
Right, so the failures are caused by a deprecation in GAE SDK #1447 . Using an older version of the SDK (the one pinned inside the Docker image used by GitHub Actions), the PR passes all tests, so I'm force-merging it.

Thanks again, @djmitche ; and sorry for the delay!

PR #1437, opened 2019-08-22
[lifecycle] Send lifecycle updates from processor

Hexcles:
This completes the phase 1 of test run lifecycle: the results processor will now send both `IN_PROGRESS` (when it starts processing a run) and `INVALID` (when an exception is raised) updates to the API.


api/receiver/routes.go
```diff
@@ -15,4 +15,7 @@ func RegisterRoutes() {
 	// PRIVATE API endpoint for creating a test run in Datastore.
 	// This API is authenticated. Only this AppEngine project has the credential.
 	shared.AddRoute("/api/results/create", "api-results-create", apiResultsCreateHandler)
+
+	// PRIVATE API endpoitn for updating the status of a pending test run
```
lukebjerring:
```suggestion
	// PRIVATE API endpoint for updating the status of a pending test run
```

api/receiver/handlers.go
```diff
@@ -33,3 +35,34 @@ func apiResultsCreateHandler(w http.ResponseWriter, r *http.Request) {
 	s := checks.NewAPI(ctx)
 	HandleResultsCreate(a, s, w, r)
 }
+
+func apiPendingTestRunUpdateHandler(w http.ResponseWriter, r *http.Request) {
+	if r.Method != "POST" {
```
lukebjerring: Not clear cut here, but this feels more like a `PATCH`, since the body provided from the Python code below only has id/status fields (and doesn't modify the other fields?)

results-processor/processor.py
```diff
@@ -242,6 +243,26 @@ def create_run(self, run_id, labels, uploader, callback_url=None):
...
 
+    def update_status(self, run_id, stage, error=None, callback_url=None):
+        assert stage, "stage cannot be empty"
+        if int(run_id) == 0:
+            _log.error('Cannot update run status: missing run_id')
+            return
+        if callback_url is None:
+            callback_url = config.project_baseurl()
+        parsed_url = urlparse(callback_url)
+        api = '%s://%s/api/status/update' % (parsed_url.scheme,
```
lukebjerring: Normally REST architecture here for and update would be a PUT to `/api/status/{id}`. See `PATCH` comment above.

Hexcles:
@lukebjerring switched to PATCH `/api/status/{id}`. PTAL.

Issue #1439, opened 2019-08-23
Azure Pipelines run manually triggered from master not labeled with master


foolip:
I triggered https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=28411 manually to retry the failed STP run in https://dev.azure.com/web-platform-tests/wpt/_build/results?buildId=28387.

It succeeded and the run is https://wpt.fyi/results/?run_id=300580012

However, the run didn't get the master label applied, so it didn't affect the front page of wpt.fyi. By now there's a new aligned run, but it's still a bug :)

I'll manually add the label in datastore for wpt.fyi, but won't bother with staging.wpt.fyi.

PR #1440, opened 2019-08-23
Fix the `IsMasterBranch()` helper for Azure Pipelines

foolip:
Right fix concluded from the logs, which said:
```
Source branch: refs/heads/master
Trigger PR branch:
Fetching https://dev.azure.com/web-platform-tests/wpt/_apis/build/builds/28411/artifacts
```

Not tested because it would just check that the string being checked
is correct, it wouldn't catch if there's a typo, and wouldn't prevent
a regression if the Azure Pipelines API changes.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1439.

### DIFF
--- a/api/azure/api.go
+++ b/api/azure/api.go
@@ -58,7 +58,7 @@ type BuildTriggerInfo struct {
 
 // IsMasterBranch returns whether the source branch for the build is the master branch.
 func (a *Build) IsMasterBranch() bool {
-	return a != nil && a.TriggerInfo.SourceBranch == "master"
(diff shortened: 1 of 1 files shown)


foolip:
Travis failed due to:
> ERROR: (gcloud.app.deploy) argument --version/-v: Bad value [azure-IsMasterBranch]: May only contain lowercase letters, digits, and hyphens. Must begin and end with a letter or digit. Must not exceed 63 characters.

Seems like I'm very good at violating the naming rules, this time I purposefully avoided a slash because I'd seen this, but still no good :)

PR #1453, opened 2019-08-29
Post a GitHub deployment instead

lukebjerring:
## Description
Tweaks the behaviour of the deployment messages to use the GitHub deployments API.

### DIFF
--- a/util/deploy-comment.sh
+++ b/util/deploy-comment.sh
@@ -33,15 +33,38 @@ POST_URL="https://api.github.com/repos/${TRAVIS_REPO_SLUG}/deployments"
 debug "${POST_URL}"
 POST_BODY="{
                 \"ref\": \"${TRAVIS_BRANCH}\",
-                \"task\": \"deploy:${APP_PATH}\",
+                \"task\": \"deploy\",
                 \"auto_merge\": false,
-                \"environment\": \"staging\",
+                \"environment\": \"${APP_PATH}\",
                 \"transient_environment\": true
             }"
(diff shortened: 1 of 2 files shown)


lukebjerring:
@foolip you can see that it "works locally" from the eventual output here https://github.com/web-platform-tests/wpt.fyi/pull/1442

util/deploy-comment.sh
```diff
@@ -20,35 +20,51 @@ if [[ -z "${TRAVIS_REPO_SLUG}" ]];
...
-          "https://api.github.com/repos/${TRAVIS_REPO_SLUG}/issues/${TRAVIS_PULL_REQUEST}/comments"
-else
-    info "Found existing comment mentioning link:\n${STAGING_URL}"
+    debug "Created deployment ${DEPLOYMENT_ID}"
 fi
+
+debug "Setting status to deployed"
+POST_BODY="{
+                \"state\": \"success\",
+                \"target_url\": \"${STAGING_URL}\",
```
foolip: This should be `environment_url` per https://developer.github.com/v3/repos/deployments/#update-a-deployment. `target_url` was renamed `log_url` it seems.

I doubt it's useful, but a log URL would be the running Travis job.

foolip:
Eventually we might want to include deployments cleanup as follow up to this, telling GitHub that a deployment is gone.

util/deploy-comment.sh
lukebjerring:
```suggestion
                \"environment_url\": \"${STAGING_URL}\",
```

PR #1454, opened 2019-08-30
Fix the environment name for GitHub deployments API calls

lukebjerring:
## Description
Fix a couple of problems with #1453 which weren't caught until it hit master.

- `$APP_PATH` wasn't forwarded from the parent script; make it an `$ENVIRONMENT` param on the comment script.
- Fail out when the create POST fails
- Clear out the `required_contexts` so that running from travis isn't blocked by travis being unfinished.


lukebjerring:
Looks like that does the trick!
https://github.com/web-platform-tests/wpt.fyi/pull/1442#event-2597413994

PR #1463, opened 2019-09-04
Add structured search atom for filtering by results that differ

### LINKED ISSUE #1460: Missing ability to filter by *any* difference to spot changes in behavior
In https://github.com/web-platform-tests/wpt/pull/11024#issuecomment-527248462 I was looking at this view:
https://wpt.fyi/results/?diff&filter=ADC&run_id=289790006&run_id=285580005

The two runs differ only by a change to testharness.js, and I'd like to see *any* difference. I suspect that `status:unknown` shows the interesting differences, but if anything else has changed without changing the overall pass rates, this wouldn't be visible.

lukebjerring:
Fixes #1460 

## Description
Adds handling of `is:different` atom, which will return true for rows where the size/cardinality of the set of result statuses is not 1.

## Review Information
- Search for is:different
- See only rows that have differing values.
- Try queries with 2, 3, and 4 runs, ensure behaves as expected.

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -55,7 +55,7 @@ Filters the results to values which possess/exhibit a given quality.
 
 #### `is:different`
 
-Filters to rows where more there is more than one resulting status for a test
+Filters to rows where there is more than one resulting status for a test
 across the runs.
 
 ### And-conjuction
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -45,6 +45,19 @@ or, negation,
 
 Where `[product]` is a product specification (e.g. `safari`, `chrome-69`).
 
+### Meta qualities
+
+> BETA: This feature is under development and may change without warning.
+
+Filters the results to values which possess/exhibit a given quality.
+
+    is:[quality]
+
+#### `is:different`
+
+Filters to rows where more there is more than one resulting status for a test
+across the runs.
+
 ### And-conjuction
 
     [query1] and [query2] [and ...]
@@ -173,3 +186,14 @@ Search untriaged issues -
 Search triaged issues -
 
     chrome:pass and link:bugs.chromium.org
+
+#### is
+
+`is` query atoms perform a search for tests that possess some meta quality.
+
+    {"is": "different"}
+
+
+
+See [#meta-qualities](Meta qualities) above for more information on other
+meta qualities than `"different"`.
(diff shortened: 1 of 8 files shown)


foolip:
Staging deployment failed, so I can't test this live.

api/query/cache/index/filter.go
```diff
@@ -189,6 +196,15 @@ func (l Link) Filter(t TestID) bool {
 	return false
 }
 
+// Filter interprets a MetadataQuality as a filter function over TestIDs.
+func (q MetadataQuality) Filter(t TestID) bool {
+	set := mapset.NewSet()
+	for _, result := range q.runResults {
+		set.Add(result.GetResult(t))
+	}
+	return set.Cardinality() != 1
```
foolip: Can this code path ever be reached with no runs, so that it matters what happens when`set.Cardinality()` is zero? Perhaps assert something if zero is impossible?

api/query/concrete_query.go
```diff
@@ -117,6 +117,15 @@ func (Link) Size() int { return 1 }
 // Size of Count is the sum of the sizes of its constituent ConcretQuery instances.
 func (c Count) Size() int { return size(c.Args) }
 
+// Size of Is depends on the quality.
```
foolip: What are these `Size()` methods for, what happens if they return the wrong value?

api/query/concrete_query.go
```diff
@@ -117,6 +117,15 @@ func (Link) Size() int { return 1 }
...
+func (q MetadataQuality) Size() int {
+	switch q {
```
foolip: `return 1` with a comment and/or assert would be better IMHO than scaffolding for extensions that may never come.

api/query/README.md
```diff
@@ -173,3 +186,14 @@ Search untriaged issues -
...
+
+#### is
+
+`is` query atoms perform a search for tests that possess some meta quality.
+
+    {"is": "different"}
+
+
+
+See [#meta-qualities](Meta qualities) above for more information on other
```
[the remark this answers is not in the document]
lukebjerring: Yep :)

api/query/cache/index/filter.go
lukebjerring:
```suggestion
	return set.Cardinality() > 1
```

api/query/concrete_query.go
lukebjerring: The "size" of the query being performed; not sure exactly when it's used. @mdittmer could answer that.

lukebjerring:
> Staging deployment failed, so I can't test this live.

Are you expecting to see the comment from wpt-pr-bot? We've migrated to GitHub deployments already :)

image

foolip:
Yes, I was expected the comments still, fail :)

PR #1466, opened 2019-09-05
Fix the processor status route

Hexcles:
Change the route from /processor to /status.

## Review Information

The "Processor" tab should now work.


Hexcles:
Travis keeps timing out. I'm force-merging this.

PR #1471, opened 2019-09-06
Add all/none root queries

### LINKED ISSUE #1467: More aliases for the `count` atom
https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/README.md#count is sweet and has aliases `one`, `two` and `three` already.

I'd like to suggest `none` and `all` queries as well. Examples I'd like to search for:
- `none(status:missing)` to exclude inconsistent subtest naming from other queries
- `all(status:error)` to find consistent harness errors

I guess `some` is not needed, because it's already implied.

lukebjerring:
Fixes #1467 
 
## Description
Adds root query options for `all` and `none` (instead of the default, `exists`).

Strangely, `count` and `seq` get wrapped in `exists`. I started to tweak that behaviour, but decided against busying up the PR. Will clean that up another time.

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -115,6 +115,49 @@ func (e AbstractExists) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	}
 }
 
+// AbstractAll represents an array of abstract queries, each of which must be
+// satifisfied by all runs. It represents the root of a structured query.
+type AbstractAll struct {
+	Args []AbstractQuery
+}
+
+// BindToRuns binds each abstract query to an and-combo of that query against
+// each specific/individual run.
+func (e AbstractAll) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
+	queries := make([]ConcreteQuery, len(e.Args))
+	for i, arg := range e.Args {
+		var query ConcreteQuery
+		byRun := make([]ConcreteQuery, 0, len(runs))
+		for _, run := range runs {
+			bound := arg.BindToRuns(run)
+			if _, ok := bound.(True); !ok { // And with True is pointless.
+				byRun = append(byRun, bound)
+			}
+		}
+		query = And{Args: byRun}
+		queries[i] = query
+	}
+	// And the overall node is true if all its exists queries are true.
+	return And{
+		Args: queries,
+	}
+}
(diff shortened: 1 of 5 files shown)


api/query/atoms.go
```diff
@@ -704,70 +799,110 @@ func MetadataQualityFromString(quality string) (MetadataQuality, error) {
...
+	}
+	{
+		var e AbstractExists
+		err := json.Unmarshal(b, &e)
+		if err == nil {
+			return e, nil
+		}
+	}
+	{
+		var a AbstractAll
```
lukebjerring: All / None re-used a/n, and the variables persisting makes for copy-pasta bugs, so I've scoped all the attempts. This (AbstractAll) and AbstractNone are added.

webapp/components/test/test-search.html
```diff
@@ -230,6 +230,26 @@
       });
     });
 
+    test('all', () => {
+      assertQueryParse('all(status:!PASS status:!OK)', {
```
foolip: Good example, I'd like to search for this :)

api/query/atoms.go
```diff
@@ -115,6 +115,49 @@ func (e AbstractExists) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
...
+// each specific/individual run.
+func (e AbstractAll) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
+	queries := make([]ConcreteQuery, len(e.Args))
+	for i, arg := range e.Args {
+		var query ConcreteQuery
+		byRun := make([]ConcreteQuery, 0, len(runs))
+		for _, run := range runs {
+			bound := arg.BindToRuns(run)
+			if _, ok := bound.(True); !ok { // And with True is pointless.
+				byRun = append(byRun, bound)
```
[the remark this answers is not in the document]
lukebjerring: There is a simplification behaviour when binding to runs (e.g. [an empty OR becomes false](https://github.com/web-platform-tests/wpt.fyi/blob/af72dd66ac74591b236170ed36cb497c21fe461f/api/query/atoms.go#L226)) which this accounts for.

lukebjerring:
Force-merging to ignore Travis timeouts.

PR #1473, opened 2019-09-09
Revert "Set up CI with Azure Pipelines"

foolip:
This hasn't been used, and by now GitHub Actions is probably a better option for cron jobs.

Reverts web-platform-tests/wpt.fyi#1117


foolip:
@Hexcles you've mentioned you're not making friends with Travis. If we were to migrate everything to another CI system, would you pick Azure Pipelines or GitHub Actions?

foolip:
That'd be a fine choice I think, the main benefit of Azure Pipeline is ability to run a custom agent pool, which you won't need for wpt.fyi.

foolip:
Travis is failing, but I doubt it's because of this change, so merging :)

Issue #1465, opened 2019-09-05
Invalid search query syntax shows "Failed to fetch test runs"


foolip:
Go to the front page of wpt.fyi and search for `is:incredible`. That shows "Failed to fetch test runs", which isn't very helpful:
![image](https://user-images.githubusercontent.com/498917/64320145-2e5b9300-cfbe-11e9-8fdd-30a13fe39772.png)

Saying that the query isn't valid would be better.

PR #1475, opened 2019-09-10
Add an API and UI for showing invalid runs

Hexcles:
API /api/status/invalid returns a list of invalid runs.

A tab bar is added to the /status interface to switch between pending
and invalid runs.

## Review Information

Go to the "processor" page (`/status`) of the staging deployment.

### DIFF
--- a/api/pending_test_runs_medium_test.go
+++ b/api/pending_test_runs_medium_test.go
@@ -73,4 +73,17 @@ func TestAPIPendingTestHandler(t *testing.T) {
 		assert.Len(t, results, 1)
 		assert.Equal(t, results[0].ID, running.ID)
 	})
+
+	t.Run("/api/status/invalid", func(t *testing.T) {
+		r, _ = i.NewRequest("GET", "/api/status/invalid", nil)
+		r = mux.SetURLVars(r, map[string]string{"filter": "invalid"})
+		resp := httptest.NewRecorder()
+		apiPendingTestRunsHandler(resp, r)
+		body, _ := ioutil.ReadAll(resp.Result().Body)
+		assert.Equal(t, http.StatusOK, resp.Code, string(body))
+		var results []shared.PendingTestRun
+		json.Unmarshal(body, &results)
+		assert.Len(t, results, 1)
+		assert.Equal(t, results[0].ID, invalid.ID)
+	})
 }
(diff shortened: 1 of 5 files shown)


webapp/components/wpt-processor.js
```diff
@@ -58,6 +73,10 @@ class WPTProcessor extends LoadingState(PolymerElement) {
       </table>
     </template>
 
+    <template is="dom-if" if="[[resultsLoadFailed]]">
+      <div>No runs found.</div>
```
foolip: Is this really true if loading *failed*? A bit like https://github.com/web-platform-tests/wpt.fyi/issues/1465.

api/pending_test_runs_medium_test.go
```diff
@@ -73,4 +73,17 @@ func TestAPIPendingTestHandler(t *testing.T) {
 		assert.Len(t, results, 1)
 		assert.Equal(t, results[0].ID, running.ID)
 	})
+
+	t.Run("/api/status/invalid", func(t *testing.T) {
+		r, _ = i.NewRequest("GET", "/api/status/invalid", nil)
+		r = mux.SetURLVars(r, map[string]string{"filter": "invalid"})
```
foolip: What does this bit do? If it adds `?filter=invalid` the result will be `/api/status/invalid?filter=invalid`, right? Or is this mocking the behavior of the URL routing code by extracting the third component as filter?

webapp/components/wpt-processor.js
```diff
@@ -74,26 +93,35 @@ class WPTProcessor extends LoadingState(PolymerElement) {
...
+        });
   }
 
-  async loadPendingRuns() {
-    const r = await fetch('/api/status/pending');
-    this.testRuns = r.ok && await r.json();
+  async loadPendingRuns(path) {
+    this.resultsLoadFailed = false;
+    const r = await fetch(path);
+    if (!r.ok || r.status !== 200) {
```
foolip: https://fetch.spec.whatwg.org/#dom-response-ok checks for status in the range 200 to 299, why do we demand that the status is 200? Also, don't we have a wrapper for this very common pattern of fetch-and-check-status?

Hexcles: Changed to checking `r.ok` only. And a quick search didn't find any wrapper. Perhaps we can refactor this into `loading-state.js`. Another day.

api/pending_test_runs_medium_test.go
Hexcles: This is because of the way we invoke the test, which bypasses the mux middleware so we'd have to set the vars ourselves as if they were populated by the router. Next time if I need to do this again I'll make a wrapper for this.

webapp/components/wpt-processor.js
Hexcles: Good point. This also isn't showing up when an empty response is successfully returned. I fixed this by splitting this into two separate error prompts.

PR #1483, opened 2019-09-13
Set up GitHub Actions to run some CI tasks

Hexcles:
Currently, running browsers is not supported because of permission
issues:

* Chrome does not like to run as root; --no-sandbox is needed.
* Firefox can run as root but requires HOME to be owned by root, too,
  which is not the case on GitHub Actions (HOME is set to /github/home,
  which is owned by a non-priviledged user).

Lint, Python, Go tests are working as expected.

Deployment should also work, but will be added later.

Travis and local development setup are not changed at the moment.


Hexcles:
Example: https://github.com/web-platform-tests/wpt.fyi/pull/1483/checks?check_run_id=222531739

Hexcles:
I plan to eventually use the same Docker image for local development as well.

foolip:
To solve the user/root problem, can't use create a user and change to it, like this:?
https://github.com/web-platform-tests/wpt/blob/bf42dca1ce568ce559d5a4cad507239035b91dcb/tools/docker/Dockerfile#L62-L76

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
+name: Continuous Integration
```
foolip: This is going to be a bit wordy at the beginning of each check. I tried a few weeks ago to use an empty name to get rid of the leading "$name / " in check names but that didn't work, then. I'd just pick something short, but it's up to you.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+on: push
+jobs:
+  lint:
+    runs-on: ubuntu-latest
```
foolip: As long as you only run docker images I guess the risk is low, but if you want to do anything on the host OS directly, I'd suggest pinning it. I noticed when looking at Dependabot that there's a mode for GitHub Actions, so I expecting bumping pinned deps will soon be automated, if it isn't already working.

PR #1500, opened 2019-09-19
Clean up count/seq root query parsing

lukebjerring:
`count` and `seq` should be siblings of `exists`, `all`, and `none`. This cleans up the redundant nesting and flattens the root query out.


lukebjerring:
Not sure if we'd call it WAI, but currently implicitly having AND for `none(status:missing) is:different` produces

    "query":{"and":[{"none":[{"status":"UNKNOWN"}]},{"exists":[{"is":"different"}]}]}

while explicitly `and`-ing with `none(status:missing) and is:different` produces

    "query":{"and":[{"none":[{"status":"UNKNOWN"}]},{"exists":[{"pattern":"and"},{"is":"different"}]}]}

This is expected given the definition of AND/OR being nesting inside a root query, and can be worked around, but we should acknowledge that we're OK with it here where it's being defined.

foolip:
@lukebjerring given that @Hexcles do you want to wait for his return, or would my review by OK?

lukebjerring:
Happy to have you review!

webapp/components/test-search.js
```diff
@@ -174,30 +176,37 @@ const QUERY_SEMANTICS = QUERY_GRAMMAR.createSemantics().addOperation('eval', {
   _terminal: function() {
     return this.sourceString;
   },
+  Root: (r) => {
+    const ps = r.eval();
+    return ps.length === 0
+      ? emptyQuery
+      : ps.length === 1
+        ? ps[0]
+        : { and: ps };
```
foolip: Is this changing the default interpretation of `q1 q2`? Looking in https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/README.md I can't spot if that's equivalent to `q1 and q2` or `q1 or q2`. I could experiment to find out, but it'd be very good to document that.

webapp/components/test-search.js
```diff
@@ -215,10 +224,12 @@ const QUERY_SEMANTICS = QUERY_GRAMMAR.createSemantics().addOperation('eval', {
         return { lessThan: parseInt(n.eval()) + 1 };
       case "<":
         return { lessThan: n.eval() };
+      case ":":
+      case "=":
```
foolip: This is something new in the syntax, right? Can you document what `=` does?

webapp/components/test-search.js
```diff
@@ -56,26 +55,28 @@ const QUERY_GRAMMAR = ohm.grammar(`
...
-      | "count:" number           -- countEq
-      | "count=" number           -- countN
+      | "count:" number           -- countN
       | "three"                   -- count3
       | "two"                     -- count2
       | "one"                     -- count1
 
-    Exists = ListOf<RootExp, space*>
+    Exists
+      = "exists(" ListOf<Exp, space*> ")" -- explicit
```
foolip: Looks like this adds an explicit `exists` form, can you document it in https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/README.md?

foolip:
Starting from https://count-root-dot-wptdashboard-staging.appspot.com/results/?run_id=239460006&run_id=237480004 I think something is still wrong.

Just `is:different` matches 550 tests. `is:different and none(status:missing)` eventually fails with "Failed to fetch test runs." Reloading shows one result and "Failed to fetch test runs."

`is:different or none(status:missing)` is the clearest signal something is wrong. It matches only 34 tests, but just `none(status:missing)` matches >34k tests. So that's not what I expected from `or`...

lukebjerring:
As I said in https://github.com/web-platform-tests/wpt.fyi/pull/1500#issuecomment-534079709 the `and` is being interpreted as a pattern. We don't define explicit and/or until we're inside a query. I could change that, but wouldn't do it as part of this PR.

So, this adds an implicit and-conjunction for multiple root queries, but doesn't support combining them with or.

webapp/components/test-search.js
lukebjerring: `count=1(...)` is equivalent to `count:1(...)`. It seems inconsistent to support `count>1` but not `count=1`. This PR just simplifies where equality is consumed.

foolip:
What does "inside a query" mean, is that inside anything with parenthesis? https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/README.md#and-conjuction makes it look like `and` and `or` would work at the top level, and I thought I'd successfully used them, but apparently not?

webapp/components/test-search.js
lukebjerring: Only for the case that either q1 or q2 is a different combination (e.g. `all(...)`), which was previously unsupported. The default is still exists as a root.

lukebjerring:
Yes, inside a root (exists, all, none, count, seq) query. By default, a query that does not contain any root queries is wrapped in an exists implicitly, so it seems like conjunction is valid at the root, but it's not.

Might pay to tweak the docs to make it obvious, though I recommend we allow different conjunction between root queries in a future PR anyway.

foolip:
Yeah, having that just work would make a lot of sense. I'm not sure I even know how I'd use it inside another query to achieve what I was trying to do...

foolip:
At this point this PR is probably just missing some docs updates. I'm not at keyboard now either so let's see who wins the race to it :)

api/query/README.md
```diff
@@ -8,7 +8,86 @@ As outlined below, the `/api/search` endpoint takes a structured query object. w
...
+By default, a search query will be implicitly treated as an `exists` query (a disjunction
+across each of the runs separately). However, there are several other root query types that
+can be invoked by wrapping the query(s), including explicitly wrapping with `exists`.
+
+If multiple root queries are used, they are combined with `AND`, i.e. each query must
+separately be true. e.g.
+
+    count>1(status:!pass) none(status:missing)
+
+Requires that more than one non-pass result is present, and none of the results is missing.
```
foolip:
```suggestion
Requires that more than one non-pass result is present, and none of the results are missing.
```

api/query/README.md
```diff
@@ -8,7 +8,86 @@ As outlined below, the `/api/search` endpoint takes a structured query object. w
...
+
+> BETA: This feature is under development and may change without warning.
+
+    count:[number]([query1] [query2])
+
+Requires that the number of results matching the given query/queries is precisely
+the given count. For example, this search atom can be used to find cases where
+exactly one result is a failure:
+
+    count:1(status:fail)
```
foolip: Do you want to document `count=` here as well?

lukebjerring: No, I think it's fine; the colon version is documented for persons reading the docs, and the equals is supported in case people infer equality from the inequality options.

PR #1501, opened 2019-09-20
Explode folder contents in UI and collapse folders in autocomplete suggestions

lukebjerring:
## Description
Changes the displayed row to explode out its contents when there's only one row, and it's a directory.

I did this to cover the behaviour of submitting a search which has results under some unique folder, because it also covers the general case of clicking/navigating down into a state with the same context (a single directory).

Additionally, contributes each separate folder of a test path as possible autocomplete suggestions, so that we can leverage the jump-on-autocomplete behaviour to achieve a different solution of the same issue.

## Review Information
- Head to the GitHub deployed environment
- Search for `web-animations/animation-model`
- See three rows, each of which is an exploded path nested under web-animations/animation-model

### DIFF
--- a/api/query/autocomplete.go
+++ b/api/query/autocomplete.go
@@ -7,6 +7,7 @@ package query
 import (
 	"encoding/json"
 	"net/http"
+	"path"
 	"sort"
 	"strings"
 	"time"
@@ -118,6 +119,14 @@ func prepareAutocompleteResponse(limit int, filters *shared.QueryFilter, testRun
 	for _, smry := range summaries {
 		for file := range smry {
 			fileSet.Add(file)
+			// Add all the dirs as options too.
+			prefix := "/"
+			dir, _ := path.Split(file)
+			bits := strings.Split(dir, "/")
+			for _, part := range bits {
+				prefix := prefix + part + "/"
+				fileSet.Add(prefix)
+			}
 		}
 	}
 
--- a/api/query/autocomplete_medium_test.go
+++ b/api/query/autocomplete_medium_test.go
@@ -37,8 +37,8 @@ func TestAutocompleteHandler(t *testing.T) {
 		},
 	}
 	summaryBytes := [][]byte{
-		[]byte(`{"/a/b/c":[1,2],"/b/c":[9,9]}`),
-		[]byte(`{"/z/b/c":[0,8],"/x/y/z":[3,4],"/b/c":[5,9]}`),
+		[]byte(`{"/a/b/c.html":[1,2],"/b/c.html":[9,9]}`),
+		[]byte(`{"/z/b/c.html":[0,8],"/x/y/z.html":[3,4],"/b/c.html":[5,9]}`),
 	}
 
 	i, err := sharedtest.NewAEInstance(true)
@@ -84,7 +84,7 @@ func TestAutocompleteHandler(t *testing.T) {
 		"/api/autocomplete?run_ids=%s&q=%s&limit=%s",
 		url.QueryEscape(fmt.Sprintf("%d,%d", testRuns[0].ID, testRuns[1].ID)),
 		url.QueryEscape(q),
-		url.QueryEscape("2"))
+		url.QueryEscape("3"))
 	r, err := i.NewRequest("GET", url, nil)
 	assert.Nil(t, err)
 	ctx := shared.NewAppEngineContext(r)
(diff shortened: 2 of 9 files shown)


foolip:
This has some funny side effects:
https://explode-dot-wptdashboard-staging.appspot.com/results/?label=master&label=experimental&aligned&q=path%3A%2Ffullscreen%2F

It shows these 3 rows:
```
fullscreen/api/
idlharness.window.html
fullscreen/rendering/
```

But idlharness.window.html is also in fullscreen/.

I think what @stephenmcgruer wanted was to just end up directly in the directory he wrote, changing the URL to https://wpt.fyi/results/web-animations/animation-model.

I think we need to:
- make the autocomplete dropdown appear faster, or autocomplete *inline*
- make web-animations/animation-model trigger autocomplete for that directory

If one does a search which has results only in a single directory, that's trickier. Changing the path in the URL means that searching again will now be scoped to the directory, which is not what I'd want most of the time.

@stephenmcgruer can you shed some light on what behavior you'd expect?

lukebjerring:
So autocompleting directories is also a good behaviour, but I'd argue that this view explosion is a good change for other non-search cases too.

foolip:
Isn't it only when filtering results that it's possible to have a problem of the type @stephenmcgruer described? Otherwise you're just looking at the directory you're in, unfiltered.

Anyway, I agree something like this is useful when filtering, but what do you think about the fullscreen example I gave? Is that the behavior you intended?

stephenmcgruer:
> @stephenmcgruer can you shed some light on what behavior you'd expect?

What you wrote was accurate, I was looking to be sent directly to `results/web-animations/animation-model` when I entered a (valid) directory that had no globs or anything (if wpt.fyi supports those). 

That said, my expectation may not be the best way to implement it. https://explode-dot-wptdashboard-staging.appspot.com/results/?label=master&label=experimental&aligned&q=web-animations%2Fanimation-model%2F is usable though visually busy. It gives me the same overview as my original expected behavior and allows me to click into the three sub-directories. However the repetition of the prefix makes it visually harder to parse. (And there are cases like Philip pointed out, but those are already beyond my wpt.fyi searching abilities ;)).

lukebjerring:
It's busy, but it actually only triggers in particular circumstances. Not everyone navigates with the search box, so I feel like it's a more generic solution. Not fussed to be overruled if others are strongly opinionated.

foolip:
So was it deliberate to only prefix the directory name to other directories?

lukebjerring:
@foolip no; that was a bug (fixed in https://github.com/web-platform-tests/wpt.fyi/pull/1501/commits/0dbb6e4129e4a2ea205276f6c632f316abc34c45)

foolip:
I see, thanks! Trying that out, I think I have only two concerns. It takes up a lot of horizontal space, and it seems to make it impossible to drill down into the common directory if it contains only files. (If there are directories one can click them and then navigate up one level, which is also not great.)

Could we instead say something like "showing results in a / b / c" where each dir is linked? It'd be the same view as if navigating to that directory, except modifying the search would apply to the higher level directory.

Maybe in the blue box?

lukebjerring:
What's the use-case for going in only one directory? We've had complaints about the current behaviour (where clicking in one dir doesn't expand the single dir, you have to click more), so I feel like "needing" that behaviour contradicts the desire to change it?

I like your suggestion about hoisting the nesting out into the message, but then we'd have three path places; the breadcrumbs (wpt / foo / bar links at the top), the message (Showing [...] in baz / qux), and the trailing last path-part that we show in the list. What about having each part of the drilled-down path be a separate link to that folder (as done in the folder splitting in the breadcrumbs at the top)?

lukebjerring:
So changing the searchcache autocomplete had no effect; I guess we reverted back to client-side computation in #679 and left the server code to rot? cc @Hexcles for future cleanup

https://github.com/web-platform-tests/wpt.fyi/blob/94a0f85448a49d185c643cad26bab74088f49544/webapp/components/test-search.js#L430 is where we need to tweak the client-side behaviour, but I'm boarding a flight!

foolip:
I would want to drill down to the directory with all the results if I wanted to share a link, since that'll be a clearer view.

> What about having each part of the drilled-down path be a separate link to that folder (as done in the folder splitting in the breadcrumbs at the top)?

That sounds good.

I could then see that as a future mode for flattening results spanning many directories, if there aren't too many results in total.

foolip:
Searching for `path:/fullscreen/` now gives "Failed to fetch test runs". Not sure if it's a regression or not.

lukebjerring:
> Failed to connect to the backend due to a timeout.

Likely taking some time to spin up the searchcache staging instance?

lukebjerring:
Seems to be happy now after a searchcache deployment.

foolip:
Is the search cache deployment status posted before it's actually done? I saw two successful deployments listed when I tried before.

foolip:
https://explode-dot-wptdashboard-staging.appspot.com/results/?label=master&label=experimental&aligned&q=path%3A%2Ffullscreen%2F looks better now, but do you still intend to link the individual path components?

lukebjerring:
The deployment will be out of date until you see a "wpt-pr-bot deployed to x" message that's _after_ a push message.

I do intend to, but I think this PR is already changing enough things at once, so would suggest a follow-up (unless you think that it's current state is not an acceptable interim?)

PR #1509, opened 2019-09-24
Add webkitgtk as a known browser:

clopez:
## Description
This adds `webkitgtk` as a known browser. We want to start sending reports with runs of this browser to staging.wpt.fyi
The product from the wptreport is `webkitgtk_minibrowser`, which is normalized to `webkitgtk`

For related pull-request adding this browser to wpt see https://github.com/web-platform-tests/wpt/pull/19227

// cc @foolip

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -555,6 +555,9 @@ def normalize_product(report: WPTReport) -> Set[str]:
     elif product == 'edgechromium':
         report.run_info['product'] = 'edge'
         return {'edge', 'edgechromium'}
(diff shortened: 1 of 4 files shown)


foolip:
@lukebjerring or @Hexcles can either of you merge at a good time? Don't want to risk breaking things because I don't know the current state of the tree and release plans.

foolip:
Both Luke and Robert are out this week, so I'll try merging this and it should be deployed to staging. Then we can see if it works there, it won't reach wpt.fyi prod without a manual prod release so the risks are low.

PR #1511, opened 2019-09-27
Support or-conjunction for root queries

### LINKED ISSUE #1507: Conjunction (`and`) with `none` query seems to not work
On [this page](https://wpt.fyi/results/?diff&filter=ADC&q=none%28status%3Amissing%29%20and%20is%3Adifferent&run_id=320340002&run_id=322080003), the query is `none(status:missing) and is:different` but the `and` doesn't seem to work, all results are returned.

lukebjerring:
## Description
Follow-up for https://github.com/web-platform-tests/wpt.fyi/pull/1500 to fix #1507 

## Review Information
- Search for `all(status:pass) or none(status:pass)`
- See expected results

### DIFF
--- a/webapp/components/test/test-search.html
+++ b/webapp/components/test/test-search.html
@@ -375,6 +375,18 @@ suite('<test-search>', () => {
           { none: [{ status: 'PASS' }] },
         ]
       });
+      assertQueryParse('idlharness all(status:fail)', {
+        and: [
+          { exists: [{ pattern: 'idlharness' }] },
+          { all: [{ status: 'FAIL' }] },
+        ]
+      });
+      assertQueryParse('2dcontext and all(status:fail)', {
+        and: [
+          { exists: [{ pattern: '2dcontext' }] },
+          { all: [{ status: 'FAIL' }] },
+        ]
+      });
     });
   });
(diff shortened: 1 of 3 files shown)


webapp/components/test/test-search.html
```diff
@@ -361,6 +361,21 @@
           }],
       });
     });
+
+    test('multi-root', () => {
+      assertQueryParse('none(status:missing) count>0(status:!pass)', {
```
foolip: Can we also support explicit and at this level? Follow-up is fine.

lukebjerring: Already had, just wasn't documented.

lukebjerring:
searchcache in staging is having issues, but I can see the produced `idlharness all(status:fail)` query is `{"run_ids":[259450014,245520018,235540020,251530023],"query":{"and":[{"exists":[{"pattern":"idlharness"}]},{"all":[{"status":"FAIL"}]}]}}` as it should be, so I'll merge anyway.

foolip:
https://github.com/web-platform-tests/wpt.fyi/pull/1511 was merged, so is this fixed then?

Issue #1499, opened 2019-09-19
Additional insights


foolip:
https://wpt.fyi/insights could be even better:
- regressions between the stable and non-stable versions of a browser. Stable->TP for Safari, but for Chrome and Firefox both stable&rarr;beta and stable&rarr;experimental can be interesting.
- regressions between browser release N and N+1, for two runs that are as close as possible in time (we don't have the infra to ensure the same wpt commit is run on both versions)
- browser-specific "anomalies" split into categories of likely severity. FAIL+PASS+PASS is the strongest signal of a problem, errors and timeouts can be interesting but often test issues, and finally missing subtests can reveal additional problems but are often OK.

@othermaciej @youennf @burg @graouts @grorg are there more cases you think might be interesting to look for?

foolip:
Regression queries suggested in https://github.com/web-platform-tests/wpt/issues/19206#issuecomment-534027278.

[Safari stable&rarr;STP](https://wpt.fyi/results/?q=seq%28status%3Aok%7Cstatus%3Apass%20status%3Aok%7Cstatus%3Apass%20status%3A%21missing%26status%3A%21ok%26status%3A%21pass%20status%3A%21missing%26status%3A%21ok%26status%3A%21pass%29&run_id=318330001&run_id=328090002&run_id=328070001&run_id=305770001 regressions) might also be interested, but could be inflated by experimental features enabled in STP.

PR #1514, opened 2019-09-30
Add "Release regressions" to insights

lukebjerring:
## Description
Suggested in #1499 

## Review Information
- Go to the insights tab
- See new "Release regressions" insight


webapp/components/channel-picker.js
```diff
@@ -0,0 +1,58 @@
...
+        notify: true,
+      },
+      channel: {
+        type: String,
+        value: 'stable',
+        notify: true,
+      },
+      channels: {
+        type: Array,
+        value: Array.from(Channels),
```
foolip: Can't Channels be used directly? A copy doesn't cost much memory, but needed?

lukebjerring: It's for global state mutation protection (common bug).

PR #1523, opened 2019-10-03
Document /admin

foolip:
I did this because it's so much work to find https://wpt.fyi//admin/results/upload for me every few months that I need it.

### DIFF
--- a/docs/admin.md
+++ b/docs/admin.md
@@ -1,0 +1,13 @@
+# Admin
+
+## Flags
+
+[`/admin/flags`](https://wpt.fyi/admin/flags) allows site-wide settings to be set, similar to the per-user [`/flags`](https://wpt.fyi/flags).
+
+## Flushing caches
+
+[`/admin/cache/flush`](https://wpt.fyi//admin/cache/flush) flushes some caches.
+
+## Uploading results
+
+[`/admin/results/upload`](https://wpt.fyi//admin/results/upload) is a form for upload results manually. Please file an issue on this project if you require credentials to upload results manually or via the API.


docs/admin.md
```diff
@@ -0,0 +1,13 @@
+# Admin
+
+## Flags
+
+[`/admin/flags`](https://wpt.fyi/admin/flags) allows site-wide settings to be set, similar to the per-user [`/flags`](https://wpt.fyi/flags).
+
+## Flushing caches
+
+[`/admin/cache/flush`](https://wpt.fyi//admin/cache/flush) flushes some caches.
```
foolip: @Hexcles does just visiting this do something? I didn't try out of fear of side effects.

docs/admin.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Uploading results
+
+[`/admin/results/upload`](https://wpt.fyi//admin/results/upload) is a form for upload results manually. Please file an issue on this project if you require credentials to upload results manually or via the API.
```
foolip: @Hexcles what is the API endpoint? I tried to find a link to it but failed.

foolip: It looks like one has to upload a gzipped report. Should that be documented inline, or here?

docs/admin.md
Hexcles: Yes, which is probably a terrible design...

docs/admin.md
Hexcles: The API is https://github.com/web-platform-tests/wpt.fyi/tree/master/api#results-creation

The file input on the upload UI has `accept=".gz"`.

PR #1524, opened 2019-10-03
Add GitHub OAuth login

lukebjerring:
## Description
Adds a button to the top right corner for letting the user log in via their GitHub account. For now, this purely recognizes the user (fetches only the public information), which allows us to identify an author when making changes to wpt-metadata on their behalf.

Sets the user via a cookie named "session" which is a map, where "user" is the user.


lukebjerring:
@KyleJu I've transferred the instance of the GitHub OAuth app that I was using for local development to you. To develop this branch locally,

- Check out this branch (`oauth`)
- Run web_server.sh as usual
- Run `make dev_data`
- Head to 0.0.0.0:8000/datastore?kind=Token
- Edit the tokens to the values of the oauth app that redirects to localhost:8080/oauth

PR #1525, opened 2019-10-07
Restructure the metadata output format

lukebjerring:
## Description
Produces an object, keyed on test path, instead of an array.
Includes subtest and status filtering for each link.


webapp/components/test-file-results.js
```diff
@@ -273,7 +277,8 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
     return new Map([...firstScreenshot, ...Object.entries(screenshots)]);
   }
 
-  computeRows(resultsTable, onlyShowDifferences) {
+  computeRows(resultsTable, metadata, onlyShowDifferences) {
+    console.log(resultsTable, metadata);
```
lukebjerring:
```suggestion
```

webapp/components/test-file-results.js
```diff
@@ -73,13 +73,17 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
         type: Array,
         value: [],
       },
+      metadata: {
+        type: Array,
```
lukebjerring:
```suggestion
        type: Object,
```

webapp/components/test-file-results.js
```diff
@@ -73,13 +73,17 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
...
+        value: [],
```
lukebjerring:
```suggestion
        value: {},
```

webapp/views/wpt-results.js
```diff
@@ -357,6 +358,7 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
         type: Array,
         computed: 'computeDisplayedTests(path, searchResults)',
       },
+      metadata: Array,
```
lukebjerring:
```suggestion
      metadata: Object,
```

lukebjerring:
@jgraham - might be worth getting your input on the data format here, since you'll be a consumer.

jgraham:
The final format is something like

```
{
  <test_id>: 
  [
    {
      "product": <product>,
      "url": <url>,
      "results": 
      [
        {
          "status": <status>, 
          "subtest": <subtest>
        }
      ]
    }
  ]
} 
```
?

That seems OK, I think. Arguably it could be keyed on `test, subtest` with subtest being `null` for links with top-level tests, but I guess that's akward in languages with only string keys. Bundling together `status` and `subtest` seems a little odd, but none of this is a blocker; a lookup by test name followed by a scan of annotations for that test seems like it's going to be necessary regardless of the details of the format.

api/azure/gomock_reflect_313602269/prog.go
```diff
@@ -0,0 +1,66 @@
+
```
KyleJu: What is this file?

webapp/views/wpt-app.js
```diff
@@ -133,7 +133,8 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
                      path="{{subroute.path}}"
                      test-runs="{{testRuns}}"
                      test-paths="{{testPaths}}"
-                     search-results="{{searchResults}}"></wpt-results>
+                     search-results="{{searchResults}}"
+                     metadata="[[metadata]]"></wpt-results>
```
KyleJu: metadata is fetched inside of wpt-metadata.js so it should be always empty at this level?

nickeldennis1980 (approved):

PR #1527, opened 2019-10-08
Expose test duration

lukebjerring:
## Description
Fixes #282 
Surfaces the `duration` field of each single test-result-file.

## Review Information
- Visit the deployed environment
- View any individual test
- See "Duration" row underneath the main status row.

PR #1528, opened 2019-10-08
Fix staging deployment

lukebjerring:
## Description
Fixes what's considered deps for the webapp deployment.


lukebjerring:
I've bundled this change with #1527 to verify it in practice.

lukebjerring:
Does the trick. TBR'ing.

PR #1530, opened 2019-10-08
Rejig go_build to prune webapp's node_modules

lukebjerring:
## Description
Ensures that the webapp/node_modules are pruned before we run `go generate`, which packs all the contents into the build (8MB when correct, >150MB when including web_component_tester deps).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,10 @@ prepush: go_build go_test lint
 python_test: python3 tox
 	cd $(WPTD_PATH)results-processor; tox
 
+# NOTE: We prune before generate, because node_modules are packr'd into the
+# binary (and part of the build).
 go_build: git mockgen packr
+	make webapp_node_modules_prune
 	cd $(WPTD_PATH); go get -v ./...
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -117,11 +120,6 @@ _go_webdriver_test: var-BROWSER java go_build_test xvfb geckodriver webserver_de
 web_components_test: xvfb firefox chrome webapp_node_modules_all psmisc
 	util/wct.sh $(USE_FRAME_BUFFER)
 
-# Dependencies for running dev_appserver.py.
-webserver_deps: webapp_deps dev_appserver_deps
-
-webapp_deps: go_build webapp_node_modules
```
Hexcles: This is also used by `deploy.sh`

lukebjerring: Good catch. That make rule is out of date anyway.

lukebjerring:
@Hexcles looks like this is needed to prevent a race condition on the node_modules being installed before the go_build, which causes a broken deployment on master. Going to merge as-is and we'll fix anything I've missed.

PR #1531, opened 2019-10-08
Update references to "live" deployment

jugglinmike:
The deployment formally available at web-platform-tests.live is now
hosted at wpt.live. Update references accordingly.

### DIFF
--- a/webapp/components/wpt-flags.js
+++ b/webapp/components/wpt-flags.js
@@ -46,7 +46,7 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'showTestRefURL',
       'structuredQueries',
       'searchPRsForDirectories',
-      'webPlatformTestsLive',
+      'wptLive',
     ];
   }
 });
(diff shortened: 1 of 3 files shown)


webapp/components/wpt-flags.js
```diff
@@ -46,7 +46,7 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'showTestRefURL',
       'structuredQueries',
       'searchPRsForDirectories',
-      'webPlatformTestsLive',
```
lukebjerring: Changing this feature name will temporarily un-launch the feature (we need to change it in data store too).

jugglinmike:
Thanks. @lukebjerring! I take it updating the data store is not something that can be done from this pull request, but let me know if I'm mistaken.

wpt.live is currently inaccessible due to some firewall restrictions (@Hexcles is on the case), but since we've also dismantled web-platform-tests.live, that may not be important.

lukebjerring:
Correct - I've done it on appengine (i.e. I added `wptLive` as an enabled flag in datastore, both staging and prod) now, so we're good to merge this when ready!

jugglinmike:
It looks like the new project is not quite ready to host WPT. We've redeployed [web-platform-tests.live](http://web-platform-tests.live) while that comes together.

@Hexcles should this patch be reverted?

Hexcles:
Sigh, yes...

PR #1532, opened 2019-10-08
Opt @foolip and @autofoolip out of staging.wpt.fyi checks

foolip:


foolip:
Sent this because I don't like seeing so many checks, and since things stabilized I never used the staging checks.

api/checks/webhook.go
```diff
@@ -338,9 +338,7 @@ func isUserWhitelisted(aeAPI shared.AppEngineAPI, login string) bool {
 		return true
 	}
 	whitelist := []string{
-		"autofoolip",
 		"chromium-wpt-export-bot",
-		"foolip",
 		"gsnedders",
```
foolip: @gsnedders @jgraham @jugglinmike @lukebjerring @Ms2ger FYI I'm opting myself out of staging checks. If anyone else doesn't want to see the extra checks either I can roll that into this PR.

lukebjerring:
FWIW, staging actually performs a different computation of regressions (using the searchcache).

foolip:
Oh, I hadn't noticed. Does it catch more cases?

lukebjerring:
Less, actually :) it requires PASS => !PASS

foolip:
Nah, I'll opt out and see if I regret it :)

PR #1534, opened 2019-10-09
Exclude wpt-env-flags in /components/ regex

lukebjerring:
## Review
- Visit the deployed environment's `/flags` endpoint
- See some client-side features are enabled (as per their environment defaults).


lukebjerring:
TBR'ing

Hexcles:
This was incorrect. #1535 is a follow-up that fixes the problem.

PR #1535, opened 2019-10-09
Fix wpt-env-flags.js

lukebjerring:
#1534 was a bad patch. This fixes that.


lukebjerring:
Change here is essentially moving the placeholder into a separate folder; seems go111 doesn't treat the first URL match as preceding the static_dir one.

dynamic-components/wpt-env-flags.js => dynamic-components/templates/wpt-env-flags.js
components/wpt-env-flags.js => dynamic-components/wpt-env-flags.js

lukebjerring:
TBR'ing this to fix bad TBR from earlier.

PR #1536, opened 2019-10-10
Revert "Update references to "live" deployment"

Hexcles:
Reverts web-platform-tests/wpt.fyi#1531

(Re-land when `wpt.live` is actually live.)


Hexcles:
Luke added a new flag instead of renaming it, so we are good.

Issue #1540, opened 2019-10-13
Salvage chrome-experimental and firefox-experimental runs


foolip:
There are a bunch of Chrome and Firefox runs with -experimental in the product name, which this code tries (but fails) to deal with:
https://github.com/web-platform-tests/wpt.fyi/blob/4a1621ee74a070d320a7b855eb902fbe7b3704ec/webapp/components/display-logo.js#L121-L129

They can't be found via /runs or /api/runs, but here are two found via the Datastore admin UI:
https://wpt.fyi/results/?run_id=4697703762100224
https://wpt.fyi/results/?run_id=4739633715871744

It should be pretty simple to fix up the product name of these and add the appropriate labels. They don't have the master label, but that could be added for the subset which are for a commit with a merge_pr_* label.

Why? To allow removing that workaround, and make more runs available for historical trends.

foolip:
@Hexcles with this like https://github.com/web-platform-tests/wpt.fyi/issues/1175 would it help if I produced a list of run IDs and what edits should be applied to them?

I ask mainly because some of these are probably also labeled incorrectly, so handling them before https://github.com/web-platform-tests/wpt.fyi/issues/1175 probably makes sense.

Hexcles:
This has already been done: https://github.com/web-platform-tests/data-migration/blob/master/tagger/browser_name.go

However, the condition might have been a bit too strict (requiring the "stable" label) so it's possible that we missed a bunch.

Hexcles:
If you have a list of run IDs, we can do fix them. There shouldn't be too many.

foolip:
Why did it require the stable label? I'd expect all of these to be Firefox Nightly and Chrome Dev...

Hexcles:
Sorry, I meant requiring the absence of the stable label, which is totally reasonable but there could be old runs that were mislabelled.

foolip:
@stephenmcgruer since you've had fun with datastore recently, do you want to own this too?

stephenmcgruer:
API links for the runs in the OP:

https://wpt.fyi/api/runs/4697703762100224
```
[...]

PR #1541, opened 2019-10-13
Remove broken code to handle '-experimental' in product

foolip:
`replace` return a new string, so this didn't do anything.


foolip:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/1540 about renaming the products to expose them without a hack like this.

PR #1544, opened 2019-10-14
Revert "Explode folder contents in UI and collapse folders in autocomplete suggestions"

foolip:
This reverts https://github.com/web-platform-tests/wpt.fyi/pull/1501.

### DIFF
--- a/api/query/autocomplete.go
+++ b/api/query/autocomplete.go
@@ -7,7 +7,6 @@ package query
 import (
 	"encoding/json"
 	"net/http"
-	"path"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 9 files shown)


foolip:
Having seen this deployed for a while now, I'm not sure I think it'll be an improvement even if the follow-up discussed in https://github.com/web-platform-tests/wpt.fyi/pull/1501#issuecomment-533856971 ("having each part of the drilled-down path be a separate link to that folder") is implemented.

https://wpt.fyi/results/webdriver?diff&filter=ADC&run_id=311390003&run_id=330280040 shows the problem pretty well I think. All tests in webdriver/ are in webdriver/tests/, and that view looks very busy.

Better perhaps would be to link directly into the deepest directory from the ancestor directory, similar to what GitHub's file listing does when there are no files in intermediate directories.

@stephenmcgruer WDYT?

foolip:
(And the reason to revert instead of suggesting the improvement of course is that @lukebjerring is no longer around to work on the follow-up or any suggested alternative.)

stephenmcgruer:
Agreed; LGTM.

PR #1551, opened 2019-10-18
Run web component tests and WebDriver tests on GitHub Actions

Hexcles:
## Review Information

See individual commit messages.

We should see three new GitHub Checks: web_components_test, go_chrome_test & go_firefox_test. All need to pass.

### DIFF
--- a/.github/actions/make-in-docker/Dockerfile
+++ b/.github/actions/make-in-docker/Dockerfile
@@ -1,6 +1,8 @@
 # vim: set expandtab sw=4
 FROM golang:1.12-buster
 
+# Create a non-priviledged user to run browsers as (Firefox and Chrome do not
+# like to run as root).
 RUN chmod a+rx $HOME && useradd --uid 9999 --user-group --create-home browser
 
 # Sort the package names!
@@ -21,7 +23,9 @@ RUN apt-get update -qqy && apt-get install -qqy --no-install-suggests \
         xvfb && \
     rm /usr/bin/firefox
 
-# Overwrite sudoers to discard the secure_path setting to make go available.
+# The base golang image adds Go paths to PATH, which cannot be inherited in
+# sudo by default because of the `secure_path` directive. Overwrite sudoers to
+# discard the setting.
 RUN echo "root ALL=(ALL:ALL) ALL" > /etc/sudoers
 
 # Provide a fake openjdk-8-jdk for google-cloud-sdk-datastore-eumlator
--- a/.github/actions/make-in-docker/Dockerfile
+++ b/.github/actions/make-in-docker/Dockerfile
@@ -1,13 +1,16 @@
 # vim: set expandtab sw=4
 FROM golang:1.12-buster
 
+RUN chmod a+rx $HOME && useradd --uid 9999 --user-group --create-home browser
+
 # Sort the package names!
+# firefox-esr: provides deps for Firefox (we don't use ESR directly)
 # openjdk-11-jdk: provides JDK/JRE to Selenium & gcloud SDK
 # python-crcmod: native module to speed up CRC checksum in gsutil
-# sudo: used in Makefile (no-op inside Docker)
 RUN apt-get update -qqy && apt-get install -qqy --no-install-suggests \
         curl \
         equivs \
+        firefox-esr \
         lsb-release \
         openjdk-11-jdk \
         python-crcmod \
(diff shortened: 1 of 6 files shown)


.github/actions/make-in-docker/Dockerfile
```diff
@@ -1,21 +1,28 @@
 # vim: set expandtab sw=4
 FROM golang:1.12-buster
 
+RUN chmod a+rx $HOME && useradd --uid 9999 --user-group --create-home browser
```
stephenmcgruer: Worth documenting why we do this here, as it otherwise isn't explained I think?

.github/actions/make-in-docker/Dockerfile
```diff
@@ -1,21 +1,28 @@
...
         python-crcmod \
         python3.7 \
         sudo \
         tox \
         wget \
-        xvfb
+        xvfb && \
+    rm /usr/bin/firefox
+
+# Overwrite sudoers to discard the secure_path setting to make go available.
```
stephenmcgruer: Sorry, bit of term overload here with `make` and `go` :D. Does this have anything to do with Makefiles, or is it just that `go` won't work when `secure_path` is set? (Which doesn't work because...?)

Makefile
```diff
@@ -95,15 +95,15 @@ _go_webdriver_test: var-BROWSER java go_build xvfb geckodriver dev_appserver_dep
...
 	cd webdriver; \
-	go test $(VERBOSE) -timeout=15m -tags=large -args \
+	COMMAND="go test $(VERBOSE) -timeout=15m -tags=large -args \
 		-firefox_path=$(FIREFOX_PATH) \
 		-geckodriver_path=$$GECKODRIVER_PATH \
 		-chrome_path=$(CHROME_PATH) \
 		-chromedriver_path=$(CHROMEDRIVER_PATH) \
 		-frame_buffer=$(USE_FRAME_BUFFER) \
 		-staging=$(STAGING) \
-		-test.timeout=30m \
```
stephenmcgruer: Why are we dropping the timeout here?

Makefile
```diff
@@ -95,15 +95,15 @@ _go_webdriver_test: var-BROWSER java go_build xvfb geckodriver dev_appserver_dep
...
-		-browser=$(BROWSER) $(FLAGS)
+		-browser=$(BROWSER) $(FLAGS)"; \
+	if [ "$$UID" == "0" ]; then sudo -u browser $$COMMAND; else $$COMMAND; fi
```
stephenmcgruer: Ok, so if we're not root we keep our current user; why not always switch to browser? For people running locally not in docker?

util/wct.sh
```diff
@@ -21,4 +21,9 @@ if [ "$USE_FRAME_BUFFER" == "true" ]; then
 fi
 
 cd webapp
-npm test
+if [ "$UID" == "0" ]; then
+  # Used in .github/actions/make-in-docker/Dockerfile
```
stephenmcgruer: I'm somewhat concerned we're going to end up sprinkling `if UID == 0 { sudo -u browser foo } else { foo }` everywhere in our scripts (or worse, miss adding it and break GH actions). I don't immediately have a suggestion on how to address this, but something to think about.

webapp/wct.conf.json
```diff
@@ -2,6 +2,20 @@
   "suites": [
     "components/test/*.html"
   ],
+  "plugins": {
+    "local": {
+      "browserOptions": {
+        "chrome": [
+          "headless",
+          "disable-gpu",
+          "no-sandbox"
```
stephenmcgruer: What does having the sandbox enabled cause to happen?

.github/actions/make-in-docker/Dockerfile
Hexcles: It's because the official Go Docker image (the base image of ours) installs Go to /usr/local/go and appends `/usr/local/go/bin` to `PATH`. Debian by default resets `PATH` to `secure_path`, which is hard coded to a list of common paths like `/bin:/usr/bin`, inside `sudo`, to prevent accidentally running a non-system binary as root.

Makefile
Hexcles: There's in fact another `-timeout` flag (set to 15min) before `-args`. This is duplicate (`-test.foo` after `-args` is equivalent to `-foo` before -`args`). And I think 15min should be enough.

util/wct.sh
Hexcles: ditto

webapp/wct.conf.json
Hexcles: Chrome refuses to start because of lack of caps, which is in fact a somewhat common problem: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker we need `--cap-add=SYS_ADMIN` or the almighty `--priviledged` when starting Docker, but we can't control how GitHub Actions starts our Docker image.

Makefile
Hexcles: On a second thought, if one attempts to run this outside of Docker, the command would fail because user `browser` does not exist. We don't officially support running the project without Docker but so far it has always been possible.

stephenmcgruer: Yeah, I suspected that was the case. Ok, lgtm with sadness :)

webapp/wct.conf.json
stephenmcgruer: Ack. Any way to document that somewhere? Worst case, in the commit message so it at least shows up if someone goes digging in history.

PR #1555, opened 2019-10-21
Let Renovate bump node dependencies as well

foolip:
The config added in https://github.com/web-platform-tests/wpt.fyi/pull/1548 leaves webdriver/package.json and webapp/package.json unmanaged. This *might* fix that.


foolip:
@Hexcles I don't know if this might also cause all of the dependencies in node_modules directories to be managed. That would be very wrong :)

PR #1558, opened 2019-10-21
Clean up Travis configurations

Hexcles:
Now that we run all of our critical tests on GitHub Actions, which is
faster and more reliable than Travis, we can remove them from Travis to
make Travis finish faster (so that PRs don't get stuck in pending status
for that long).

We still run deployment and cleanup jobs on Travis. This commit also
cleans up the remaining job definitions a bit:
1. Upgrade from Ubuntu Xenial (16.04 LTS) to Bionic (18.04 LTS).
2. Change the deprecated keyword `matrix` to `jobs`.
3. Use conditional stages to simplify job definitions.
4. Run the cleanup job in a different stage to prevent race conditions.
5. Allow experimental Puppeteer tests to fail.

## Review Information

Check Travis. We should only see a Puppeteer job.

Unfortunately we need to merge it to master to test the remaining job definitions, but it's trivial to rollback this config change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-dist: xenial
+dist: bionic
 
 notifications:
   email:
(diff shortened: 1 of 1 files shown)


Hexcles:
I hate "smart" capitalization: https://travis-ci.com/web-platform-tests/wpt.fyi/builds/132881863

I wrote "deploy PR", and Travis showed it as "Deploy pr", what the...

PR #1563, opened 2019-10-22
Also run tests on PRs originated from a fork

Hexcles:
If a PR is from a fork, we don't receive the push event.

https://help.github.com/en/github/automating-your-workflow-with-github-actions/workflow-syntax-for-github-actions#on


Hexcles:
Actually, this is quite bad... We know have two duplicate sets of checks for PRs from this repo.

Let me try just `pull_request`.

Hexcles:
Alright, this is much better!

Issue #1579, opened 2019-10-23
Deploy 688384c to production


Hexcles:
Last deployment: #1520; changelist: df81019...688384c

The major change since then is the Go 1.11 migration (which encompasses migration to `go.mod`). It is a rather significant change so let's hope nothing breaks.

There is no schema change or observable UX/API changes.

Hexcles:
webapp 0%: https://688384c-dot-wptdashboard.appspot.com

stephenmcgruer:
> There is no schema change or observable UX/API changes.

Dunno if this counts, but I think the "Duration" is new, right? https://688384c-dot-wptdashboard.appspot.com/results/WebCryptoAPI/derive_bits_keys/ecdh_keys.https.any.worker.html?label=experimental&label=master&aligned vs https://wpt.fyi/results/WebCryptoAPI/derive_bits_keys/ecdh_keys.https.any.worker.html?label=master&label=experimental&aligned

Hexcles:
Thanks, Stephen -- yeah I missed that when skimming the changelist.

Hexcles:
processor 100%: 688384c-dot-processor-dot-wptdashboard.appspot.com
searchcache 100%: 688384c-dot-searchcache-dot-wptdashboard.appspot.com

Hexcles:
Promoted webapp to 50%

Hexcles:
webapp at 100%

Hexcles:
The announcer error has been fixed by #1580, and I deployed 8e62d2e-dot-announcer-dot-wptdashboard.appspot.com to prod.

PR #1580, opened 2019-10-23
Fix the WPT repo URL in announcer

Hexcles:
This should fix the problem found in https://github.com/web-platform-tests/wpt.fyi/issues/1579#issuecomment-545581342

Test: Travis should successfully deploy a staging version of this branch.


Hexcles:
I also did a global grep of `w3c/web-platform-tests` and didn't find anything else that needed to be fixed.

Hexcles:
It worked!

Issue #1352, opened 2019-06-12
Web component tests randomly hang


Hexcles:
The tests randomly hang about half way through and every time it might get stuck at a different location.

Hexcles:
This issue seems to have reappeared, this time on GitHub Actions. e.g.

https://github.com/web-platform-tests/wpt.fyi/pull/1554/checks?check_run_id=271841178#step:3:708

Hexcles:
cc @stephenmcgruer

Hexcles:
https://github.com/web-platform-tests/wpt.fyi/pull/1584/checks?check_run_id=274075821#step:3:994 showed something interesting:
```
A content process crashed and MOZ_CRASHREPORTER_SHUTDOWN is set, shutting down
```
But I can't reproduce the crash locally.

Hexcles:
I still can't reproduce the hang locally. For a moment, I thought the root cause was a big test file (`test-runs-query-builder.js`), but no I cannot prove conclusively that splitting the test makes the hang less often.

My best guess is that we are seeing the content process crashing, potentially due to lack of resources (memory?) on GitHub Actions.

PR #1584, opened 2019-10-24
Refresh WCT and disable WCT on Firefox

Hexcles:
This is a hacky workaround for #1352 .

In the end, I'm not able to even reproduce the hang locally, let lone finding the root cause, so I've decided that it's the right tradeoff to **temporarily disable WCT** on Firefox until we find a better testing framework. At least that way we can still have some coverage (on Chrome) and make the status sign meaningful (it's been a while that we always have an error sign).

However, I did discover a few problems in our WCT suite:

1. Big test file: `test-runs-query-builder.js` contained two components and its test was noticeably slower than other tests. Perhaps coincidentally, Firefox tends to hang when running this test, but I don't have a conclusive answer as everything is flaky.
2. Leaky network requests: a few tests (those for `views/`) leaked external fetches.

And I also made some drive-by code-health improvements: upgrade Firefox, simplify CLI args, and remove dead code.

Please review the PR commit by commit and see individual commit messages for details. And finally, please rebase merge.


Hexcles:
Damn, I got one pass. And after retry, Firefox hang again...

Hexcles:
@stephenmcgruer this has been a hard decision to me. I feel bad to only test on Chrome especially given the nature of this project... PTAL and let me know what you think.

Hexcles:
My best guess is that we are seeing the content process (of Firefox) crashing, potentially due to lack of resources (memory?) on GitHub Actions.

webapp/wct.conf.json
```diff
@@ -11,12 +11,14 @@
           "no-sandbox"
         ],
         "firefox": [
-          "--headless"
```
stephenmcgruer: Did this flag actually change in Firefox 70?

webapp/components/test/wpt-app.html
```diff
@@ -18,16 +19,43 @@
...
   setup(() => {
-    appFixture = fixture('wpt-app-fixture');
-    appFixture.path = '/';
+    sandbox = sinon.sandbox.create();
+    // Spoof an empty result for APIs to speed up tests.
+    // This is done in suiteSetup/suiteTeardown to avoid async fetches slipping
+    // through between tests.
+    const ignore = new RegExp('/api/(shas|versions|interop)');
+    sandbox.stub(window, 'fetch', url => {
+      if (url === undefined) {
```
stephenmcgruer: This is likely fine, but if we didn't check this wouldn't we just throw anyway when we accessed `url.pathname` (which is also fine)?

webapp/components/test/wpt-app.html
```diff
@@ -18,16 +19,43 @@
...
+  let sandbox;
 
   setup(() => {
-    appFixture = fixture('wpt-app-fixture');
-    appFixture.path = '/';
+    sandbox = sinon.sandbox.create();
+    // Spoof an empty result for APIs to speed up tests.
+    // This is done in suiteSetup/suiteTeardown to avoid async fetches slipping
+    // through between tests.
+    const ignore = new RegExp('/api/(shas|versions|interop)');
```
stephenmcgruer: Nit; I think I would view this from a different angle; these are not 'ignored' URLs, these are 'handled' URLs to me. Everything else is unhandled and should throw. But your call :).

webapp/components/test/wpt-interop.html
```diff
@@ -16,27 +18,34 @@
 import { WPTInterop } from '../../views/wpt-interop.js';
 import { PolymerElement } from '../../node_modules/@polymer/polymer/polymer-element.js';
 
-suite('WPTInterop', () => {
-  const fetches = {};
-  let sandbox;
+const fetches = {};
```
stephenmcgruer: Could you add a line of documentation about what `fetches` is for?

webapp/components/product-builder.js
```diff
@@ -0,0 +1,261 @@
+/**
```
stephenmcgruer: FYI, I only scanned the files in this commit (Split product-builder.js out) since I assume its a straight-forward code heist. Lmk if I should take a closer look.

webapp/wct.conf.json
Hexcles: No. `firefox --help` has always shown the single-dash form, but takes the double-dash form implicitly (at least on Linux). I changed it to avoid relying on undocumented behaviours.

webapp/components/product-builder.js
Hexcles: Yes it is a simple copy-paste, followed by removing unused imports.

webapp/components/test/wpt-app.html
Hexcles: I'm leaving this as is for now.

webapp/components/test/wpt-app.html
Hexcles: Good point! I copied this over without giving it much thought.

PR #1593, opened 2019-10-30
Remove the @google-cloud/datastore dependency in webdriver/

foolip:
It is never `require()`'d.

Introduced in https://github.com/web-platform-tests/wpt.fyi/pull/1304.


foolip:
Spotted via https://github.com/web-platform-tests/wpt.fyi/pull/1569 where I thought it seems strange that this dependency is needed. CI will reveal what I've missed :)

Issue #1598, opened 2019-11-01
Taskcluster webhook failing because `target_url` changes


foolip:
wpt.fyi/runs has no recent Chrome and Firefox runs:
![image](https://user-images.githubusercontent.com/498917/68026201-30fc0100-fc6c-11e9-9fa2-c511e4508d46.png)

Looking at https://github.com/web-platform-tests/wpt/settings/hooks/53634544 (needs admin access) you can see:
![image](https://user-images.githubusercontent.com/498917/68026240-4ec96600-fc6c-11e9-959f-a09ec0a93f30.png)

Key bit:
> unrecognized target_url: https://tools.taskcluster.net/groups/F0JZQ3ZXSdm52wokJv0XXg

The problematic code is here:
https://github.com/web-platform-tests/wpt.fyi/blob/446d5d8e21905a1c66b0468bc7780ad79eb15873/api/taskcluster/webhook.go#L87-L90

What's changes is that the old payloads had `target_url`s on the form "https://tools.taskcluster.net/task-group-inspector/#/SVjeHXGNRauh0ac-tQ4klQ" while a new one looks like "https://tools.taskcluster.net/groups/VFI9FCVDTLW5rkvwdyM_yQ".

foolip:
The comments around the patterns are interesting because the `inspectorURLRegex` form was actually the one that had no match before, it was used for the Statuses API as well:
https://github.com/web-platform-tests/wpt.fyi/blob/446d5d8e21905a1c66b0468bc7780ad79eb15873/api/taskcluster/webhook.go#L30-L36

This may cause this issue to be entangled in https://github.com/web-platform-tests/wpt.fyi/issues/1240, but hopefully not.

Hexcles:
For the record, the root cause was a change on the Taskcluster side: https://github.com/taskcluster/taskcluster/pull/1489

Hexcles:
I have resent all failed webhooks. Runs are coming back into staging.

Hexcles:
Production has been backfilled, too (many runs are still [being processed](https://wpt.fyi/status) so it might take a short while).

PR #1599, opened 2019-11-01
Drop the /tasks/ part of the expected `target_url` for Taskcluster

foolip:
Sample URL from a real webhook payload:
https://tools.taskcluster.net/groups/VFI9FCVDTLW5rkvwdyM_yQ

May fix https://github.com/web-platform-tests/wpt.fyi/issues/1598.


foolip:
Happily, the tests failed:
https://github.com/web-platform-tests/wpt.fyi/pull/1599/checks?check_run_id=284765012#step:3:922

Unfortunately, I can't work any more on this today.

PR #1601, opened 2019-11-01
[UI] Add WebKitGTK as a product

stephenmcgruer:
This adds WebKitGTK as a product to the UI. This should not affect API endpoints.


stephenmcgruer:
(This is WIP)

stephenmcgruer:
@Hexcles PTAL. I couldn't think of much in the way of testing for this change, but (very) happy to do any if you can think of something worthwhile. (Maybe just a dumb test that setting `products` via templating does override the rendered result properly?)

webapp/components/product-info.js
```diff
@@ -97,6 +109,13 @@ function parseProduct(name) {
 const ProductInfo = (superClass) => class extends superClass {
   static get properties() {
     return {
+      // Polymer templates can only access variables in the scope of the owning
+      // class. Forward some declarations so that subclasses can use them in
+      // template parameters.
+      allProducts: {
+        type: Array,
+        value: AllProducts,
```
Hexcles: Nit: mark this as `readOnly: true` to prevent accidental modification.

stephenmcgruer: Nice, didn't know that was an available property. Done!

PR #1607, opened 2019-11-04
[UI] Clarify github repos on the about page

stephenmcgruer:
Previously it seemed like wpt.fyi was the one repo for both running and
serving the tests. Split that section into two.

### DIFF
--- a/webapp/components/wpt-header.js
+++ b/webapp/components/wpt-header.js
@@ -59,7 +59,6 @@ class WPTHeader extends WPTFlags(PolymerElement) {
           <a href="/status">Processor</a>
         </template>
         <a href="/about">About</a>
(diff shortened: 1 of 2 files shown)


stephenmcgruer:
Looks like this:

hGGPzTRoWA9

Hexcles:
Could you also change the link and perhaps the wording of the "GitHub Source" link in the header? I see a few options:
1. Change it to "WPT Docs" and link to https://web-platform-tests.org
2. Remove it altogether

I don't think we need a top-level link to our GitHub repo.

stephenmcgruer:
Removed it. I don't think it adds value; people will probably click through to About instead now and hopefully be less likely to be misled.

Issue #1605, opened 2019-11-04
Accept GitHub status events from the new TC deployment


jgraham:
TaskCluster is changing instances; we will be on the community instance. This means that the root URL will change to `https://community-tc.services.mozilla.com`. We need to update the wpt.fyi configuration to match; deploying this should presumably be coordinated with actually switching wpt over to the new instance.

For full details see https://bugzilla.mozilla.org/show_bug.cgi?id=1574668

foolip:
@Hexcles are you able to manage this transition this week? If not, @stephenmcgruer?

stephenmcgruer:
@jgraham - is the new root URL already up and running? Assuming so, seems we should just land and deploy a change in wpt.fyi that accepts either URL in `processTaskClusterBuild`, and then migrate WPT.

Looks like the deadline is November 9th from the Mozilla bug.

jgraham:
Yes it's already running. If we can accept both that's perfect (we've taken the same approach with the wpt code itself).

djmitche:
Please ping me if there's anything I can do to help.  From a quick refresher on status, I think I'm out of the critical path now?

Hexcles:
@jgraham suggested to accept both domains. I took a quick look and it's indeed possible if we assume the domain in the GitHub status event from Taskcluster is always the same as the domain of the API server. I created #1610 for this.

And @djmitche , no it's on me and James' coordinating the switchover. We plan to do it tomorrow 3:30pm UTC.

djmitche:
Awesome!  Note, too, that the context in the status event has a different prefix for the two integrations.

Hexcles:
@djmitche Just to confirm, if I use the prefix extracted from the context in a GitHub status as the rootURL to construct API URL using `tcurl.API`, I should be able to query the task group in the context, right?

djmitche:
I think the contexts are strings like "Taskcluster (push)" or "Community-TC (push)", so you'd need to index some array of rootUrls to determine the rootUrl.  But with that modification, yes.

djmitche:
Here's a commit with the new deployment (click the checkmark)

Hexcles:
Ahh *that* context. Sorry I meant the target URL.

And yes, I need to accept the new context string. Thanks for the heads-up!

Hexcles:
Alright, #1610 should do it.

Hexcles:
Reopening this issue to keep track of the rollout and switchover process

[...]

PR #1612, opened 2019-11-06
[webhook] Fix URL handling for old Taskcluster

Hexcles:
Old Taskcluster instance, which uses subdomains for different services
and we need to strip the subdomain away. For example, GitHub events from
the old Taskcluster instance have "tools.taskcluster.net" in their
target URLs, but the root URL should be "taskcluster.net".

#1605

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -226,7 +226,7 @@ func parseTaskclusterURL(targetURL string) (rootURL, taskGroupID, taskID string)
 		rootURL = matches[1]
 		taskGroupID = matches[2]
 	} else if matches := taskURLRegex.FindStringSubmatch(targetURL); len(matches) > 2 {
-		if len(matches) > 2 {
+		if len(matches) > 3 {
 			rootURL = matches[1]
 			taskGroupID = matches[2]
 			taskID = matches[3]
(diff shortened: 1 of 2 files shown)


api/taskcluster/webhook.go
```diff
@@ -223,15 +223,24 @@ func shouldProcessStatus(log shared.Logger, processAllBranches bool, status *sta
 
 func parseTaskclusterURL(targetURL string) (rootURL, taskGroupID, taskID string) {
 	if matches := inspectorURLRegex.FindStringSubmatch(targetURL); len(matches) > 2 {
-		return matches[1], matches[2], ""
-	}
-	if matches := taskURLRegex.FindStringSubmatch(targetURL); len(matches) > 2 {
+		rootURL = matches[1]
+		taskGroupID = matches[2]
+	} else if matches := taskURLRegex.FindStringSubmatch(targetURL); len(matches) > 2 {
 		if len(matches) > 2 {
```
LukeZielinski: Maybe a Go-newb question, but don't we already check `len(matches) > 2)` on the prev line? If so, seems like the `else` below won't ever execute?

Hexcles:
```suggestion
		if len(matches) > 3 {
```

Hexcles: Fixed.

stephenmcgruer:
Drive-by; that latest commit fixed a bug but no test failed over it (I think). Worth making sure its tested.

Hexcles:
@stephenmcgruer great point!

Turns out, an optional capturing group will always produce a "match" in Golang -- it'll be empty when it does not match anything. In other words, `len(matches)` is always 3 in that branch.

Switched to testing `matches != nil` per the [Golang docs](https://golang.org/pkg/regexp/#Regexp.FindStringSubmatch).

PR #1615, opened 2019-11-07
Hide the auto-complete drop-down arrow on Chrome/Safari

stephenmcgruer:
The arrow is visually unpleasant on Safari (always visible, strangely
vertical blue box). On Chrome it's not bad (only visible on mouseover,
just a grey arrow), but the webkit selector used to hide it hits both
browsers and I would rather not add UA sniffing.


stephenmcgruer:
Hah, this seems to turn out to work in Chrome but *not* Safari (which was the whole point). Sigh.

PR #1617, opened 2019-11-08
[API] Add `is:optional` command

stephenmcgruer:
Similar to `is:tentative`, this atom filters for tests marked as optional (currently based on file name).

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -155,6 +155,11 @@ across the runs.
 Filters to tests that are marked as tentative (currently based on [file
 name](https://web-platform-tests.org/writing-tests/file-names.html)).
 
+##### `is:optional`
+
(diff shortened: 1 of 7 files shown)


foolip:
Thanks for doing this! Can you add a TODO to also handle the `may` and `should` flags for CSS with this atom? Documentation: https://web-platform-tests.org/writing-tests/css-metadata.html#requirement-flags

stephenmcgruer:
I didn't know those were already a thing! Added a note, and filed https://github.com/web-platform-tests/wpt.fyi/issues/1619.

I took a quick look at `FileContentsQuery` but the approach is pretty bad performance wise I suspect. We should investigate tagging the tests in the METADATA (or otherwise pre-building an index during METADATA production) as a potential alternative.

PR #1621, opened 2019-11-08
Update web-platform-tests.live to wpt.live

foolip:
web-platform-tests.live is no longer online. wpt.live is.


Ms2ger:
Thanks for fixing these as well!

Issue #1414, opened 2019-08-14
Fetch the browser logos using npm?


alrra:
The [browser logos](https://github.com/alrra/browser-logos) are available on [`npm`](https://www.npmjs.com/org/browser-logos). They are usually keep up to date and highly optimized (e.g. see the changelogs for [`Firefox`](https://github.com/alrra/browser-logos/blob/b92e3a4717d9f120c7be5484d087afc00947de91/src/firefox/CHANGELOG.md#readme), [`Chrome`](https://github.com/alrra/browser-logos/blob/master/src/chrome/CHANGELOG.md#readme), [`Edge Dev`](https://github.com/alrra/browser-logos/blob/b92e3a4717d9f120c7be5484d087afc00947de91/src/edge-dev/CHANGELOG.md#readme), etc.).

This would make things easier in cases such as: #98, #1347.

---

Here are some examples of usage: [v8.dev](https://github.com/v8/v8.dev/commit/542d0b520e8331f0e73aed6c6f8a9a1c2dd30356#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), [WHATWG](https://github.com/whatwg/whatwg.org/commit/0018d3affaf9fc18ef471be098a08feb86b603ba).

PR #1625, opened 2019-11-08
Fetch `webkitgtk` logos from `npm`

alrra:
Related to: #1414 and #1601


webapp/package.json
```diff
@@ -22,6 +22,7 @@
     "@browser-logos/safari": "1.0.6",
     "@browser-logos/safari-technology-preview": "1.0.7",
     "@browser-logos/uc": "1.0.5",
+    "@browser-logos/web": "2.0.3",
```
Hexcles: The PR title says `webkitgtk`, but this is `web` i.e. Epiphany?

alrra: > but this is web i.e. Epiphany?

[Yes](https://github.com/alrra/browser-logos/tree/ed5b24d8a66365416f1e69d1979a55c9324b2367/src/web#web-fka-epiphany).

> The PR title says webkitgtk

Because `webkitgtk` is the name this project uses, `web` is the name `browser-logos` uses for that browser.

See also: https://wiki.gnome.org/Apps/Web.

Hexcles: Yeah I'm aware of that, but they are not the same and in fact we run both browsers.

@clopez do we have a dedicated logo for webkitgtk now?

alrra: > Yeah I'm aware of that, but they are not the same and in fact we run both browsers.

Ah ok. The logo this project uses is the same as the one `web` has. 🤷🏻‍♂️

clopez: No, we don't have a dedicated logo for webkitgtk. Using the logo from Web (Epiphany) is fine for now I guess (actually we are already doing that since #1601). Also note that for the moment we don't run tests with Web/Epiphany automatically.

Hexcles: Thanks!

PR #1637, opened 2019-11-13
[processor] Add experimental label to canary channel

### LINKED ISSUE #1635: Should Edge Canary have the 'experimental' label?
Forked from https://github.com/web-platform-tests/wpt.fyi/issues/1539. Copying some comments over:

https://github.com/web-platform-tests/wpt.fyi/issues/1539#issuecomment-552049114
"Do we have an issue for the fact that Edge Canary doesn't seem to be tagged with experimental? (Or is there a valid reason it shouldn't be)."

https://github.com/web-platform-tests/wpt.fyi/issues/1539#issuecomment-552073962
"There’s no issue for that AFAIK, and I don’t recall how it came about. @mustjab do you know why it’s Edge Dev and not Canary that shows on wpt.fyi by default?"

https://github.com/web-platform-tests/wpt.fyi/issues/1539#issuecomment-552133209
"The code reason, incidentally, would seem to be that results-processor/wptreport.py has a CHANNEL_TO_LABEL map that doesn't include 'canary' (it maps ['release', 'stable'] --> 'stable', ['beta'] --> 'beta', and ['dev', 'experimental', 'nightly', 'preview'] --> 'experimental'), and that the report from azure doesn't include 'experimental' either."

Assigning to mustjab to answer the query about whether Edge Canary should have the experimental label.

Hexcles:
Fixes #1635 

This must land right after https://github.com/web-platform-tests/wpt/pull/20240 .

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -38,6 +38,7 @@ CHANNEL_TO_LABEL = {
     'stable': 'stable',
     'beta': 'beta',
     'dev': 'experimental',
+    'canary': 'experimental',
     'experimental': 'experimental',
     'nightly': 'experimental',
     'preview': 'experimental',
--- a/results-processor/wptreport_test.py
+++ b/results-processor/wptreport_test.py
@@ -577,15 +577,24 @@ class HelpersTest(unittest.TestCase):
             prepare_labels(r, '', 'blade-runner'),
             {'blade-runner', 'dev', 'experimental', 'firefox'}
         )
+
         r._report['run_info']['browser_channel'] = 'nightly'
         self.assertSetEqual(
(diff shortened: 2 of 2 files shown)


Hexcles:
Yeah that's a great point that I didn't realize... Thanks, @foolip !

Short of a customizable homepage, I'd still prefer to show Edge Dev instead of Edge Canary by default for better alignment with Chrome. We should change the default query to accommodate this.

foolip:
Approved so that it can be merged and released when the time is right.

Labeling both as experimental and changing the default query would solve most of the problem, but there may be people who have queries against the experimental label that they expect to be a consistent set of configurations. Not sure how to deal with that really.

Hexcles:
Dropping this in favour of #2081

Issue #123, opened 2018-05-07
Add utility for cleaning up AppEngine versions


lukebjerring:
We're currently at 57 versions, and the limit is 200 [[1]](https://cloud.google.com/appengine/docs/standard/python/an-overview-of-app-engine#limits).

    gcloud app versions list

Shows the current versions, including comical testing names like `pants`.

[1] - https://cloud.google.com/appengine/docs/standard/python/an-overview-of-app-engine#limits

PR #1638, opened 2019-11-13
[deploy] Filter all illegal chars in branch names

### LINKED ISSUE #1578: Renovate PRs cannot be deployed to staging
Branches generated by Renovate have super long names with characters that cannot be used in GAE version names, so they usually cannot be deployed. e.g.

https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/248537768#L1225

```
ERROR: (gcloud.app.deploy) argument --version/-v: Bad value [renovate-pluralize-8.x]: May only contain lowercase letters, digits, and hyphens. Must begin and end with a letter or digit. Must not exceed 63 characters.
```

Hexcles:
Switch to a more sophisticated two-pass `tr` translation to clean up
branch names:
1. Convert all uppercase to lowercase.
2. Convert all non-alphanumeric chars to hyphens.

Fixes #1578.

### DIFF
--- a/util/deploy.sh
+++ b/util/deploy.sh
@@ -51,7 +51,7 @@ if [[ "${APP_PATH}" == "webapp" ]]; then
 fi
 
 # Create a name for this version
-VERSION_BRANCH_NAME="$(echo ${BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"} | tr /_ - | cut -c 1-28)"
+VERSION_BRANCH_NAME="$(echo ${BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"} | tr [:upper:] [:lower:] | tr -c [:alnum:] - | cut -c 1-28)"
 USER="$(git remote -v get-url origin | sed -E 's#(https?:\/\/|git@)github.com(\/|:)##' | sed 's#/.*$##')-"
 if [[ "${USER}" == "web-platform-tests-" ]]; then USER=""; fi
 


util/deploy.sh
```diff
@@ -51,7 +51,7 @@ if [[ "${APP_PATH}" == "webapp" ]]; then
 fi
 
 # Create a name for this version
-VERSION_BRANCH_NAME="$(echo ${BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"} | tr /_ - | cut -c 1-28)"
+VERSION_BRANCH_NAME="$(echo ${BRANCH_NAME:-"$(git rev-parse --abbrev-ref HEAD)"} | tr [:upper:] [:lower:] | tr -c [:alnum:] - | cut -c 1-28)"
```
stephenmcgruer: This seems like it will replace all non-alphanumeric characters with one `-` for each - is that what we want? I'm not familiar with what branch names we tend to deal with.

Hexcles: Yes, e.g. `Abc1_2.3-4` to `abc1-2-3-4`. Any concerns about that?

stephenmcgruer: That case looks fine. I was thinking of something like `Abc@#123`, which would just be a little weird (double-hyphen). But if that's rare/never happens, SGTM.

Hexcles: Yeah hopefully no one does that...

stephenmcgruer:
I think this is possibly related to a build failure for my PR: https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/257541465

```
ERROR: (gcloud.app.deploy) argument --version/-v: Bad value [smcgruer-webkitgtk-search-]: May only contain lowercase letters, digits, and hyphens. Must begin and end with a letter or digit. Must not exceed 63 characters.
```

I would guess that echo is adding a newline:

```
smcgruer@penguin:~/github/wpt.fyi$ echo smcgruer/webkitgtk-search | tr [:upper:] [:lower:] | tr -c [:alnum:] -
smcgruer-webkitgtk-search-smcgruer@penguin:~/github/wpt.fyi$
```

Hexcles:
You are exactly right. I didn't notice this because my branch name was too long and the trailing newline got cut.

This is yet another example of shell trying to be smart but creating a footgun:

```bash
foo=$(echo -n hello)
echo ${#foo}
bar=$(echo hello)
echo ${#bar}
```

Both produce 5. Sigh...

Issue #1656, opened 2019-11-20
Do not use Ephiphany logo for WebKitGTK


stephenmcgruer:
Igalia [have requested](https://github.com/web-platform-tests/wpt.fyi/issues/1576#issuecomment-556024432) that we don't use the current Epiphany logo for our WebKitGTK runs, due to concerns about people getting confused between the products.

Unfortunately, however, there is also still not a logo for WebKitGTK, so for now we'll have to find some generic (and free-use) logo.

stephenmcgruer:
I believe https://en.wikipedia.org/wiki/File:Internet-web-browser.svg will do

Hexcles:
@clopez said in https://github.com/web-platform-tests/wpt.fyi/pull/1625#discussion_r344914986

> No, we don't have a dedicated logo for webkitgtk. Using the logo from Web (Epiphany) is fine for now I guess (actually we are already doing that since #1601). Also note that for the moment we don't run tests with Web/Epiphany automatically.

stephenmcgruer:
I'm not sure what quoting that comment is meant to say, @Hexcles ? It is true that previously @clopez said it was ok to use the Epiphany logo, but @mcatanzaro then saw and disagreed strongly with that. I'll leave it to those two to resolve that difference, but in the meantime it seems reasonable to locate a different icon to use (and shouldn't be much work to swap it out).

mcatanzaro:
The proposed Tango icon looks good to me.

The shared heritage with the Epiphany icon is pretty clear too. Ours is a redesign of the Adwaita web-browser icon, which I presume is a redesign of the Tango icon.

Hexcles:
Oh I got the timing wrong. I thought @clopez 's comment was newer so I got confused. Never mind.

PR #1657, opened 2019-11-20
Switch WebKitGTK logo to tango icon

stephenmcgruer:
Logo from https://en.wikipedia.org/wiki/File:Internet-web-browser.svg

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/1656


Hexcles:
For the record, this asset is in the public domain.

LGTM, but one optional nit: I'd have preferred to split this into two commits: first a pure revert of 11ae764c2b770541d3892642e9e0b69271020e76 and then a replacement of the assets, so that we are sure nothing is missed by mistake. But feel free to merge this as is if you've checked.

stephenmcgruer:
A good idea, as it caught two places I forgot to update! It also reverts the gitignore which is a bit less good, but I think worth it for the clean revert.

PR #1658, opened 2019-11-20
Collect exceptions using Google Analytics

Hexcles:
Also tweak templates to place GA snippets before any other scripts
instead of at the end as suggested by
https://developers.google.com/analytics/devguides/collection/analyticsjs

### DIFF
--- a/webapp/templates/_ga.html
+++ b/webapp/templates/_ga.html
@@ -18,5 +18,5 @@ window.onerror = function(message, source, lineno, colno) {
   }
   // https://developers.google.com/analytics/devguides/collection/analyticsjs/exceptions
   ga('send', 'exception', {'exDescription': description});
-});
+};
 </script>
--- a/webapp/templates/_ga.html
+++ b/webapp/templates/_ga.html
@@ -5,5 +5,18 @@ m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
 })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
 ga('create', 'UA-89387706-1', 'auto');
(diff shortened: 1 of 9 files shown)


webapp/templates/_ga.html
```diff
@@ -5,5 +5,18 @@
 })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
 
 ga('create', 'UA-89387706-1', 'auto');
+// self-navigator.js sends more pageview events for in-page navigations.
 ga('send', 'pageview');
+
+window.addEventListener('error', function(message, source, lineno, colno) {
```
Hexcles:
```suggestion
window.onerror = function(message, source, lineno, colno) {
```

FYI @stephenmcgruer : it turns out I didn't read MDN closely enough. The callback for `addEventListener` only takes a single argument `event`, the form of callback I'm using is for `window.onerror`.

webapp/templates/_ga.html
```diff
@@ -5,5 +5,18 @@
...
+  let description;
+  // https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
+  if (message.toLowerCase().indexOf('script error') > -1) {
+    description = `External error: ${message}`;
+  } else {
+    description = `${message} at ${source}:${lineno}:${colno}`;
+  }
+  // https://developers.google.com/analytics/devguides/collection/analyticsjs/exceptions
+  ga('send', 'exception', {'exDescription': description});
+});
```
Hexcles:
```suggestion
};
```

webapp/templates/_ga.html
stephenmcgruer: And thus turns out that I did not review well enough, sorry :(.

Still LGTM.

PR #1665, opened 2019-11-23
Fix stale version cleanup

Hexcles:
##  Description

The version cleanup script deletes version older than 1 day without a corresponding branch in origin. Previously, this was done by checking the *local* refs fetched from the remote. There was a problem: Travis uses a shallow clone so we don't have all the refs, so I "worked around" it using `--unshallow`: #1290 . However, I didn't actually test it. Turns out the fix is not working. `git clone --depth=N` also implies `--single-branch`, so even with `--unshallow` we'd only only fetch the full history of the master branch, not the other branches.

This PR fixes the problem properly using `git ls-remote`, which directly queries the remote end without fetching. And this time, I tweaked the Travis setup to actually test it.

## Review Information

Example dry run (before reverting the debug commits): https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/259816161

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,11 @@ env:
 # In jobs.include, each job will inherit the stage of the previous job and the
 # default stage is "test".
 stages:
-  - name: clean up versions
-    if: type = pull_request
   - name: deploy PR
-    if: (type = push) AND (branch = master)
+    if: type = pull_request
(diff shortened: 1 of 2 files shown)


Hexcles:
cc @KyleJu the cleanup script does have a bug.

PR #1669, opened 2019-11-25
Retrieve wpt.fyi Github Bot Client Instance

KyleJu:
Create a method to retrieve an instance of wpt.fyi Github Bot client, for the implementation of /api/metadata PATCH endpoint. See [design](https://docs.google.com/document/d/1iRkaK6cGgXp3DKbNbPMVsYGMaOHO-5CfqEuLPUR_2HM/edit#heading=h.tx8rv4daaxk9)

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -100,16 +100,12 @@ func (a appEngineAPIImpl) GetSlowHTTPClient(timeout time.Duration) (*http.Client
 
 func (a *appEngineAPIImpl) GetGitHubClient() (*github.Client, error) {
 	if a.githubClient == nil {
-		ds := NewAppEngineDatastore(a.ctx, false)
-		secret, err := GetSecret(ds, "github-api-token")
+		client, err := a.getGithubClientFromKey("github-api-token")
 		if err != nil {
 			return nil, err
 		}
 
-		oauthClient := oauth2.NewClient(a.ctx, oauth2.StaticTokenSource(&oauth2.Token{
-			AccessToken: secret,
-		}))
-		a.githubClient = github.NewClient(oauthClient)
+		a.githubClient = client
 	}
 	return a.githubClient, nil
 }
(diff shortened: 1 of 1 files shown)


KyleJu:
@Hexcles Would you mind adding wpt.fyi@chromium.org to the WPT org and give it the write/read access to wpt-metadata?

shared/appengine.go
```diff
@@ -195,3 +200,18 @@ func getURL(host, path string, filter TestRunFilter) *url.URL {
 	detailsURL.RawQuery = filter.ToQuery().Encode()
 	return detailsURL
 }
+
+func (a appEngineAPIImpl) getGithubClientFromKey(key string) (*github.Client, error) {
```
stephenmcgruer: Nit; for consistency the parameter here is a `token`. In this context, a `key` is what we get back from the datastore.

shared/appengine.go
```diff
@@ -185,6 +181,15 @@ func (a appEngineAPIImpl) GetRunsURL(filter TestRunFilter) *url.URL {
 	return getURL(a.GetHostname(), "/runs", filter)
 }
 
+func (a appEngineAPIImpl) GetWPTFYIGithubBot() (*github.Client, error) {
```
stephenmcgruer: I note that `GetGitHubClient` caches the client, whereas this version just retrieves it each time. That may be a reasonable choice given usage, just wanted to make sure we were aware of the difference.

wptfyibot: Right, we can look at the /api/metadata usage on the dashboard after it launches.

KyleJu: Right, we can look at the /api/metadata usage on the dashboard after it launches.

KyleJu:
Chatted with @stephenmcgruer: (1) granted wpt.fyi@chromium.org read/write access to wpt-metadata (2) added its auth secrete to the Datastore with a github-wpt-fyi-bot-token token name

PR #1671, opened 2019-11-28
Update web_server.sh util script to use right make target

stephenmcgruer:


stephenmcgruer:
@Hexcles I noticed this a while back (2 weeks?) - I think you changed the Makefile. Not sure if this naive rename is 100% right since you changed more than just the name, but it seems to work for me locally (have been dragging around this diff for 2 weeks and finally got around to uploading it :D)

PR #1690, opened 2019-12-02
GitHub Login Integration

KyleJu:
## Description
On the UI side, this PR adds a button to the top right corner for letting the user log in/log out via their GitHub account.

On the server side, this PR implements /login, /oauth and /logout to facilitate the Github Oauth flow. Upon a successful login, it sets the user handle, user email and a Github token via a cookie named "session".

For more information, see [GitHub Login Integration on wpt.fyi](https://docs.google.com/document/d/1iRkaK6cGgXp3DKbNbPMVsYGMaOHO-5CfqEuLPUR_2HM/edit)
## Review Information
1. Open view deployment
2. Go to /flags and check Enable GitHub OAuth login
3. Go back to home page and click on Sign In with Github
4. Once login, your Github handle should appear on the top right corner.

## TODO
Need unit testing


go.mod
```diff
@@ -14,11 +14,12 @@ require (
 	github.com/gobuffalo/packr/v2 v2.7.1
 	github.com/golang/groupcache v0.0.0-20191002201903-404acd9df4cc // indirect
 	github.com/golang/mock v1.3.1
-	github.com/google/go-cmp v0.3.1 // indirect
+	github.com/google/go-github v17.0.0+incompatible
```
Hexcles: We should use only one version of `go-github` unless there's reason that we have to.

go.sum
```diff
@@ -115,10 +117,16 @@ github.com/google/uuid v1.1.1/go.mod h1:TIyPZe4MgqvfeYDBFedMoGGpEw/LqOeaOT+nhxU+
 github.com/googleapis/gax-go/v2 v2.0.4/go.mod h1:0Wqv26UfaUD9n4G6kQubkQ+KchISgw+vpHVxEJEs9eg=
 github.com/googleapis/gax-go/v2 v2.0.5 h1:sjZBwGj9Jlw33ImPtvFviGYvseOtDM7hkSKB7+Tv3SM=
 github.com/googleapis/gax-go/v2 v2.0.5/go.mod h1:DWXyrwAJ9X0FpwwEdw+IPEYBICEFu5mhpdKc/us6bOk=
+github.com/gorilla/handlers v1.4.0 h1:XulKRWSQK5uChr4pEgSE4Tc/OcmnU9GJuSwdog/tZsA=
+github.com/gorilla/handlers v1.4.0/go.mod h1:Qkdc/uu4tH4g6mTK6auzZ766c4CA0Ng8+o/OAirnOIQ=
 github.com/gorilla/handlers v1.4.2 h1:0QniY0USkHQ1RGCLfKxeNHK9bkDHGRYGNDFBCS+YARg=
 github.com/gorilla/handlers v1.4.2/go.mod h1:Qkdc/uu4tH4g6mTK6auzZ766c4CA0Ng8+o/OAirnOIQ=
+github.com/gorilla/mux v1.7.2 h1:zoNxOV7WjqXptQOVngLmcSQgXmgk4NMz1HibBchjl/I=
+github.com/gorilla/mux v1.7.2/go.mod h1:1lud6UwP+6orDFRuTfBEV8e9/aOM/c4fVVCaMa2zaAs=
```
Hexcles: These will probably go away after a `go mod tidy`.

webapp/components/gh-login.js
```diff
@@ -0,0 +1,79 @@
+/**
+ * Copyright 2018 The WPT Dashboard Project. All rights reserved.
```
Hexcles: Nit: let's make it 2019 :)

Also nit: I don't think it's worthwhile saving four characters here. Let's call this file and the component `github-login`.

webapp/templates/index.html
```diff
@@ -1,13 +1,14 @@
...
+    {{ template "_head_common.html" }}
+    <script type="module" src="/views/wpt-app.js"></script>
+    <script type="module" src="/components/test-runs-query.js"></script>
+  </head>
+  <body>
+    <div id="content">
+      <wpt-app {{if .User}}user="{{.User}}"{{end}}
+               {{- template "_test_run_ui_query_params.html" . }}></wpt-app>
+    </div>
+  </body>
```
Hexcles: Looks like a bad merge?

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+
+import (
+	"context"
+	"crypto/rand"
+	"encoding/base64"
+	"encoding/gob"
+	"net/http"
+	"net/url"
+
+	"github.com/google/go-github/github"
```
Hexcles:
```suggestion
	"github.com/google/go-github/github/v28"
```

This is how we use a versioned module in Go. Without specifying the version explicitly, you'd get the `+incompatible` marker in `go.mod`.

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/gorilla/securecookie"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"golang.org/x/oauth2"
+	ghOAuth "golang.org/x/oauth2/github"
+	"google.golang.org/appengine"
+)
+
+// Copyright 2019 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
Hexcles: License header should be at the very top.

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+
+func init() {
+	gob.Register(map[string]interface{}{})
```
Hexcles: It's unclear to me why we'd need `gob`. Could you add some comments?

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+type User struct {
+	GitHubHandle string
+	GithuhEmail  string
+}
+
+func loginHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	aeAPI := shared.NewAppEngineAPI(ctx)
+	if !aeAPI.IsFeatureEnabled("githubLogin") {
+		http.Error(w, "Feature not implemented", http.StatusNotImplemented)
```
Hexcles: Nit
```suggestion
		http.Error(w, "Feature not enabled", http.StatusNotImplemented)
```

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+	returnURL := r.FormValue("return")
+	redirect := returnURL
+	log := shared.GetLogger(ctx)
+	if user == nil || token == nil {
+		log.Infof("Initiating a new user login.")
+		conf := getGithubOAuthConfig(ctx)
+		conf.RedirectURL = getCallbackURI(returnURL, r)
+		state, err := generateRandomState(32)
+		if err != nil {
+			log.Errorf("Failed to generate random state")
```
Hexcles:
```suggestion
			log.Errorf("Failed to generate random state: %v", err)
```

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+	ctx := shared.NewAppEngineContext(r)
+	aeAPI := shared.NewAppEngineAPI(ctx)
+	if !aeAPI.IsFeatureEnabled("githubLogin") {
+		http.Error(w, "Feature not implemented", http.StatusNotImplemented)
+		return
+	}
+
+	user, token := getUserFromCookie(r)
+	returnURL := r.FormValue("return")
+	redirect := returnURL
```
Hexcles: Nit: I'd prefer setting `redirect` to an empty string here (effectively only declaring `redirect`), as it'd be set to completely different values in the two branches below.

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+func loginHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	aeAPI := shared.NewAppEngineAPI(ctx)
+	if !aeAPI.IsFeatureEnabled("githubLogin") {
+		http.Error(w, "Feature not implemented", http.StatusNotImplemented)
+		return
+	}
+
+	user, token := getUserFromCookie(r)
+	returnURL := r.FormValue("return")
```
Hexcles: What would happen if `returnURL` is empty?

webapp/login.go
```diff
@@ -0,0 +1,255 @@
...
+		GitHubHandle: ghUser.GetLogin(),
+		GithuhEmail:  ghUser.GetEmail(),
+	}
+	setSession(ctx, user, &token.AccessToken, w)
+	log.Infof("User %s logged in", user.GitHubHandle)
+
+	ret := r.FormValue("return")
+	encodedState := r.FormValue("state")
+	stateFromCookie := getState(r)
+	if encodedState != stateFromCookie {
```
Hexcles: We should verify the state as the very first step.

In the current implementation, if the states don't match, we'd already set session cookies via `setSession`, and `http.Error` does not unset those cookies, so the user would see a 500 error page but with their cookies set -- in other words, if they refresh the homepage, they'd be logged in.

webapp/login.go
KyleJu: See https://github.com/gorilla/securecookie (search gob); this is needed in order to for securecookies to encode/decode custom types.

webapp/login.go
KyleJu: Set to "/" if empty. However, since the request is made on the wpt.fyi page, this shouldn't be empty

webapp/login.go
KyleJu: Good catch!

webapp/login.go
Hexcles: `User` is a basic type. If I'm reading the docs correctly, one would only need to register a concrete type for an interface.

KyleJu: If I don't register User I got "ERROR: Failed to set session cookie: securecookie: error - caused by: securecookie: error - caused by: gob: type not registered for interface: webapp.User".

My understanding is the concrete type is being encoded and decoded as an interface here https://github.com/gorilla/securecookie/blob/86450627d8e6b88e00ea41f228489b89d08a40de/securecookie.go#L449. Plus User is being wrapped around a struct, it should be a custom type?

Hexcles: > My understanding is the concrete type is being encoded and decoded as an interface here https://github.com/gorilla/securecookie/blob/86450627d8e6b88e00ea41f228489b89d08a40de/securecookie.go#L449.

Ahh I see. That's why `securecookie` needs to register all custom types, even though they are concrete (`gob` itself only requires you to register interfaces). Can we have a comment here?

webapp/login.go
```diff
@@ -0,0 +1,323 @@
...
+	"github.com/google/go-github/v28/github"
+	"github.com/gorilla/securecookie"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"golang.org/x/oauth2"
+	ghOAuth "golang.org/x/oauth2/github"
+	"google.golang.org/appengine"
+)
+
+func init() {
+	// Need RegisterName for securecookie encoding - for local packages, Register appends main[0-9]{5}.User
```
Hexcles: I'm confused by the latter half of this comment, which seems like an implementation detail of `gob`. Perhaps we can just leave it out. And make it clear that we need to register all custom types stored in `securecookie`.

webapp/login.go
```diff
@@ -0,0 +1,323 @@
...
+			log.Errorf("Failed to generate random state: %v", err)
+			http.Error(w, "Error creating a random state for login", http.StatusInternalServerError)
+			return
+		}
+
+		redirect = conf.AuthCodeURL(state, oauth2.AccessTypeOnline)
+		err = setState(ctx, state, w)
+		if err != nil {
+			log.Errorf("Failed to set state cookie: %s", err.Error())
+			http.Error(w, "Error setting state cookie for login", http.StatusInternalServerError)
```
Hexcles: This is missing a `return`.

webapp/routes.go
```diff
@@ -32,6 +32,11 @@ func init() {
 
 // RegisterRoutes adds the route handlers for the webapp.
 func RegisterRoutes() {
+	// Ensure login in (with GitHub)
```
Hexcles:
```suggestion
	// GitHub OAuth login
```

webapp/login.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	http.SetCookie(response, cookie)
+}
+
+func getGithubOAuthConfig(ctx context.Context) *oauth2.Config {
+	store := shared.NewAppEngineDatastore(ctx, false)
+	log := shared.GetLogger(ctx)
+	clientID, err := shared.GetSecret(store, "github-oauth-client-id")
+	if err != nil {
+		log.Errorf("Failed to get github-oauth-client-id secret: %s", err.Error())
```
Hexcles: When this happens, the returned `Config` will be unusable (without a `clientID`), but the caller couldn't tell unless they inspect the members. Since this is a fatal error, we generally want to return the error, too, so that the caller can act accordingly (fail early).

webapp/login.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	secret, err := shared.GetSecret(store, "github-oauth-client-secret")
+	if err != nil {
+		log.Errorf("Failed to get github-oauth-client-secret: %s", err.Error())
```
Hexcles: Ditto.

webapp/login.go
KyleJu: Add in the testing PR

KyleJu:
Note: need to create a Github App for staging instance domain and production domain for this to work

PR #1691, opened 2019-12-02
Implement /api/metadata CORS policy

KyleJu:
TODO: Need to change the origin to the URL of Mozilla dashboard origin

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -71,5 +71,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/screenshot/{screenshot:.*}", "api-screenshot", shared.WrapPermissiveCORS(apiScreenshotRedirectHandler))
 
 	// API endpoint for searching Metadata for the products.
-	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapRestrictiveCORS(apiMetadataHandler, []string{"https://developer.mozilla.org"}, []string{"GET", "POST", "PATCH", "OPTIONS"}))
+	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
+
+	// API endpoint for modifying Metadata.
+	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, []string{"https://developer.mozilla.org"}, []string{"PATCH"}))
 }
--- a/api/routes.go
+++ b/api/routes.go
@@ -71,5 +71,5 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/screenshot/{screenshot:.*}", "api-screenshot", shared.WrapPermissiveCORS(apiScreenshotRedirectHandler))
 
 	// API endpoint for searching Metadata for the products.
-	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
+	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapRestrictiveCORS(apiMetadataHandler, []string{"https://developer.mozilla.org"}, []string{"GET", "POST", "PATCH", "OPTIONS"}))
 }
--- a/shared/routing.go
+++ b/shared/routing.go
@@ -42,15 +42,23 @@ func WrapHSTS(h http.HandlerFunc) http.HandlerFunc {
 // WrapPermissiveCORS wraps the given handler func in one that sets an
 // all-permissive CORS header on the response.
 func WrapPermissiveCORS(h http.HandlerFunc, methods ...string) http.HandlerFunc {
-	return WrapRestrictiveCORS(h, []string{"*"}, methods)
+	opts := []handlers.CORSOption{
+		handlers.AllowedOrigins([]string{"*"}),
+		handlers.AllowedHeaders([]string{"Content-Type"}),
+	}
+	if len(methods) > 0 {
+		opts = append(opts, handlers.AllowedMethods(methods))
+	}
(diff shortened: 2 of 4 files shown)


shared/routing.go
```diff
@@ -42,8 +42,14 @@ func WrapHSTS(h http.HandlerFunc) http.HandlerFunc {
...
 func WrapPermissiveCORS(h http.HandlerFunc, methods ...string) http.HandlerFunc {
+	return WrapRestrictiveCORS(h, []string{"*"}, methods)
+}
+
+// WrapRestrictiveCORS wraps the given handler func in one that sets
+// a CORS header with specified origins and methods on the response.
+func WrapRestrictiveCORS(h http.HandlerFunc, origins []string, methods []string) http.HandlerFunc {
 	opts := []handlers.CORSOption{
-		handlers.AllowedOrigins([]string{"*"}),
+		handlers.AllowedOrigins(origins),
```
Hexcles: For the use case of triaging metadata, we also need `AllowCredentials` to allow cookies, but we probably don't want that for other requests, so I suggest renaming this method to `WrapTrustedCORS` (because it is not the opposite of `WrapPermissiveCORS` -- it's both more restrictive in terms of origins and more relaxed in terms of credentials), and making `WrapPermissiveCORS` **not** call this method (but duplicating the code instead).

webapp/web/routes_test.go
```diff
@@ -148,6 +169,13 @@ func assertHSTS(t *testing.T, path string) {
 		fmt.Sprintf("%s", res.Header["Strict-Transport-Security"]))
 }
 
+func sendHttptestRequest(req *http.Request) *httptest.ResponseRecorder {
+	rr := httptest.NewRecorder()
+	handler, _ := http.DefaultServeMux.Handler(req)
+	handler.ServeHTTP(rr, req)
+	return rr
```
Hexcles: Nit: might as well return `rr.Result()` here.

webapp/web/routes_test.go
```diff
@@ -121,6 +121,27 @@ func TestAdminCacheFlushBound(t *testing.T) {
...
+	assert.Equal(t, http.StatusOK, rr.Result().StatusCode)
+	assert.Equal(t, "Content-Type", rr.Result().Header.Get("Access-Control-Allow-Headers"))
+	assert.Equal(t, "PATCH", rr.Result().Header.Get("Access-Control-Allow-Methods"))
+	assert.Equal(t, "https://developer.mozilla.org", rr.Result().Header.Get("Access-Control-Allow-Origin"))
+
+	invalidOriginReq := httptest.NewRequest("PATCH", "/api/metadata", nil)
+	invalidOriginReq.Header.Add("Origin", "https://foo")
+	rr = sendHttptestRequest(invalidOriginReq)
+
+	assert.Equal(t, "", rr.Result().Header.Get("Access-Control-Allow-Origin"))
```
Hexcles: We want to test the following cases.

* For the read-only/permissive API
    * A simple GET request from an untrusted origin: the response should contain `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Headers: Content-Type`.
    * A preflight OPTIONS request for POST with `Access-Control-Request-Headers: Content-Type` from an untrusted origin: the response should contain `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Headers: Content-Type`.
    * A POST request with `Content-Type: application/json` from an untrusted origin: the response should contain `Access-Control-Allow-Origin: *`.
* For the mutation/restrictive API
    * A preflight OPTIONS request for PATCH from an untrusted origin: the response should **not** contain `Access-Control-Allow-Origin`.
    * A PATCH request from an untrusted origin: the response should **not** contain `Access-Control-Allow-Origin`.
    * A preflight OPTIONS request for PATCH with `Access-Control-Request-Headers: Content-Type` from a trusted origin: the response should contain `Access-Control-Allow-Origin` matching the origin, `Access-Control-Allow-Headers: Content-Type`, and `Access-Control-Allow-Credentials`.
    * A PATCH request from a trusted origin: the response should contain `Access-Control-Allow-Origin` matching the origin.

api/routes.go
```diff
@@ -71,5 +71,5 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/screenshot/{screenshot:.*}", "api-screenshot", shared.WrapPermissiveCORS(apiScreenshotRedirectHandler))
 
 	// API endpoint for searching Metadata for the products.
-	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
+	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapRestrictiveCORS(apiMetadataHandler, []string{"https://developer.mozilla.org"}, []string{"GET", "POST", "PATCH", "OPTIONS"}))
```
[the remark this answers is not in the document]
KyleJu: Thanks! I split the endpoint according to the methods it can receive. 

Regarding whether to list OPTIONS as an accepted header - I do see numerous examples [ search Access-Control-Allow-Methods on this page](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) on MDN which list OPTIONS, and gorilla does not seem to set OPTIONS either.

webapp/web/routes_test.go
KyleJu: Thanks for the clarification. I didn't understand the CORS correctly for simple requests when we chatted offline. Due to the fact that these test requests does not use browsers (it doesn't trigger preflights automatically and gorilla doesn't response to any non-OPTIONS methods), it would be hard to test GET/POST/PATCH requests, without creating mocks for loggers, httpClients.

Hexcles: Let's leave a TODO here for now.

PR #1693, opened 2019-12-03
Tweak Renovate configurations

Hexcles:
* Run npm dedupe after npm update.
* Group browser-logos manually
* Auto merge minor/linters/types updates

// Thanks for the pointer, @foolip !


Hexcles:
Also FWIW, I sent a patch to Renovate upstream to add the `@browser-logos` group as a preset: https://github.com/renovatebot/presets/pull/152

Once that's merged, we can switch to the preset instead.

PR #1702, opened 2019-12-10
Improve error handling in results-processor and fix the missing SHAs in invalid runs

Hexcles:
This PR contains two relatively independent commits, so it's best to review them separately:

1. Improvement: When there are conflicting fields in run_info, try to preserve as much information as possible by delaying raising exceptions.
2. Bug fix: The field name for the SHA in report.run_info should be `revision`; `full_revision_hash` is the field name used in Datastore on the server (due to historical reasons).

There's nothing to "preview". We rely on unit tests in results-processor, which are rather comprehensive.

### DIFF
--- a/results-processor/processor_test.py
+++ b/results-processor/processor_test.py
@@ -10,6 +10,7 @@ from werkzeug.datastructures import MultiDict
 import test_util
 import wptreport
 from processor import Processor, process_report
+from test_server import AUTH_CREDENTIALS
 
 
 class ProcessorTest(unittest.TestCase):
@@ -192,7 +193,11 @@ class MockProcessorTest(unittest.TestCase):
         mock.create_run.assert_not_called()
 
 
-class ProcessorServerTest(unittest.TestCase):
+class ProcessorDownloadServerTest(unittest.TestCase):
+    """This class tests behaviours of Processor related to downloading
+    artifacts (e.g. JSON reports) from an external server. test_server is used
+    to emulate the success and failure modes of an external server.
+    """
     def setUp(self):
         self.server, self.url = test_util.start_server(False)
 
@@ -232,3 +237,49 @@ class ProcessorServerTest(unittest.TestCase):
             # artifact_test.zip as the filename.
             path = p._download_single(self.url + '/download/attachment')
             self.assertTrue(path.endswith('.zip'))
+
+
+class ProcessorAPIServerTest(unittest.TestCase):
+    """This class tests API calls from Processor to webapp (e.g.
+    /api/results/create, /api/status). test_server is used to emulate webapp
+    and verify credentials and payloads.
+    """
+    def setUp(self):
+        self.server, self.url = test_util.start_server(True)
+
+    def tearDown(self):
+        if self.server.poll() is None:
+            self.server.kill()
+
+    def test_update_status(self):
+        with Processor() as p:
+            p._auth = AUTH_CREDENTIALS
+            p.report.update_metadata(
+                revision='21917b36553562d21c14fe086756a57cbe8a381b')
+            p.update_status(
+                run_id='12345', stage='INVALID',
+                error='Sample error', callback_url=self.url)
+        self.server.terminate()
+        _, err = self.server.communicate()
+        self.assertEqual(
+            err,
+            b'{"id": 12345, "stage": "INVALID", "error": "Sample error", '
+            b'"full_revision_hash": "21917b36553562d21c14fe086756a57cbe8a381b"}\n')
+
+    def test_create_run(self):
+        api = self.url + '/api/results/create'
+        with Processor() as p:
+            p._auth = AUTH_CREDENTIALS
+            p.report.update_metadata(
+                browser_name='chrome',
+                browser_version='70',
+                os_name='Linux',
+                revision='21917b36553562d21c14fe086756a57cbe8a381b')
+            p.create_run('12345', '', 'blade-runner', callback_url=api)
+            # p.test_run_id is set based on the response from the API, which in
+            # turn is set according to the request. Hence this verifies that we
+            # pass the run ID to the API correctly.
+            self.assertEqual(p.test_run_id, 12345)
+        self.server.terminate()
+        # This is needed to close the stdio pipes.
+        self.server.communicate()
--- a/results-processor/test_server.py
+++ b/results-processor/test_server.py
@@ -11,6 +11,9 @@ import time
 
 import flask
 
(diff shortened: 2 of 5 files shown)


Hexcles:
A bit more about the test setup: to test that results-processor calls APIs correctly, we set up a simple test server written in Flask, which runs in a separate child process. `test_util.py` is the module imported by tests, responsible for managing the server process and optionally capturing its output (e.g. echo back the request payload for verification); `test_server.py` is the server itself, which is a standalone executable script.

results-processor/processor_test.py
```diff
@@ -232,3 +233,42 @@ def test_download_content_disposition(self):
...
             path = p._download_single(self.url + '/download/attachment')
             self.assertTrue(path.endswith('.zip'))
+
+
+class ProcessorAPIServerTest(unittest.TestCase):
+    def setUp(self):
+        self.server, self.url = test_util.start_server(True)
+
+    def tearDown(self):
+        if self.server.poll() is None:
```
stephenmcgruer: What does it mean for poll() to return None, or to return a non-None value? (I.e. I'm trying to understand when we will/won't call kill())

results-processor/processor_test.py
```diff
@@ -232,3 +233,42 @@ def test_download_content_disposition(self):
             # artifact_test.zip as the filename.
             path = p._download_single(self.url + '/download/attachment')
             self.assertTrue(path.endswith('.zip'))
+
+
+class ProcessorAPIServerTest(unittest.TestCase):
```
stephenmcgruer: Nit; can you add a brief comment explaining what this class is testing?

results-processor/processor_test.py
```diff
@@ -192,7 +193,7 @@ def test_params_plumbing_duplicate(self, MockProcessor):
         mock.create_run.assert_not_called()
 
 
-class ProcessorServerTest(unittest.TestCase):
+class ProcessorDownloadServerTest(unittest.TestCase):
```
stephenmcgruer: Nit; can you add a brief comment explaining what this class is testing?

results-processor/processor_test.py
```diff
@@ -232,3 +233,42 @@ def test_download_content_disposition(self):
...
+        api = self.url + '/api/results/create'
+        with Processor() as p:
+            p._auth = AUTH
+            p.report.update_metadata(
+                browser_name='chrome',
+                browser_version='70',
+                os_name='Linux',
+                revision='21917b36553562d21c14fe086756a57cbe8a381b')
+            p.create_run('12345', '', 'blade-runner', callback_url=api)
+            self.assertEqual(p.test_run_id, 12345)
```
stephenmcgruer: It's not very clear to me what we're testing here. We call `create_run` and then check the value of a field in the Processor object we have. It may be worth a brief comment noting that if `create_run` succeeds it sets this field.

(Side-bar for my education; does this impl a Processor object handles exactly one run? Or one run at a time?)

results-processor/test_server.py
```diff
@@ -11,6 +11,9 @@
 
 import flask
 
+# Exported credentials for authenticated APIs
+AUTH = ('TEST_USERNAME', 'TEST_PASSWORD')
```
stephenmcgruer: I would suggest a more descriptive name than `AUTH`. Perhaps `AUTH_CREDENTIALS`.

results-processor/wptreport.py
```diff
@@ -152,33 +152,42 @@ def _add_chunk(self, chunk: RawWPTReport) -> None:
...
-
+            conflicts = []
             for key in chunk['run_info']:
-                update_property(
+                conflict = update_property(
                     key,
                     cast(Dict, chunk['run_info']),
                     cast(Dict, self._report['run_info']),
-                    ignore_conflict if key in IGNORED_CONFLICTS else None,
+                    lambda _1, _2: None,  # Set conflicting fields to None.
```
stephenmcgruer: No idea if this is better, but thought it was cute:

`lambda *unused: None`
or even
`lambda *_: None`

results-processor/wptreport.py
```diff
@@ -152,33 +152,42 @@ def _add_chunk(self, chunk: RawWPTReport) -> None:
...
                 )
+                # Delay raising exceptions even when conflcits are not ignored,
```
stephenmcgruer: 'conflicts'

results-processor/wptreport.py
```diff
@@ -152,33 +152,42 @@ def _add_chunk(self, chunk: RawWPTReport) -> None:
...
+                # so that we can set as many metadata as possible.
```
stephenmcgruer: Nit; would be more correct to say 'set as many pieces of metadata', or (colloquially), 'set as much metadata'. This is fine though.

results-processor/processor_test.py
Hexcles: This is a white-box test, replying on the fact that we set this field (`p.test_run_id`) based on the response from `/api/results/create`, which in turn uses the ID provided by the caller of that API (`p.create_run`). This essentially tests that `create_run` passes the run ID correctly to the server. I'll write a comment.

(Yes, one run at a time / per request.)

results-processor/processor_test.py
Hexcles: `poll` returns the `returncode` of a terminated process. It is pretty much `wait(2)` with `WNOHANG`. If the process is still running, it returns `None`.

If you notice the `tearDown` methods are different in the two new classes, it is intentional. In this class, we capture the output of the test server process using `communicate`, which would block until the process terminates, so we call `terminate` in the tests before `communicate`. (There are other ways to read stdout/err, but this is the easiest way IMO for our use case.) This however creates the possibility of never calling `terminate` (if tests fail prematurely), so we have to double check and kill the server if needed in `tearDown`.

results-processor/wptreport.py
Hexcles: I'd prefer being explicit about the number of arguments, in case I ever change how `conflict_func` is used in `update_property`.

results-processor/wptreport.py
Hexcles:
```suggestion
                # Delay raising exceptions even when conflicts are not ignored,
```

Thanks!

results-processor/wptreport.py
Hexcles:
```suggestion
                # so that we can set as much metadata as possible.
```

(I thought datum was singular and data was plural, and same would go for meta{datum,data}. Human languages are weird...)

Issue #1706, opened 2019-12-16
Collect/Publish results from Servo


jgraham:
Servo now has daily runs e.g. https://community-tc.services.mozilla.com/tasks/groups/ZeyukzlSSd6ATXzhw-Pmqw we should add servo to the frontend and populate it with that data.

Hexcles:
Pretty sure this has been fixed: https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=servo&product=firefox&aligned

And yeah I like that logo ;)

PR #1707, opened 2019-12-16
Add Servo support

jgraham:
I have literally just copied the relevant-looking bits of webkitgtk code, so this may not be right :)

This is the UI part of #1706 .


jgraham:
@jdm does servo have an official logo these days? :)

jdm:
We're actually kicking off the logo process again right now. Until then, it's https://github.com/servo/servo/blob/master/resources/servo.png.

Hexcles:
Well, I wouldn't mind ;) Pushed.

PR #1708, opened 2019-12-16
Include and display product info in pending runs

Hexcles:
Currently, a pending test run does not include product-related metadata, even though we sometimes do have the data in the JSON report. This change plugs said data into the `PendingTestRun` entity whenever it's available to make debugging much easier.

## Review

Again, the changes are split into multiple commits logically, so it'd be easier to review commit by commit.

Demo: https://pending-runs-dot-wptdashboard-staging.appspot.com/status
(Switch to "invalid runs" and you'll see an invalid run uploaded by me manually.)

### DIFF
--- a/api/receiver/create_run_test.go
+++ b/api/receiver/create_run_test.go
@@ -45,27 +45,28 @@ func TestHandleResultsCreate(t *testing.T) {
 	assert.Nil(t, err)
 	req := httptest.NewRequest("POST", "/api/results/create", strings.NewReader(string(body)))
 	req.SetBasicAuth("_processor", "secret-token")
-	testRunIn := &shared.TestRun{
-		ID:        12345,
-		TimeStart: time.Date(2018, time.June, 21, 18, 39, 54, 218000000, time.UTC),
-		TimeEnd:   time.Date(2018, time.June, 21, 20, 3, 49, 0, time.UTC),
-		Labels:    []string{"foo", "bar"},
-		ProductAtRevision: shared.ProductAtRevision{
-			Product: shared.Product{
-				BrowserName:    "firefox",
-				BrowserVersion: "59.0",
-				OSName:         "linux",
-				OSVersion:      "4.4",
-			},
-			Revision:         sha[:10],
-			FullRevisionHash: sha,
+	pAtR := shared.ProductAtRevision{
+		Product: shared.Product{
+			BrowserName:    "firefox",
+			BrowserVersion: "59.0",
+			OSName:         "linux",
+			OSVersion:      "4.4",
 		},
+		Revision:         sha[:10],
+		FullRevisionHash: sha,
+	}
+	testRunIn := &shared.TestRun{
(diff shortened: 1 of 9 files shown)


Hexcles:
Give it a few more minutes to let Travis finish the deployment. (The current staging deployment has some small UI issues.)

api/receiver/create_run_test.go
```diff
@@ -45,27 +45,28 @@ func TestHandleResultsCreate(t *testing.T) {
...
-		ProductAtRevision: shared.ProductAtRevision{
-			Product: shared.Product{
-				BrowserName:    "firefox",
-				BrowserVersion: "59.0",
-				OSName:         "linux",
-				OSVersion:      "4.4",
-			},
-			Revision:         sha[:10],
-			FullRevisionHash: sha,
+	pAtR := shared.ProductAtRevision{
```
stephenmcgruer: `pAtR`... I do not love this name. I do not have a better suggestion that is still 'golang short', but... damn, quite the jumble of letters :P

results-processor/processor.py
```diff
@@ -257,9 +257,16 @@ def update_status(self, run_id, stage, error=None, callback_url=None):
         payload = {'id': int(run_id), 'stage': stage}
         if error:
             payload['error'] = error
-        revision = self.report.run_info.get('revision')
-        if revision:
-            payload['full_revision_hash'] = revision
+        if self.report.run_info.get('revision'):
```
stephenmcgruer: There are a variety of renames here (revision --> full_revision_hash, product --> browser_name, etc). Are these commonly taken to be well known why they happen, or should they be documented at all?

webapp/components/wpt-processor.js
```diff
@@ -73,23 +74,27 @@ class WPTProcessor extends LoadingState(PolymerElement) {
         <vaadin-grid-column auto-width header="GitHub Check Run" hidden>
           <template>[[item.check_run_id]]</template>
         </vaadin-grid-column>
-        <vaadin-grid-column auto-width header="SHA">
-          <template>[[item.full_revision_hash]]</template>
+        <vaadin-grid-column auto-width header="Product">
+          <template>[[_product(item)]]</template>
+        </vaadin-grid-column>
+        <!-- Only show the prefix of a SHA, but still allow finding in page. -->
```
stephenmcgruer: Is showing the prefix enforced via the explicit width? That wasn't 100% clear to me, might suggest:

`<!-- Set explicit width to only show the prefix of a SHA, but still allow find-in-page. -->`

api/receiver/create_run_test.go
Hexcles: To be honest, me neither, just trying to drink the Go kool aid here.

results-processor/processor.py
Hexcles: This is largely historical. And it turns out we never really document the mapping apart from in the source code. Filed #1716 .

PR #1718, opened 2019-12-18
Release engineering for results-processor: gcloudignore and requirements.txt

Hexcles:
Please read the commit message and the new section in `README.md` for details.

This change should not cause any user observable changes.


results-processor/README.md
```diff
@@ -26,3 +26,28 @@ virtualenv itself).
 deactivate  # or in a different shell
 tox
 ```
+
+## Managing dependencies
+
+We pin all dependencies recursively in `requirements.txt`, and use
+[`pyup`](../.pyup.yml) to automatically upgrade them. However,
+`requirements.txt` is a flat list and does not differentiate direct
+dependencies from transitive ones, so when a direct dependency drops a sub
```
foolip: If this is a problem, how about using requirements.in as in https://github.com/foolip/ecosystem-infra-rotation/blob/master/requirements.in? Doesn't pyup support that?

Hexcles:
Pyup doesn't seem to support that, at least according to the docs I found.
Another popular solution is to list direct dependencies in setup.py (even
if we are not publishing a package) but pyup doesn't seem to support that,
either. Maybe it's time to explore Renovate for python.

PR #1720, opened 2019-12-19
Merge Tests to Oauth Implementation Branch

KyleJu:
## Description 
Encapsulates implementation details of GitHub OAuth flow into an interface for setting up mockgen. Integration tests for /login and /oauth endpoints.


webapp/login.go
```diff
@@ -32,6 +34,91 @@ type User struct {
 	GithuhEmail  string
 }
 
+type GithubOAuth interface {
```
Hexcles: Need comment.

webapp/login_test.go
```diff
@@ -0,0 +1,106 @@
...
+	mockgo.EXPECT().Datastore().AnyTimes().Return(mockStore)
+
+	handleLogin(mockgo, w, req)
+
+	resp := w.Result()
+	cookies := w.Header().Get("Set-Cookie")
+	assert.Equal(t, http.StatusTemporaryRedirect, resp.StatusCode)
+	assert.Equal(t, "https://redirect?", w.Header().Get("Location"))
+
+	assert.Equal(t, "a", cookies)
```
Hexcles: This looks like a typo?

Hexcles:
Nit; there are a few misspelled "ouath" in the code.

KyleJu:
> This generally looks good; the idea is solid. Great job on figuring out `gomock`!
> 
> I have two high-level suggestions that would make the code simpler and clearer:
> 
> 1. Extract the `GitHubOAuth` (by the way, let's use `GitHub` not `Github`) interface into a new file, preferably `shared/github_oauth.go` and the mock will be `shared/sharedtest/github_oauth_mock.go`. This will make `login.go` a lot cleaner and make your life much easier when you start implementing access control in your API endpoint (you'd also need this interface there, and that's why it should be "shared").
> 2. There's no need to have getters for `Datastore` and `Context`. The easier pattern to do this is to turn them into parameters of `handleLogin` and `handleOauth`. This will save you some plumbing.

When I remove getters for Datastore and Context, I find myself often passing in context and datastore together through parameters/return. I think I prefer having datastore and ctx as the fields of githubOAuthImp because the setup seems to be more structured. What do you think?

I have implemented other suggested changes.

webapp/login.go
```diff
@@ -31,6 +31,81 @@ type User struct {
 	GithuhEmail  string
 }
 
+type githubOAuthImp struct {
```
Hexcles: This impl should also go to `shared/github_oauth.go`.

webapp/login.go
```diff
@@ -31,6 +31,81 @@ type User struct {
...
+func (g *githubOAuthImp) GetGitHubUser(client *github.Client) (*github.User, error) {
+	ghUser, _, err := client.Users.Get(g.ctx, "")
+	if err != nil {
+		return nil, err
+	}
+
+	return ghUser, nil
+}
+
+func newGitHubOAuth(ctx context.Context) (shared.GitHubOAuth, error) {
```
Hexcles: This, too (and it'd need to be exported).

webapp/login_test.go
```diff
@@ -0,0 +1,104 @@
...
+	assert.True(t, strings.Contains(cookies, "HttpOnly"))
+	assert.True(t, strings.Contains(cookies, "SameSite=Lax"))
+}
+
+func TestHandleOauth(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+	ctx := sharedtest.NewTestContext()
+	w := httptest.NewRecorder()
+	sessionVal := "MTU3Njc4MDA0N3xpSXVpeXJyWFRFZUZBRTBDdGVISG00Q1h3YWlqM3ZmdkFORmk1Z1pWTFQtd24yQjlXM0ZsTkF5Ti1sS1ozZS1EWjZub1dSRXVyMHd5TnprZV9ZeHF0Zz09fNRqZ-8jwop8p39BpJTXlpNrRsfMeWMTH4CuRfA0QS0e"
```
Hexcles: Add a comment here to explain what is the decrypted value (plaintext) and how to generate this.

webapp/login_test.go
```diff
@@ -0,0 +1,104 @@
...
+	defer mockCtrl.Finish()
+	ctx := sharedtest.NewTestContext()
+	w := httptest.NewRecorder()
+	req := httptest.NewRequest("GET", "https://foo/login", nil)
+
+	mockStore := sharedtest.NewMockDatastore(mockCtrl)
+	mockStore.EXPECT().NewNameKey("Token", gomock.Any()).AnyTimes().Return(nil)
+	mockStore.EXPECT().Get(gomock.Any(), gomock.Any()).AnyTimes().Return(nil).Do(func(key shared.Key, dst interface{}) {
+		token, _ := dst.(*shared.Token)
+		// Need to create a 16 bytes fake secrete for securecookie.
```
Hexcles: Clarify that the length is the important bit, and the actual value doesn't matter.

webapp/login_test.go
```diff
@@ -0,0 +1,104 @@
...
+		Name:  "state",
+		Value: sessionVal,
+		Path:  "/",
+	})
+
+	mockStore := sharedtest.NewMockDatastore(mockCtrl)
+	mockStore.EXPECT().NewNameKey("Token", gomock.Any()).AnyTimes().Return(nil)
+	mockStore.EXPECT().Get(gomock.Any(), gomock.Any()).AnyTimes().Return(nil).Do(func(key shared.Key, dst interface{}) {
+		token, _ := dst.(*shared.Token)
+		// Need to create a 16 bytes fake secrete for securecookie.
```
Hexcles: Ditto.

PR #1595, opened 2019-10-31
[WIP]Triage endpoint

KyleJu:
This PR introduces /api/metadata PATCH endpoint to automatically create PR and check in wpt failures/flakiness to wpt-metadata repositories. 

This CL should be landed after TODOs are complete and Github oauth flow is created. 
TODO: Improving logging

Testing Instructions: 
curl --header "Content-Type: application/json" -d '{\"/a/b.html\":[{\"url\":\"foo\", \"product\":\"chrome\"}]}'  --request PATCH https://triage-endpoint-dot-wptdashboard-staging.appspot.com/api/metadata?

Result:
https://github.com/web-platform-tests/wpt-metadata/pull/44/files

Debug Info:
https://pantheon.corp.google.com/logs/viewer?project=wptdashboard-staging&resource=gae_app%2Fmodule_id%2Fdefault%2Fversion_id%2Ftriage-endpoint&minLogLevel=0&expandAll=false&timestamp=2019-11-12T20:54:33.321000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeStart=2019-11-12T19:04:56.066Z&interval=PT1H&logName=projects%2Fwptdashboard-staging%2Flogs%2Fstderr&logName=projects%2Fwptdashboard-staging%2Flogs%2Fappengine.googleapis.com%252Frequest_log&scrollTimestamp=2019-11-12T20:53:23.587861000Z&dateRangeUnbound=forwardInTime

PR #1724, opened 2020-01-02
Refactor login.go in Preparation for /api/metadata PATCH

KyleJu:
Refactor login.go in Preparation for /api/metadata PATCH, taken from #1595.

### DIFF
--- a/shared/github_oauth.go
+++ b/shared/github_oauth.go
@@ -19,7 +19,7 @@ import (
 )
 
 func init() {
-	// All custom types stored in securecookie are needed to be registered.
+	// All custom types stored in securecookie need to be registered.
 	gob.RegisterName("User", User{})
 }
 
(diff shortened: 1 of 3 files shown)


shared/github_oauth.go
```diff
@@ -8,12 +8,27 @@ package shared
...
 
 	"github.com/google/go-github/v28/github"
+	"github.com/gorilla/securecookie"
 	"golang.org/x/oauth2"
 	ghOAuth "golang.org/x/oauth2/github"
+	"google.golang.org/appengine"
 )
 
+func init() {
+	// All custom types stored in securecookie are needed to be registered.
```
stephenmcgruer:
```suggestion
	// All custom types stored in securecookie need to be registered.
```

shared/github_oauth.go
```diff
@@ -100,3 +115,51 @@ func NewGitHubOAuth(ctx context.Context) (GitHubOAuth, error) {
 
 	return &githubOAuthImp{ctx: ctx, conf: oauth, ds: store}, nil
 }
+
+var secureCookie *securecookie.SecureCookie
+
+// GetSecureCookie returns an instance of securecookie.
```
stephenmcgruer: This documentation just tells me a shallow version of what this method does, which I can also get just by looking at the return type. It would be more useful to have something like:

```
// GetSecureCookie returns the securecookie instance for wpt.fyi. This instance can
// be used to encode and decode cookies set by wpt.fyi, to avoid forgery attacks.
```

shared/github_oauth.go
```diff
@@ -100,3 +115,51 @@ func NewGitHubOAuth(ctx context.Context) (GitHubOAuth, error) {
 
 	return &githubOAuthImp{ctx: ctx, conf: oauth, ds: store}, nil
 }
+
+var secureCookie *securecookie.SecureCookie
```
stephenmcgruer: I know this is just moving code we already had, but do you happen to know why we store `secureCookie` in a global, rather than create it anew each time `GetSecureCookie` is called? Do we have reason to expect it to be a slow operation?

shared/github_oauth.go
```diff
@@ -100,3 +115,51 @@ func NewGitHubOAuth(ctx context.Context) (GitHubOAuth, error) {
...
+			log.Errorf("Failed to get secure-cookie-blockkey secret: %s", err.Error())
+			return nil, err
+		}
+
+		secureCookie = securecookie.New([]byte(hashKey), []byte(blockKey))
+	}
+	return secureCookie, nil
+}
+
+// GetUserFromCookie returns User and GitHub Oauth token from a session cookie.
```
stephenmcgruer: Nit:

```
// GetUserFromCookie extracts the User and GitHub OAuth token from a request's
// session cookie, if it exists. If the cookie does not exist or cannot be decoded, nil
// is returned for both.
```

shared/github_oauth.go
KyleJu: Good question. I think the original purpose is to avoid the datastore lookup repeatedly from `GetSecret` methods but turns out that they are memcached after the initial lookup. Having a global secureCookie will not speed things up.

PR #1725, opened 2020-01-02
Create Triage Handler Helper Class

KyleJu:
Create Triage Handler Helper Class, taken from #1595 

This PR creates a helper class that merges wpt metadata into the existing ones in wpt-metadata repositories and creates a PR using the GitHub client provided.

Tested in https://github.com/web-platform-tests/wpt-metadata/pull/44/files


api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (tm triageMetadata) mergeToGithub(triagedMetadataMap map[string][]byte, log shared.Logger) error {
+	ref, err := tm.getRef()
+	if err != nil {
+		log.Errorf("Unable to get/create the commit reference: %s\n", err)
+		return err
+	}
+
+	if ref == nil {
+		log.Errorf("No error returned but the reference is nil")
```
stephenmcgruer: Does this need a `\n`?

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	pr, _, err := client.PullRequests.Create(tm.ctx, prRepoOwner, prRepo, newPR)
+	if err != nil {
+		return err
+	}
+
+	log.Infof("PR created: %s\n", pr.GetHTMLURL())
+	return nil
+}
+
+func (tm triageMetadata) mergeToGithub(triagedMetadataMap map[string][]byte, log shared.Logger) error {
```
stephenmcgruer: This doesn't actually merge the change to GitHub, does it? It stops after creating the PR.

Actually, this has me confused a bit. As noted above, we use the same `commitBranch` for all calls to this code. Doesn't that mean calls to `triage(..)` will overlap? E.g.:

`code initializes`
- `commitBranch` set to, say, auto-triage-branch1234

1st call to `triage(..)`
- creates refs/heads/auto-triage-branch1234, pointing at master SHA `s1`
- creates a tree for the changes
- creates a new commit, `s2`, with parent `s1`, and updates refs/heads/auto-triage-branch1234 to point at `s2`
- creates a github PR from refs/heads/auto-triage-branch1234 to master

2nd call to `triage(..)`
- grabs refs/heads/auto-triage-branch1234, which points at `s2`
- creates a tree for the changes
- creates a new commit, `s3`, with parent `s2`, and updates refs/heads/auto-triage-branch1234 to point at `s3` <-- doesn't this change the *existing* PR from above?
- creates a github PR from refs/heads/auto-triage-branch1234 to master <-- but this new PR will contain *two* commits not in master?

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+
+	return nil
+}
+
+// Add Metadata into the existing Metadata YML files and only return modified files.
+func (tm triageMetadata) addToFiles(metadata shared.MetadataResults, filesMap map[string]shared.Metadata) map[string][]byte {
+	res := make(map[string][]byte)
+	for test, links := range metadata {
+		folderName, _ := shared.SplitWPTTestPath(test)
+		appendTestName(test, metadata)
```
stephenmcgruer: It's very unclear to me what this function (appendTestName) does or why it is being called here. I tried reading the method definition, but it's digging deep into the MetadataResults which makes it hard to understand.

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+		log.Errorf("Error while creating the pull request: %s", err)
+		return err
+	}
+
+	return nil
+}
+
+// Add Metadata into the existing Metadata YML files and only return modified files.
+func (tm triageMetadata) addToFiles(metadata shared.MetadataResults, filesMap map[string]shared.Metadata) map[string][]byte {
+	res := make(map[string][]byte)
```
stephenmcgruer: `res` is not used until the second loop, and declaring it here confused me into thinking it was relevant for the first loop. Please move down.

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	for test, links := range metadata {
```
stephenmcgruer: Once I understood that this loop is *not* for filling `res`, I realized it would be useful to have a comment explaining what it *is* for. The overall purpose appears to be updating filesMap with the new information in metadata, so that we can then (afterwards) just walk the new metadata again and grab the YAML bytes from filesMap?

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	"github.com/google/go-github/v28/github"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"gopkg.in/yaml.v2"
+)
+
+var (
+	sourceOwner   = "web-platform-tests"
+	sourceRepo    = "wpt-metadata"
+	commitMessage = "Commit New Metadata"
+	commitBranch  = "auto-triage-branch" + generateRandomInt()
```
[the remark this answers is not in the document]
KyleJu: Good catch! The purpose of `generateRandomInt()` is to help create a new checkout from master branch every time. 

It is a bug because I didn't consider the lifecycle of a package-level global variables. Global variables should be generally avoided due to concurrent issues, unless its ready-only. I can make all of them constant variables besides commitBranch, but the neater way is probably to define a struct, turn them into local variables and pass it along.

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	baseBranch    = "master"
+	prRepoOwner   = sourceOwner
+	prRepo        = sourceRepo
+	prBranch      = baseBranch
+	prSubject     = "Automatically Triage New Metadata"
+	prDescription = "PR for metadata triaged through /api/metadata/triage endpoint. See <insert a doc> for more information about how to use this service."
```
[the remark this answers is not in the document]
KyleJu: I will leave TODO at this line and write a doc when landing the triage metadata handler (next PR)

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+
+func (tm triageMetadata) getRef() (ref *github.Reference, err error) {
+	client := tm.githubClient
+	if ref, _, err = client.Git.GetRef(tm.ctx, sourceOwner, sourceRepo, "refs/heads/"+commitBranch); err == nil {
+		return ref, nil
+	}
+
+	// We consider that an error means the branch has not been found and needs to
+	// be created.
+	if commitBranch == baseBranch {
```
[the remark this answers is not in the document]
KyleJu: This check is no longer useful. Removed

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	if baseRef, _, err = client.Git.GetRef(tm.ctx, sourceOwner, sourceRepo, "refs/heads/"+baseBranch); err != nil {
+		return nil, err
+	}
+	newRef := &github.Reference{Ref: github.String("refs/heads/" + commitBranch), Object: &github.GitObject{SHA: baseRef.Object.SHA}}
+	ref, _, err = client.Git.CreateRef(tm.ctx, sourceOwner, sourceRepo, newRef)
+	return ref, err
+}
+
+// getTree generates the tree to commit based on the given files and the commit
+// of the ref you got in getRef.
```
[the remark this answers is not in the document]
KyleJu: Updated. The API doesn't require ref to work but it wouldn't work without ref in this context. IIUC, without ref, the tree created will only include the changes in triageMetadataMap and delete the rest of content in the repo; see https://developer.github.com/v3/git/trees/#create-a-tree.

I would prefer passing `getRef` instead because (1) ref is shared by `pushCommit` and (2)each method represents the flow of creating a Git Objects - create blobs using files, create a branch out of master, create a tree using blobs and the branch ref, create a commit using tree and the branch ref.

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+func (tm triageMetadata) getTree(ref *github.Reference, triagedMetadataMap map[string][]byte) (tree *github.Tree, err error) {
+	client := tm.githubClient
+	// Create a tree with what to commit.
+	entries := []github.TreeEntry{}
+
+	for folderPath, content := range triagedMetadataMap {
+		dest := shared.GetMetadataFilePath(folderPath)
+		entries = append(entries, github.TreeEntry{Path: github.String(dest), Type: github.String("blob"), Content: github.String(string(content)), Mode: github.String("100644")})
+	}
```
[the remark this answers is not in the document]
KyleJu: A valid super nit

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+
+// createCommit creates the commit in the given reference using the given tree.
+func (tm triageMetadata) pushCommit(ref *github.Reference, tree *github.Tree) (err error) {
+	client := tm.githubClient
+	// Get the parent commit to attach the commit to.
+	parent, _, err := client.Repositories.GetCommit(tm.ctx, sourceOwner, sourceRepo, *ref.Object.SHA)
+	if err != nil {
+		return err
+	}
+	// This is not always populated, but is needed.
```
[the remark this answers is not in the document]
KyleJu: IIUC, when there is not commit in the repo at all. It means the repo is an empty one.

api/triage_metadata.go
```diff
@@ -0,0 +1,231 @@
...
+	// Create the commit using the tree.
+	date := time.Now()
+	author := &github.CommitAuthor{Date: &date, Name: &tm.authorName, Email: &tm.authorEmail}
+	commit := &github.Commit{Author: author, Message: &commitMessage, Tree: tree, Parents: []github.Commit{*parent.Commit}}
+	newCommit, _, err := client.Git.CreateCommit(tm.ctx, sourceOwner, sourceRepo, commit)
+	if err != nil {
+		return err
+	}
+
+	// Attach the commit to the master branch.
```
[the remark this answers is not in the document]
KyleJu: Correct. Removed!

api/triage_metadata.go
KyleJu: Related to the comment above. commitBranch is meant to be a fresh checkout of master every time so it doesn't modify any existing PR. In case of conflicts, we have to resolve merge conflicts ourselves.

api/triage_metadata.go
KyleJu: Removed all \n. Good catch

api/triage_metadata.go
KyleJu: Added!

api/triage_metadata.go
KyleJu: It is due to the format of metadata being accepted by /api/metadata/triage. Its format follows https://github.com/web-platform-tests/wpt.fyi/tree/master/api#apimetadata, where `metadata stored in the wpt-metadata repository, with the (normally file-sharded) data all flattened into a JSON object which is keyed by test name.` So all the `TestPath` is missing in `metadata` because its a flattened structure.

shared/metadata_test.go
```diff
@@ -221,3 +247,46 @@ func TestConstructMetadataResponse_WithEmptyProductSpec(t *testing.T) {
...
+	assert.Equal(t, "foo/bar", folderPath)
+	assert.Equal(t, "foo1", testPath)
+
+	folderPath, testPath = SplitWPTTestPath("/foo/bar")
+	assert.Equal(t, "foo", folderPath)
+	assert.Equal(t, "bar", testPath)
+}
+
+func TestGetMetadataFilePath(t *testing.T) {
+	actual := GetMetadataFilePath("foo")
```
stephenmcgruer: Nit; also worth testing an empty input?

shared/product_spec.go
```diff
@@ -164,3 +164,8 @@ func (p *ProductSpec) UnmarshalYAML(unmarshal func(interface{}) error) (err erro
 	*p, err = ParseProductSpec(s)
 	return err
 }
+
+// MarshalYAML marshals ProductSpec in string.
```
stephenmcgruer:
```suggestion
// MarshalYAML serializes a ProductSpec into a YAML string.
```

shared/statuses.go
```diff
@@ -185,3 +185,8 @@ func (s *TestStatus) UnmarshalYAML(unmarshal func(interface{}) error) (err error
 	}
 	return err
 }
+
+// MarshalYAML marshals TestStatus in string.
```
stephenmcgruer:
```suggestion
// MarshalYAML serializes TestStatus into a YAML string.
```

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+	"math/rand"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/google/go-github/v28/github"
+	"gopkg.in/yaml.v2"
+)
+
+// TriageMetadataInterface encapsulates triage() method for testing purpose.
```
stephenmcgruer:
```suggestion
// TriageMetadataInterface encapsulates the Triage() method for testing.
```

(or 'for testing purposes')

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+type TriageMetadataInterface interface {
+	Triage(metadata MetadataResults) error
+}
+
+type triageMetadata struct {
```
stephenmcgruer: Would be nice to have a little documentation on triageMetadata, metadataGithub, and wptmetadataGitHubInfo

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+	prRepoOwner   string
+	prRepo        string
+	prBranch      string
+	prSubject     string
+	prDescription string
+}
+
+func getNewCommitBranchName(ctx context.Context, client *github.Client, sourceOwner string, sourceRepo string) string {
+	commitBranch := "auto-triage-branch" + generateRandomInt()
+	// If the commitBranch name already exist, genereate a new one. We consider that an error means the branch has not been found and needs to
```
stephenmcgruer:
```suggestion
	// If the commitBranch name already exists, generate a new one. We consider an error to mean that the branch has not been found and needs to
```

Also nit; can you split this comment a little earlier? I know this is golang and so line length is... err... optional, but since we're already splitting for 'be created' here maybe we could split further back and get a more readable split. (E.g. split somewhere around 'branch')

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+	// be created.
+	for ref, _, err := client.Git.GetRef(ctx, sourceOwner, sourceRepo, "refs/heads/"+commitBranch); err == nil && ref != nil; {
```
stephenmcgruer: Note that this could spin forever if we are getting real errors back from `client.Git.GetRef` (what if we hit an API limit?). I would suggest defensively only trying this loop a few times and bailing if we hit the limit. Real conflicts should be pretty rare (10,000 numbers to choose uniformly from, and in theory only a handful of PR branches exist at any time).

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+		commitBranch = "auto-triage-branch" + generateRandomInt()
+	}
+
+	return commitBranch
+}
+
+func getWptmetadataGitHubInfo(ctx context.Context, client *github.Client) wptmetadataGitHubInfo {
```
stephenmcgruer: I'll leave this up to you, but a bunch of this is duplicated constants (which I don't see changing) for every query that comes in. I would personally have gone for:

```
const (
    sourceOwner = "web-platform-tests"
    sourceRepo = "wpt-metadata"
    baseBranch = "master"
    commitMessage = "Commit New Metadata"
    prSubject = "Automatically Triage New Metadata"
    prDescription = "PR for metadata triaged through /api/metadata/triage endpoint. See <insert a doc> for more information about how to use this service."
)
```

And then `wptmetadataGitHubInfo` wouldn't exist, `metadataGithub` would just have a `commitBranch` member.

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+		if err != nil {
+			logger.Errorf("Error from marshal %s: %s", folderName, err.Error())
+			continue
+		}
+		res[folderName] = metadataBytes
+	}
+	return res
+}
+
+// appendTestName populate TestPath name of metadata from test.
```
stephenmcgruer:
```suggestion
// appendTestName populates the TestPath name of metadata from test.
```

That said, this really doesn't tell me what the purpose of the method is, it just summarizes what the code does. Something like your reply to me in the comment is more useful:

```
// The metadata triage API end-point accepts metadata in a flattened JSON structure. To re-create
// the file-sharded structure of the wpt-metadata repository, we have to re-fill in the TestPath field
// for each new test added.
```

shared/triage_metadata.go
```diff
@@ -0,0 +1,257 @@
...
+	_, testName := SplitWPTTestPath(test)
+	for linkIndex, link := range links {
+		for resultIndex := range link.Results {
+			metadata[test][linkIndex].Results[resultIndex].TestPath = testName
+		}
+	}
+}
+
+func generateRandomInt() string {
+	rand.Seed(time.Now().UnixNano())
```
stephenmcgruer: No need to seed `rand` every time you use it; this can be in an init method.

Also note that this is not cryptographically secure of course, but that isn't important here.

shared/triage_metadata_test.go
```diff
@@ -0,0 +1,227 @@
...
+	assert.Equal(t, "foo", actual.Links[0].URL)
+	assert.Equal(t, 1, len(actual.Links[0].Results))
+	assert.Equal(t, "b.html", actual.Links[0].Results[0].TestPath)
+	assert.Equal(t, "firefox", actual.Links[1].Product.BrowserName)
+	assert.Equal(t, "https://bug.com/item", actual.Links[1].URL)
+	assert.Equal(t, "chrome", actual.Links[2].Product.BrowserName)
+	assert.Equal(t, "foo1", actual.Links[2].URL)
+}
+
+func TestGenerateRandomInt(t *testing.T) {
```
stephenmcgruer: Uh.... this will fail, occasionally. Actually, it's very unlikely to fail, but it might.

It's not worth testing random number generation. What you could test is `getNewCommitBranchName`, with appropriate mocking of github.

shared/triage_metadata.go
KyleJu: By looking at the [client.Git.GetRef method](https://github.com/google/go-github/blob/master/github/git_refs.go#L59), I suspect that there will never be a case where an error occurs but (err == nil && ref != nil) is true. But it certainly doesn't hurt to code defensively here. 

Since the GitHub API rate limit is 5000/hour and it is rare to hit a real conflict, I will only allow this method to loop 10

shared/triage_metadata.go
KyleJu: Right I would have made the same choice, besides there is one annoying hiccup - constant variables are not addressable, [see here](https://play.golang.org/) and [explanation](https://stackoverflow.com/questions/35146286/find-address-of-constant-in-go). I have to create extra variables to indirectly get the address

shared/metadata_test.go
KyleJu: Good catch!

shared/triage_metadata.go
stephenmcgruer: Wow, TIL. And I see that a bunch of the GitHub APIs require pointers (apparently to make distinguishing 'unset' easy: https://godoc.org/github.com/google/go-github/github#hdr-Creating_and_Updating_Resources)

Thanks for the explanation!

KyleJu: np. We also use it in the metadata.go to distinguish [unset fields](https://github.com/web-platform-tests/wpt.fyi/blob/master/shared/metadata.go#L47)

PR #1745, opened 2020-01-15
Support Display for Metadata Test Name with a Wildcard

KyleJu:
Feature request from https://github.com/web-platform-tests/wpt-metadata/issues/49. Displaying test names with wildcards.

Test name with a wildcard will be displayed in all tests of the current directory and its subdirectories.

### DIFF
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -77,7 +77,11 @@ class WPTMetadataNode extends PolymerElement {
 
   computeTestHref(path, metadataNode) {
     const currentUrl = window.location.href;
-    return currentUrl.replace(path, metadataNode.test);
+    let testname = metadataNode.test;
+    if (testname.endsWith('/*')) {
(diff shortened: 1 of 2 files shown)


webapp/components/wpt-metadata.js
```diff
@@ -221,6 +225,13 @@ class WPTMetadata extends LoadingState(PolymerElement) {
     this.shadowRoot.querySelector('#metadata-toggle').hidden = true;
     this.shadowRoot.querySelector('#metadata-collapsible').opened = true;
   }
+
+  checkPath(testname, path) {
+    if (testname.endsWith('/*')) {
+      return path.startsWith(testname.substring(0, testname.length - 1));
```
stephenmcgruer: Does the path always end with a trailing slash?

KyleJu: Good catch! 

I don't think it ends with a trailing slash at all. I use `testname.length -1` to avoid the case when path is /foobar but testname is /foo/*. Added one more check

PR #1385, opened 2019-07-11
[WIP] Create a UI component for Amend metadata

KyleJu:
## Description
Add UI component for amending metadata.

Create a toast that shows the instructions to triage a failing test

## Review Information
1. Go to staging instance and any sub-folder.
2. Click on cells that belong to a failing test

## TODO
Plumb the test failure from somehwere

PR #1746, opened 2020-01-16
Implement Triage Metadata Single Select UI

KyleJu:
This PR builds on top of https://github.com/web-platform-tests/wpt.fyi/pull/1385 and integrates with the new Triage Metadata API.

### Review Info
1. Check Show Triage Metadata UI on the wpt.fyi result page in /flags
2. Log in on wpt.fyi

### Follow-up PR
1. Add support for triaging subtest and folders(wildcards)
2. Check if a test has been triaged, except subtests
3. Multiple-select UI


KyleJu:
@stephenmcgruer let me think of a way to make login work without breaking staging.wpt.fyi but ready for review

stephenmcgruer:
Its claiming the deployment is out of date?

image

KyleJu:
> Its claiming the deployment is out of date?
> 
> image

The deployment still works. I did a rebase and force push. That could be why?

stephenmcgruer:
Firstly, major apologies for not getting to this PR until now. In the future, please ping me earlier (and repeatedly) if I drop the ball like this.

I'm not used to reviewing UX changes, but I'm starting with a review of using the feature on the deployed environment. I reckon it doesn't make a lot of sense to review the code until we have that behavior agreed upon.

The below are my major points from examining the UX. The earlier ones are more confident requests, then it heads towards more opinion things to discuss :).

1. Users do not care what 'metadata' is; the pop-up title should probably be 'Triage Test' or 'Triage Failing Test' instead.
1. It is not clear what the text box is meant to contain. Can we use a label or placeholder attribute (they have slightly different behavior) with something like 'Bug URL'.
1. The URL text box does not auto-focus when the triage pop-up appears; doing so would allow the user to just start typing.
1. When I submitted a test (by filling in the URL and clicking 'triage'), the dialog did not dismiss, even though I am now done with it. This also meant I wasn't sure if the triage had been submitted or not.
1. It feels like it would be good (along with the auto-focusing above) to map pressing the enter key when the textbox is focused to 'submit triage'. This may result in accidental submissions, but an interface of 'click on test, ctrl+v, enter' would be really nice.
1. The submitted pop-up disappears too fast.
1. The submitted pop-up doesn't actually say what it is, just shows a URL. Would be nice to have something like `Created triage [pull request](url here)`.
1. The submitted pop-up is actually hard to notice, being way down in the corner. I'm not sure this would be a correct change, but maybe experiment with having the submitted pop-up also be a central overlay? Maybe clicking the triage button would gray it out but leave the dialog up until we submit, at which point the dialog would change to show the URL and just have a dismiss button?
1. The UX does not appear to be available for folders (which would create a wildcard entry). Not sure if we want to do this outside of the 'multi-select triage mode' however.
1. It is still very unclear that the UX is available, and may be surprising when it does appear (especially given that the pointer is still a cursor selector but clicking causes something to happen!)

KyleJu:
> Firstly, major apologies for not getting to this PR until now. In the future, please ping me earlier (and repeatedly) if I drop the ball like this.
> 
> I'm not used to reviewing UX changes, but I'm starting with a review of using the feature on the deployed environment. I reckon it doesn't make a lot of sense to review the code until we have that behavior agreed upon.
> 
> The below are my major points from examining the UX. The earlier ones are more confident requests, then it heads towards more opinion things to discuss :).
> 
> 1. Users do not care what 'metadata' is; the pop-up title should probably be 'Triage Test' or 'Triage Failing Test' instead.
> 2. It is not clear what the text box is meant to contain. Can we use a label or placeholder attribute (they have slightly different behavior) with something like 'Bug URL'.
> 3. The URL text box does not auto-focus when the triage pop-up appears; doing so would allow the user to just start typing.
> 4. When I submitted a test (by filling in the URL and clicking 'triage'), the dialog did not dismiss, even though I am now done with it. This also meant I wasn't sure if the triage had been submitted or not.
> 5. It feels like it would be good (along with the auto-focusing above) to map pressing the enter key when the textbox is focused to 'submit triage'. This may result in accidental submissions, but an interface of 'click on test, ctrl+v, enter' would be really nice.
> 6. The submitted pop-up disappears too fast.
> 7. The submitted pop-up doesn't actually say what it is, just shows a URL. Would be nice to have something like `Created triage [pull request](url here)`.
> 8. The submitted pop-up is actually hard to notice, being way down in the corner. I'm not sure this would be a correct change, but maybe experiment with having the submitted pop-up also be a central overlay? Maybe clicking the triage button would gray it out but leave the dialog up until we submit, at which point the dialog would change to show the URL and just have a dismiss button?
> 9. The UX does not appear to be available for folders (which would create a wildcard entry). Not sure if we want to do this outside of the 'multi-select triage mode' however.
> 10. It is still very unclear that the UX is available, and may be surprising when it does appear (especially given that the pointer is still a cursor selector but clicking causes something to happen!)

Let me take a stab at 1-7 & 9. I agree with 10, and we could possibly change cursor selector from the pointer to a hand when hovering, indicting that this cell can be selected.

For number 8, I thought it would create a more consistent experience if the pop-up shows up on the  bottom left because it's also the behaviour elsewhere in wpt.fyi, e.g. [comparing runs](https://wpt.fyi/runs?label=master&from=2020-02-19T00%3A00) or when runs are being cached. I can certainly bring it up in the meeting with Chrome UX.

KyleJu:
@stephenmcgruer Implemented all the feedback except number 8 and 9. For 9 I will leave it to a separate CL, along with triaging subtest.

stephenmcgruer:
Thanks for addressing those! Love the changes, especially the UI for triaging something now (the highlight is really nice).

Overall the UX experience lgtm. A few nits below. I'm OOO soon, so if its ok I'm going to let @Hexcles handle the code review (and they may have UX opinions of their own); sorry about that.

1. Can we add a tooltip when the user hovers over a x/y test that can be triaged, just to explain what clicking will do?
1. When the triage fails, it just says 'Fail to triage failing tests'. This is both grammatically incorrect and not very helpful; I think we need to try to give the user some sort of understanding of what failed (e.g. authentication error, etc).
1. The box still does not dismiss on pressing 'enter' (the url clears and thats it).
1. The pop-up with the triage result still disappears too fast.
1. When I clicked 'triage', the box disappears but I have no idea if anything happened. Perhaps adding a pop-up of 'triaging...' that stays until we get the result to show?

KyleJu:
> Thanks for addressing those! Love the changes, especially the UI for triaging something now (the highlight is really nice).
> 
> Overall the UX experience lgtm. A few nits below. I'm OOO soon, so if its ok I'm going to let @Hexcles handle the code review (and they may have UX opinions of their own); sorry about that.
> 
> 1. Can we add a tooltip when the user hovers over a x/y test that can be triaged, just to explain what clicking will do?
> 2. When the triage fails, it just says 'Fail to triage failing tests'. This is both grammatically incorrect and not very helpful; I think we need to try to give the user some sort of understanding of what failed (e.g. authentication error, etc).
> 3. The box still does not dismiss on pressing 'enter' (the url clears and thats it).
> 4. The pop-up with the triage result still disappears too fast.
> 5. When I clicked 'triage', the box disappears but I have no idea if anything happened. Perhaps adding a pop-up of 'triaging...' that stays until we get the result to show?

Due to the sheer number of failing tests, I feel like having a tooltip doesn't deliver the best experience. Plus it is also explained in the multi-select mode.

@Hexcles  for 5, I wonder if we have a similar mechanism in [LadingState](https://github.com/web-platform-tests/wpt.fyi/blob/0beb408a4951af7d6af9ac9a30d0b9b648c259a6/webapp/components/loading-state.js#L8)?

webapp/views/wpt-results.js
```diff
@@ -90,6 +91,14 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
...
         white-space: nowrap;
         color: black;
+        opacity: 1;
+      }
+      td.triage {
+        cursor: pointer;
+        transition: 0.2s;
+      }
+      td.triage:hover {
+        opacity: 0.7;
```
Hexcles:
```suggestion
        opacity: 0.7;
        box-shadow: 5px 5px 5px;
```
Nit: just to make this a bit more prominent.

webapp/views/wpt-results.js
```diff
@@ -90,6 +91,14 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
       td.numbers {
         white-space: nowrap;
         color: black;
+        opacity: 1;
```
Hexcles:
```suggestion
```
Nit: this is a no-op.

webapp/views/wpt-results.js
```diff
@@ -297,6 +317,7 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
         </section>
       </template>
     </template>
+    <wpt-amend-metadata path="[[ path ]]" products="[[products]]" test="[[node.path]]" product-index="[[i]]"></wpt-amend-metadata>
```
Hexcles: Are `node` and `i` actually defined here?

webapp/views/wpt-results.js
```diff
@@ -477,6 +498,14 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
 
   constructor() {
     super();
+    this.openAmendMetadata = (i, node) => {
+      return () => {
+        const amend = this.shadowRoot.querySelector('wpt-amend-metadata');
```
Hexcles: Nit: if you assign an `id` (e.g. `amend`) to `wpt-amend-metadata`, you could access it via `this.$.amend` which also saves the cost of querying the DOM every time.

webapp/views/wpt-results.js
```diff
@@ -823,10 +852,16 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
     }
   }
 
-  platformID({browser_name, browser_version, os_name, os_version}) {
+  platformID({ browser_name, browser_version, os_name, os_version }) {
     return `${browser_name}-${browser_version}-${os_name}-${os_version}`;
   }
 
+  canAmendMetadata(node, index, testRun) {
```
Hexcles: We probably want a test for this.

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    const toast = this.shadowRoot.querySelector('#showPR');
+    const prLink = this.shadowRoot.querySelector('#prLink');
+    this.load(
+      window.fetch(url, fetchOpts)
+        .then(r => {
+          if (!r.ok || r.status !== 200) {
+            this.pr = '';
+            prLink.text = r.status + ': ' + r.statusText;
```
Hexcles: Instead of `statusText` (which is empty by default), we want to show the actual body of the response (`r.text()` I think) instead. Try triaging without logging in, users should see "not logged in" in the toast (which is the error message from the API).

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+      <paper-dialog>
+        <h3>Triage Failing Test</h3>
+        <div class="metadataEntry">
+          <img class="browser" src="[[displayLogo(product)]]">
+          &nbsp; >> [[test]] : &nbsp;
+          <paper-input label="Bug URL" value="{{url}}" autofocus></paper-input>
+        </div>
+        <div class="buttons">
+        <paper-button onclick="[[close]]">Dismiss</paper-button>
+        <paper-button onclick="[[triage]]" dialog-confirm>Triage</paper-button>
```
Hexcles:
```suggestion
          <paper-button onclick="[[close]]">Dismiss</paper-button>
          <paper-button onclick="[[triage]]" dialog-confirm>Triage</paper-button>
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+    return productVal[productIndex];
+  }
+
+  getTriagedMetadataMap(product, test) {
+    var link = {};
+    link[test] = [{ 'url': this.url, 'product': product }];
+    return link;
+  }
+
+  displayLogo(product) {
```
Hexcles: Pretty sure this is already implemented somewhere else (`ProductInfo` mixin?).

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+
+  constructor() {
+    super();
+    this.triage = this.handleTriage.bind(this);
+    this.close = this.close.bind(this);
+    this.enter = this.triageOnEnter.bind(this);
+  }
+
+  get dialog() {
+    return this.shadowRoot.querySelector('paper-dialog');
```
Hexcles: Same here, you could assign an ID to `paper-dialog` and access it via `$`.

webapp/components/wpt-amend-metadata.js
KyleJu: It is in the (display-logo.js)[https://github.com/web-platform-tests/wpt.fyi/blob/982ffdb3ba3d901f48513f89d7e323bf5973ad32/webapp/components/display-logo.js#L103]. What if we move displayLogo to ProductInfo?

Hexcles: Yeah let's move it into `ProductInfo`.

webapp/views/wpt-results.js
KyleJu: This method is incomplete because we need to add the ability to triage directory (wildcard) and subtest. Let me add a TODO and come back to this in the following PR

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+          <img class="browser" src="[[getLogo(product)]]">
+          &nbsp; >> [[test]] : &nbsp;
+          <paper-input label="Bug URL" value="{{url}}" autofocus></paper-input>
+        </div>
+        <div class="buttons">
+          <paper-button onclick="[[close]]">Dismiss</paper-button>
+          <paper-button onclick="[[triage]]" dialog-confirm>Triage</paper-button>
+        </div>
+      </paper-dialog>
+      <paper-toast id="showPR" duration="10000"><a id="prLink" class="link" target="_blank" href="[[pr]]"></a></paper-toast>
```
Hexcles: Nit: for element IDs, we tend to use dashes, e.g. `show-pr`, `pr-link`.

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        return text;
+      })
+      .then(text => {
+        this.pr = text;
+        prLink.text = 'Created traige ' + text;
+        toast.open();
+      }).catch(error => {
+        prLink.text = error.message;
```
Hexcles: Putting error messages inside `<a>` is a bit weird. I'd suggest adding an empty `<span>` within `<paper-toast>`.

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,164 @@
...
+    return productVal[productIndex];
+  }
+
+  getTriagedMetadataMap(product, test) {
+    var link = {};
+    link[test] = [{ 'url': this.url, 'product': product }];
+    return link;
+  }
+
+  getLogo(product) {
```
Hexcles: I don't think this is necessary. Could you try calling `displayLogo` directly with only one argument in the template?

webapp/components/wpt-metadata.js
```diff
@@ -85,11 +86,8 @@ class WPTMetadataNode extends PolymerElement {
     return currentUrl.replace(path, testname);
   }
 
-  displayLogo(product) {
-    if (!product) {
-      return;
-    }
-    return `/static/${product}_64x64.png`;
+  getLogo(product) {
```
Hexcles: Same here.

webapp/components/wpt-amend-metadata.js
```diff
@@ -0,0 +1,161 @@
...
+          <img class="browser" src="[[displayLogo(product)]]">
+          &nbsp; >> [[test]] : &nbsp;
+          <paper-input label="Bug URL" value="{{url}}" autofocus></paper-input>
+        </div>
+        <div class="buttons">
+          <paper-button onclick="[[close]]">Dismiss</paper-button>
+          <paper-button onclick="[[triage]]" dialog-confirm>Triage</paper-button>
+        </div>
+      </paper-dialog>
+      <paper-toast id="show-pr" duration="10000"><span id="msg"></span><a id="prLink" class="link" target="_blank"]"></a></paper-toast>
```
Hexcles: `]`? Seems like a typo here?`

This is really a static template and we can directly put data binding in it instead of using the DOM API to overwrite its content.

```suggestion
      <paper-toast id="show-pr" duration="10000"><span>[[errorMessage]]</span><a id="pr-link" class="link" target="_blank" href="[[prLink]]">[[prText]]</a></paper-toast>
```

And then you can have three new properties `errorMessage`, `prLink` & `prText`. Data binding is one of the top reasons people use Polymer instead of rolling their own custom elements. This makes updating logic and data flow super clear.

PR #1749, opened 2020-01-20
Update metadata triage CORS rule for gecko dashboard

jgraham:
This is currently hosted on my github.io but may move in the future.


stephenmcgruer:
Can you update `./webapp/web/routes_test.go`? It currently checks (in a few places) the old URL and so fails.

PR #1752, opened 2020-01-21
Fix a securecookie test (TestHandleOauth)

Hexcles:
See each individual commit message. Only the last commit fixes the actual bug, whereas the first two help investigating the cause.

With the first two commits, I was able to see the underlying error: "failed to decode state cookie: securecookie: expired timestamp", which led me to the root cause.

### DIFF
--- a/webapp/login.go
+++ b/webapp/login.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
+	"fmt"
 	"net/http"
 	"net/url"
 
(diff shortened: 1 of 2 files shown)


webapp/login.go
```diff
@@ -47,16 +48,18 @@ func handleLogin(g shared.GitHubOAuth, w http.ResponseWriter, r *http.Request) {
...
+			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
 		redirect = g.GetAuthCodeURL(state, oauth2.AccessTypeOnline)
 		err = setState(ctx, ds, state, w)
 		if err != nil {
-			log.Errorf("Failed to set state cookie: %s", err.Error())
-			http.Error(w, "Error setting state cookie for login", http.StatusInternalServerError)
+			err = fmt.Errorf("failed to set state cookie for OAuth: %v", err)
```
stephenmcgruer: I have a concern with this re leaking internal info. What if sometime down the line, someone refactors `setState` to return an error with privileged information (or, maybe `setState` calls another method that can error, and returns that error), not realizing that we send the entire thing back to the random user who sent us a HTTP request?

Hexcles: That's a legit concern! Although I personally think that secrets shouldn't even be logged, let alone included in the responses, as logs tend to be readable by more people (e.g. exposed in error reporting), it's fair to employ defense by depth here.

In lieu of returning the error detail, we could fix our logging in tests properly -- instead of using a nil handler, we can wrap `testing.Log` as a handler and use it in tests. I'm going to send a follow-up PR for that.

Hexcles:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/1754

PR #1765, opened 2020-01-28
Support '*' in META.yml when searching `link`

stephenmcgruer:


Hexcles:
(Note: please either add "[WIP]" in the subject or create the PR as a draft.)

Hexcles:
Yet the PR description still says "WIP" :)

stephenmcgruer:
I am having trouble getting the deployment of this to work as expected again (there appears to be no searchcache instance up). I'm going to merge this, then compare performance on staging vs prod. If there's a noticeable difference, I may revert.

Hexcles:
> I am having trouble getting the deployment of this to work as expected again (there appears to be no searchcache instance up).

The branch name was too long, and this was before my fix #1773 .

stephenmcgruer:
> LGTM. Using a trie only speeds up a constant factor (the average depth of test)

For the record, without a trie searching `link:bugs.webkit.org` on staging.wpt.fyi takes ~1 second now (previously 800ms, no idea of the noise level here). So it seems 'ok' as is now, without a trie.

PR #1763, opened 2020-01-27
Accessibility fixes as suggested by Lighthouse

Hexcles:
See individual commits for details.

Lighthouse a11y scores before the change:
* homepage: 75
* file result: 71

Lighthouse a11y scores after the change:
* homepage: 81
* file result: 76

PR #1767, opened 2020-01-29
Set up Lighthouse CI

Hexcles:
## Description

This PR sets up Lighthouse CI on Travis. Currently it will only run on master against the newly deployed version on staging, because it's easier to set up.

I cherry-picked a few assertions as errors -- these are the ones I just fixed in #1763 so we don't want regressions. We are still far away from being able to use the "recommended" preset.

## Review Information

For testing purposes, I triggered a run on this PR branch:
https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/281682765

See the commit status of e61631d2810798974e9e37cc9bfd9b55edf4ca11 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,10 +45,13 @@ jobs:
         - MAKE_TEST_TARGET=puppeteer_chrome_test
 
   include:
-    # All the other tests run on GitHub Actions now. We haven't migrated
-    # Puppeteer tests because it is experimental and has little coverage.
+    # These are the default "test" stage, which will run at the *end*.
+    # All the other tests run on GitHub Actions. We haven't migrated Puppeteer
+    # tests and Lighthouse CI because they need to wait for deployments.
     - env:
         - MAKE_TEST_TARGET=puppeteer_chrome_test
+    - env:
+        - MAKE_TEST_TARGET=lighthouse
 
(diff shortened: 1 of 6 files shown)


Hexcles:
@stephenmcgruer one could run Lighthouse on their branch deployment from Chrome Devtools, which is actually more actionable than using the `lhci` CLI (you can directly jump to problematic elements, etc.).

stephenmcgruer:
> @stephenmcgruer one could run Lighthouse on their branch deployment from Chrome Devtools, which is actually more actionable than using the lhci CLI (you can directly jump to problematic elements, etc.).

Right, but you don't get a diff that way. Running the CLI with the same config as master means that as long as master is green, any reported failures are newly added by your change :)

Hexcles:
> Right, but you don't get a diff that way. Running the CLI with the same config as master means that as long as master is green, any reported failures are newly added by your change :)

Good point. However, Lighthouse does not have diffing support anyway. I was trying to get a snapshot of the report, and assert that we cannot regress, but that doesn't seem possible, so I added those assertions manually as a stopgap. The recommended way is in fact to get an almost perfect score and turn off a few checks for known issues. I think I'll get there within this sprint, so this is just temporary.

PR #1768, opened 2020-01-29
More a11y improvements

Hexcles:
1. Add labels to the expand toggle button and the search box.
2. Get rid of tiny fonts.
3. Remove empty `<a>` elements.

This brings the Lighthouse a11y scores over 90.

### DIFF
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -44,7 +44,7 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
 
     <div class="right">
       <paper-toggle-button checked="{{isVerbose}}">
-        Expand
+        Show Details
       </paper-toggle-button>
     </div>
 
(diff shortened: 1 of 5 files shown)


webapp/components/test-file-results.js
```diff
@@ -43,8 +43,8 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
     </style>
 
     <div class="right">
-      <label class="pad">Expand</label>
-      <paper-toggle-button class="pad" checked="{{isVerbose}}">
+      <paper-toggle-button checked="{{isVerbose}}">
+        Expand
```
stephenmcgruer: This seems like a chance to better explain what this toggle does. `Show Details`? `Detailed Output`? `Expand Failures`?

Hexcles:
```suggestion
        Show Details
```

"Show Details" sounds good!

Issue #1776, opened 2020-01-30
/api/metadata exceeds GitHub quota a few times a day


Hexcles:
We are seeing an increased number of 500s in the webapp recently. Upon closer investigation, most of them are from `/api/metadata`, which received 403 when trying to download the `wpt-metadata` source tarball. [GitHub docs](https://developer.github.com/v3/#rate-limiting) indicate that this is most likely due to rate limiting.

We might want to cache this artifact instead of caching our responses to `/api/metadata` requests (which change more often because of the URL params).

This will likely become much more severe once we enable showing metadata on the UI by default.

PR #1779, opened 2020-02-03
[Searchcache, webapp] Create Caching for WPT Metadata

KyleJu:
Fix for #1776 in searchcache and webapp. This change polls Metadata update in the background in searchcache and update Metadata through memcache in webapp on a 10-minute interval.

### DIFF
--- a/api/metadata_cache.go
+++ b/api/metadata_cache.go
@@ -1,0 +1,75 @@
+// Copyright 2020 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package api
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"time"
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"google.golang.org/appengine/memcache"
+)
+
+const metadataCacheKey = "WPT-METADATA"
+
+type webappMetadataFetcher struct {
+	ctx    context.Context
+	client *http.Client
+	log    shared.Logger
+	url    string
+}
+
+func (f webappMetadataFetcher) Fetch() (res map[string][]byte, err error) {
+	metadataMap, err := getMetadataFromMemcache(f.ctx, f.log, f.client, f.url)
+	if err == nil && metadataMap != nil {
+		return metadataMap, nil
+	}
+
+	return shared.CollectMetadataWithURL(f.client, f.url)
+}
+
+func getMetadataFromMemcache(ctx context.Context, log shared.Logger, client *http.Client, url string) (res map[string][]byte, err error) {
+	cached, err := memcache.Get(ctx, metadataCacheKey)
+
+	if err != nil && err != memcache.ErrCacheMiss {
+		log.Errorf("Error from getting Metadata in memcache: %s", err.Error())
+		return nil, err
+	}
+
+	if err == nil && cached != nil {
+		// Caches hit; update Metadata.
+		var metadataMap map[string][]byte
+		err = json.Unmarshal(cached.Value, &metadataMap)
+		if err != nil {
+			log.Errorf("Error from unmarshaling Metadata in memcache: %s", err.Error())
+			return nil, err
+		}
+
+		return metadataMap, nil
+	}
+
+	// Caches missed.
+	metadataByteMap, err := shared.CollectMetadataWithURL(client, url)
+	if err != nil {
+		log.Errorf("Error from CollectMetadataWithURL in a cache miss: %s", err.Error())
+		return nil, err
+	}
+
+	body, err := json.Marshal(metadataByteMap)
+	if err != nil {
+		log.Errorf("Error from marshaling metadataByteMap in a cache miss: %s", err.Error())
+		return metadataByteMap, nil
+	}
+
+	item := &memcache.Item{
+		Key:        metadataCacheKey,
+		Value:      body,
+		Expiration: time.Minute * 10,
+	}
+	memcache.Set(ctx, item)
+	return metadataByteMap, nil
+}
--- a/api/metadata_cache.go
+++ b/api/metadata_cache.go
@@ -19,12 +19,11 @@ const metadataCacheKey = "WPT-METADATA"
 type webappMetadataFetcher struct {
 	ctx    context.Context
 	client *http.Client
-	log    shared.Logger
 	url    string
 }
 
 func (f webappMetadataFetcher) Fetch() (res map[string][]byte, err error) {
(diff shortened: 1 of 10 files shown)


KyleJu:
@Hexcles One question I have is whether I put the polling in [webapp](https://github.com/web-platform-tests/wpt.fyi/blob/master/webapp/web/main.go) or searchcache

Hexcles:
Sorry for the late reply.

Both webapp and searchcache need metadata, but only searchcache has a long running daemon where you can do the polling, so we'd need to have a polling loop in a goroutine in searchcache that periodically updates the metadata tarball in memcache, and the webapp should then read from memcache.

shared/metadata_util.go
```diff
@@ -14,17 +14,34 @@ import (
 	"strings"
 )
 
-const metadataArchiveURL = "https://api.github.com/repos/web-platform-tests/wpt-metadata/tarball"
+// metadataMapCached is the local cache of Metadata.
+var metadataMapCached map[string][]byte = nil
```
Hexcles: We not only need this in-memory cache, but more importantly need to put it into memcache for use by webapp.

Hexcles:
Actually, I just realized that memcache isn't available in AppEngine Flex: https://cloud.google.com/appengine/docs/flexible/go/migrating#memcache

So we need two different caching strategies:
* In webapp, we need to cache the metadata tarball in memcache and fetch a new one upon expriy.
* In searchcache, what you have now (polling + in-memory cache) should be good.

api/query/cache/service/main.go
```diff
@@ -313,6 +313,12 @@ func main() {
 	// after backfilling was started.
 	go poll.KeepRunsUpdated(store, logger, *updateInterval, *updateMaxRuns, idx)
 
+	var netClient = &http.Client{
+		Timeout: time.Second * 5,
+	}
+	// Polls Metadata update every 10 minutes.
+	go shared.KeepMetadataUpdated(netClient, logger, time.Second*10)
```
Hexcles:
```suggestion
	go shared.KeepMetadataUpdated(netClient, logger, time.Minute*10)
```

Also this probably shouldn't live in `shared` as it's only used by searchcache.

shared/poll.go
```diff
@@ -0,0 +1,36 @@
...
+	"net/http"
+	"time"
+)
+
+// KeepMetadataUpdated implements updates to metadataMapCached via simple polling every
+// interval duration.
+func KeepMetadataUpdated(client *http.Client, logger Logger, interval time.Duration) {
+	// Start by waiting polling interval.
+	logger.Infof("Starting Metadata update via polling; waiting polling interval first...")
+	time.Sleep(interval)
```
Hexcles: Why does it start by waiting first? Shall we fetch metadata immediately instead?

KyleJu:
> Actually, I just realized that memcache isn't available in AppEngine Flex: https://cloud.google.com/appengine/docs/flexible/go/migrating#memcache
> 
> So we need two different caching strategies:
> 
> * In webapp, we need to cache the metadata tarball in memcache and fetch a new one upon expriy.
> * In searchcache, what you have now (polling + in-memory cache) should be good.

Per our discussion offline, metadata caching for standard webapp will be done using memcache in a separate PR. This is ready for review

shared/poll.go
```diff
@@ -0,0 +1,36 @@
...
+	logger.Infof("Metadata update via polling started")
+
+	for {
+		start := time.Now()
+		err := UpdatedMetadata(client, logger)
+		if err != nil {
+			logger.Errorf("Error fetching Metadata for update: %v", err)
+		}
+
+		wait(start, interval)
```
Hexcles: I think this is a bit overly complicated. We can simply `time.Sleep(interval)` again and disregard how much time the download takes.

api/metadata_cache.go
```diff
@@ -0,0 +1,75 @@
...
+func (f webappMetadataFetcher) Fetch() (res map[string][]byte, err error) {
+	metadataMap, err := getMetadataFromMemcache(f.ctx, f.log, f.client, f.url)
+	if err == nil && metadataMap != nil {
+		return metadataMap, nil
+	}
+
+	return shared.CollectMetadataWithURL(f.client, f.url)
+}
+
+func getMetadataFromMemcache(ctx context.Context, log shared.Logger, client *http.Client, url string) (res map[string][]byte, err error) {
```
KyleJu: Note to the reviewer: This is taken from #1784 and have tested separately

KyleJu:
@Hexcles ready for review

api/metadata_handler.go
```diff
@@ -35,8 +34,8 @@ func apiMetadataHandler(w http.ResponseWriter, r *http.Request) {
 	ctx := shared.NewAppEngineContext(r)
 	client := shared.NewAppEngineAPI(ctx).GetHTTPClient()
 	logger := shared.GetLogger(ctx)
-	metadataURL := shared.MetadataArchiveURL
-	delegate := MetadataHandler{logger, client, metadataURL}
+	fetcher := webappMetadataFetcher{ctx: ctx, client: client, log: logger, url: shared.MetadataArchiveURL}
+	delegate := MetadataHandler{logger, fetcher}
 
 	// Serve cached with 5 minute expiry. Delegate to Metadata Handler on cache miss.
 	shared.NewCachingHandler(
```
Hexcles: We shouldn't use CachingHandler any more, which adds yet another layer of memcache (keyed by URL) that's redundant now.

api/metadata_cache.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/web-platform-tests/wpt.fyi/shared"
+	"google.golang.org/appengine/memcache"
+)
+
+const metadataCacheKey = "WPT-METADATA"
+
+type webappMetadataFetcher struct {
+	ctx    context.Context
+	client *http.Client
+	log    shared.Logger
```
Hexcles: Optional: we don't really need to have this field. `Fetch` can simply get a logger from `ctx` and pass it to `getMetadataFromMemcache`. This also works in tests: `TestContext` has a nil logger associated with it already.

api/query/metadata_cache.go
```diff
@@ -0,0 +1,27 @@
...
+	"net/http"
+
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+// MetadataMapCached is the local cache of Metadata in searchcache.
+var MetadataMapCached map[string][]byte = nil
+
+type searchcacheMetadataFetcher struct {
+	client *http.Client
```
Hexcles: Constructing a `http.Client` is really cheap. I think we can construct a `&http.Client{Timeout: time.Second * 5}` directly in `Fetch` every time, which also has the extra benefit of having a single timeout definition.

PR #1804, opened 2020-02-20
[DO NOT MERGE] Repro of failure when calling search API

stephenmcgruer:


Hexcles:
Actually, let's keep this open for repo.

stephenmcgruer:
Oh right, to give us a PR that is based before your fix. Good idea :)

Issue #1747, opened 2020-01-17
Migrate Go AppEngine to use cloud SDK and eventually to golang 1.12


Hexcles:
## Goal

Upgrade standard AppEngine Go runtime to 1.12+ early to avoid last-minute rush before deprecation as it's a relatively big refactoring.

It also has additional benefits:
* Greatly simplify our code base (unifying flex and standard runtimes)
* Open up access to more Cloud SDK (no longer need special AppEngine versions)
* Modernize our toolchain to benefit from new Go features and bring the prod environment closer to our default dev setup on workstations (which has the latest Go)

## Overview

The biggest blocker for upgrading to 1.12+ for us is:
> The Go 1.12+ runtimes do not provide access to proprietary App Engine APIs. You can use the Google Cloud client library or third party libraries to access Google Cloud services. You can find suggested alternatives to App Engine APIs in the Migrating from the App Engine Go SDK section.

as we've done the [rest](https://cloud.google.com/appengine/docs/standard/go/go-differences#changes_in_the_app_engine_go_112_runtime) when upgrading to Go 1.11.

Here's the full list of APIs for reference:
https://cloud.google.com/appengine/docs/standard/go/go-differences#migrating-appengine-sdk

Note that we will be able to migrate gradually, **in place** -- we can keep using the Go 1.11 runtime while switching individual APIs to the new Cloud SDK, since most of them are available in Go 1.11 as well:
https://cloud.google.com/appengine/docs/standard/go111/go-differences#migrating-appengine-sdk

## Breakdown

* [x] Clean up remaining urlfetch (tiny) 
* [x] AppEngine Modules API -> google-api-go-client (small)
* [x] TaskQueue -> Cloud Tasks (small)
* [x] AppEngine logging -> Cloud Logging (small) @KyleJu 
* [x] AppEngine Datastore -> Cloud Datastore (big, need to figure out standalone Datastore emulator) (refactoring by @KyleJu, switchover by @Hexcles )
* [x] AppEngine Context -> request.Context (small) @KyleJu 
* [x] AppEngine memcache -> Memorystore (medium)
* [x] AppEngine Users API -> GitHub OAuth login (big)
* [x] Remove AppEngine search (never turned on by default) (small)
* [ ] Local testing: `dev_appserver` -> `go run` (small)

Hexcles:
Note that memcache is going away and we need to move to Memorystore.

Hexcles:
[...]

PR #1811, opened 2020-02-21
Migrate from appengine/urlfetch to net/http

### LINKED ISSUE #1808: staging searchcache is returning non-JSON data for structured queries
Reproduction: https://staging.wpt.fyi/results/?label=experimental&label=master&aligned&q=none%28link%3Abugs.chromium.org%29

For some reason, staging searchcache appears to be returning data that includes a few lines of what might be partial SHAs before the json. An example output (obtained from debugging going on in https://github.com/web-platform-tests/wpt.fyi/pull/1804) is:

```cdae1
{ "runs": [ ...
```

It's not clear why the searchcache would be doing this. It's only a few commits ahead of prod (all of which seem at first glance to be inconsequential), and prod is fine. I was unable to reproduce with a local build of searchcache.

Hexcles:
We still don't know why #1808 is happening, but based on experiments
switching to net/http fixes #1808, and it is recommended anyway (also
part of #1747).

This unfortunately changes the signature of a widely used method.

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -18,7 +18,6 @@ import (
 	"github.com/google/go-github/v28/github"
 	"golang.org/x/oauth2"
 	"google.golang.org/appengine"
-	"google.golang.org/appengine/urlfetch"
 	"google.golang.org/appengine/user"
 )
 
@@ -28,9 +27,9 @@ type AppEngineAPI interface {
 
 	// GitHub OAuth client
 	GetGitHubClient() (*github.Client, error)
-	// urlfetch cilents
+	// http.Client
 	GetHTTPClient() *http.Client
-	GetSlowHTTPClient(time.Duration) (*http.Client, context.CancelFunc)
+	GetHTTPClientWithTimeout(time.Duration) *http.Client
 
 	// AppEngine User API
 	IsAdmin() bool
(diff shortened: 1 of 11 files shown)


shared/appengine.go
```diff
@@ -42,14 +41,18 @@ type AppEngineAPI interface {
 	// GetHostname returns the canonical hostname for the current AppEngine project,
 	// i.e. staging.wpt.fyi or wpt.fyi.
 	GetHostname() string
-	// GetVersionedHostname returns the AppEngine hostname for the current version,
+	// GetVersionedHostname returns the AppEngine hostname for the current version of the default service,
 	// i.e. version-dot-wptdashboard{,-staging}.appspot.com.
+	// Note: if the default service does not have the current version,
```
stephenmcgruer: Nit, but these seem very inconsistently wrapped. Some lines look like they wrap around 100 or 120 chars, then a few wrap around 80.

shared/appengine.go
```diff
@@ -161,15 +159,7 @@ func (a appEngineAPIImpl) GetVersionedHostname() string {
 }
 
 func (a appEngineAPIImpl) GetServiceHostname(service string) string {
```
stephenmcgruer: I guess we had no tests for this then? Not sure if its worth adding much now that the method is just a simple `Sprintf` wrapper though

Hexcles: Right, we don't actually have a test for this whole module, which is supposed to be a thin wrapper of appengine APIs.

PR #1812, opened 2020-02-21
[searchcache] Improve error handling

Hexcles:
Previously we had a bunch of cases where the API handler did not return
after encountering fatal errors. In some cases, we would even try to
rewrite the status code of the response after writing out the error
message, which is impossible.

There is no bulletproof way to prevent missing returns, but it still
helps to extract the bulk of the handler into a new function that
explicitly returns an error which is then handled in one single place.

This is largely a no-op refactor.

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -6,10 +6,8 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"runtime"
@@ -19,12 +17,10 @@ import (
 	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/datastore"
 	"github.com/sirupsen/logrus"
-	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/backfill"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/monitor"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/poll"
-	cq "github.com/web-platform-tests/wpt.fyi/api/query/cache/query"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/api/option"
 	mrpb "google.golang.org/genproto/googleapis/api/monitoredres"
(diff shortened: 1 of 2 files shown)


Hexcles:
I discovered the issues while investigating the corrupted JSON from searchcache. Although they are not causing any immediate harm (at least for the happy code path), we should still fix them.

api/query/cache/service/main.go
```diff
@@ -67,166 +63,12 @@ func readinessCheckHandler(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
-		log.Errorf("Failed to marshal results: %s", err.Error())
-		http.Error(w, "Failed to marshal results to JSON", http.StatusInternalServerError)
-		return
-	}
-	if len(missing) != 0 {
-		w.WriteHeader(http.StatusUnprocessableEntity)
+		log := shared.GetLogger(r.Context())
+		log.Errorf(err.Error())
+		http.Error(w, err.Message, err.Code)
```
stephenmcgruer: My concern here is that because `w` has been passed into `searchHandlerImpl`, we have no guarantee that `w` has not already been written to. What happens if you call `http.Error` after calling (say) `w.Write(...)` - does it panic?

api/query/cache/service/search.go
```diff
@@ -0,0 +1,220 @@
+// Copyright 2018 The WPT Dashboard Project. All rights reserved.
```
stephenmcgruer: `2018`, huh? ;)

api/query/cache/service/search.go
```diff
@@ -0,0 +1,220 @@
...
+	"net/http"
+
+	"github.com/web-platform-tests/wpt.fyi/api/query"
+	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
+	cq "github.com/web-platform-tests/wpt.fyi/api/query/cache/query"
+	"github.com/web-platform-tests/wpt.fyi/shared"
+)
+
+type searchErr struct {
+	// Detail is the internal (inner) error that should not be exposed.
```
stephenmcgruer:
```suggestion
	// Detail is the internal error that should not be exposed to the end-user.
```

This just seems to be defining the word 'internal'?

api/query/cache/service/search.go
```diff
@@ -0,0 +1,220 @@
...
+			Detail:  err,
+			Message: "Failed to create query plan",
+			Code:    http.StatusInternalServerError,
+		}
+	}
+
+	results := plan.Execute(runs, opts)
+	res, ok := results.([]shared.SearchResult)
+	if !ok {
+		return &searchErr{
```
stephenmcgruer: Nice catch realizing that we used to log the wrong error (!), but in general it seems concerning that `plan.Execute` could return either nil or not a `[]shared.SearchResult` and we would get no details about that :/.

api/query/cache/service/search.go
Hexcles:
```suggestion
// Copyright 2020 The WPT Dashboard Project. All rights reserved.
```

api/query/cache/service/main.go
Hexcles: No, it does not. It'd fail to update the status code; technically it's a behaviour not defined in Golang docs.

api/query/cache/service/search.go
Hexcles: Yeah, this is not good, but it should be fixed in another PR (and probably when we run into real problems).

PR #1823, opened 2020-02-28
Implement Manifest.ContainsTest

Hexcles:
Rename `Contains` to `ContainsFile` to make it clear that it checks the
existence of a file path (or directory path), and implement a new
`ContainsTest` method to check the existence of a test URL.

`ContainsTest` supports both multi-global tests and test variants.

## Review Information

I'm not quite happy about some of the names (especially the exported one). Please feel free to give suggestions.

### DIFF
--- a/shared/manifest.go
+++ b/shared/manifest.go
@@ -114,7 +114,6 @@ func (m *Manifest) ContainsTest(testURL string) (bool, error) {
 	testURL = strings.TrimLeft(testURL, "/")
 	path, query := ParseTestURL(testURL)
 	parts := strings.Split(path, "/")
-	// parts=["foo", "bar", "test.any.js"]
 	for _, trie := range m.imap {
 		leaf, ok := findNode(trie, parts).([]interface{})
 		if !ok {
@@ -128,7 +127,7 @@ func (m *Manifest) ContainsTest(testURL string) (bool, error) {
 			return false, ErrInvalidManifest
 		}
 		for _, v := range leaf[1:] {
-			// variant=[url, extra]
+			// variant=[url, extras...]
 			variant, ok := v.([]interface{})
 			if !ok || len(variant) < 2 {
 				return false, ErrInvalidManifest
@@ -234,7 +233,7 @@ func ExplodePossibleFilenames(filePath string) []string {
 // ParseTestURL parses a WPT test URL and returns its file path and query
 // components. If the test is a multi-global (auto-generated) test, the
 // function returns the underlying file name of the test.
-// e.g. testURL="foo/bar/test.any.worker.html?varaint"
+// e.g. testURL="foo/bar/test.any.worker.html?variant"
 //      filepath="foo/bar/test.any.js"
 //      query="?variant"
 func ParseTestURL(testURL string) (filePath, query string) {
@@ -244,10 +243,10 @@ func ParseTestURL(testURL string) (filePath, query string) {
 		query = testURL[qPos:]
 	}
 	for _, i := range implosions() {
-		tSuffix := i[0]
-		fSuffix := i[1]
-		if strings.HasSuffix(filePath, tSuffix) {
-			filePath = strings.TrimSuffix(filePath, tSuffix) + fSuffix
+		testSuffix := i[0]
+		fileSuffix := i[1]
+		if strings.HasSuffix(filePath, testSuffix) {
+			filePath = strings.TrimSuffix(filePath, testSuffix) + fileSuffix
 			break
 		}
 	}
--- a/shared/manifest.go
+++ b/shared/manifest.go
@@ -71,7 +71,7 @@ func (m *Manifest) unmarshalAll() error {
 	return nil
 }
 
-func trieContains(t interface{}, parts []string) interface{} {
+func findNode(t interface{}, parts []string) interface{} {
 	if len(parts) == 0 {
 		return t
 	}
@@ -81,7 +81,7 @@ func trieContains(t interface{}, parts []string) interface{} {
 	if !ok {
 		return nil
 	}
-	return trieContains(trie[parts[0]], parts[1:])
+	return findNode(trie[parts[0]], parts[1:])
 }
 
 // ContainsFile checks whether m contains a file path (including directories).
@@ -97,7 +97,7 @@ func (m *Manifest) ContainsFile(path string) (bool, error) {
 	}
 	parts := strings.Split(path, "/")
 	for _, items := range m.imap {
-		if trieContains(items, parts) != nil {
+		if findNode(items, parts) != nil {
 			return true, nil
 		}
 	}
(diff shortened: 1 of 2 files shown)


shared/manifest.go
```diff
@@ -68,27 +71,95 @@ func (m *Manifest) unmarshalAll() error {
...
 	return false, nil
 }
 
+// ContainsTest checks whether m contains a full test URL.
+func (m *Manifest) ContainsTest(testURL string) (bool, error) {
+	if err := m.unmarshalAll(); err != nil {
+		return false, err
+	}
+
+	// URLs in the manifest do not include the leading slash (url_base).
```
stephenmcgruer: I would suggest extracting this part:

```
	parts, query := ParseTestURL(testURL) 
	for _,  items := range m.imap {
		...
```

shared/manifest.go
```diff
@@ -68,27 +71,95 @@ func (m *Manifest) unmarshalAll() error {
...
+		query = testURL[qPos:]
+	}
+	path = RecoverTestFilename(path)
+	parts := strings.Split(path, "/")
+	// e.g. testURL="/foo/bar/test.any.html?varaint" would become
+	// testURL="foo/bar/test.any.html?varaint"
+	// path="foo/bar/test.any.js"
+	// query="?variant"
+	// parts=["foo", "bar", "test.any.js"]
+	for _, items := range m.imap {
```
stephenmcgruer: Not clear to me why the variable is called `items` here. Isn't `imap` a map from e.g. "testharness" to the (unmarshalled) trie for "testharness" tests? (E.g. isn't it `item`?)

shared/manifest.go
```diff
@@ -68,27 +71,95 @@ func (m *Manifest) unmarshalAll() error {
...
+		item, ok := trieContains(items, parts).([]interface{})
```
stephenmcgruer: So `trieContains` actually finds the node in the trie that `parts` encodes, and then we assume that if that can be cast to a list then it is a leaf node.

I would suggest renaming `trieContains` to `findNode` or something like that, and rename `item` to `leaf` or `node`.

shared/manifest.go
```diff
@@ -68,27 +71,95 @@ func (m *Manifest) unmarshalAll() error {
...
+		if trieContains(items, parts) != nil {
 			return true, nil
 		}
 	}
 	return false, nil
 }
 
+// ContainsTest checks whether m contains a full test URL.
+func (m *Manifest) ContainsTest(testURL string) (bool, error) {
+	if err := m.unmarshalAll(); err != nil {
```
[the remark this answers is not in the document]
Hexcles: This is a good question. The reasons are twofold:

1. I could use the same logic as filtering, but the major difference between them is the use case. Filtering is usually only done once (e.g. in an API request to get a subtree), but existence tests tend to be done multiple times in a row (e.g. in wpt-metadata to check whether a bunch of paths exist). If I use the exact same logic, then we're likely to parse the same subtrees multiple times. Parsing the whole trie recursively is an easy optimization based on use cases.
2. In theory I could implement a true JIT cache so that each subtree is parsed at most once, but that'd require a big type switch (to check whether each `interface{}` is a parsed map or a `RawMessage`). And I think at that point it'd be premature optimization.

shared/manifest.go
Hexcles: `imap` is a map from test types (e.g. "testharness") to a (marshalled) trie of test items. Each test file (leaf node) is an "item". I can rename `items` to `trie`, which probably makes more sense.

shared/manifest.go
Hexcles: I'll call it `leaf`.

shared/manifest.go
Hexcles: On second thought, I think this should be folded into `RecoverTestFilename` (and of course it needs to be renamed). A sane API should encapsulate all the details about variants (query strings) and multi-global tests; the caller only needs to in a test URL and gets all the necessary information.

stephenmcgruer: So this is fine, but I would personally roll in the trimming of testURL and splitting of path into query. Your call though.

shared/manifest.go
```diff
@@ -68,27 +71,87 @@ func (m *Manifest) unmarshalAll() error {
...
+		return false, err
+	}
+
+	// URLs in the manifest do not include the leading slash (url_base).
+	testURL = strings.TrimLeft(testURL, "/")
+	path, query := ParseTestURL(testURL)
+	parts := strings.Split(path, "/")
+	// parts=["foo", "bar", "test.any.js"]
```
stephenmcgruer: Nit; I don't think this adds much.

shared/manifest.go
```diff
@@ -68,27 +71,87 @@ func (m *Manifest) unmarshalAll() error {
...
+		}
+		// A leaf node represents a test file, and has at least two
+		// elements: [SHA, variants...].
+		if len(leaf) < 2 {
+			return false, ErrInvalidManifest
+		}
+		for _, v := range leaf[1:] {
+			// variant=[url, extra]
+			variant, ok := v.([]interface{})
+			if !ok || len(variant) < 2 {
```
stephenmcgruer: Can a variant have len > 2?

shared/manifest.go
```diff
@@ -154,15 +231,25 @@ func ExplodePossibleFilenames(filePath string) []string {
 	return nil
 }
 
-func trieContains(t interface{}, parts []string) bool {
-	if len(parts) == 0 {
-		return t != nil
+// ParseTestURL parses a WPT test URL and returns its file path and query
+// components. If the test is a multi-global (auto-generated) test, the
+// function returns the underlying file name of the test.
+// e.g. testURL="foo/bar/test.any.worker.html?varaint"
```
stephenmcgruer:
```suggestion
// e.g. testURL="foo/bar/test.any.worker.html?variant"
```

shared/manifest.go
```diff
@@ -154,15 +231,25 @@ func ExplodePossibleFilenames(filePath string) []string {
...
+		filePath = testURL[:qPos]
+		query = testURL[qPos:]
 	}
-
-	// t could be nil (e.g. if the previous part does not exist in the map), in which case casting will fail.
-	trie, ok := t.(map[string]interface{})
-	if !ok {
-		return false
+	for _, i := range implosions() {
+		tSuffix := i[0]
```
stephenmcgruer: Eugh, I hate golang. Would it be so dreadful to name this `testSuffix` and `fileSuffix`?

shared/manifest.go
Hexcles: Agreed. And with the example moved to ParseTestURL, this doesn't make much sense, either.

```suggestion
```

shared/manifest.go
Hexcles: Yes, reftests do. Let me expand the comment above a bit.

shared/manifest.go
Hexcles: They'd be `ts` and `fs` in true Golang style. :)

And no personally I don't think it's worthwhile to save a few keystrokes, either...

shared/manifest.go
```diff
@@ -68,27 +71,87 @@ func (m *Manifest) unmarshalAll() error {
...
+			// is not a list (i.e. not a leaf).
+			continue
+		}
+		// A leaf node represents a test file, and has at least two
+		// elements: [SHA, variants...].
+		if len(leaf) < 2 {
+			return false, ErrInvalidManifest
+		}
+		for _, v := range leaf[1:] {
+			// variant=[url, extra]
```
Hexcles:
```suggestion
			// variant=[url, extras...]
```

shared/manifest.go
```diff
@@ -154,15 +231,25 @@ func ExplodePossibleFilenames(filePath string) []string {
...
-
-	// t could be nil (e.g. if the previous part does not exist in the map), in which case casting will fail.
-	trie, ok := t.(map[string]interface{})
-	if !ok {
-		return false
+	for _, i := range implosions() {
+		tSuffix := i[0]
+		fSuffix := i[1]
+		if strings.HasSuffix(filePath, tSuffix) {
+			filePath = strings.TrimSuffix(filePath, tSuffix) + fSuffix
```
Hexcles:
```suggestion
		testSuffix := i[0]
		fileSuffix := i[1]
		if strings.HasSuffix(filePath, testSuffix) {
			filePath = strings.TrimSuffix(filePath, testSuffix) + fileSuffix
```

PR #1826, opened 2020-03-02
[webapp] Add an Endpoint for User's Log-in Status

KyleJu:
Add an endpoint that checks user's log-in status. Only open to trusted third-parties and simple requests.

### DIFF
--- a/api/routes.go
+++ b/api/routes.go
@@ -77,5 +77,5 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, CORSList, []string{"PATCH"}))
 
 	// API endpoint for checking a user's login status.
-	shared.AddRoute("/api/user", "login-status", shared.WrapApplicationJSON(shared.WrapTrustedCORS(apiUserHandler, CORSList, nil)))
+	shared.AddRoute("/api/user", "api-user", shared.WrapApplicationJSON(shared.WrapTrustedCORS(apiUserHandler, CORSList, nil)))
 }
--- a/api/routes.go
+++ b/api/routes.go
@@ -74,5 +74,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
 
 	// API endpoint for modifying Metadata.
-	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, shared.CORSList, []string{"PATCH"}))
+	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, CORSList, []string{"PATCH"}))
+
+	// API endpoint for checking a user's login status.
+	shared.AddRoute("/api/user", "login-status", shared.WrapApplicationJSON(shared.WrapTrustedCORS(apiUserHandler, CORSList, nil)))
 }
--- a/api/routes.go
+++ b/api/routes.go
@@ -74,5 +74,5 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
 
 	// API endpoint for modifying Metadata.
-	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, []string{"https://jgraham.github.io"}, []string{"PATCH"}))
+	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, shared.CORSList, []string{"PATCH"}))
 }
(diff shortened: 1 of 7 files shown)


webapp/routes.go
```diff
@@ -34,6 +34,7 @@ func init() {
 func RegisterRoutes() {
 	// GitHub OAuth login
 	shared.AddRoute("/login", "login", loginHandler)
+	shared.AddRoute("/login/status", "login-status", shared.WrapTrustedCORS(loginStatusHandler, []string{"https://jgraham.github.io"}, []string{}))
```
Hexcles: Use a constant to store the CORS list.

api/routes.go
```diff
@@ -74,5 +74,5 @@ func RegisterRoutes() {
 	shared.AddRoute("/api/metadata", "api-metadata", shared.WrapPermissiveCORS(apiMetadataHandler))
 
 	// API endpoint for modifying Metadata.
-	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, []string{"https://jgraham.github.io"}, []string{"PATCH"}))
+	shared.AddRoute("/api/metadata/triage", "api-metadata-triage", shared.WrapTrustedCORS(apiMetadataTriageHandler, shared.CORSList, []string{"PATCH"}))
```
KyleJu: @Hexcles  Looks like `/api/metadata/triage` is not returning JSON here and this API returns a PR URL in text (as w.write() detects the return type automatically). If I wrap shared.WrapApplicationJSON around it, it would possibly break jgraham's code?

shared/CORS_list.go
```diff
@@ -0,0 +1,8 @@
+// Copyright 2020 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+// CORSList is a list of trusted third-party origins.
+var CORSList = []string{"https://jgraham.github.io"}
```
Hexcles: It probably makes more sense to have it in `api/routes.go`. CORS is specific to webapp APIs and I don't expect this to be "shared".

webapp/login.go
```diff
@@ -16,6 +17,34 @@ import (
...
+func loginStatusHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := shared.NewAppEngineContext(r)
+	aeAPI := shared.NewAppEngineAPI(ctx)
+	if !aeAPI.IsFeatureEnabled("githubLogin") {
+		http.Error(w, "Feature not enabled", http.StatusNotImplemented)
+		return
+	}
+
+	if cookie, err := r.Cookie("session"); err != nil || cookie == nil {
+		respond, _ := json.Marshal("User is not logged in")
```
Hexcles: Hmm, this will produce `"User is not logged in"` (with the quotes), which is technically JSON but not really more ergonomic, especially when you look at the success case: callers of this API would have to parse that string to get the username. If GitHub allows spaces in usernames, this would be even more brittle.

I was thinking we could return a struct like
```go
type response struct {
  User *shared.User `json:,omitempty`
  Error string `json:,omitempty`
}
```

That way, it's very easy to check whether the user is logged in (based on which field is present) and extract information.

api/routes.go
Hexcles: We shouldn't wrap a plaintext response with `shared.WrapApplicationJSON`, but sorry I don't quite see your point here?

KyleJu: Sorry, /login/status only returns plaintext in my original implementation until your [comment](https://github.com/web-platform-tests/wpt.fyi/pull/1826#discussion_r386674858), and I thought you suggested that all APIs should return in JSON format regardless.

shared/CORS_list.go
KyleJu: Right, but this `CORSList` is shared by `webapp` and `api`. By having it in `api/routes.go`, `api` dependency is introduced in `webapp`. What if I move this login/status endpoint to `api`?

webapp/login.go
KyleJu: Thanks for the suggestion and I agree with you here! Perhaps we can have `Error *string` instead, so the error message is omitted when the response is success.

Hexcles: No we don't have to; empty strings are omitted.

KyleJu:
@Hexcles friendly ping :)

webapp/login.go
```diff
@@ -16,6 +17,48 @@ import (
 	"golang.org/x/oauth2"
 )
 
+type loginResponse struct {
+	User  *shared.User `json:"user,omitempty"`
```
Hexcles: Note: we should add json annotation to `shared.User`, too; otherwise, it'd have capitalized field names, e.g. `user.GitHubEmail`, which looks inconsistent.

api/user.go
```diff
@@ -0,0 +1,54 @@
...
+		marshalled, err := json.Marshal(response)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.WriteHeader(http.StatusUnauthorized)
+		w.Write(marshalled)
+		return
+	}
```
Hexcles: Isn't this basically the same as the check inside `GetUserFromCookie`?

KyleJu: Well similar, originally I thought having this check would deliver a clear message  - whether a user is not logged in due to an empty session value vs internal serval error (corrupted cookie value, cookie expiry, etc). After a second thought, they should be treated the same because users have to log in again.

PR #1831, opened 2020-03-03
Add the missing Content-Type header back

### LINKED ISSUE #1830: Webhooks fail to send results to the receiver
**Current status:** this is currently broken on master and staging, but hadn't been discovered. Prod was broken for a short while following today's deployment #1825 , but was shortly reverted -- in other words, prod is currently not affected.

**Discovery:** staging hasn't been receiving any new runs since around Feb 24th, but we don't monitor staging. The bug surfaced when today, a week since Feb 24th, [a WebDriver test](https://travis-ci.com/web-platform-tests/wpt.fyi/jobs/293165433#L2879) that checks https://staging.wpt.fyi/runs started to fail on CI. Because of the lack of new runs, the recent runs page has become empty (by default we show recent runs in the past week). This bug would have been discovered in another way tomorrow by our rotation as the prod no longer stops receiving new runs for a day.

With bisecting, I was able to confirm the culprit change was https://github.com/web-platform-tests/wpt.fyi/commit/9831b916d1edbf5baab7e021398867a554b93ac4 . After this change, the results receiver started to receive empty payload from webhooks, confirmed by checking server logs from the `/api/results/upload` endpoint.

Hexcles:
This attempts to fix #1830.

Missing Content-Type: application/x-url-form-encoded when posting an
HTTP form is clearly a bug, but it is unclear why it only started to
surface after #1811.

A preventative measure is also added to return an error when the upload
payload is empty, along with some minor improvements to logging.

### DIFF
--- a/api/receiver/receive_results.go
+++ b/api/receiver/receive_results.go
@@ -62,15 +62,16 @@ func HandleResultsUpload(a API, w http.ResponseWriter, r *http.Request) {
 
 	log := shared.GetLogger(a.Context())
 	var results, screenshots []string
-	if r.MultipartForm != nil && r.MultipartForm.File != nil && len(r.MultipartForm.File["result_file"]) > 0 {
+	if f := r.MultipartForm; f != nil && f.File != nil && len(f.File["result_file"]) > 0 {
 		// result_file[] payload
-		files := r.MultipartForm.File["result_file"]
-		sFiles := r.MultipartForm.File["screenshot_file"]
+		files := f.File["result_file"]
+		sFiles := f.File["screenshot_file"]
 		log.Debugf("Found %d result files, %d screenshot files", results, screenshots)
 		var err error
 		results, screenshots, err = saveToGCS(a, uploader, files, sFiles)
 		if err != nil {
-			http.Error(w, err.Error(), http.StatusInternalServerError)
+			log.Errorf("Failed to save files to GCS: %v", err)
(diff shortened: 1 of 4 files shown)


Hexcles:
This seems to work. As an experiment, I promoted the staging branch to the default version, so we should soon start seeing new runs on staging.

api/receiver/receive_results.go
```diff
@@ -62,33 +62,38 @@ func HandleResultsUpload(a API, w http.ResponseWriter, r *http.Request) {
...
 		log.Debugf("Found Azure URL: %s", azureURL)
 		extraParams["azure_url"] = azureURL
-	} else {
+	} else if len(r.PostForm["result_url"]) > 0 {
 		// General result_url[] payload
 		results = r.PostForm["result_url"]
 		screenshots = r.PostForm["screenshot_url"]
 		log.Debugf("Found %d result URLs, %d screenshot URLs", results, screenshots)
+	} else {
+		log.Errorf("No results found")
```
stephenmcgruer: Will it be clear on the logs side what this refers to? (e.g. if it shows up as `receive_results.go: No results found` thats probably clear, but if it just shows up as `[timestamp] No results found` thats quite hard to debug from!

Hexcles: Yeah we should probably improve the underlying log handler to include file names, but that'd be another PR.

PR #1833, opened 2020-03-03
Avoid dereferencing null in webdriver tests

Hexcles:
`e.shadowRoot` might be null if the custom element hasn't been defined, in which case we should not try to query elements in the shadow DOM (and retry instead).

Example panic:
```
panic: javascript error: javascript error: Cannot read property 'querySelectorAll' of null
  (Session info: chrome=80.0.3987.132)
  (Driver info: chromedriver=80.0.3987.106 (f68069574609230cf9b635cd784cfb1bf81bb53a-refs/branch-heads/3987@{#882}),platform=Linux 5.0.0-1032-azure x86_64) [recovered]
	panic: javascript error: javascript error: Cannot read property 'querySelectorAll' of null
  (Session info: chrome=80.0.3987.132)
  (Driver info: chromedriver=80.0.3987.106 (f68069574609230cf9b635cd784cfb1bf81bb53a-refs/branch-heads/3987@{#882}),platform=Linux 5.0.0-1032-azure x86_64)
```

### DIFF
--- a/webdriver/webdriver.go
+++ b/webdriver/webdriver.go
@@ -126,7 +126,7 @@ func FindShadowElements(
 		result, err := d.ExecuteScriptRaw(
 			fmt.Sprintf(`return Array.from(arguments)
 				.reduce((s, e) => {
-					return s.concat(Array.from(e.shadowRoot.querySelectorAll('%s')))
+					return e.shadowRoot ? s.concat(Array.from(e.shadowRoot.querySelectorAll('%s'))) : s
 				}, [])`,
 				selector),
 			interfaces)


webdriver/search_test.go
```diff
@@ -109,7 +109,7 @@ func getSearchElement(wd selenium.WebDriver) (selenium.WebElement, error) {
 
 func getPathPartElements(wd selenium.WebDriver, element string) ([]selenium.WebElement, error) {
 	e, err := wd.FindElement(selenium.ByTagName, "wpt-app")
-	if err != nil {
+	if e == nil || err != nil {
```
stephenmcgruer: Is this meant to happen? All the uses of `FindElement` I can see in the upstream project seem to assume that if `err == nil` then the element is non-nil (e.g. https://github.com/tebeka/selenium/blob/a49cf4b98a36c2b21b1ccb012852bd142d5fc04a/example_test.go#L80, https://github.com/tebeka/selenium/blob/e24f7b7726ca0d40f51b3ba4c5d5c4ddcb4bae80/remote.go#L749).

Hexcles: Good point. "No such element" is an error: https://github.com/tebeka/selenium/blob/e24f7b7726ca0d40f51b3ba4c5d5c4ddcb4bae80/remote.go#L28

Let me take another look.

Hexcles: Thanks a lot for spotting this! I indeed didn't read the stacktrace closely enough. The bug we are seeing is that `e.shadowRoot` is null, instead of `e` being null (it wouldn't).

PR #1835, opened 2020-03-04
Fix a panic in diff.go

Hexcles:
panic: runtime error: invalid memory address or nil pointer dereference
goroutine 16 [running]:
google.golang.org/appengine/panic
	/usr/local/go/src/runtime/panic.go:513
github.com/web-platform-tests/wpt.fyi/api.handleAPIDiffGet
	/tmp/staging/srv/api/diff.go:83
github.com/web-platform-tests/wpt.fyi/api.apiDiffHandler
	/tmp/staging/srv/api/diff.go:25

The direct cause is trying to dereference an incorrect error (err
instead of parseErr). However, the whole function body was convoluted
and error-prone, so I've decided to refactor it (extract `loadDiffRuns`
out) to make error handling clearer.

Also remove an unused type definition.

### DIFF
--- a/api/diff.go
+++ b/api/diff.go
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-type diffResult struct {
-	Diff    shared.ResultsDiff `json:"diff"`
-	Renames map[string]string  `json:"renames,omitempty"`
+func loadDiffRuns(store shared.Datastore, q url.Values) (shared.TestRuns, error) {
+	if runIDs, err := shared.ParseRunIDsParam(q); err != nil {
+		return nil, err
+	} else if len(runIDs) > 0 {
+		runs, err := runIDs.LoadTestRuns(store)
+		// If all errors are NoSuchEntity, we don't treat it as an error.
+		// If err is nil, the type conversion will fail.
+		if multiError, ok := err.(appengine.MultiError); ok {
+			all404s := true
+			for _, err := range multiError {
+				if err != datastore.ErrNoSuchEntity {
+					all404s = false
+					break
+				}
+			}
+			if all404s {
+				return nil, nil
+			}
+		}
+		// err many be non nil here.
+		return runs, err
+	}
+
+	// NOTE: We use the same params as /results, but also support
+	// 'before' and 'after' and 'filter'.
+	runFilter, err := shared.ParseTestRunFilterParams(q)
+	if err != nil {
+		return nil, err
+	}
+	if beforeAndAfter, err := shared.ParseBeforeAndAfterParams(q); err != nil {
+		return nil, err
+	} else if len(beforeAndAfter) > 0 {
+		runFilter.Products = beforeAndAfter
+	}
+	runsByProduct, err := LoadTestRunsForFilters(store, runFilter)
+	if err != nil {
+		return nil, err
+	}
+	return runsByProduct.AllRuns(), nil
 }
 
 func handleAPIDiffGet(w http.ResponseWriter, r *http.Request) {
 	ctx := shared.NewAppEngineContext(r)
-
-	q := r.URL.Query()
-	runIDs, err := shared.ParseRunIDsParam(q)
-	if err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
-		return
-	}
-
-	var diffFilter shared.DiffFilterParam
-	var paths mapset.Set
-	if diffFilter, paths, err = shared.ParseDiffFilterParams(q); err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
-		return
-	}
-
 	store := shared.NewAppEngineDatastore(ctx, true)
-	var runs shared.TestRuns
-	if len(runIDs) > 0 {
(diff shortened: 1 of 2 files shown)


Hexcles:
FYI @stephenmcgruer (don't review this and enjoy your time off)

api/diff.go
```diff
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-type diffResult struct {
-	Diff    shared.ResultsDiff `json:"diff"`
-	Renames map[string]string  `json:"renames,omitempty"`
+func loadDiffRuns(store shared.Datastore, q url.Values) (shared.TestRuns, error) {
+	if runIDs, err := shared.ParseRunIDsParam(q); err != nil {
```
KyleJu: Don't think `run_id` param is documented anywhere in [api/diff](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md#apidiff). Would you mind doing a drive-by change on the README too?

api/diff.go
```diff
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
...
-		runsByProduct, err = LoadTestRunsForFilters(store, runFilter)
-		if err != nil {
-			runs = runsByProduct.AllRuns()
-		}
+	if len(runs) != 2 {
+		http.Error(w, fmt.Sprintf("Diffing requires exactly 2 runs, but found %v", len(runs)), http.StatusBadRequest)
+		return
 	}
 
+	diffFilter, paths, err := shared.ParseDiffFilterParams(q)
```
KyleJu: I think I would rather have it in the beginning of the handler, so if this checks fails, it doesn't do all the extra work necessarily (e.g. `LoadTestRuns`). And it might be more clearer on the param requirements

api/diff.go
KyleJu: Also I think this is wrong - IIUC, if `ParseRunIDsParam` returns an error, you can still get `runs` through `runFilter` in line 58. This results in ignoring `filter` information

api/diff.go
```diff
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
...
+func loadDiffRuns(store shared.Datastore, q url.Values) (shared.TestRuns, error) {
+	if runIDs, err := shared.ParseRunIDsParam(q); err != nil {
+		return nil, err
+	} else if len(runIDs) > 0 {
+		runs, err := runIDs.LoadTestRuns(store)
+		// If all errors are NoSuchEntity, we don't treat it as an error.
+		// If err is nil, the type conversion will fail.
+		if multiError, ok := err.(appengine.MultiError); ok {
+			all404s := true
+			for _, err := range multiError {
```
KyleJu: Could you `err` overwrites the `err` from runIDs.LoadTestRuns(store) when `all404s` is false?

Hexcles: No, and that's exactly the reason I'm trying to use `:=` as much as possible.

api/diff.go
```diff
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-type diffResult struct {
-	Diff    shared.ResultsDiff `json:"diff"`
-	Renames map[string]string  `json:"renames,omitempty"`
+func loadDiffRuns(store shared.Datastore, q url.Values) (shared.TestRuns, error) {
+	if runIDs, err := shared.ParseRunIDsParam(q); err != nil {
+		return nil, err
+	} else if len(runIDs) > 0 {
```
KyleJu: Not particularly fond of the fact that `ParseRunIDsParam` could return `nil, nil` instead of `[], nil`, and `len(nil)` would throw an error here. See [here](https://github.com/web-platform-tests/wpt.fyi/blob/76e6b108d32efba7d38ae96e1bf0334b25c64e2a/shared/params.go#L475)

api/diff.go
KyleJu: Optional - Ignore when `err` is nil here and let `ParseTestRunFilterParams` (line 58) fetches the runs as a fallback?

api/diff.go
Hexcles: Thanks for checking this! This won't be an error because the `nil` here is typed: https://play.golang.org/p/EECu3SNLH4q

api/diff.go
```diff
@@ -30,77 +30,72 @@ func apiDiffHandler(w http.ResponseWriter, r *http.Request) {
...
+					all404s = false
+					break
+				}
+			}
+			if all404s {
+				return nil, nil
+			}
+		}
+		// err many be non nil here.
+		return runs, err
```
KyleJu: Retuning `runs` and `err` together is a bit confusing. What if we add a check on line 51 `else if (err !=nil) {return nil, err)` and return `runs, nil` here?

api/diff.go
Hexcles: Good point! Done!

api/diff.go
Hexcles: Done.

api/diff.go
Hexcles: Moved this along with the error handling to the beginning.

PR #1842, opened 2020-03-09
Cleanup .gitignore and .gcloudignore

Hexcles:
* Some files no longer exist.
* Simplify rules (e.g. `**/foo` and `foo/*` are equivalent to `foo/`).
* Sort rules by categories.

### DIFF
--- a/webapp/web/.gcloudignore
+++ b/webapp/web/.gcloudignore
@@ -1,3 +1,7 @@
+# Files matching the following rules won't be uploaded to AppEngine (webapp).
+# The syntax is the same as `.gitignore`; we do not use `.gitignore` directly
+# because we need to update some generated files.
+
 # Tests
 *_test.go
 test/
--- a/webapp/web/.gcloudignore
+++ b/webapp/web/.gcloudignore
@@ -4,10 +4,10 @@
 
 # Tests
 *_test.go
+/webapp/static/wptd-metrics/
+/webapp/static/24278ab617/
+/webdriver/
 test/
-webdriver/
-static/24278ab617/
-static/wptd-metrics/
 
 # Configurations
 .eslintrc.json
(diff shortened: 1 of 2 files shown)


webapp/web/.gcloudignore
```diff
@@ -1,28 +1,35 @@
-# Don't upload test data, package config, etc. to AppEngine.
```
KyleJu: I think this line is still useful?

Hexcles: Hmm, yeah. I broke it into categories, but I should keep a summary line for this file.

webapp/web/.gcloudignore
```diff
@@ -1,28 +1,35 @@
...
-**/*_test.go
+# Tests
```
Hexcles:
```suggestion
# Files matching the following rules won't be uploaded to AppEngine (webapp).
# The syntax is the same as `.gitignore`; we do not use `.gitignore` directly
# because we need to update some generated files.

# Tests
```

webapp/web/.gcloudignore
```diff
@@ -1,28 +1,35 @@
...
+# Other services
+api/query/cache/service/
 results-processor/
-revisions/
-static/24278ab617/
-static/wptd-metrics/
-test/
-wct.conf.json
-webdriver/
+util/
```
KyleJu: util/ will ignore https://github.com/web-platform-tests/wpt.fyi/blob/master/webapp/components/test/util/helpers.js too?

Hexcles:
You're right. Although tests are already ignored anyway, this seems too
brittle and will cause confusion in the future. Let me add some leading
slashes to the rules.

webapp/web/.gcloudignore
KyleJu: actually that is a part of test/.

PR #1866, opened 2020-03-29
[UI] Triage Metadata Multi Select UI

KyleJu:
Triage Metadata Multi Select UI. Followup for #1746. Also filtering empty URLs before sending requests to /api/metadata/triage, per #1870

Follow-up:
1. Add the ability to triage subtest
2. Return a warning if a test is triaged


webapp/components/github-login.js
```diff
@@ -55,13 +67,19 @@ class GitHubLogin extends PolymerElement {
         type: String,
         value: null,
       },
+      isTriagedMode: {
```
KyleJu: Note to reviewers: I haven't decided what to do here. Ideally I want to make it work like a client-side flag in flags/.

KyleJu:
@Hexcles @stephenmcgruer any feedback is welcomed

KyleJu:
Seeing a [test](https://github.com/web-platform-tests/wpt.fyi/blob/e76420c06cd54520e3d1ec1c5486f762f66b5099/webdriver/search_test.go#L61) timeout but don't think it is related.
--- FAIL: TestSearch (150.13s)
    --- FAIL: TestSearch/wpt-results (136.24s)
        --- PASS: TestSearch/wpt-results/search-input (5.78s)
        --- FAIL: TestSearch/wpt-results/search-param (130.45s)
            search_test.go:67: 
                	Error Trace:	search_test.go:67
                	Error:      	Expected nil, but got: &errors.errorString{s:"timeout after 10.053547908s"}
                	Test:       	TestSearch/wpt-results/search-param
            search_test.go:84: 
                	Error Trace:	search_test.go:84
                	            				search_test.go:68
                	Error:      	Expected exactly 1 results
                	Test:       	TestSearch/wpt-results/search-param

KyleJu:
> Doing my high-level-pass-first approach again, hope that's ok :). Again from most concrete to least:
> 
> 1. There is no clear way to know whether you are in Triage Mode or not. When you click the button in the `?` overlay, it just clicks and then... nothing. This should have state; either a slider or a button that stays depressed (maybe changes color? I think I prefer a slider).
>    
>    * Sub-note: Note that whether or not triage mode is on, you see the same thing when you hover over a `1/2` test (because single-select triage is always on), which is confusing. I'm not yet sure whether we should disable single-select triage mode or change the UI when hovering, however.
> 2. In fact, it doesn't seem possible to leave triage mode?
> 3. Triage Mode should not be selectable (either greyed out or hidden entirely) when the user is not logged in.
> 4. I don't think reducing the opacity of a field is the right way to indicate it is selected. The most common pattern would be to give it some sort of clearly-differentiated border, right?
> 5. I think 'TRIAGE SELECTED' or just 'TRIAGE' would be clearer than 'VIEW TRIAGE' (the latter uses 'triage' as a noun, which implies the triage has already happened. The former uses it is as a verb).
> 6. Should there be a 'CANCEL' button on the 'X bugs triaged' bottom-left popup as well? Also, that reminds me; 'VIEW TRIAGE' isn't clearly a button which is probably bad for accessibility.
> 7. Can we be smart about what is selected (if possible) and change 'test' to 'directory' when the user selects a wildcard directory? Maybe this doesn't work because there are situations where you can select both a directory and a test...
> 8. I still think that making the user enter a bug URL for each test they've selected is the wrong flow. I'm open to landing this without fixing that, but if we do that I want to see a user-study shortly afterwards, and I think it would be a source of frustration for users.
> 9. I still think that hiding the triage mode button inside a `?` is not going to be great for people finding it. Again, open to landing it but would want a user study to follow. I suspect that when you say "Ok, enter triage mode now" they won't know where to look!
> 10. Triages go away when changing directory. Is that ok? Probably?
> 11. Should we error in some way if things are already triaged? Not sure? Probably not?

Comment 1-3: Done. Implemented a slider instead. One question is: the page needs to be refreshed after the triage mode is selected/de-selected. There are a few ways to do it - through data binding to refresh the layout or refresh the page entirely.
4:  Done. Changed it to a black border, but maybe there is better choice for colour? using Lighthouse?
5, 8 : Done. 

6: Not sure about this one. I didn't see a cancel button in [recent runs](https://wpt.fyi/runs?label=master&from=2020-03-30T00%3A00)
7, 9, 10: Not implemented for now.
11: Maybe we can do it on the server side?

stephenmcgruer:
> Comment 1-3: Done. Implemented a slider instead. One question for myself: the page needs to be refreshed after the triage mode is selected/de-selected. There are a few ways to do it - through data binding to refresh the layout or refresh the page entirely.

We should probably refresh the data bindings, refreshing the page is rather painful for the user :).

> 4: Done. Changed it to a black border, better choice for colour? using Lighthouse?

Great question. I had a look around a few articles online, but most used a 'tickbox' solution to select the entire row. What you have now might be ok; it looks good on a directory level view, but I think its affecting layout. I'm not sure what other options we have; CSS outline may be useful but will cause a double-thickness line if you select two cells above one another :/

> 5, 8 : Done.

Your solution to 8 looks great - and love that it supports multiple browsers too! Great job.

webapp/components/wpt-flags.js
```diff
@@ -101,8 +108,27 @@ wpt.FlagsClass = (superClass, readOnly, useLocalStorage) => class extends superC
...
+    super.ready();
+    if (useLocalStorage) {
+      for (const feature of wpt.ClientSideWriteableFeatures) {
+        this._createMethodObserver(`valueChangedInLocalStorage(${feature}, '${feature}')`);
+      }
+    }
+  }
+
+  valueChangedInLocalStorage(value, feature) {
+    window.location.reload();
```
KyleJu, line 127: Note to reviewer: this line ties to [this comment](https://github.com/web-platform-tests/wpt.fyi/pull/1866#issuecomment-609848180). Ideally we want to only refresh the layout of the page but there is no easy way to do, due to the fact that these flags are passed through mixins, instead of data bindings. Any suggestion is welcomed here

webapp/components/github-login.js
```diff
@@ -4,44 +4,64 @@
...
+import '../node_modules/@polymer/paper-toggle-button/paper-toggle-button.js';
+import { WPTFlags } from '../components/wpt-flags.js';
 
-class GitHubLogin extends PolymerElement {
+class GitHubLogin extends WPTFlags(PolymerElement) {
   static get template() {
     return html`
     <style>
-      paper-button {
+      .log-in-button {
```
Hexcles, line 22: Maybe `.login-button`?

webapp/components/github-login.js
```diff
@@ -4,44 +4,64 @@
...
+        vertical-align: baseline;
+      }
+      .triage-toggle {
+        display: inline;
+      }
     </style>
     <template is="dom-if" if="[[!user]]">
-      <paper-button raised onclick="[[logIn]]">
-      <iron-icon src="/static/github.svg"></iron-icon>
+     <iron-icon class="help" src="/static/help.svg" onclick="[[openHelpDialog]]"></iron-icon>
```
Hexcles, line 40:
```suggestion
      <iron-icon class="help" src="/static/help.svg" onclick="[[openHelpDialog]]"></iron-icon>
```

webapp/components/github-login.js
```diff
@@ -4,44 +4,64 @@
...
+      <paper-button class="log-in-button" raised onclick="[[logIn]]">
+      <iron-icon class="github-icon" src="/static/github.svg"></iron-icon>
       Sign in with GitHub
```
Hexcles, line 43:
```suggestion
        <iron-icon class="github-icon" src="/static/github.svg"></iron-icon>
        Sign in with GitHub
```

webapp/components/github-login.js
```diff
@@ -4,44 +4,64 @@
...
+      </paper-button>
     </template>
     <template is="dom-if" if="[[user]]">
-      <iron-icon src="/static/github.svg"></iron-icon>
-      [[user]]
-      <paper-icon-button title="Sign out" icon="exit-to-app" onclick="[[logOut]]">
+      <div>
+        <paper-toggle-button class="triage-toggle" checked="{{multiselectTriageUI}}">
+          Triage Mode
+         </paper-toggle-button>
```
Hexcles, line 50:
```suggestion
        </paper-toggle-button>
```

webapp/components/github-login.js
```diff
@@ -4,44 +4,64 @@
...
+        <iron-icon class="github-icon" src="/static/github.svg"></iron-icon>
+          [[user]]
```
Hexcles, line 52:
```suggestion
        [[user]]
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -36,19 +36,29 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
...
-          <paper-input label="Bug URL" value="{{url}}" autofocus></paper-input>
-        </div>
+        <h3>Triage Failing Tests</h3>
+        <template is="dom-repeat" items="[[displayedMetadata]]" as="node">
+          <div class="metadataEntry">
+            <img class="browser" src="[[displayLogo(node.product)]]">
+            &nbsp; : &nbsp; <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
+          </div>
+          <template is="dom-repeat" items="[[node.tests]]" as="test">
+           <li>[[test]]</li>
```
Hexcles, line 59:
```suggestion
            <li>[[test]]</li>
```

webapp/components/github-login.js
```diff
@@ -62,6 +82,7 @@ class GitHubLogin extends PolymerElement {
     super();
     this.logIn = this.handleLogIn.bind(this);
     this.logOut = this.handleLogOut.bind(this);
+    this.openHelpDialog = this.openHelpDialog.bind(this);
```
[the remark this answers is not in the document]
KyleJu: Good idea and worked! https://github.com/web-platform-tests/wpt.fyi/issues/1888

webapp/components/wpt-amend-metadata.js
KyleJu: I double checked other indentations and you caught them all. Thanks!

webapp/components/wpt-flags.js
KyleJu: Option 2 is less messy and easier, especially when there are multiple writeable client-side flags. But after some unsuccessful attempts doing the following in `wpt.results.js`

`   
  ready() {
    super.ready();
    this.addEventListener('storage', function(e) {
      if (e.key === 'multiselectTriageUI') {
        this.multiselectTriageUI = e.newValue;
        this.reloadData();
      }
    });
  }
`
I realized in https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API/Using_the_Web_Storage_API, it says

> This won't work on the same page that is making the changes — it is really a way for other pages on the domain using the storage to sync any changes that are made. Pages on other domains can't access the same storage objects.

and [another example](https://stackoverflow.com/questions/3562929/how-can-i-get-an-event-to-fire-every-time-localstorage-is-updated-in-safari-5). Guess I will go with option one for now until we

KyleJu:
@Hexcles ready for another review, except [the flexbox comment](https://github.com/web-platform-tests/wpt.fyi/pull/1866#discussion_r404425971). Let me play with flexbox a little bit first.

webapp/views/wpt-results.js
```diff
@@ -99,6 +100,9 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
         opacity: 0.7;
         box-shadow: 5px 5px 5px;
       }
+      td[selected] {
+        border: 2px solid #000000;
```
KyleJu, line 104: Note to myself: I have seen missing parts of borders before but couldn't reproduce today. Check again

webapp/views/wpt-results.js
```diff
@@ -404,9 +416,30 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
...
+      },
+      isTriagable: {
+        type: Boolean,
+        observer: 'isTriagableUpdated',
+      }
     };
   }
 
+  isTriagableUpdated(isTriagable) {
+    this.reloadData();
```
Hexcles, line 440: Why do we need to reload data here?

webapp/components/wpt-flags.js
```diff
@@ -101,8 +108,26 @@ wpt.FlagsClass = (superClass, readOnly, useLocalStorage) => class extends superC
   static get properties() {
     const props = {};
```
Hexcles, line 109: The `WPTFlags` mixin should have a base `isTriageMode` property, instead of defining it in every class. In fact, it kind of already has one: `multiselectTriageUI`, which should probably just be renamed to `isTriageMode` and saves a property.

webapp/components/wpt-flags.js
```diff
@@ -101,8 +108,26 @@ wpt.FlagsClass = (superClass, readOnly, useLocalStorage) => class extends superC
...
     makeFeatureProperties(props, wpt.ClientSideFeatures, readOnly, useLocalStorage);
+    if (useLocalStorage) {
```
Hexcles, line 111: Do we still need to condition on localStorage?

webapp/components/wpt-flags.js
```diff
@@ -101,8 +108,26 @@ wpt.FlagsClass = (superClass, readOnly, useLocalStorage) => class extends superC
...
+      makeFeatureProperties(props, wpt.ClientSideWriteableFeatures, false, useLocalStorage);
+    }
     return props;
   }
+
+  ready() {
+    super.ready();
+    if (useLocalStorage) {
+      for (const feature of wpt.ClientSideWriteableFeatures) {
+        this._createMethodObserver(`valueChangedInLocalStorage(${feature}, '${feature}')`);
```
Hexcles, line 121: This should be removed I think?

webapp/views/wpt-results.js
KyleJu: It is because we have to turn on/off `on-click` to each cell when the value of `isTriagable` changes

webapp/components/wpt-flags.js
KyleJu: Without it, `WPTEnvironmentFlagsEditor` (I believe that's the page for server-side flags) will also try to define this method, which is not that useful because it's a client-side flag only and we don't have to manage it?

webapp/components/wpt-flags.js
KyleJu: Without this line, the value of `isTriageMode` won't be saved when a user refresh the page?

Hexcles: Do we want that? I think we can either store every state (including the cells selected), or revert back to default (triage mode off, nothing selected) upon refresh. Remembering just the triage mode state seems not particularly useful and potentially confusing.

webapp/components/wpt-flags.js
KyleJu: Yep we shouldn't define it every time. AS for the naming, `multiselectTriageUI` is used for local storage by observing its values, and `isTriageMode` is used to trigger the `reload` through data bindings. If we use the same name for both, isn't it a conflict?

webapp/components/wpt-flags.js
KyleJu: Right, I can see that the triage mode should only be on when users trying to triage at the moment. Having it on by default doesn't make sense anyways (confusing on the UI)

webapp/components/wpt-flags.js
KyleJu: why don't we remove all the changes in `wpt-flags.js`, since we don't need storage anyways?

Hexcles: We can. This is arguably a different kind of "flag" (or rather, "state") that can go into a different mixin class.

webapp/components/github-login.js
```diff
@@ -4,44 +4,62 @@
...
 import { PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
 import '../node_modules/@polymer/paper-button/paper-button.js';
 import '../node_modules/@polymer/paper-menu-button/paper-menu-button.js';
 import '../node_modules/@polymer/iron-icon/iron-icon.js';
 import '../node_modules/@polymer/paper-styles/color.js';
+import '../node_modules/@polymer/paper-toggle-button/paper-toggle-button.js';
+import { WPTFlags } from '../components/wpt-flags.js';
 
-class GitHubLogin extends PolymerElement {
+class GitHubLogin extends WPTFlags(PolymerElement) {
```
Hexcles, line 18: Does this class till need to inherit from `WPTFlags`?

webapp/components/wpt-amend-metadata.js
```diff
@@ -36,19 +36,29 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
...
-        <div class="metadataEntry">
-          <img class="browser" src="[[displayLogo(product)]]">
-          &nbsp; >> [[test]] : &nbsp;
-          <paper-input label="Bug URL" value="{{url}}" autofocus></paper-input>
-        </div>
+        <h3>Triage Failing Tests</h3>
+        <template is="dom-repeat" items="[[displayedMetadata]]" as="node">
+          <div class="metadataEntry">
+            <img class="browser" src="[[displayLogo(node.product)]]">
+            &nbsp; : &nbsp; <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
```
Hexcles, line 56: Nit: are you trying to work around the whitespace folding to add more visual space here? Perhaps it's better to add some padding/margin in CSS instead.

webapp/views/wpt-results.js
```diff
@@ -1004,6 +1033,46 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
...
+    const td = e.target.closest('td');
+    var test = td.getAttribute('data-test');
+    const browser = this.products[td.getAttribute('data-index')].browser_name;
+    if (this.computePathIsASubfolder(test)) {
+      test = test + '/*';
+    }
+
+    if (this.selectedMetadata.find(s => s.test === test && s.product === browser)) {
+      this.selectedMetadata = this.selectedMetadata.filter(s => !(s.test === test && s.product === browser));
+      td.removeAttribute('selected');
```
Hexcles, line 1062: Shall we remove this cell from `this.selectedCells` here? I know it doesn't make a difference (we'll simply try to unset the property again in `clearSelectedCells`), but this can be a footgun in the future if one starts to rely on this property being up to date.

webapp/views/wpt-results.js
```diff
@@ -497,13 +530,9 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
...
-    this.openAmendMetadata = (i, node) => {
-      return () => {
-        const amend = this.$.amend;
-        amend.test = node.path;
-        amend.productIndex = i;
-        amend.open();
-      };
+    this.openAmendMetadata = () => {
+      const amend = this.$.amend;
+      amend.open();
```
Hexcles, line 535:
```suggestion
      this.$.amend.open();
```

In fact, now that `openAmendMetadata` does not take any parameters, you can define this as a class method and use `on-click` instead.

webapp/views/wpt-results.js
```diff
@@ -218,7 +230,7 @@ class WPTResults extends WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase
 
                 <template is="dom-repeat" items="{{testRuns}}" as="testRun">
                   <template is="dom-if" if="[[ canAmendMetadata(node, index, testRun) ]]">
-                    <td class\$="numbers triage [[ testResultClass(node, index, testRun, 'passes') ]]" onclick="[[openAmendMetadata(index, node)]]">
+                    <td class\$="numbers triage [[ testResultClass(node, index, testRun, 'passes') ]]" data-index$="[[index]]" data-test$="[[node.path]]" on-click="handleSelectMetadata">
```
Hexcles, line 233: `handleSelectMetadata` should probably be defined the same way as the previous `openAmendMetadata`: a function factory that takes `index` and `node` and returns a function, this avoids having to store lots of data attributes on the DOM.

Hexcles:
@KyleJu as discussed earlier in our 1:1, let's add some tests for those methods that transform/generate data (e.g. `getTriagedMetadataMap`, `populateDisplayData`). And then this is good to go!

KyleJu:
> @KyleJu as discussed earlier in our 1:1, let's add some tests for those methods that transform/generate data (e.g. `getTriagedMetadataMap`, `populateDisplayData`). And then this is good to go!

Sorry I am having difficulty rebase the last few commits (all tests). Somehow I am getting a bunch of conflicts on go.mod, even though I pulled from the master. Ready for the final stamp

webapp/components/test/test-wpt-amend-metadata.html
```diff
@@ -0,0 +1,59 @@
...
+
+import '../wpt-amend-metadata.js';
+suite('<wpt-amend-metadata>', () => {
+  let appFixture = null;
+  setup(() => {
+    appFixture = fixture('wpt-amend-metadata-fixture');
+  });
+
+  test('getTriagedMetadataMap(displayedMetadata)', () => {
+    let displayedMetadata = [
```
Hexcles, line 29:
```suggestion
    const displayedMetadata = [
```

webapp/components/test/test-wpt-amend-metadata.html
```diff
@@ -0,0 +1,59 @@
...
+      { product: 'firefox', url: 'foo', tests: ['testA', 'testB'] },
+      { product: 'chrome', url: 'bar', tests: ['testA', 'testB'] }
+    ];
+    let expected = {
```
Hexcles, line 33:
```suggestion
    const expected = {
```

webapp/components/test/test-wpt-amend-metadata.html
```diff
@@ -0,0 +1,59 @@
...
+    assert.deepEqual(appFixture.getTriagedMetadataMap(displayedMetadata), expected);
+  });
+  test('populateDisplayData()', () => {
+    appFixture.selectedMetadata = [
+      { product: 'firefox', test: 'testA' },
+      { product: 'firefox', test: 'testB' },
+      { product: 'chrome', test: 'testA' },
+      { product: 'chrome', test: 'testB' }
+    ];
+    let expected = [
```
Hexcles, line 47:
```suggestion
    const expected = [
```

webapp/components/test/test-wpt-amend-metadata.html
```diff
@@ -0,0 +1,59 @@
+<!doctype html>
```
Hexcles, line 1: This file should be named simply `wpt-amend-metadata.html`, without the `test-` prefix (which is already implied by the directory name.)

PR #1883, opened 2020-04-08
Collect crashtest results from Taskcluster

Hexcles:
We forgot to add crashtest into the regex which is used to filter only
the recognized task names (and exclude e.g. Python unit tests).

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -28,7 +28,7 @@ const flagPendingChecks = "pendingChecks"
 var (
 	// This should follow https://github.com/web-platform-tests/wpt/blob/master/.taskcluster.yml
 	// with a notable exception that "*-stability" runs are not included at the moment.
-	taskNameRegex = regexp.MustCompile(`^wpt-(\w+-\w+)-(testharness|reftest|wdspec|results|results-without-changes)(?:-\d+)?$`)
+	taskNameRegex = regexp.MustCompile(`^wpt-(\w+-\w+)-(testharness|reftest|wdspec|crashtest|results|results-without-changes)(?:-\d+)?$`)
 	// Taskcluster has used different forms of URLs in their Check & Status
 	// updates in history. We accept all of them.
 	// See TestExtractTaskGroupID for examples.
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -26,7 +26,7 @@ const flagTaskclusterAllBranches = "taskclusterAllBranches"
 const flagPendingChecks = "pendingChecks"
 
 var (
-	// This should follow https://github.com/web-platform-tests/wpt/blob/master/.taskcluster.yml
+	// This should follow https://github.com/web-platform-tests/wpt/blob/master/tools/ci/tc/tasks/test.yml
 	// with a notable exception that "*-stability" runs are not included at the moment.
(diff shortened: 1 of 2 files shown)


api/taskcluster/webhook.go
```diff
@@ -28,7 +28,7 @@ const flagPendingChecks = "pendingChecks"
 var (
 	// This should follow https://github.com/web-platform-tests/wpt/blob/master/.taskcluster.yml
```
stephenmcgruer, line 29: This file no longer defines the types of tests, if that is what this comment is meant to imply

Hexcles:
```suggestion
	// This should follow https://github.com/web-platform-tests/wpt/blob/master/tools/ci/tc/tasks/test.yml
```

Hexcles:
> Any thoughts on a way we could avoid this in the future? Maybe log a warning somewhere we'd see if it we see a TC entry we don't recognize?

That's a great question! I read the code again and I think we can actually match the following pattern to get all WPT tasks (i.e. filtering out Python unit tests etc.) and then skip the `stability` task:

`^wpt-([a-z]+-[a-z]+)-([a-z]+(?:-[a-z]+)*)(?:-\d+)?$`

which can match all of the following:
* `wpt-chrome-dev-stability`
* `wpt-chrome-dev-results-without-change`
* `wpt-chrome-dev-testharness-1`

I know I have a tendency of writing complicated regex as a vim user... Does this regex seem too much, @stephenmcgruer ?

stephenmcgruer:
> `^wpt-([a-z]+-[a-z]+)-([a-z]+(?:-[a-z]+)*)(?:-\d+)?$`

Does this accept `wpt-foo-bar--1`, or am I misreading? Also `wpt-foo-bar-`. That might be ok, but wanted to poke at the edge cases to see.

> I know I have a tendency of writing complicated regex as a vim user... Does this regex seem too much, @stephenmcgruer ?

I think it is 'fine', but does it really answer the goal of 'get all WPT tasks'? All it takes is TC sending `wpt-foobar` (or, say `wpt-chrome-testharness-1`, I think?) and we'd miss it.

Now, this is a totally grey area - it's all a question of how defensive we want to be in detecting 'unknown' TC tasks. If we think that all the TC tasks we care about are going to be `wpt-browser-name-task-name-1`, or close variants thereof, then this is fine. If we want to be extremely defensive we should look at even the unittest cases and see.

Also if we're being very defensive (and if we don't do this already) we could look at the reverse and warn if there are tasks that we expect that don't appear. But that may not work due to e.g. TC runs from PRs.

Hexcles:
> Does this accept `wpt-foo-bar--1`, or am I misreading? Also `wpt-foo-bar-`. That might be ok, but wanted to poke at the edge cases to see.

No, it does not accept either. https://play.golang.org/p/Dn90TZBz_Cm And this is partly why I was worried it'd be a bit excessive; it's not obvious at all. Maybe I can include these examples in the unit test.

> I think it is 'fine', but does it really answer the goal of 'get all WPT tasks'? All it takes is TC sending `wpt-foobar` (or, say `wpt-chrome-testharness-1`, I think?) and we'd miss it.

Right. It's not bulletproof but I think it might strike a better balance that handles new types of tests with the possibility of accepting false positives, which hopefully would be rare.

stephenmcgruer:
> No, it does not accept either. https://play.golang.org/p/Dn90TZBz_Cm And this is partly why I was worried whether the regex would be a bit excessive; it's not obvious at all. Maybe I can include these examples in the unit test.

Ah, I missed on first pass that the middle group isn't optional.

I think it would be fine, with some documentation nearby on what it intends to match.

PR #1889, opened 2020-04-14
Remove the "wpt.fyi is a work in progress" warning

foolip:
Continuing to show this caveat prominently isn't really helpful.


stephenmcgruer:
I support this, but I leave the final decision to @Hexcles .

PR #1894, opened 2020-04-15
[webapp] Remove wpt-flags from document head

KyleJu:


webapp/templates/admin_flags.html
```diff
@@ -2,8 +2,6 @@
 <head>
   <meta charset="utf-8">
   <title>wpt.fyi - Feature flags</title>
-
-  <script type="module" src="/components/wpt-flags.js"></script>
```
Hexcles, line 6: I think you still need this. This template does not include `_head_common.html`.

KyleJu: I don't see why we need this. This template is used for [admin_handler](https://github.com/web-platform-tests/wpt.fyi/blob/1d7c7247617cd127a3f791e9e20997ece9eb82fb/webapp/admin_handler.go#L44) and I thought we don't need wpt-flags in `<head>`

Hexcles: Without this, the `/admin/flags` page is broken -- `<wpt-environment-flags-editor>` does not exist.

KyleJu: You are right. wpt-flags.js reference is missing in that case.

Issue #1834, opened 2020-03-04
[Login] OAuth Application API is Deprecating 


KyleJu:
[OAuth Application API is Depreciating](https://developer.github.com/changes/2020-02-14-deprecating-oauth-app-endpoint/). We need to move to the new endpoint by July 1st

KyleJu:
Plan to tackle it in Q2 as a P1.

KyleJu:
This issue has been resolved in #1839.

GitHub is replacing `GET /applications/:client_id/tokens/:access_token` with `POST /applications/:client_id/token`, which affects https://github.com/web-platform-tests/wpt.fyi/blob/111641e0cefbfee98cda2c83a7ba360bf7d31668/shared/github_oauth.go#L151. However the underlying [go-github library](https://github.com/google/go-github) has implemented [this change](https://github.com/google/go-github/issues/1331) and included in release [v29.0.3](https://github.com/google/go-github/compare/v29.0.3...master).

There is no breaking changes in the [implementation](https://github.com/google/go-github/pull/1461/files#diff-fa118b5047b081bf147238afc4f38b36R146) and hence this issue is resolved.

KyleJu:
@LukeZielinski

Hexcles:
> On April 14th, 2020 at 08:38 (PDT), your application (wpt-metadata (staging)) issued a request using the deprecated form of OAuth Application API that includes access tokens in the URL path. GitHub has deprecated these endpoints and replaced them with a version that accepts access tokens in the request body.

I wonder what is still using this API then? #1839 landed in March.

LukeZielinski:
@Hexcles Where is that message coming from? Logs? Email?

Hexcles:
That was from an email from GitHub (delivered to my chromium.org account and I'm not quite sure why -- maybe I registered this GitHub OAuth application originally).

I did a little digging. This is the change we want: https://github.com/google/go-github/commit/419f07c399bbced679719c419ae683618018556c And it's in v31 according to the tag.

@KyleJu could you try rolling `go-github` to v31? Note that according to the [release logs](https://github.com/google/go-github/releases) there might be some breaking changes.

Hexcles:
I think the confusion here was caused by a misunderstanding in the [revision selection](https://git-scm.com/docs/gitrevisions) syntax of git.

[...]

PR #1898, opened 2020-04-16
Upgrade go-github from v29 to v31

LukeZielinski:
The intent is to pull in the new oauth implementation.

Ran the following commands:
$ find . -name *.go | xargs sed -i 's|github.com/google/go-github/v29/github|github.com/google/go-github/v31/github|'
$ go get -t -u -v -d ./...
$ go mod tidy


shared/triage_metadata.go
```diff
@@ -108,10 +108,11 @@ func (tm triageMetadata) getCommitBranchRef() (ref *github.Reference, err error)
 func (tm triageMetadata) getTree(ref *github.Reference, triagedMetadataMap map[string][]byte) (tree *github.Tree, err error) {
 	client := tm.githubClient
 
-	entries := []github.TreeEntry{}
+	entries := []*github.TreeEntry{}
 	for folderPath, content := range triagedMetadataMap {
 		dest := GetMetadataFilePath(folderPath)
-		entries = append(entries, github.TreeEntry{Path: github.String(dest), Type: github.String("blob"), Content: github.String(string(content)), Mode: github.String("100644")})
+		newEntry := github.TreeEntry{Path: github.String(dest), Type: github.String("blob"), Content: github.String(string(content)), Mode: github.String("100644")}
```
LukeZielinski, line 114: Not too sure whether newEntry is needed here - I was getting "literal is not a type" errors when trying to inline creation of this TreeEntry point inside the call to append, and I couldn't really make sense of it. Open to suggestions.

Hexcles: That's interesting. Is that the full error message?

For example, this works: https://play.golang.org/p/aNPerJmBrRS

LukeZielinski: Ahh, ok, I was just thinking of this in C++ style (where taking the address of a stack-allocated variable is bad news). I was calling `append(entries, new(github.TreeEntry{...` which was throwing that error. Updated to your suggestion, which seems to work.

Hexcles: Right, `new` only takes a type. Stack and heap are basically hidden away in Golang by GC and the memory model is closer to Java. Even if the object was indeed created on the stack, GC would be responsible for moving it out and updating all referencing pointers.

Just keep in mind don't cast the pointer to untyped `uintptr`, at which point there'll be no guarantees.

KyleJu:
This should fix https://github.com/web-platform-tests/wpt.fyi/issues/1834

PR #1914, opened 2020-04-24
Fix "too many runs" in /api/runs

### LINKED ISSUE #1913: /api/runs?product=safari-100%20preview returns no runs
There are thousands of `safari-100 preview` runs, but this API returns 0 with no errors.

`/api/runs?product=safari-105%20preview` works, as well as `/api/runs?product=safari-100%20preview&label=master`.

Hexcles:
## Description

This fixes #1913 .

The root cause is that when the filter is too broad (e.g. without `label=master`), we will find thousands of candidates for further filtering in `clientSideFilter`, which exceeds the limit of `GetMulti` (up to 1000 entities per call). This PR sets a limit on the number of candidates, and adds a warning to logs.

Note that this is not a perfect fix. It is possible that we will miss runs. Due to Datastore limitations (inequality can only be used on one field), we cannot filter by timestamps here if we are already filtering by some other criterias, so we really need to get *all* runs matching all the other filters to make sure we don't miss any in the time range. However, this was already broken before and I don't see a good way to fix it properly without refactoring (once again) the query system.

## Review Information

The first commit is somewhat unrelated. I originally thought the bug was in parsing `safari-104 preview`, so I added a bunch of test cases which all passed, but they are still good to have anyway.

The root problem is somewhat difficult to test (it'd probably require adding thousands of runs to the test server). However, the bug would have been discovered easily through server errors (500) if we had checked the error returned by `clientSideFilter`. This error check is now added to prevent future regressions.

https://too-many-runs-dot-wptdashboard-staging.uk.r.appspot.com/api/runs?product=safari-100%20preview&max-count=10
https://wpt.fyi/api/runs?product=safari-100%20preview&max-count=10

### DIFF
--- a/shared/datastore.go
+++ b/shared/datastore.go
@@ -14,6 +14,10 @@ import (
 // ErrEntityAlreadyExists is returned by Datastore.Insert when the entity already exists.
 var ErrEntityAlreadyExists = errors.New("datastore: entity already exists")
 
+// MaxKeysPerLookup is the max number of keys allowed per lookup (e.g. GetMulti).
+// https://cloud.google.com/datastore/docs/concepts/limits
+const MaxKeysPerLookup = 1000
+
 // Key abstracts an int64 based datastore.Key
 type Key interface {
 	IntID() int64
--- a/shared/test_run_query.go
+++ b/shared/test_run_query.go
@@ -223,6 +223,8 @@ func clientSideFilter(
 	for key := range productIDFilter.Iter() {
 		keys = append(keys, store.NewIDKey("TestRun", key.(int64)))
 		if len(keys) == capacity {
+			// FIXME: This might produce incomplete results.
+			// https://github.com/web-platform-tests/wpt.fyi/pull/1914
 			break
 		}
 	}
--- a/shared/test_run_query.go
+++ b/shared/test_run_query.go
@@ -171,6 +171,9 @@ func (t testRunQueryImpl) LoadTestRunKeys(
 		var keys []Key
 		if productIDFilter != nil {
 			keys, err = clientSideFilter(t.store, product, productIDFilter, from, to, limit)
+			if err != nil {
+				return nil, err
+			}
 		} else {
 			// Otherwise, just run a "GetAll" filter. Expensive.
 			log.Debugf("Falling back to GetAll datastore query.")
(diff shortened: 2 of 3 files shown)


stephenmcgruer:
@KyleJu do you have cycles to review this?

KyleJu:
> @KyleJu do you have cycles to review this?

Happy to do so

shared/datastore.go
```diff
@@ -14,6 +14,10 @@ import (
 // ErrEntityAlreadyExists is returned by Datastore.Insert when the entity already exists.
 var ErrEntityAlreadyExists = errors.New("datastore: entity already exists")
 
+// MaxKeysPerLookup is the max number of keys allowed per lookup (e.g. GetMulti).
+// https://cloud.google.com/datastore/docs/concepts/limits
```
KyleJu, line 18: This link sends me to `Firestore in Datastore mode limits` page. I don't think we use Firestore?

shared/test_run_query.go
```diff
@@ -210,10 +213,18 @@ func clientSideFilter(
 	to *time.Time,
 	limit *int) (keys []Key, err error) {
 	log := GetLogger(store.Context())
-	log.Debugf("Loading %v viable runs to filter them.", productIDFilter.Cardinality())
-	keys = make([]Key, 0, productIDFilter.Cardinality())
+	capacity := productIDFilter.Cardinality()
+	if productIDFilter.Cardinality() > MaxKeysPerLookup {
+		log.Warningf("Too many viable runs: %d>%d", productIDFilter.Cardinality(), MaxKeysPerLookup)
```
KyleJu, line 218:
```suggestion
		log.Warningf("%d viable runs exceed lookup limit %d", productIDFilter.Cardinality(), MaxKeysPerLookup)
```

KyleJu: how about something like this?

shared/datastore.go
Hexcles: No we don't use native Firestore. However, https://cloud.google.com/datastore/docs/firestore-or-datastore#automatic_upgrade_to_datastore_mode so this is the technically correct term for Datastore now.

shared/test_run_query.go
```diff
@@ -210,10 +213,18 @@ func clientSideFilter(
...
+	capacity := productIDFilter.Cardinality()
+	if productIDFilter.Cardinality() > MaxKeysPerLookup {
+		log.Warningf("%d viable runs exceed the lookup limit %d", productIDFilter.Cardinality(), MaxKeysPerLookup)
+		capacity = MaxKeysPerLookup
+	}
+	log.Debugf("Loading %d viable runs to filter.", capacity)
+	keys = make([]Key, 0, capacity)
 	for key := range productIDFilter.Iter() {
 		keys = append(keys, store.NewIDKey("TestRun", key.(int64)))
+		if len(keys) == capacity {
```
KyleJu, line 225: Maybe add a note here that runs might be missing?

Hexcles:
```suggestion
		if len(keys) == capacity {
			// FIXME: This might produce incomplete results.
			// https://github.com/web-platform-tests/wpt.fyi/pull/1914
```

Yeah. Ideally we should tell the users, but short of that we need at least a comment here.

Hexcles:
> What stops us from using store.GetMulti(keys, runs) twice? That would consider as two lookups?

That's actually a good question!

We could in theory get all the runs by querying one batch of 1000 at a time. However, we might end up fetching tens of thousands of runs just to return up to 500 runs (the maximum allowed `max-count` is 500), which is a perfect recipe for amplification DoS.

Let me think about how/if we can refactor this part more properly and take this quick fix for now.

PR #1925, opened 2020-04-29
[taskcluster] Allow underscore (_) for the browser name identifier.

clopez:
## Description
Since commit b8fac0b the webkitgtk runs from taskcluster are not longer updated to wpt.fyi
This is caused because the browser identifier on TC has an underscore.
This fixes: https://github.com/web-platform-tests/wpt/issues/23313

## Review Information
Tested by:
```
go test -tags=small -v  ./... -run TestTaskNameRegex
```


clopez:
I think the CI failure on ```Integration / web_components_test (pull_request)``` its unrelated.
The `go_test` one passed

PR #1927, opened 2020-04-30
[Triage Metadata] Fix race condition due to wrong parent SHA

### LINKED ISSUE #1890: Triaging can overwrite existing metadata
https://github.com/web-platform-tests/wpt-metadata/pull/147 overwrote the changes from https://github.com/web-platform-tests/wpt-metadata/pull/146. It looks it just created the META.yml file and replaced what was in the repo.

KyleJu:
Fix #1890. In addition, it changes the test setup for metadata-related tests by mocking MetadataFetcher interface.

This PR changes the signature of `Fetcher` interface to return a GitHub master `sha` and plumbs it to `webapp`. `searchcache` remains unchanged.

In terms of the test setup, it changes the old setup where each test creates a local server that serves gzip_testfile.tar.gz.


api/metadata_cache.go
```diff
@@ -17,52 +18,79 @@ import (
...
 	}
 
-	return shared.CollectMetadataWithURL(f.client, f.url)
+	shaKey, err := f.gitHubUtil.GetWPTMetadataMasterSHA()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// CollectMetadataWithURL retrieves the content of the wpt-metadata repo from master by default.
+	res, err = shared.CollectMetadataWithURL(f.client, f.url)
```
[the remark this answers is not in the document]
Hexcles: In addition, I know this was already here before, but would we ever hit this code? `getMetadataFromMemcache` also calls `CollectMetadataWithURL` if cache is missed, so it seems to me we have two duplicate fallback paths.

api/query/metadata_cache.go
```diff
@@ -18,13 +18,14 @@ type searchcacheMetadataFetcher struct {
...
-		return MetadataMapCached, nil
+		return nil, MetadataMapCached, nil
 	}
 
 	var netClient = &http.Client{
 		Timeout: time.Second * 5,
 	}
-	return shared.CollectMetadataWithURL(netClient, f.url)
+	res, err = shared.CollectMetadataWithURL(netClient, f.url)
+	return nil, res, err
```
Hexcles, line 30: We should at least add a TODO here.

shared/github_util.go
```diff
@@ -0,0 +1,41 @@
...
+	baseRef, _, err := g.client.Git.GetRef(g.ctx, sourceOwner, sourceRepo, "refs/heads/"+baseBranch)
+	if err != nil {
+		return nil, err
+	}
+
+	return baseRef.Object.SHA, nil
+}
+
+// GetGitHubUtil returns an instance of the GitHubUtil interface.
+func GetGitHubUtil(ctx context.Context, client *github.Client) GitHubUtil {
```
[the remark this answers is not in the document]
Hexcles: And I just realized you also had `GetGitHubClient`; we should also rename that, too.

api/metadata_cache.go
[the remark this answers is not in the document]
KyleJu: Using `https://developer.github.com/v3/repos/contents/#get-archive-link` was also my origin consideration, however there are a slight complications here:

1. IIUC, this API only takes ref, which is a branch name. Even if the race condition happens, we have no way to refer to the `master` we get from `GetWPTMetadataMasterSHA` because `master` has changed. Also there is no way to get an archive from a `sha` for this [reason](https://stackoverflow.com/questions/26135216/why-isnt-there-a-git-clone-specific-commit-option/26135822#26135822)

2.Retrieving metadata this way kind of circles back to the origin issue with testing in searchcache. In metadata [search atoms](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/atoms.go#L241), there is no way to pass in a mock object. We certainly don't have to make `webapp` and `searchcahe` work the same way (leave searchcache as it is), but I also don't like the inconsistency.

3. The race condition is also very unlikely to happen. These functions are almost called side-by-side. But of course, I would rather not have it.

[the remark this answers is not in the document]
KyleJu: For your second comment, I think if it hits any of the `err != nil` case, such as [this one](https://github.com/web-platform-tests/wpt.fyi/pull/1927/files/74efe85420669ae0721eea02829e8ade45f6e762#diff-2813b30cfdffd1a9bf51da1d3bca7e06R49), this "second" fall back will hit. I would guess, the cache hits but the data is corrupted? So it's not a miss but `Unmarshal` would fail

api/query/metadata_cache.go
KyleJu: Just to clarify, `TODO` for getting the `sha` right? because we are not dealing with the race condition in the searchcache yet.

api/metadata_cache.go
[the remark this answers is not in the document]
Hexcles: A SHA is also a valid ref. This seems to work: https://api.github.com/repos/web-platform-tests/wpt-metadata/tarball/69157cf699ab7a279b3c7a3a546ed9674c290996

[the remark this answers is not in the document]
Hexcles: > 

For your second comment, I think if it hits any of the err != nil case, such as this one, this "second" fall back will hit. I would guess, the cache hits but the data is corrupted? So it's not a miss but Unmarshal would fail

Good point, but I think these two fallbacks should and can be consolidated.

api/query/metadata_cache.go
Hexcles: Right.

shared/github_util.go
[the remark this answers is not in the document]
KyleJu: `GetGitHubClient` used in [too many places](https://github.com/web-platform-tests/wpt.fyi/search?q=GetGitHubClient&type=). I will do it in a separate PR

shared/metadata.go
```diff
@@ -57,7 +57,7 @@ func GetMetadataResponse(testRuns []TestRun, log Logger, fetcher MetadataFetcher
 		productSpecs[i] = ProductSpec{ProductAtRevision: run.ProductAtRevision, Labels: run.LabelsSet()}
 	}
 
-	metadata, err := GetMetadataByteMap(log, fetcher)
+	_, metadata, err := GetMetadataByteMap(log, fetcher)
```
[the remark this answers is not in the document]
KyleJu: Agree. It will be hard to add SHA with the current format, (`MetadataResults` is a map). I have filed https://github.com/web-platform-tests/wpt.fyi/issues/1938 and added TODOs

api/metadata_cache.go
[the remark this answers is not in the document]
KyleJu: I actually spent lots of time googling ref vs SHA but should have actually just tried it. Thanks for it! Addressed both of your comments.

shared/github_util.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/google/go-github/v31/github"
+)
+
+const sourceOwner string = "web-platform-tests"
+const sourceRepo string = "wpt-metadata"
+const baseBranch string = "master"
+
+// GitHubUtil encapsulates the implementation of accessing the information in the wpt-metadata repo.
+type GitHubUtil interface {
```
Hexcles, line 20: I think this interface should probably be moved into `metadata_util.go`, named as `MetadataUtil`, and include the existing `CollectMetadataWithURL` function, too. These two functions are both utility functions for the `wpt-metadata` repo so it's weird to have them scattered in different places.

shared/triage_metadata.go
```diff
@@ -265,14 +257,14 @@ func generateRandomInt() string {
 }
 
 func (tm triageMetadata) Triage(metadata MetadataResults) (string, error) {
-	filesMap, err := GetMetadataByteMap(tm.logger, tm.fetcher)
+	sha, filesMap, err := GetMetadataByteMap(tm.logger, tm.fetcher)
```
Hexcles, line 260: We talked about force refreshing the cache before creating the triage PR to minimize the chance of race conditions. Do you still plan to do that?

api/metadata_cache.go
```diff
@@ -18,16 +18,19 @@ import (
...
 type webappMetadataFetcher struct {
-	ctx        context.Context
-	client     *http.Client
-	url        string
-	gitHubUtil shared.GitHubUtil
+	ctx           context.Context
+	client        *http.Client
+	url           string
+	gitHubUtil    shared.GitHubUtil
+	isFetchMaster bool
```
KyleJu, line 25: isFetchMaster/forceUpdate/g

shared/triage_metadata.go
KyleJu: See next [commit](https://github.com/web-platform-tests/wpt.fyi/pull/1927/commits/aeb0f22529d93f6553b8380db765b20a8d96d940)

KyleJu:
Note to the reviewer @Hexcles:
The quickest way to review it is one commit at a time after your [LGTM comment](https://github.com/web-platform-tests/wpt.fyi/pull/1927#pullrequestreview-406211849). Before change 4a458a5, I addressed all your comments, except [merging GitHubUtil into MetadataUtil](https://github.com/web-platform-tests/wpt.fyi/pull/1927#discussion_r420481048). 4a458a5 addresses discussions in our 1:1, which are in summary

- It changes the metadata-related test setup using MetadataFetcher mock, instead of a local test server.

-It removes the underlying MetadataUtil interface and changed them to public methods

shared/metadata_util.go
```diff
@@ -2,27 +2,52 @@
...
+
+	return baseRef.Object.SHA, nil
 }
 
-// CollectMetadataWithURL iterates through wpt-metadata repository and returns a
-// map that maps a test path to its META.yml file content, using a given URL.
-func CollectMetadataWithURL(client *http.Client, url string) (res map[string][]byte, err error) {
+// GetWPTMetadataRepoData iterates through wpt-metadata repository and returns a
+// map that maps a test path to its META.yml file content, using a given ref.
+func GetWPTMetadataRepoData(client *http.Client, ref *string) (res map[string][]byte, err error) {
```
Hexcles, line 44: Nit: maybe call this GetWPTMetadataArchive

api/metadata_handler.go
```diff
@@ -45,19 +51,29 @@ func apiMetadataTriageHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	githubBotClient, err := shared.GetGithubClientFromToken(ctx, "github-wpt-fyi-bot-token")
+	wptfyiBotClient, err := shared.GetGithubClientFromToken(ctx, "github-wpt-fyi-bot-token")
 	if err != nil {
-		http.Error(w, "Unable to get Github Client: "+err.Error(), http.StatusInternalServerError)
+		http.Error(w, "Unable to get wptfyiBotClient: "+err.Error(), http.StatusInternalServerError)
```
Hexcles, line 56:
```suggestion
		http.Error(w, "Unable to get GitHub client for wpt-fyi-bot: "+err.Error(), http.StatusInternalServerError)
```

api/metadata_handler.go
```diff
@@ -45,19 +51,29 @@ func apiMetadataTriageHandler(w http.ResponseWriter, r *http.Request) {
...
+	gitHubClient, err := shared.NewAppEngineAPI(ctx).GetGitHubClient()
+	if err != nil {
+		http.Error(w, "Unable to get Github client: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	fetcher := webappMetadataFetcher{
+		ctx:          ctx,
+		httpClient:   aeAPI.GetHTTPClient(),
+		gitHubClient: gitHubClient,
```
Hexcles, line 72: Why don't we just use `wptfyiBotClient` here? It has broader scopes but the fetcher is read-only anyway.

api/metadata_handler_test.go
```diff
@@ -109,55 +109,36 @@ func TestHandleMetadataTriage_InvalidBody(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, w.Code)
 }
 
-func TestFilterMetadataHanlder_Success(t *testing.T) {
-	handler := func(w http.ResponseWriter, r *http.Request) {
-		http.ServeFile(w, r, "../shared/metadata_testdata/gzip_testfile.tar.gz")
-	}
-	server := httptest.NewServer(http.HandlerFunc(handler))
-	defer server.Close()
+func TestMetadataHanlder_GET_Success(t *testing.T) {
```
Hexcles, line 112: But now we no longer have tests for the fetcher itself including the decompression and deserialization logic. Could you move this test with a server to `metadata_util_test.go`?

api/metadata_handler.go
KyleJu: Yea that would work here. I was worried that at peak wptfyiBotClient is more likely to reach the GitHub request limit

api/metadata_handler_test.go
KyleJu: Good point!

PR #1935, opened 2020-05-04
[UI] Add UI components to triage subtests

KyleJu:
A follow-up for https://github.com/web-platform-tests/wpt.fyi/pull/1866.

- Add the ability to triage subtests. 

- Refactor and create a mixin for wpt-amend-metadata related properties and methods. 

- Remove the wpt-result reloading behavior when enable Metadata Triage UI toggle.


webapp/components/test-file-results-table.js
```diff
@@ -116,16 +132,29 @@ class TestFileResultsTable extends WPTColors(PathInfo(TestRunsBase)) {
...
+                <a class="ref-button" href="[[ computeAnalyzerURL(result.screenshots) ]]">
+                  <iron-icon icon="image:compare"></iron-icon>
+                  COMPARE
+                </a>
+              </template>
+            </td>
+          </template>
+
+          <template is="dom-if" if="[[ canAmendMetadata(result.status) ]]">
+            <td class$="[[ colorClass(result.status) ]]" onclick="[[handleSelectMetadata(index, row.name, result.status)]]">
```
Hexcles, line 148: I know this is how we did it previously, but since this is the second time around I want to take some time to minimize the code duplication.

Is the event handler the only difference between the two branches? We could always bind the event listener and move the branch into the function.

webapp/components/test-file-results-table.js
```diff
@@ -184,6 +214,23 @@ class TestFileResultsTable extends WPTColors(PathInfo(TestRunsBase)) {
...
+        value: [],
+      },
+      testPlural: {
+        type: String,
+        computed: 'computeTestPlural(selectedMetadata)',
+      },
+      isTriageMode: {
+        type: Boolean,
+        observer: 'isTriageModeUpdated',
+      },
```
Hexcles, line 233: How many of these new properties and methods are the same as the ones in `wpt-results.js`? Could we create a mixin?

webapp/components/wpt-amend-metadata.js
```diff
@@ -84,6 +85,11 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
         type: Array,
         value: []
       },
+      // This testStatusValues mapping is defined at shared/statuses.go.
+      testStatusValues: {
+        type: Object,
+        value: { 'FAIL': 6, 'TIMEOUT': 4, 'ERROR': 3 }
```
Hexcles, line 91: The triage API expects a numerical value for test status? These numbers are supposed to be internal.

KyleJu: Per our 1:1, the `status` field is optional so I will investigate the original motives of using numeral values vs strings and add it in a separate PR.

KyleJu:
@Hexcles Cannot think of a better way to deal with the `rebase` issue. I am happy to create a new branch if that makes things easier.

Also addressed [your comment](https://github.com/web-platform-tests/wpt.fyi/pull/1949#issuecomment-627349943) from this PR and add tests for `handleSelectBehaviours`

webapp/views/wpt-results.js
```diff
@@ -416,30 +406,9 @@ class WPTResults extends Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(Tes
...
-      },
-      isTriageMode: {
-        type: Boolean,
-        observer: 'isTriageModeUpdated',
-      }
     };
   }
 
-  isTriageModeUpdated(isTriageMode) {
-    this.reloadData();
```
Hexcles, line 440: Looks like `reloadData` is no longer used. Maybe remove it?

webapp/components/wpt-amend-metadata.js
```diff
@@ -10,7 +10,77 @@ import '../node_modules/@polymer/paper-toast/paper-toast.js';
 import { html, PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
 import { LoadingState } from './loading-state.js';
 import { ProductInfo } from './product-info.js';
-import { PathInfo } from '../components/path.js';
+import { PathInfo } from './path.js';
+
+const AmendMetadataUtil = (superClass) => class extends superClass {
```
Hexcles, line 15: Maybe call it a `Mixin` instead.

webapp/views/wpt-results.js
KyleJu: It's still used [elsewhere](https://github.com/web-platform-tests/wpt.fyi/blob/master/webapp/views/wpt-results.js#L1007)

PR #1949, opened 2020-05-11
[UI] Unable to deselect test directories

KyleJu:
Fix for #1951. Users are unable to deselect test directories in the Triage Metadata UI.

The issue is caused by the fact that the `test` is being overwritten with extra `/*` inside of `handleSelectMetadata`

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -1045,8 +1045,10 @@ class WPTResults extends Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(Tes
     }
   }
 
-  handleSelectMetadata(index, test) {
+  handleSelectMetadata() {
     return (e) => {
+      // Prevent accidentally modifying these captured parameters in the closure.
+      const [index, test] = arguments;
       const td = e.target.closest('td');
       const browser = this.products[index].browser_name;
 
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -1046,9 +1046,9 @@ class WPTResults extends Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(Tes
   }
 
   handleSelectMetadata() {
+    // Prevent accidentally modifying these captured parameters in the closure.
+    const [index, test] = arguments;
     return (e) => {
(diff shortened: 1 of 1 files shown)


KyleJu:
@Hexcles The `test` shouldn't be passed by reference or a data binding. I cannot seem to understand why

Hexcles:
Hmm what's this bug? Is this #1951 that I just filed?

KyleJu:
> Hmm what's this bug? Is this #1951 that I just filed?

Correct

Hexcles:
So I suspect the `path` property is passed in, not a simple string. Could you log `typeof test` and `test.prototype`?

KyleJu:
> test.prototype



> So I suspect the `path` property is passed in, not a simple string. Could you log `typeof test` and `test.prototype`?

I debugged in the devtool and I got "string" and undefined, which seem right

KyleJu:
Since we will do a refactoring [here](https://github.com/web-platform-tests/wpt.fyi/pull/1935#discussion_r421216680), I will add the tests in #1935

webapp/views/wpt-results.js
```diff
@@ -1049,16 +1049,18 @@ class WPTResults extends Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(Tes
     return (e) => {
```
Hexcles, line 1049:
```js
handleSelectMetadata() {
  // Prevent accidentally modifying these captured parameters in the closure.
  const [index, test] = arguments;
```

KyleJu:
friendly ping @Hexcles

webapp/views/wpt-results.js
```diff
@@ -1045,20 +1045,24 @@ class WPTResults extends Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(Tes
     }
   }
 
-  handleSelectMetadata(index, test) {
+  handleSelectMetadata() {
     return (e) => {
+      // Prevent accidentally modifying these captured parameters in the closure.
+      const [index, test] = arguments;
```
Hexcles, line 1051: We should capture the arguments in the parent function instead.

```suggestion
    // Prevent accidentally modifying these captured parameters in the closure.
    const [index, test] = arguments;
    return (e) => {
```

PR #1955, opened 2020-05-13
Improve logging for check_run and check_suite paths

stephenmcgruer:
Previously we logged debug information for check_run immediately upon receiving them,
but check_suites only after filtering them for 'allowed' checks. This unifies the two codepaths
to both log immediately, so that we can better understand calls we receive.

In addition, this commit also fixes a misleading log line in handleCheckRunEvent (which used
'check_suite' instead of 'check_run'), and adds the GitHub App Name to the debug info.


stephenmcgruer:
Please double check that I'm not missing something obvious, but this looks like a copy/paste error? It has already confused me once:

image

stephenmcgruer:
Actually, I want to add a bit more logging here since I'm having trouble tracking some stuff down. Let me add that and ping for re-revew.

PR #1959, opened 2020-05-14
Add href that searches test in Monorail

KyleJu:
Add a href that links to a test search in Monorail.

### Test
Enable the Triage Metadata toggle, click on failing chrome tests and click on Triage. The tests on the triage view are hyperlinks to Monorail

### DIFF
--- a/webapp/components/wpt-amend-metadata.js
+++ b/webapp/components/wpt-amend-metadata.js
@@ -7,11 +7,13 @@
 import '../node_modules/@polymer/paper-dialog/paper-dialog.js';
 import '../node_modules/@polymer/paper-input/paper-input.js';
 import '../node_modules/@polymer/paper-toast/paper-toast.js';
+import '../node_modules/@polymer/paper-tooltip/paper-tooltip.js';
 import { html, PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
 import { LoadingState } from './loading-state.js';
 import { ProductInfo } from './product-info.js';
+import { PathInfo } from '../components/path.js';
 
-class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
+class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement))) {
   static get is() {
     return 'wpt-amend-metadata';
   }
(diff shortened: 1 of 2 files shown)


webapp/components/wpt-amend-metadata.js
```diff
@@ -135,6 +146,21 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
...
+
+  getSearchURLHref(testName) {
+    if (this.computePathIsATestFile(testName)) {
+      // Remove name flags and extensions: https://web-platform-tests.org/writing-tests/file-names.html
+      testName = testName.split('.')[0];
+    } else {
+      testName = testName.replace(/((\/\*)?$)/, '');
+    }
+
+    return 'https://bugs.chromium.org/p/chromium/issues/list?q=' + testName;
```
Hexcles, line 161: This needs to be quoted. Compare the search results [with](https://bugs.chromium.org/p/chromium/issues/list?q=%22%2FFileAPI%2Ffile%2FFile-constructor%22&can=2) and [without](https://bugs.chromium.org/p/chromium/issues/list?q=%2FFileAPI%2Ffile%2FFile-constructor&can=2) quotes. And you can use a template string.

```suggestion
    return `https://bugs.chromium.org/p/chromium/issues/list?q="${testName}"`;
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -59,7 +61,16 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
             <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
           </div>
           <template is="dom-repeat" items="[[node.tests]]" as="test">
-            <li>[[test]]</li>
+            <template is="dom-if" if="[[hasHref(node.product)]]">
+              <li>
+                <a href="[[getSearchURLHref(test)]]" target="_blank">[[test]]</a>
+                <paper-tooltip offset="0" animation-delay="0">Search this test in Monorail</paper-tooltip>
```
Hexcles, line 67: BTW, this should probably be inlined instead of in a tooltip. It's confusing what this linkified path is when you are just looking at the popup.

KyleJu: How about a search-ish icon that has tooltip? Some test names tend to be quite long so I don't want the line even longer

webapp/components/wpt-amend-metadata.js
KyleJu: Yup I agree

webapp/components/wpt-amend-metadata.js
Hexcles: You could use `text-overflow: ellipsis` https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow to make the line doesn't get too lon.

KyleJu: Thanks! I set the word limit to 100 characters~

webapp/components/wpt-amend-metadata.js
```diff
@@ -59,7 +69,12 @@ class AmendMetadata extends LoadingState(ProductInfo(PolymerElement)) {
             <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
           </div>
           <template is="dom-repeat" items="[[node.tests]]" as="test">
-            <li>[[test]]</li>
+            <li>
+              <div class="testname"> [[test]] </div>
+              <template is="dom-if" if="[[hasHref(node.product)]]">
+                <a href="[[getSearchURLHref(test)]]" target="_blank"> [Search in Monorail] </a>
```
Hexcles, line 74:
```suggestion
                <a href="[[getSearchURLHref(test)]]" target="_blank"> [Search on crbug] </a>
```

Not everyone knows what Monorail is.

PR #1966, opened 2020-05-19
Metadata-related cleanup regarding "GitHub"

Hexcles:
I found a few typos, and along the way, a redundant exported struct as well.


shared/appengine.go
```diff
@@ -100,7 +100,7 @@ func (a appEngineAPIImpl) GetHTTPClientWithTimeout(timeout time.Duration) *http.
 
 func (a *appEngineAPIImpl) GetGitHubClient() (*github.Client, error) {
```
KyleJu, line 101: nit: should we change `GetGitHubClient` to `NewGitHubClient`? Same goes for `GetGitHubClientFromToken`? Although most of the functions in this file start with `Get*`

Hexcles: Renamed `GetGitHubClientFromToken` to `NewGitHubClientFromToken`, but decided to keep `GetGitHubClient` because it's not a constructor but an actual getter.

KyleJu:
Thanks Rob!

PR #1969, opened 2020-05-21
Abstract Taskcluster tasks into their own struct

stephenmcgruer:
This is a prepatory refactor for the Taskcluster Checks migration.
Instead of using tcqueue.TaskDefinitionAndStatus directly, we extract
the information we actually need into a new taskInfo struct.

The goal is to ultimately share the logic in extractArtifactURLs and
processTaskclusterBuild between the GitHub Status and the (upcoming)
GitHub Checks paths.

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -152,16 +152,16 @@ func (s statusEventPayload) IsOnMaster() bool {
 // taskInfo is an abstraction of a Taskcluster task, containing the necessary
 // information for us to process the task in wpt.fyi.
 type taskInfo struct {
-	Name   string
-	TaskID string
-	State  string
+	name   string
+	taskID string
+	state  string
 }
 
 // taskGroupInfo is an abstraction of a Taskcluster task group, containing the
-// necessary information ofr us to process the group in wpt.fyi.
+// necessary information for us to process the group in wpt.fyi.
 type taskGroupInfo struct {
-	TaskGroupID string
-	Tasks       []taskInfo
+	taskGroupID string
+	tasks       []taskInfo
 }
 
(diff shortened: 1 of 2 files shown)


stephenmcgruer:
It will likely be helpful to see where these changes are going in context when reviewing this. The following two PRs - https://github.com/web-platform-tests/wpt.fyi/pull/1970 and https://github.com/web-platform-tests/wpt.fyi/pull/1971 - need a bit more work but give an idea of where I'm heading with this.

api/taskcluster/webhook.go
```diff
@@ -143,21 +149,30 @@ func (s statusEventPayload) IsOnMaster() bool {
 	return false
 }
 
-func processTaskclusterBuild(aeAPI shared.AppEngineAPI, rootURL, taskGroupID, taskID string, sha string, labels ...string) (bool, error) {
+// taskInfo is an abstraction of a Taskcluster task, containing the necessary
+// information for us to process the task in wpt.fyi.
+type taskInfo struct {
+	Name   string
```
KyleJu, line 155: Variables used within the package should be lowercased

api/taskcluster/webhook.go
```diff
@@ -143,21 +149,30 @@ func (s statusEventPayload) IsOnMaster() bool {
...
+	TaskID string
+	State  string
+}
+
+// taskGroupInfo is an abstraction of a Taskcluster task group, containing the
+// necessary information ofr us to process the group in wpt.fyi.
```
KyleJu, line 161:
```suggestion
// necessary information for us to process the group in wpt.fyi.
```

api/taskcluster/webhook.go
```diff
@@ -143,21 +149,30 @@ func (s statusEventPayload) IsOnMaster() bool {
...
+type taskGroupInfo struct {
+	TaskGroupID string
```
KyleJu, line 163: ditto

KyleJu:
Looks good to me otherwise!

PR #1970, opened 2020-05-21
Introduce an EventInfo struct for api/taskcluster/webhook.go

stephenmcgruer:
This struct abstracts the information we need to process a GitHub
event relating to Taskcluster tasks finishing. The goal is to be
able to reuse processTaskclusterBuild between the existing GitHub
Status path, and the future GitHub Checks path.


api/taskcluster/webhook.go
```diff
@@ -37,6 +39,92 @@ var (
...
+//
+// It has two return values:
+//   * An EventInfo struct; this will be non-nil if and only if the event
+//     should be processed.
+//   * An error; if non-nil, then this event has caused an error that should be
+//     reported to the user.
+func GetEventInfo(payload []byte, log shared.Logger, api API) (*EventInfo, error) {
+	var status StatusEventPayload
+	if err := json.Unmarshal(payload, &status); err != nil {
+		// TODO: This now gives InternalServerError rather than StatusBadRequest; handle.
```
stephenmcgruer, line 85: Highlighting this as something that needs resolved before this gets merged.

stephenmcgruer:
Ping for review @KyleJu @Hexcles

api/taskcluster/webhook.go
Hexcles: It could be a bit tricky to handle here. You might want to move the unmarshalling out of this function and check the error there, which would lead to duplication of ~5 lines of code in the two code paths that I think is acceptable

api/taskcluster/webhook.go
```diff
@@ -63,47 +151,20 @@ func tcStatusWebhookHandler(w http.ResponseWriter, r *http.Request) {
...
-			if sender != "" {
-				labels.Add(shared.GetUserLabel(sender))
-			}
-
-			log.Debugf("Taskcluster task group %s", taskGroupID)
-			group, err := getTaskGroupInfo(rootURL, taskGroupID)
-			if err != nil {
-				return false, err
-			}
+	processed := false
```
KyleJu, line 157: Again, I think we should first check `if err != nil` before doing anything with the value. If you check err != nil, its safe to assume that event is not nil and remove `if event != nil` check

api/taskcluster/webhook.go
```diff
@@ -37,6 +39,92 @@ var (
...
+type apiImpl struct {}
+
+// GetEventInfo turns a GitHub Status event payload into an EventInfo struct.
+//
+// It has two return values:
+//   * An EventInfo struct; this will be non-nil if and only if the event
+//     should be processed.
+//   * An error; if non-nil, then this event has caused an error that should be
+//     reported to the user.
+func GetEventInfo(payload []byte, log shared.Logger, api API) (*EventInfo, error) {
```
KyleJu, line 82: Might be up for debate, but I would rather return `EventInfo`.  A pointer is not needed (*event used for processTaskclusterBuild) and returning an empty struct is not inherently bad. Once an error is returned, you shouldn't use non-error return values for anything else.

KyleJu: ties to [this comment] (https://github.com/web-platform-tests/wpt.fyi/pull/1970/files#r434037566)

api/taskcluster/webhook.go
```diff
@@ -37,6 +39,92 @@ var (
...
+// necessary information for us to process the group in wpt.fyi.
+type TaskGroupInfo struct {
+	TaskGroupID string
+	Tasks       []TaskInfo
+}
+
+// EventInfo is an abstraction of a GitHub Status event, containing the
+// necessary information for us to process the event in wpt.fyi.
+type EventInfo struct {
+	Sha     string
```
KyleJu, line 60: lowercase

KyleJu: NM, it is used in the tests.

api/taskcluster/webhook.go
stephenmcgruer: Done, albeit I'm not happy where we end up. We're back in the realm of a far too large (imo) if/else switch, see https://github.com/web-platform-tests/wpt.fyi/pull/1971/files#diff-5ed03f025b533061a381c9d5a256ae2cR201

I don't understand why the creators of golang made their language so adversarial to proper abstraction -_-

stephenmcgruer: (Also we lose a bunch of tests, which makes me sad...)

KyleJu: Yea we cannot help it unless we pass in `http` into the function, which makes it harder to keep track of the return statuses.

api/taskcluster/webhook.go
```diff
@@ -227,11 +291,11 @@ func parseTaskclusterURL(targetURL string) (rootURL, taskGroupID, taskID string)
 	return rootURL, taskGroupID, taskID
 }
 
-func getTaskGroupInfo(rootURL string, groupID string) (*taskGroupInfo, error) {
+func (api *apiImpl) GetTaskGroupInfo(rootURL string, groupID string) (*TaskGroupInfo, error) {
```
KyleJu, line 294: Pointer is not needed here

stephenmcgruer: I assume you mean a pointer receiver is not needed here (versus the pointer return type)? Fixed.

KyleJu: Sorry about the confusion. Right a pointer receiver is not needed. You also need to remove the `&` [here](https://github.com/web-platform-tests/wpt.fyi/pull/1970/commits/1575cc0f3b8cd3a98fed89fc40766ae139fa0e5b#diff-5ed03f025b533061a381c9d5a256ae2cR148)

stephenmcgruer: Sorry, I totally missed this comment before clicking the merge button... but somehow the code builds? 

@Hexcles do you know why it seems to be ok to pass a `&apiImpl{}` to a function which requires `api API`?

stephenmcgruer:
Eugh, we also had a conflict with the other PR I landed yesterday -_-. (master is now red). Sending a fixup for that shortly.

api/taskcluster/webhook.go
KyleJu: https://play.golang.org/p/vhzr2Qz2Fgk. I think pointers are dereferenced automatically in this case, but it doesn't work other way around (https://play.golang.org/p/_yKicUCK730)

Hexcles: Because `API` is an interface, which simply defines some function signatures; and your `apiImpl` now has no function that takes a pointer type as the receiver, so both the `apiImpl` struct and a pointer to it have all the functions attached and hence satisfy the `API` interface, whereas in your original implementation with `(api *apiImpl) GetTaskGroupInfo`, the struct `apiImpl{}` does NOT have `GetTaskGroupInfo` so it can't be used as `API`.

Issue #1899, opened 2020-04-16
Make it easier to see which failures have been triaged


foolip:
Currently, no bug links are shown by default, but enabling "Show metadata Information on the wpt.fyi result page." in /flags shows a list at the end.

In a directory like [css/css-images/](https://wpt.fyi/results/css/css-images?label=master&label=experimental&aligned) it's not very easy to figure out which failures have already been triaged. I ended up triaging one test that I didn't know was already triaged in https://github.com/web-platform-tests/wpt-metadata/pull/160.

The UX of this might be tricky, it will easily become cluttered, but something like a bug icon next to the test name or in the individual cells would make it easier to tell what's wrong. Maybe the space that this occupies could also become the thing you click to link new information.

KyleJu:
This requires changes in the UI side and the server side. 

On the server, every time a user triages a test,  the metadata cache in webapp should be updated to reflect the up-to-date metadata information.

On the UI side, we plan to display triaged metadata next to passing/total numbers inside of cells in a directory, similar to reftest screenshot icons:
<img width="920" alt="Screen Shot 2020-05-01 at 12 30 21 PM" src="https://user-images.githubusercontent.com/8611520/80822033-da826780-8ba7-11ea-9e4f-14d7e3cbdf4e.png">
In the subtest view, the metadata triage information will be shown through a row, like the row `harness status`

KyleJu:
The propose change on the server side has been implemented through a [force cached update](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/metadata_cache.go#L29)

PR #1982, opened 2020-05-28
[UI] Add metadata bug icon

KyleJu:
Fix for https://github.com/web-platform-tests/wpt.fyi/issues/1899. 

Question to discuss:
How should I display wildcards - might clutter the UI.

Follow-up PR:
Display triaged metadata under subtest view, both inline and on the bottom of the page.

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -234,7 +234,7 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
                     /
                     <span class\$="total [[ testResultClass(node, index, testRun, 'total') ]]">{{ getNodeResultDataByPropertyName(node, index, testRun, 'total') }}</span>
 
-                    <template is="dom-if" if="[[hasMetadata(index, node.path)]]">
+                    <template is="dom-if" if="[[shouldDisplayMetadata(index, node.path)]]">
                       <a href="[[ getMetadataUrl(index, node.path) ]]" target="_blank">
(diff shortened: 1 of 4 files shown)


KyleJu:
@Hexcles in our previous 1:1, we decided on a UI where the inlined bug icon will anchor link to the displayed metadata on the bottom. However it turns out [ShadowDOM doesn't support it](https://github.com/w3c/webcomponents/issues/66). The bug icon will instead link to the metadata URL. One question is whether we should/how to display the wildcard without cluttering the view.

webapp/views/wpt-results.js
```diff
@@ -381,7 +386,7 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
         type: Array,
         computed: 'computeDisplayedTests(path, searchResults)',
       },
-      metadata: Object,
+      metadataMap: Object,
```
KyleJu, line 389: I have a bit concern for a race condition here, when metadataMap is fetched after the table cell is populated and `hasMetadata` will return false. But it has never happened in my experiment.

KyleJu:
@Hexcles ready for review

webapp/views/wpt-results.js
Hexcles: (The function in question is actually `shouldDisplayMetadata`.)

I think there *is* a race condition here. One way to fix it is to add `metadataMap` as a parameter to `shouldDisplayMetadata`, so that `shouldDisplayMetadata` will be re-evaluated every time `metadataMap` is  updated.

PR #1992, opened 2020-06-04
Strip the master label for pr_head and pr_base changes

stephenmcgruer:
We have seen cases where Community-TC is triggering pull request runs
for master commits, which confuses wpt.fyi if we mis-tag them.


Hexcles:
For the record, I've removed the `master` label from all the affected runs (those with both `master` and `pr_head` labels, all created on June 4th):
* 555410001
* 561460004
* 561460005
* 561460007
* 565210001
* 567250001
* 572720002
* 572730003
* 572730005
* 574890001
* 576630002
* 580550001
* 585980001
* 587990001

We are also fairly certain this was a transient hiccup in GitHub/Taskcluster integration.

Issue #2006, opened 2020-06-09
Chrome flakiness in the insight tab crashing in prod


KyleJu:
Reproduce [here](https://wpt.fyi/results/css?label=master&label=experimental&max-count=10&product=chrome&q=seq%28%28status%3APASS%7Cstatus%3AOK%29%20%28status%3A%21PASS%26status%3A%21OK%26status%3A%21unknown%29%29%20seq%28%28status%3A%21PASS%26status%3A%21OK%26status%3A%21unknown%29%20%28status%3APASS%7Cstatus%3AOK%29%29)

Error message:
Uncaught TypeError: Cannot read property 'browser_name' of undefined
    at HTMLElement.getMetadataUrl (wpt-results.js:1051)

KyleJu:
Failing on index-out-of-bound. In `this.products[index].browser_name`, it expects this.products is an array of size 10 because the search atoms compare the recent 10 chrome runs

KyleJu:
@Hexcles perhaps we can turn off the `Show metadata Information on the wpt.fyi result page.` flag by default for now until a fix comes in

stephenmcgruer:
I have turned this off in the admin flags for wpt.fyi (it was already off for staging.wpt.fyi). Checking in an incognito window implies that it has worked.

KyleJu:
Same issue as https://github.com/web-platform-tests/wpt.fyi/issues/1896

KyleJu:
This issue should be fixed in the next wpt.fyi release.

KyleJu:
#2031. In prod

Issue #1896, opened 2020-04-16
Wrong browser triaged after deleting a column


foolip:
https://github.com/web-platform-tests/wpt-metadata/pull/160 associated a bug with Firefox instead of Safari.

Steps to reproduce:
1. Start at https://wpt.fyi/results/css/css-images?label=experimental&label=master&aligned
2. Edit the browsers shown to delete Edge, then copy the permalink and follow it
3. You will be at https://wpt.fyi/results/css/css-images?run_id=488890002&run_id=483120004&run_id=505970001
4. Click any cell with failures in the 3rd (Safari) column

You will see something like this:
![image](https://user-images.githubusercontent.com/498917/79435882-8173da80-7fd0-11ea-8780-28539cbb96ec.png)

Note the Firefox icon here instead of Safari.

It seems like there's some hardcoded mapping from column to browser that doesn't care about what the actual order is. Swapping the order illustrates this problem clearly, try https://wpt.fyi/results/css/css-images?run_id=483120004&run_id=505970001&run_id=488890002 and note that the first column is treated as Chrome, second is Edge, third is Firefox. In this example, all are wrong.

stephenmcgruer:
cc @Hexcles 

> It seems like there's some hardcoded mapping from column to browser that doesn't care about what the actual order is.

I doubt that is true. It's more likely that this is polymer data binding going wrong - I wouldn't be surprised if the metadata component binds to the first array of `products` and isn't updated if that changes.

> https://wpt.fyi/results/css/css-images?run_id=483120004&run_id=505970001&run_id=488890002 

This one is interesting, but https://wpt.fyi/results/css/css-images?label=master&product=firefox&product=safari&product=chrome has the correct browsers in the correct columns when triaging, which implies there is no such hard-coded order.

KyleJu:
I believe @stephenmcgruer is correct. Even with https://wpt.fyi/results/css/css-images?label=master&product=firefox&product=safari&product=chrome, if you click on `edit` to edit the browsers, you will see a mismatch.

<img width="1286" alt="Screen Shot 2020-04-16 at 10 39 40 AM" src="https://user-images.githubusercontent.com/8611520/79469886-bb43e180-7fce-11ea-9aaf-b15656bbb662.png">

The problem is that the displayed browsers on the first row of the table don't have `products` array to indicate selected browsers, but `testRuns`. So there is a mismatch every time you specify `run_ids` or `product` parameters in the URL

[...]

PR #2007, opened 2020-06-09
Products and TestRuns mismatch

KyleJu:
Fix for https://github.com/web-platform-tests/wpt.fyi/issues/2006, https://github.com/web-platform-tests/wpt.fyi/issues/1896

The mismatch between `products` and `testRuns` lies at the use of `/api/run` and `api/search`. When `run_ids` and `q` are used directly in `api/search`, `products` becomes outdated. However, updating `products` immediately after `testRuns` is fetched will trigger another cycle of `testRuns` update.

IMO, `products` should be named something like `queryProducts` because it's exclusively used for the query parameters of `/api/runs`. For https://github.com/web-platform-tests/wpt.fyi/issues/1896, the question is when a user deletes/adds a product, are they deleting a `queryProducts` or a `displayedProduct`?

### DIFF
--- a/webapp/components/product-info.js
+++ b/webapp/components/product-info.js
@@ -106,6 +106,16 @@ function parseProduct(name) {
   };
 }
 
+function createProduct(browserName, browserVersion, labels, revision) {
+  const product = {
+    browser_name: browserName,
+    browser_version: browserVersion,
+    labels: labels,
+    revision: revision,
+  };
+  return product;
+}
+
 // eslint-disable-next-line no-unused-vars
 const ProductInfo = (superClass) => class extends superClass {
   static get properties() {
@@ -232,4 +242,5 @@ export {
   ProductInfo,
   parseProductSpec,
   parseProduct,
+  createProduct,
 };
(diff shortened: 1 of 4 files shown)


KyleJu:
@Hexcles Not ready for review. There is an issue with the observer on `products` that causes `testRuns` to be fetched multiple times unnecessarily.

KyleJu:
Per discussions from Robert & me 1:1, this PR addresses the mismatch between `products` and the products returned from `api/search`. The cause of the mismatch is due to `max-count` or fetching through `run_ids` directly. However, there are still issues with the construction of permalink and edit when fetching directly by `run_ids`

webapp/components/product-info.js
```diff
@@ -106,6 +106,16 @@ function parseProduct(name) {
   };
 }
 
+function createProduct(browserName, browserVersion, labels, revision) {
```
Hexcles, line 109: This helper function should probably take a run instead. It'd make the call sites much simpler.
```suggestion
function productFromRun(run) {
```
You can simply `testRuns.map(productFromRun)`.

PR #2009, opened 2020-06-09
Upgrade go dependencies

Hexcles:
Trying my luck to see if this addresses the `go get` issue on GitHub
Actions.

go get -t -u -d ./...
go mod tidy


Hexcles:
That did not work.

Looks like we're experiencing https://github.com/googleapis/google-cloud-go/issues/2409 and I guess we'll just have to wait.

Hexcles:
Due to some reasons that I don't fully understand yet, go 1.12 resolves our `go.mod` differently and produces two more indirect dependencies than go 1.14.

It could be that there are version-specific dependencies or the dependency resolution algorithm has been tweaked.

Either way, since we have go 1.14 on our workstation, the `go.mod` we checked in does not contain the two dependencies in question, and CI (which runs 1.12) fails.

PR #2013, opened 2020-06-12
[Taskcluster] Read repo name/owner from correct struct member

stephenmcgruer:
A CheckSuite has a Repository member, but GitHub actually sends the
repository information in the CheckSuiteEvent.Repo member instead.


stephenmcgruer:
> Grrhh why....

Looking at comments in go-github, it seems like it may depend on whether you're getting a webhook or a GitHub App event. Which is just... wow.

PR #2025, opened 2020-06-23
[Triage Metadata] Inline bug icons for subtests

### LINKED ISSUE #1899: Make it easier to see which failures have been triaged

KyleJu:
A follow-up for #1982 and fix #1899.

### DIFF
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -144,7 +144,7 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
             <code>[[ subtestMessage(result, verbose) ]]</code>
 
             <template is="dom-if" if="[[shouldDisplayMetadata(index, row.name, metadataMap)]]">
-              <a href="[[ getMetadataUrl(index, row.name, metadataMap) ]]" target="_blank">
+              <a href="[[ getMetadataUrlForSubtest(index, row.name, metadataMap) ]]" target="_blank">
                 <iron-icon class="bug" icon="bug-report"></iron-icon>
               </a>
             </template>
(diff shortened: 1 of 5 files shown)


Hexcles:
Web component tests failed?

Hexcles:
The UI is a bit misaligned...
image

webapp/components/test-file-results-table.js
```diff
@@ -412,6 +419,28 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
...
   }
+
+  shouldDisplayMetadata(index, subtestname, metadataMap) {
+    if (!metadataMap) {
+      return false;
+    }
+    return this.displayMetadata && this.getMetadataUrl(index, subtestname, metadataMap) !== '';
+  }
+
+  getMetadataUrl(index, subtestname, metadataMap) {
```
Hexcles, line 430: Is this copied from `wpt-results.js`? Can we share them?

KyleJu:
> The UI is a bit misaligned...
> image

Duration row is not supposed to display bug icons. Just fixed it!

webapp/components/test-file-results-table.js
KyleJu: It is a bit different. `metadataMap` is a nested map with a format {`key1`: {`key2`: url}} where `key1` is the test name and `key2` is a subtest name or `/`. So the way it retrieves the URL is different between wpt-result.js and test-file-results-table.js.

PR #2045, opened 2020-07-07
Receive 100 check-runs per page

stephenmcgruer:
This is a quick fix to workaround the missing shards from WPT results.
Longer term we should properly deal with pagination, but for now we use
the fact that we know there are less than 100 results.

### DIFF
--- a/api/checks/runs.go
+++ b/api/checks/runs.go
@@ -86,7 +86,12 @@ func getExistingCheckRuns(ctx context.Context, suite shared.CheckSuite) ([]*gith
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


Hexcles:
I suggest applying this option to here as well

https://github.com/web-platform-tests/wpt.fyi/blob/a64f9c2124b15cdc7c4cead0179b018587e1e5d8/api/checks/runs.go#L89

to eliminate potential inconsistencies / future griefs. And I also think there is a real concern of creating duplicate wpt.fyi custom checks because we now have so many check runs from Taskcluster.

PR #2046, opened 2020-07-07
Add metadata bug icons for wildcards

KyleJu:
Bug fix for https://github.com/web-platform-tests/wpt.fyi/issues/1899

### DIFF
--- a/webapp/components/path.js
+++ b/webapp/components/path.js
@@ -78,6 +78,14 @@ const PathInfo = (superClass) => class extends superClass {
     return path && path === '/';
   }
 
+  isParentDir(path, dir) {
+    if (dir.startsWith(path)) {
+      const relativePath = dir.substring(path.length);
+      return relativePath.split('/').filter(p => p).length === 1;
+    }
+    return false;
+  }
+
   splitPathIntoLinkedParts(inputPath) {
     // Remove the leading slash.
     const encoded = this.encodeTestPath(inputPath).slice(1);
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -266,7 +266,8 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
       if (this.pathIsASubfolder) {
         curPath = curPath + '/';
       }
-      return curPath.startsWith(testname.substring(0, testname.length - 1)) || this.isParentDir(curPath, testname.substring(0, testname.length - 2));
+      const dirname = testname.substring(0, testname.length - 1);
+      return curPath.startsWith(dirname) || this.isParentDir(curPath, dirname);
     }
     return testname.startsWith(path) && testResultSet.has(testname);
   }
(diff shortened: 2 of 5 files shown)


stephenmcgruer:
Does this... work? What's the goal of this change?

`css-contain/` doesn't show an icon here: https://wildcards-icon-dot-wptdashboard-staging.uk.r.appspot.com/results/css?label=master&label=experimental&aligned&q=triaged%3Asafari ?

KyleJu:
> Does this... work? What's the goal of this change?
> 
> `css-contain/` doesn't show an icon here: https://wildcards-icon-dot-wptdashboard-staging.uk.r.appspot.com/results/css?label=master&label=experimental&aligned&q=triaged%3Asafari ?

Yep there is an issue. Meant to see how it works on staging. The purpose of it is to add an icon to triaged directories (wildcards), but not on the subdirectories/files of the triaged directory

KyleJu:
> LGTM
> 
> A test would be great.

I will add some tests

KyleJu:
@Hexcles Ready for another look

webapp/components/path.js
```diff
@@ -78,6 +78,14 @@ const PathInfo = (superClass) => class extends superClass {
     return path && path === '/';
   }
 
+  isParentDir(path, dir) {
+    if (dir.startsWith(path)) {
+      const relativePath = dir.substring(path.length);
+      return relativePath.split('/').filter(p => p).length === 1;
```
Hexcles, line 84: What's the point of `filter(p => p)`?

webapp/components/wpt-metadata.js
```diff
@@ -266,7 +266,7 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
       if (this.pathIsASubfolder) {
         curPath = curPath + '/';
       }
-      return curPath.startsWith(testname.substring(0, testname.length - 1));
+      return curPath.startsWith(testname.substring(0, testname.length - 1)) || this.isParentDir(curPath, testname.substring(0, testname.length - 2));
```
Hexcles, line 269: IIUC, this change now shows wildcard metadata in the immediate parent. E.g., if you navigate `foo/`, you'll now see `foo/bar/*` at the bottom.

I like the new behaviour but the implementation seems a bit brittle regarding the trailing slash. Can we make `isParentDir` work both with and without the trailing slash, so that we can use `testname.substring(0, testname.length - 1)` (preferably through a new temporary variable) in both places?

webapp/components/path.js
KyleJu: e.g. '/abd/'.split('/')) returns ['', 'abd, ''].`filter(p =>p)` allows us to remove empty strings

webapp/components/wpt-metadata.js
KyleJu: yea I had a feeling that my handling of trailing slash is brittle to begin with. But for your comment, it actually works with and without the trailing slash. See [test cases](https://github.com/web-platform-tests/wpt.fyi/pull/2046/files#diff-46af02d1b3b0dda7bc52ab60e2815e9eR88). I will use `testname.substring(0, testname.length - 1)` for consistency

PR #2049, opened 2020-07-08
Handle paginated check_runs API in api/taskcluster/webhook.go

stephenmcgruer:


stephenmcgruer:
WDYT? I could definitely see trying to pull some logic out of ListCheckRuns, to make the pagination testable. Happy to put the work in to do that, if it feels necessary.

api/taskcluster/webhook.go
```diff
@@ -431,13 +431,37 @@ func (api apiImpl) GetTaskGroupInfo(rootURL string, groupID string) (*TaskGroupI
...
+	hitEnd := false
+	for i := 0; i < 5; i++ {
+		result, response, err := api.ghClient.Checks.ListCheckRunsCheckSuite(api.ctx, owner, repo, checkSuiteID, &options)
+		if err != nil {
+			return runs, err
+		}
+
+		runs = append(runs, result.CheckRuns...)
+
+		// Setup for the next call.
```
Hexcles, line 453: Move this comment after the if block.

api/taskcluster/webhook.go
```diff
@@ -431,13 +431,37 @@ func (api apiImpl) GetTaskGroupInfo(rootURL string, groupID string) (*TaskGroupI
...
+		if response.NextPage >= response.LastPage {
+			hitEnd = true
```
Hexcles, line 455: You could simply return runs here and you won't need `hitEnd`.

api/taskcluster/webhook.go
```diff
@@ -431,13 +431,37 @@ func (api apiImpl) GetTaskGroupInfo(rootURL string, groupID string) (*TaskGroupI
 	return &group, nil
 }
 
-func (api apiImpl) ListCheckRuns(owner string, repo string, checkSuiteID int64) (*github.ListCheckRunsResults, *github.Response, error) {
+func (api apiImpl) ListCheckRuns(owner string, repo string, checkSuiteID int64) ([]*github.CheckRun, error) {
+	var runs []*github.CheckRun
```
Hexcles, line 435: Nit: maybe add a comment here saying 100 is the maximum allowed.

api/taskcluster/webhook.go
stephenmcgruer: Lol, duh, thanks 🤣

PR #2052, opened 2020-07-09
Handle paginated check_runs API in api/checks/runs.go

stephenmcgruer:


api/checks/runs.go
```diff
@@ -86,16 +87,33 @@ func getExistingCheckRuns(ctx context.Context, suite shared.CheckSuite) ([]*gith
...
 			PerPage: 100,
 		},
 	}
-	runs, _, err := client.Checks.ListCheckRunsForRef(ctx, suite.Owner, suite.Repo, suite.SHA, &options)
-	if err != nil {
-		return nil, err
+
+	// As a safety-check, we will not do more than 5 iterations (at 100
+	// check runs per page, this gives us a 500 run upper limit).
+	for i := 0; i < 5; i++ {
```
stephenmcgruer, line 101: This is may be not enough, 0f3f2b950d26ab0c88084c1537143e38712cb4b9 has 227 check_runs and it would have more if `epochs/daily` wasn't currently failing due to the sink-task dependency problem.

stephenmcgruer: http get https://api.github.com/repos/web-platform-tests/wpt/commits/0f3f2b950d26ab0c88084c1537143e38712cb4b9/check-runs Accept:application/vnd.github.antiope-preview+json, FTR

Hexcles: I'd opt to double the limit; let's give it 10 pages.

Hexcles:
FWIW, I think the "correct" way (as GitHub intends) to do this is to store a mapping of refs -> check suites/runs on our side. Maybe we can file an issue and/or leave a TODO here.

stephenmcgruer:
> FWIW, I think the "correct" way (as GitHub intends) to do this is to store a mapping of refs -> check suites/runs on our side. Maybe we can file an issue and/or leave a TODO here.

Ack, I'll file an issue.

Issue #2058, opened 2020-07-13
[UI] Metadata bug icon sometimes has broken link


stephenmcgruer:
Reproduction: Take a look at the bug link for font-language-override-02.html on https://staging.wpt.fyi/results/css/css-fonts?label=master&label=experimental&product=chrome%5Btaskcluster%5D&product=firefox%5Btaskcluster%5D&product=safari%5Bazure%5D&aligned&q=triaged%3Achrome%20count%3A2%28status%3Apass%29%20none%28status%3Amissing%7Cstatus%3Anotrun%29%20firefox%3Apass%20safari%3Apass

It goes to https://staging.wpt.fyi/results/css/crbug.com/481430 , because the underlying metadata is probably `crbug.com/418430` (no http or https). 

Possibly we should only linkify the icons if they go to things that look like URLs?

Hexcles:
The field is called "url", right? If so then they should all be URLs and maybe we should prepend `http://`.

KyleJu:
> `http://`.

I don't see a Metadata URL starting with `http://`. I would prefer prepending `https://` just because users will feel safer clicking on it?

PR #2060, opened 2020-07-13
[UI] Prepend https to WPT Metadata URLs

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2058


KyleJu:
Test failures are not related

KyleJu:
> I'd suggest dropping `https://` from more tests and perhaps use only one test case (either reuse an existing one or create a new one) to test the mixture of URLs with and without the prefix.
> 
> Also, we shouldn't add the https:// prefix if there's already a http:// prefix.

Thanks for the comment! I mixed `http://`, `https://` and without prefix in one of the cases and dropped the prefix in the rest.

Issue #2063, opened 2020-07-13
[UI] Metadata with wildcards are not filtered based on test results


KyleJu:
See [this search](https://staging.wpt.fyi/results/css/css-fonts?label=master&label=experimental&product=chrome%5Btaskcluster%5D&product=firefox%5Btaskcluster%5D&product=safari%5Bazure%5D&aligned&q=font-language-override-02.html). `/css/css-fonts/font-display/* ` shouldn't be present under the `wpt /css/css-fonts` directory

PR #2066, opened 2020-07-15
Filter wildcards using searchResults

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2063

### DIFF
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -194,8 +194,9 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
 
     const testResultSet = new Set();
     for (const result of searchResults) {
-      var test = result.test;
-      // getDirname eventually terminates with an empty string.
+      let test = result.test;
+      // Add all ancestor directories of test into testResultSet.
+      // getDirname eventually returns an empty string at the root to terminate the loop.
       while (test !== '') {
         testResultSet.add(test);
         test = this.getDirname(test);
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -278,7 +278,7 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
       }
       const dirname = testname.substring(0, testname.length - 1);
       const dirnameWithoutSlash = testname.substring(0, testname.length - 2);
-      return curPath.startsWith(dirname) || (this.isParentDir(curPath, dirname) && testResultSet.has(dirnameWithoutSlash));
+      return (curPath.startsWith(dirname) || this.isParentDir(curPath, dirname) && testResultSet.has(dirnameWithoutSlash);
     }
     return testname.startsWith(path) && testResultSet.has(testname);
   }
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -212,7 +212,7 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
 
     let metadataMap = {};
     let displayedMetadata = [];
-    for (const test of Object.keys(metadata).filter(k => this.checkPath(k, path, testResultSet))) {
+    for (const test of Object.keys(metadata).filter(k => this.shouldShowMetadata(k, path, testResultSet))) {
       const seenURLs = new Set();
       seenURLs.add(''); // Avoids accepting empty URLs.
       for (const link of metadata[test]) {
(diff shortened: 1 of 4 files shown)


KyleJu:
Deployment is working and will add some tests.

Hexcles:
`PathInfo.getDirname` is not a function?

KyleJu:
> `PathInfo.getDirname` is not a function?

its still outdated. I have fixed that issue in the latest commit `ad828d6`

Hexcles:
I wonder if something's wrong with draft PRs...

KyleJu:
> I wonder if something's wrong with draft PRs...

I add some tests and push another commit. Hope it fixes the issue

webapp/components/wpt-metadata.js
```diff
@@ -267,7 +276,8 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
         curPath = curPath + '/';
       }
       const dirname = testname.substring(0, testname.length - 1);
-      return curPath.startsWith(dirname) || this.isParentDir(curPath, dirname);
+      const dirnameWithoutSlash = testname.substring(0, testname.length - 2);
+      return curPath.startsWith(dirname) || (this.isParentDir(curPath, dirname) && testResultSet.has(dirnameWithoutSlash));
```
Hexcles, line 280: Is there a way to always use one of `dirname` and `dirnameWithoutSlash`? IIRC `isParentDir` accepts both?

webapp/components/wpt-metadata.js
```diff
@@ -192,7 +192,16 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
       return;
     }
 
-    return new Set(searchResults.map(r => r.test));
+    const testResultSet = new Set();
+    for (const result of searchResults) {
+      var test = result.test;
+      // Dir terminates with ''
```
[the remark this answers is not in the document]
KyleJu: Oops I clarified the comment. Most of getDirname APIs terminates with `/` so I thought I would leave a comment here.

webapp/components/wpt-metadata.js
KyleJu: Correct, `isParentDir` accepts both. However, if using `dirnameWithoutSlash` for all of them, there is an edge case when `curpath` is `/abcd.html and `dirname` is `/abc`; using `dirname` for `testResultSet.has(dirnameWithoutSlash)` makes the implementation of `getDirname` more complicated. I thought it wasn't worth the work there

Hexcles: But you could add `test + '/'` to `testResultSet` without having to change `getDirname`?

KyleJu: Right but testResultSet has both directories and test name. To do `test + '/'`, we need to add a check when test is a testname where we don't want to add `/` (or `testResultSet.add(test)` before going into the loop) . Same goes for cases when test is `/`.  These make line 200 and 201 less elegant. But I am open to it

Hexcles: Could you remind me what the purpose of `curPath.startsWith(dirname)` is? If it's not necessary, can we just keep the second part and always use `dirnameWithoutSlash`?

KyleJu: The purpose is that a metadata with wildcard shows up in all children directories and I do think it is necessary

KyleJu:
@Hexcles friendly ping

webapp/components/wpt-metadata.js
Hexcles: Reading this change again, I wonder if the condition should be

```suggestion
      return (curPath.startsWith(dirname) || this.isParentDir(curPath, dirname) && testResultSet.has(dirnameWithoutSlash);
```

And we should probably make the names more meaningful (e.g. `testname` -> `metadataTestName`, `checkPath` -> `shouldShowMetadata`) and add some comments to each clause of this boolean expression.

Hexcles:
There's also a merge conflict.

webapp/components/wpt-metadata.js
KyleJu: Sound good. The fact that you are able to see `curPath` implies `testResultSet.has(dirnameWithoutSlash)` is true but this is more rigid nevertheless

KyleJu:
@Hexcles conflict resolved but I feel like navigating in [this wpt.fyi staging](https://filter-results-for-wil-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=master&label=experimental&aligned) is visibly slower? Or its already slow

Hexcles:
@KyleJu got a specific example (maybe a search?) where this staging deployment is slower?

KyleJu:
> @KyleJu got a specific example (maybe a search?) where this staging deployment is slower?

e.g. clicking `css-fonts/` [on this staging](https://filter-results-for-wil-dot-wptdashboard-staging.uk.r.appspot.com/results/css?label=experimental&label=master&aligned) vs [wpt.fyi ](https://wpt.fyi/results/css?label=experimental&label=master&aligned) without enabling `Show metadata Information on the wpt.fyi result page.` I feel like the former is a second slower

Hexcles:
I tried capturing some performance profiles.

wpt.fyi without showing metadata took about 300ms while both wpt.fyi and this staging instance with showing metadata took about 500ms for this navigation, so I don't think this change is causing much performance regression, but rather showing the metadata is a bit slow.

KyleJu:
GitHub checks still failing after rerunning 3 times. https://github.com/web-platform-tests/wpt.fyi/issues/2076

Hexcles:
admin-merged

Issue #2076, opened 2020-07-21
go_chrome_test started failing


Hexcles:
Example from https://github.com/web-platform-tests/wpt.fyi/pull/2073: https://github.com/web-platform-tests/wpt.fyi/pull/2073/checks?check_run_id=891861401#step:3:1926

```
2020/07/20 22:18:55 Failed to start selenium remote with capabilites:
2020/07/20 22:18:55 map[browserName:firefox moz:firefoxOptions:{/usr/bin/firefox []  <nil> map[]}]
2020/07/20 22:18:55 session not created: Failed to start browser /usr/bin/firefox: no such file or directory
2020/07/20 22:18:55 Failed to create webdriver: session not created: Failed to start browser /usr/bin/firefox: no such file or directory
--- FAIL: TestQueryBuilder_MasterCheckedForMasterLabelQuery (23.83s)
panic: session not created: Failed to start browser /usr/bin/firefox: no such file or directory [recovered]
	panic: session not created: Failed to start browser /usr/bin/firefox: no such file or directory
```

This is a bit puzzling. go_chrome_test shouldn't need to start Firefox in the first place. And it's passing on master: https://github.com/web-platform-tests/wpt.fyi/runs/885542114?check_suite_focus=true

Hexcles:
This is now failing on master, too. https://github.com/web-platform-tests/wpt.fyi/runs/894964478

stephenmcgruer:
Able to reproduce locally via `make go_chrome_test`.

`webdriver/webdriver.go` is not receiving the `-browser` flag and is just using the default, which is firefox.

Now, `go test` is meant to do the flag parsing for us (https://blog.jbowen.dev/2019/08/using-go-flags-in-tests/) , and the other flags (e.g. firefox_path) seem to work, so this is very strange.

stephenmcgruer:
It looks like something in the `geckodriver_path` is causing the rest of the flags to not be parsed. The below diff causes the `browser` to be parsed properly but then claims that `chrome_path` isn't set.

```
$ git diff
diff --git a/Makefile b/Makefile
index eb733e1..839f092 100644
--- a/Makefile
+++ b/Makefile
@@ -95,12 +95,12 @@ _go_webdriver_test: var-BROWSER java go_build xvfb geckodriver dev_appserver_dep
        cd webdriver; \
        COMMAND="go test $(VERBOSE) -timeout=15m -tags=large -args \
[...]

PR #2081, opened 2020-07-28
[results-processor] Handle Chrome Nightly

Hexcles:
This change refactors the logic for adding release channel labels based
on browser_channel so that the mapping is browser-dependent, with two
changes in behaviours:

1. Treat Chrome Nightly differently from Firefox Nightly. Chrome Nightly
   is in fact Chromium ToT (recently added in
   https://github.com/web-platform-tests/wpt/pull/24702), and we don't
   want to add the experimental label to it which would mix it with the
   official Chrome Dev.
2. Add a "nightly" label to Chrome/Edge Canary (we only have Edge Canary
   runs as of now) to allow easy comparison between e.g. Chrome Nightly
   (Chromium ToT) and Edge Canary.

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -559,6 +559,8 @@ def prepare_labels(report: WPTReport,
     elif not (labels & RELEASE_CHANNEL_LABELS):
         # Default to "stable" if no channel label or browser_channel is present
         # TODO(Hexcles): remove this fallback default eventually.
+        _log.warn('Test run does not have browser_channel or any channel label,'
+                  ' assumed stable.')
         labels.add('stable')
 
     # Remove any empty labels.
@@ -615,6 +617,8 @@ def create_test_run(report, run_id, labels_str, uploader, auth,
     """
     if callback_url is None:
         callback_url = config.project_baseurl() + '/api/results/create'
+    _log.info('Creating run %s from %s using %s',
+              run_id, uploader, callback_url)
 
     labels = prepare_labels(report, labels_str, uploader)
     assert len(labels) > 0
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -499,7 +499,10 @@ class WPTReport(object):
 
 
 def _channel_to_labels(browser: str, channel: str) -> Set[str]:
-    """Maps browser-specific channels to well-known labels.
+    """Maps a browser-specific channel to labels.
+
+    The original channel is always preserved as a label. In addition,
+    well-known aliases of browser-specific channels are added.
 
     This aligns channels to RELEASE_CHANNEL_LABELS so that different browsers
     can be compared meaningfully on wpt.fyi. A few other aliases are added for
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -562,8 +562,8 @@ def prepare_labels(report: WPTReport,
     elif not (labels & RELEASE_CHANNEL_LABELS):
         # Default to "stable" if no channel label or browser_channel is present
         # TODO(Hexcles): remove this fallback default eventually.
-        _log.warn('Test run does not have browser_channel or any channel label,'
-                  ' assumed stable.')
+        _log.warn('Test run does not have browser_channel or any channel label'
+                  ', assumed stable.')
         labels.add('stable')
 
     # Remove any empty labels.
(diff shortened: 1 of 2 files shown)


Hexcles:
go_chrome_test failure is unrelated (#2076 )

results-processor/wptreport.py
```diff
@@ -505,6 +498,33 @@ def serialize_gzip(self, filepath):
         self.write_gzip_json(filepath, self._report)
 
 
+def _channel_to_labels(browser: str, channel: str) -> Set[str]:
+    """Maps browser-specific channels to well-known labels.
```
stephenmcgruer, line 502: Bit of a nit, but given that we put `channel` into `labels` to begin with, and could end up just exiting without doing anything more (say for `channel=foo`) it's not really 'to well-known labels'. It more attempts to extend the set of labels to include well-known ones.

It also doesn't align them to RELEASE_CHANNEL_LABELS - canary maps to nightly, and neither of those are in RELEASE_CHANNEL_LABELS ? (To be fair, there's an issue for those so maybe thats ok).

results-processor/wptreport.py
```diff
@@ -523,20 +543,22 @@ def prepare_labels(report: WPTReport,
     Returns:
         A set of strings.
     """
+    browser = report.run_info['product']
+    # browser_channel is an optional field.
+    channel = report.run_info.get('browser_channel')
```
stephenmcgruer, line 548: Instead of the comment, could use `get('browser_channel', None)`, which shows it may not exist and what we want in that situation.

results-processor/wptreport.py
```diff
@@ -523,20 +543,22 @@ def prepare_labels(report: WPTReport,
...
-        labels.add(report.run_info['browser_channel'])
-        if report.run_info['browser_channel'] in CHANNEL_TO_LABEL:
-            labels.add(CHANNEL_TO_LABEL[report.run_info['browser_channel']])
-    elif not any([i in labels for i in set(CHANNEL_TO_LABEL.values())]):
-        # Default to "stable".
+    if channel:
+        labels |= _channel_to_labels(browser, channel)
+    elif not (labels & RELEASE_CHANNEL_LABELS):
+        # Default to "stable" if no channel label or browser_channel is present
+        # TODO(Hexcles): remove this fallback default eventually.
```
stephenmcgruer, line 561: Do we have a way to track that this branch is happening? How will we know when it is safe to remove?

results-processor/wptreport.py
Hexcles: That's the same as `get(...)` (`get` returns the value, or None if not found). The comment was meant to point out why the two fields are looked up differently.

stephenmcgruer: Right, I was suggesting it as a way of documenting it via-code. But maybe future-us would just wander along and simplify it back to `get('browser_channel')` anyway :). Consider the suggestion very-optional-and-maybe-wrong :D

results-processor/wptreport.py
Hexcles: > Bit of a nit, but given that we put channel into labels to begin with, and could end up just exiting without doing anything more (say for channel=foo) it's not really 'to well-known labels'. It more attempts to extend the set of labels to include well-known ones.

Maybe

```suggestion
    """Maps a browser-specific channel to labels.

    The original channel is always preserved as a label. In addition,
    well-known aliases of browser-specific channels are added.
```

> It also doesn't align them to RELEASE_CHANNEL_LABELS - canary maps to nightly, and neither of those are in RELEASE_CHANNEL_LABELS ? (To be fair, there's an issue for those so maybe thats ok).

No, it doesn't, and that's why I didn't use the same term (but "well-known" instead). This is point 2 in the PR description, which can be split into a separate change.

stephenmcgruer: SGTM

results-processor/wptreport.py
Hexcles: Great point! I've added some logging to see if this still happens.

results-processor/wptreport.py
```diff
@@ -505,6 +498,36 @@ def serialize_gzip(self, filepath):
         self.write_gzip_json(filepath, self._report)
 
 
+def _channel_to_labels(browser: str, channel: str) -> Set[str]:
```
Hexcles, line 501:
```suggestion
def _channel_to_labels(browser: str,
                       channel: str) -> Set[str]:
```

Issue #2091, opened 2020-08-03
`/api/webhook/taskcluster` doesn't ignore partially-failed Taskcluster runs


stephenmcgruer:
Run 631210001 has only ~17k test results (should be ~34k).

https://community-tc.services.mozilla.com/tasks/groups/aK62uKihTDaEs6lIm2HILQ is its Taskcluster run. Three of the Firefox tasks (`wpt-firefox-stable-testharness-{12, 13, 14}`) failed, so wpt.fyi should ignore the entire run as failed.

Instead, the logs show that we think the runs were complete: [appengine log link](https://pantheon.corp.google.com/logs/viewer?project=wptdashboard&folder&organizationId&minLogLevel=0&expandAll=false&timestamp=2020-08-03T18:50:06.403000000Z&customFacets=&limitCustomFacetWidth=true&dateRangeEnd=2020-08-03T18:50:01.808Z&interval=PT1H&resource=gae_app%2Fmodule_id%2Fdefault&logName=projects%2Fwptdashboard%2Flogs%2Fstderr&logName=projects%2Fwptdashboard%2Flogs%2Fappengine.googleapis.com%252Frequest_log&scrollTimestamp=2020-08-03T01:41:17.513019000Z&filters=text:e35de4d284f32&advancedFilter=resource.type%3D%22gae_app%22%0Aresource.labels.project_id%3D%22wptdashboard%22%0Aresource.labels.module_id%3D%22default%22%0Aresource.labels.version_id%3D%22rev-1d47c8b%22%0Aresource.labels.zone%3D%22us14%22%0Atimestamp%3D%222020-08-03T01:41:17.513019000Z%22%0AinsertId%3D%225f276b41000cb6bab147e7a4%22&dateRangeUnbound=backwardInTime)

`Adding task: wpt-firefox-stable-testharness-12, id: JHLMIxUNQ3erXs-0Nv45XQ, status: completed`

I suspect that the API we use for the status (GetStatus()) is returning just finished/not-finished, rather than finished-and-passing).

stephenmcgruer:
Yep, we read from `run.GetStatus()`, which is documented as (https://developer.github.com/v3/checks/runs/#parameters):

`The current status. Can be one of queued, in_progress, or completed. Default: queued`

We then check for `!= 'completed'` to tell if something is wrong with the task-group (because for statuses we grab the status from Taskcluster and there `failed` and `completed` are distinct statuses).

[...]

PR #2092, opened 2020-08-03
[Taskcluster] Reject tasks that have non-success state

stephenmcgruer:
During the migration to the GitHub Checks API for Taskcluster, we
mistakenly interpreted the 'status' field of a CheckRun as equivalent to
the 'status' field of a Taskcluster task. This is not the case; in the
Taskcluster case 'status' captures the success/failure state of a task,
but for CheckRuns it merely captures the finished/not-finished state.

Instead, CheckRuns should use the 'conclusion' field to get the result
of the task.

To keep things compatible with the existing logic that assumes
Taskcluster status results, we map 'success' --> 'completed'. In the
future we should remove the old status code and switch to just using the
conclusion.

Issue: https://github.com/web-platform-tests/wpt.fyi/issues/2091

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -165,10 +165,11 @@ func GetCheckSuiteEventInfo(checkSuite github.CheckSuiteEvent, log shared.Logger
 
 		log.Debugf("Adding task: %s, id: %s, conclusion: %s", run.GetName(), taskID, run.GetConclusion())
 
-		// Match the GitHub Check Runs API to the old Taskcluster status API.
-		state := "completed"
-		if run.GetConclusion() != "success" {
-			state = run.GetConclusion()
+		// Reconstruct Taskcluster TaskInfo from the check run without calling Taskcluster API.
+		state := run.GetConclusion()
+		if state == "success" {
+			// Checked in ExtractArtifactURLs.
+			state = "completed"
 		}
 
 		group.Tasks = append(group.Tasks, TaskInfo{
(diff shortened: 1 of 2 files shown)


api/taskcluster/webhook.go
```diff
@@ -163,12 +163,18 @@ func GetCheckSuiteEventInfo(checkSuite github.CheckSuiteEvent, log shared.Logger
 			group.TaskGroupID = taskID
 		}
 
-		log.Debugf("Adding task: %s, id: %s, status: %s", run.GetName(), taskID, run.GetStatus())
+		log.Debugf("Adding task: %s, id: %s, conclusion: %s", run.GetName(), taskID, run.GetConclusion())
+
+		// Match the GitHub Check Runs API to the old Taskcluster status API.
```
Hexcles, line 168: This is a bit off. `TaskInfo` is what Taskcluster API returns (see `GetTaskGroupInfo`). In the old world, we always call the API and deserialize the response (partially) ourselves. In the new world, if we were to call the API, the response would still be the same; however, we have enough information from the GitHub webhook payload to reconstruct `TaskInfo` without calling the API, so perhaps

```suggestion
		// Reconstruct Taskcluster TaskInfo from the check run without calling Taskcluster API.
```

api/taskcluster/webhook.go
```diff
@@ -163,12 +163,18 @@ func GetCheckSuiteEventInfo(checkSuite github.CheckSuiteEvent, log shared.Logger
...
+		state := "completed"
+		if run.GetConclusion() != "success" {
+			state = run.GetConclusion()
+		}
```
Hexcles, line 172: Nit (though I know `GetConclusion` is cheap)

```suggestion
		state := run.GetConclusion()
		if state == "success" {
			// Checked in ExtractArtifactURLs.
			state = "completed"
		}
```

PR #2099, opened 2020-08-05
[WIP] Fix go_chrome_test

stephenmcgruer:


stephenmcgruer:
I do not like this fix :(. In one case, we're saying 'wct shouldn't install drivers' by specifying NOSELENIUM. But then that requires it to install them late in the `npm test` call, which then fails since its running as `browser`.

That really _should_ be fixed by us installing chromedriver and geckodriver for wct, but when I tried to do that it **still** tried to install them anyway since it was looking for a specific version (which as we learned whilst investigating this, is not the version that `npx selenium-standalone install --singleDriverInstall=firefox` gives).

I **think** we could then fix that by adding `--skip-selenium-install` to the `npm test` command in package.json, but I haven't tried that yet... what do you think of that change?

Hexcles:
Just want to clarify: you tried installing the webdrivers explicitly in Makefile before running `wct.sh`, with `NOSELENIUM`set, but `npm test` in `wct.sh` still tried to install those binaries (of a specific version)?

If so, it seems reasonable to try `skipSeleniumInstall` https://github.com/Polymer/tools/commit/cc477c062183a5dbf20febc0fa31d8614501cde5
and see if `wct` can reuse the installed binaries (albeit of a different version)

stephenmcgruer:
Yep; NOSELENIUM is only used during `webapp/node_modules/wct-local/scripts/postinstall.js`, whereas the install when running `npm test` comes from the plugin's `onPrepare` in `webapp/node_modules/wct-local/lib/plugin.js`.

It gets worse; wct-local actually requires *specific* versions of selenium-server and the drivers, specified inside `webapp/node_modules/wct-local/selenium-overrides.js`. This can be overridden by specifying your own config file via the `SELENIUM_OVERRIDES_CONFIG` env variable, but at the heart of the matter is that wct-local is very opinionated when it comes to what server/drivers it is using and we're going to be fighting that the whole way.

Hexcles:
How does `npx selenium-standalone install --singleDriverInstall=firefox` decide which version it installs, and how/when did it change? We've pinned our version of `selenium-standalone` to 6.17.0 for a long time. Yet somehow it started to download a newer version of geckodriver.

My concern is that we'd not only need to override the config for wct-local but also for selenium-standalone.

stephenmcgruer:
> We've pinned our version of selenium-standalone to 6.17.0 for a long time. Yet somehow it started to download a newer version of geckodriver.

Do we? It's not in package.json, only package-lock.json as 6.17.0. But npm doesn't guarantee package-lock.json, so:

```
$ pwd
/home/smcgruer/github/wpt.fyi/webapp
$ rm -r node_modules/
$ npm install
...
$ grep -r "\"version\":" node_modules/selenium-standalone/package.json
  "version": "6.17.0"
$ rm -r node_modules/
$ npm install selenium-standalone
$ grep -r "\"version\":" node_modules/selenium-standalone/package.json
  "version": "6.18.0"
```

So if we're using selenium-standalone directly (which we are), we should pin it in package.json I believe.

stephenmcgruer:
(Or use `npm ci` rather than `npm install`, see https://docs.npmjs.com/cli/ci.html)

Hexcles:
Sigh I never really understood what `package-lock.json` can and cannot do... Yeah we should pin it in `package.json`.

stephenmcgruer:
> Sigh I never really understood what package-lock.json can and cannot do... Yeah we should pin it in package.json.

It's... really complicated :D. They initially wanted package-lock.json to be what it's name implied, a lock. But then they saw people putting "^1.0.0" in their package.json, but checking in their package-lock.json and thus locking themselves to 1.0.0 - and being upset about that. So ultimately they decided to make `npm install` 'attempt' to satisfy both requirements, but not be beholden to package-lock.json. And then later, after all the *other* people complained, they introduced `npm ci` which only listens to package-lock.json.

So... yeah.

Anyway, I'll send the PR momentarilly.

stephenmcgruer:
... I dunno why Travis is failing.

```
ERROR: (gcloud.auth.activate-service-account) There was a problem refreshing your current auth tokens: ('invalid_grant: Invalid grant: account not found', '{"error":"invalid_grant","error_description":"Invalid grant: account not found"}')
Please run:
  $ gcloud auth login
```

Hexcles:
Ahh shoot I deleted the service account.

It was a long story. The service account was created in the wptdashboard project, before we had wptdashboard-staging. Later we granted it access to wptdashboard-staging and disabled it on wptdashboard as we stopped auto-deploying to prod. The service account was "disabled" for almost a year so I cleaned it up recently without realizing that Travis was still using this service account to deploy to staging (surprise, "disabled" doesn't really disable an account completely)...

I'll fix the service account.

PR #2100, opened 2020-08-06
[webapp] Show Chromium logo for Chrome "nightly"

Hexcles:
Add a special case to the front-end to show the Chromium logo for Chrome
"nightly" runs, which are in fact Chromium ToT.

The logo comes from
chromium/src/chrome/app/theme/chromium/product_logo_64.png

Drive-by: remove a workaround as the issue has been fully resolved.

## Review Information

Compare
https://staging.wpt.fyi/results/?diff&filter=ADC&run_id=267800022&run_id=273170022
https://chromium-logo-dot-wptdashboard-staging.uk.r.appspot.com/results/?diff&filter=ADC&run_id=267800022&run_id=273170022


webapp/components/product-info.js
```diff
@@ -146,9 +146,11 @@ const ProductInfo = (superClass) => class extends superClass {
...
       return;
     }
-    // TODO: Remove the special case for Servo when it has per-channel logos.
-    if (name !== 'servo' && labels) {
-      labels = new Set(labels);
+    labels = new Set(labels);
+    // Special case for Chrome nightly, which is in fact Chromium ToT:
+    if (name === 'chrome' && labels.has('nightly')) {
+      name = 'chromium';
+    } else if (name !== 'servo' && labels) {  // Servo does not have per-channel logos.
```
KyleJu, line 153: Although it doesn't make a difference, just note that `if (new Set(undefined))` returns `true` here

Hexcles: Good point! Reading the source code again, I think we can simply drop this, since the final outcome would be the same (an empty set would not contain any of the known channel labels).

```suggestion
    } else if (name !== 'servo') {  // Servo does not have per-channel logos.
```

PR #2107, opened 2020-08-11
Fix timestamps on /status page

### LINKED ISSUE #2106: wpt.fyi/status page is broken
```
Uncaught RangeError: Incorrect locale information provided
    at Date.toLocaleDateString (<anonymous>)
    at eval (eval at _timestamp (wpt-processor.js:181), <anonymous>:1:3)
    at HTMLElement._timestamp (wpt-processor.js:181)
```

Nothing has changed around this so I wonder why this started to break...

Hexcles:
Fixes #2106

## Review Information
Click the "processor" link on the staging deployment and switch to the "invalid" tab. You should see runs, and no errors in DevTools.

### DIFF
--- a/webapp/components/wpt-processor.js
+++ b/webapp/components/wpt-processor.js
@@ -178,7 +178,7 @@ class WPTProcessor extends ProductInfo(LoadingState(PolymerElement)) {
       dateStyle: 'short',
       timeStyle: 'medium',
     };
-    return new Date(date).toLocaletring('en-US', opts);
+    return new Date(date).toLocaleString('en-US', opts);
   }
 
   refreshContextMenu(e) {
--- a/webapp/components/wpt-processor.js
+++ b/webapp/components/wpt-processor.js
@@ -178,7 +178,7 @@ class WPTProcessor extends ProductInfo(LoadingState(PolymerElement)) {
       dateStyle: 'short',
       timeStyle: 'medium',
     };
-    return new Date(date).toLocaleDateString('en-US', opts);
+    return new Date(date).toLocaletring('en-US', opts);
   }
 
   refreshContextMenu(e) {


stephenmcgruer:
So right now on https://wpt.fyi/status I see runs in the 'invalid' tab, and no console errors. So I'm not clear what is meant to be broken?

Hexcles:
This is apparently caused by https://bugs.chromium.org/p/v8/issues/detail?id=10613

KyleJu:
> This is apparently caused by https://bugs.chromium.org/p/v8/issues/detail?id=10613

Should we abandon this fix instead then?

Hexcles:
Oh sorry I wasn't clear: the "regression" was caused by a change in Chrome's behaviour. We still need this fix to make it work in Chrome 85.

webapp/components/wpt-processor.js
```diff
@@ -178,7 +178,7 @@ class WPTProcessor extends ProductInfo(LoadingState(PolymerElement)) {
       dateStyle: 'short',
       timeStyle: 'medium',
     };
-    return new Date(date).toLocaleDateString('en-US', opts);
+    return new Date(date).toLocaletring('en-US', opts);
```
KyleJu, line 181: Should be `toLocaleString`

KyleJu:
IIUC, `toLocaleDateString(`) converts the date of a Date object into a string while `toLocaleString()` converts date and time into a string, so I don't think [missing timestamps](https://github.com/web-platform-tests/wpt.fyi/issues/2106#issuecomment-672124576) is the issue here. It is likely that `date` is misformatted and threw a `RangeError` exception. What if we catch the `RangeError `exception instead?

webapp/components/wpt-processor.js
Hexcles, line 181:
```suggestion
    return new Date(date).toLocaleString('en-US', opts);
```

Hexcles:
Ahh sorry again for the lack of context. And fixed the typo :)

What I described in #2106 was a change in the browser's behaviour: starting from Chrome 85, using `toLocaleDateString` with the `timeStyle` option, or `toLocaleTimeString` with the `dateStyle` option would cause an exception (both of which were previously no-op), which would cause the date/time stamps to be missing from the page. In our code, we had `toLocaleDateString` with both `timeStyle` and `dateStyle`, so that's not good. The solution is to either remove `timeStyle` or use `toLocaleString` so that both options are meaningful. I decided to switch to `toLocaleString` to show not just the dates but also the time as I think that'd be more useful.

Issue #2118, opened 2020-08-20
Navigating directly to a file-interop view doesn't load


stephenmcgruer:
Reproduction: visit https://wpt.fyi/interop/webdriver/tests/perform_actions/pointer.py?label=experimental&label=master&aligned directly.

The results table won't load, and in the console it says:

```
Uncaught TypeError: Cannot read property 'map' of null
    at HTMLElement.computeDisplayedProducts (test-file-results-table.js:285)
    at runMethodEffect (property-effects.js:1038)
    at Object.runComputedEffect [as fn] (property-effects.js:634)
    at runEffects (property-effects.js:140)
    at runComputedEffects (property-effects.js:457)
    at HTMLElement._propertiesChanged (property-effects.js:1910)
    at HTMLElement._flushProperties (properties-changed.js:384)
    at HTMLElement.ready (property-effects.js:1873)
    at HTMLElement.ready (element-mixin.js:694)
    at HTMLElement.ready (test-runs-query.js:56)
```

If you then navigate up to https://wpt.fyi/interop/webdriver/tests/perform_actions?label=experimental&label=master&aligned (by clicking the directory name) and back down to pointer.py, it shows the results table correctly.

stephenmcgruer:
Note: This was reported by AutomatedTester on IRC, who said that Lan sent him the link. That implies that it may be worth syncing with Lan as someone who may be using the interop view (which we've long considered fairly defunct...)

PR #2119, opened 2020-08-20
[Triage Metadata] testRuns null issue

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2118

For computed properties, "the computing function is not invoked until at least one dependency is defined (!== undefined)". In this case, testRuns is defined but assigned to null instead of a real value.

### DIFF
--- a/webapp/components/test/test-file-results-table.html
+++ b/webapp/components/test/test-file-results-table.html
@@ -54,29 +54,30 @@ suite('TestFileResultsTable', () => {
         assert.notEqual(tfrt.colorClass('OK'), tfrt.colorClass('PASS'));
       });
     });
+
+    suite('computeDisplayedProducts', () => {
+      test('null testRuns', () => {
+        assert.deepEqual(tfrt.computeDisplayedProducts(null), []);
+      });
+      test('simple testRuns', () => {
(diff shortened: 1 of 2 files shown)


KyleJu:
Travis failures seem unrelated

KyleJu:
> LGTM
> 
> Can we add a simple test for this in `webapp/components/test/test-file-results-table.html`?

Done! web_components_test (pull_request) failures are unrelated.

webapp/components/test/test-file-results-table.html
```diff
@@ -55,6 +55,30 @@
...
+    test('null testRuns', () => {
+      assert.deepEqual(tfrt.computeDisplayedProducts(null), []);
+    });
+    test('simple testRuns', () => {
+      const testRuns = [{
+        browser_name: 'firefox',
+        browser_version: 1,
+        labels: ['labelA'],
+        time_start: '2018-01-12T12:00:00Z',
+        time_end: '2018-01-12T13:20:00Z',
```
Hexcles, line 69: Nit: we don't really need these two fields here
```suggestion
```

Hexcles (approved):

PR #2136, opened 2020-09-01
Debug WCT failures

Hexcles:
Does enable verbose logging really "fix" WCT???


Hexcles:
So it's indeed flaky after all, sigh...

PR #2147, opened 2020-09-10
Remove remaining uses of urlfetch

Hexcles:
Part of #1747.

Drive-by: fix a warning discovered by the latest version of golang:
  api/query/cache/service/search.go:110:40: conversion from RunID (int64) to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)


KyleJu (approved):
LGTM

PR #2149, opened 2020-09-11
Get application identity from env vars and APIs

Hexcles:
Migrate away from deprecated google.golang.org/appengine; #1747.

There should be no observable changes, except one when running locally: stop showing the current local git branch as the version, as the added convenience don't justify the complexities.

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -40,11 +40,11 @@ func init() {
 		runtimeIdentity.AppID = proj
 		runtimeIdentity.Service = os.Getenv("GAE_SERVICE")
 		if runtimeIdentity.Service == "" {
-			panic("Missing GAE_SERVICE")
+			panic("Missing environment variable: GAE_SERVICE")
 		}
 		runtimeIdentity.Version = os.Getenv("GAE_VERSION")
 		if runtimeIdentity.Version == "" {
-			panic("Missing GAE_VERSION")
+			panic("Missing environment variable: GAE_VERSION")
 		}
 		service, err := apps.NewService(context.Background())
 		if err != nil {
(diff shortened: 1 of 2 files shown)


Hexcles:
@stephenmcgruer could you take a quick look? Kyle is OOO and I have some other changes blocked by this.

shared/appengine.go
```diff
@@ -11,16 +11,53 @@ import (
...
+
+func init() {
+	// Env vars available on GAE:
+	// https://cloud.google.com/appengine/docs/standard/go/runtime#environment_variables
+	// Note: the "region code" part of GAE_APPLICATION is NOT location ID.
+	if proj := os.Getenv("GOOGLE_CLOUD_PROJECT"); proj != "" {
+		runtimeIdentity.AppID = proj
+		runtimeIdentity.Service = os.Getenv("GAE_SERVICE")
+		if runtimeIdentity.Service == "" {
+			panic("Missing GAE_SERVICE")
```
stephenmcgruer, line 43: Although we'll get a stack-trace anyway if we panic, can we be more specific about the problem here? I.e. `GAE_SERVICE environment variable not set` or similar.

shared/appengine.go
```diff
@@ -11,16 +11,53 @@ import (
...
+	apps "google.golang.org/api/appengine/v1"
+
+	// TODO(#1747): Deprecate this library.
 	"google.golang.org/appengine/user"
 )
 
+// runtimeIdentity contains the identity of the current AppEngine service when
+// running on GAE, or empty when running locally.
+var runtimeIdentity struct {
+	LocationID  string
```
stephenmcgruer, line 28: This appears unused in this change; is it used in a dependent change?

stephenmcgruer (approved):

shared/appengine.go
Hexcles: Yeah, Cloud Tasks requires this in the API call.

PR #2153, opened 2020-09-14
Migrate from TaskQueue to Cloud Tasks

Hexcles:
This is a refactor without behavioural changes. Reviewing by commits might be easier (and skip the mockgen commit).

Part of #1747 .


KyleJu:
@Hexcles I will review it by the end of the day

api/checks/api.go
```diff
@@ -44,31 +42,27 @@ type API interface {
...
+
 	queue string
 }
 
 // NewAPI returns a real implementation of the API
 func NewAPI(ctx context.Context) API {
-	return checksAPIImpl{
-		ctx:   ctx,
-		queue: CheckProcessingQueue,
+	return &checksAPIImpl{
```
KyleJu, line 52: why is `&` added here?

shared/appengine.go
```diff
@@ -25,18 +27,29 @@ import (
...
+	// Internal details of the application identity
 	application *apps.Application
 }
 
+// clients contains heavyweight (e.g. with connection pools) clients that
+// should be bound to the runtime instead of each request in order to be
+// reused. They are initialized and authenticated at startup using the
+// background context; each request should use its own context.
+var clients struct {
+	cloudtasks *cloudtasks.Client
```
KyleJu, line 44: Confusing to have the same name as the library. Maybe `taskClient`?

shared/appengine.go
```diff
@@ -25,18 +27,29 @@ import (
...
+
+	// Internal details of the application identity
 	application *apps.Application
 }
 
+// clients contains heavyweight (e.g. with connection pools) clients that
+// should be bound to the runtime instead of each request in order to be
+// reused. They are initialized and authenticated at startup using the
+// background context; each request should use its own context.
+var clients struct {
```
KyleJu, line 43: `clients` here seems to refer multiple clients but it is in fact just one. How about just `client`?

shared/appengine.go
```diff
@@ -46,15 +59,19 @@ func init() {
...
+			if runtimeIdentity.application, err = service.Apps.Get(proj).Do(); err != nil {
+				panic(err)
+			}
 		}
-		runtimeIdentity.application, err = service.Apps.Get(proj).Do()
-		if err != nil {
+		runtimeIdentity.LocationID = runtimeIdentity.application.LocationId
+
+		var err error
+		if clients.cloudtasks, err = cloudtasks.NewClient(ctx); err != nil {
```
KyleJu, line 72: Would adding `defer client.Close()` somewhere make a difference? you mentioned it is bound to the runtime instead of each request. I assume there is one runtime per version of a service? I guess my question is, in the standard environment without a long-running background process, `clients.cloudtasks` won't get initialized multiple times without being `Close()`?

shared/appengine.go
Hexcles: But the whole name would then be `clients.tasksClient`, which sound a bit redundant. We almost never use the field name alone (except in the initializer).

shared/appengine.go
Hexcles: You're right correct; there's going to be more coming soon :)

api/checks/api.go
Hexcles: Without `&`, we are returning a struct, which does not satisfy the `API` interface (or more specifically, the `shared.AppEngineAPI` interface), as the embedded `shared.appEngineAPIImpl` has a few functions using the *pointer* receiver type (those functions need to modify the content of `appEngineAPIImpl`, e.g. for caching).

shared/appengine.go
Hexcles: That's a great point!

These clients are indeed bound to the runtime / server process. Even in a standard environment, there's still a server process being started and terminated on demand (*not* for each request, but automatically managed by AppEngine); it's just that we cannot rely on this server process to run indefinitely.

We do need to close the clients before exiting though, in order to prevent leaking any resources (connections). I've added that in the latest commit.

Hexcles:
@KyleJu done

api/checks/api.go
KyleJu: But the `&` is used next to the `checksAPIImpl` and I don't see any `checksAPIImpl` functions use [pointer receiver type](https://github.com/web-platform-tests/wpt.fyi/blob/38ce5b2bae65f59f15d14283806fb2a4835e81d2/api/checks/api.go#L36)?

Hexcles: Yeah actually you're right. I'm embedding the interface `AppEngineAPI` here, which is initialized to a pointer `& appEngineAPIImpl` created by `NewAppEngineAPI`, so there's our pointer required by the methods on `*appEngineAPIImpl`. `checksAPIImpl` itself doesn't have to be pointer.

```suggestion
	return checksAPIImpl{
```

shared/appengine.go
```diff
@@ -24,6 +24,31 @@ import (
...
+
+// Init initializes all clients in Clients. If an error is encountered, it
+// returns immediately without trying to initialize the remaining clients.
+func (c *clientsImpl) Init(ctx context.Context) (err error) {
+	c.cloudtasks, err = cloudtasks.NewClient(ctx)
+	return err
+}
+
+// Close closes all clients in Clients. It must be called once and only once
+// before the server exits. Do not use any AppEngine APIs afterwards.
```
KyleJu, line 49: You mean Google Cloud client library ?

Hexcles: Hmm I was trying to refer to any API in this file, just to be safe (though some APIs don't use these clients for now). Let me rephrase it

```suggestion
// before the server exits. Do not use AppEngineAPI afterwards.
```

KyleJu (approved):
LGMT :)

shared/appengine.go
```diff
@@ -12,50 +12,46 @@ import (
...
+		return nil
 	}
+	c.cloudtasks, err = cloudtasks.NewClient(ctx)
+	return err
+}
+
+// Close closes all clients in Clients. It must be called once and only once
+// before the server exits. Do not use AppEngineAPI afterwards.
+func (c *clientsImpl) Close() error {
+	err := c.cloudtasks.Close()
```
KyleJu, line 52: when `runtimeIdentity.AppID == ""` and return `nil`, `defer shared.Clients.Close()` will still be called. c.cloudtasks.Close() might cause a nil pointer exception

Hexcles: Aha great find! Could you send a PR to fix it up?

Issue #2158, opened 2020-09-17
Deploy 35b4e74 to production


Hexcles:
Previous deployment was #2129 (02189b2)

Changelist 02189b2...35b4e74

Major changes:
- [Triage Metadata] testRuns null issue (#2119) 
- [UI] Show instructions after toggling the triage mode (#2114)
- [results-processor] Use HTTP header to verify requests (#2146)
- Get application identity from env vars and APIs (#2149) 

This push is happening as part of the regular weekly push.

Pushing all three services - webapp, processor, and searchcache.

Hexcles:
Also manually deployed `webapp/web/queue.yaml`

Hexcles:
Rolled back due to elevated 5xx in searchcache

Hexcles:
More details:

The initial symptom wasn't searchcache, but elevated 5xx in `/api/checks/...`, with no useful error messages (**action**: add more logs in this code path). The checks API needs to use searchcache, so out of a hunch I tried searching `status:...` on wpt.fyi and indeed encountered issues, which led us to searchcache.

Current status: 

The new searchcache deployment returns a 404 from **GAE** front end, not from us! Somehow the version seems to be offline or incomplete or does not exist (despite the green checkmark in GCP dashboard).

Hexcles:
Attempt to overwrite the broken version failed, but deleting and redeploying the version worked. The new version has now again been promoted to 100%.

I'll take the action item above and add some logs.

PR #2159, opened 2020-09-18
[checks] Fix logging

Hexcles:
Make sure to log errors for 5xx responses and always return after
setting the error code.

This is a follow-up to the outage in #2158 .


api/checks/update.go
```diff
@@ -64,8 +64,8 @@ func updateCheckHandler(w http.ResponseWriter, r *http.Request) {
 		msg := "Could not find runs to compare"
 		if err != nil {
```
KyleJu, line 65: maybe remove this?

KyleJu (approved):
Thanks for doing this!

Issue #2157, opened 2020-09-17
Any GitHub user (without being in the org) can triage metadata


Hexcles:
https://github.com/web-platform-tests/wpt.fyi/blob/35b4e74dc5625de763ffb25a27468fff649d92bd/shared/github_oauth.go#L157-L164

Here we are passing an empty string for `user` to `GetOrgMembership`, which means we are checking the org membership of the authenticated user (of `gaci`), which is our bot.

KyleJu:
Fixed in #2160

PR #2160, opened 2020-09-18
Refactor part of GitHub OAuth

Hexcles:
This is in part a refactor to pave the way for migrating away from AppEngine User API to GitHub User API (part of #1747 ) and also a fix for #2157.

It'd be easier to review by commits.

### DIFF
--- a/api/metadata_handler.go
+++ b/api/metadata_handler.go
@@ -66,7 +66,11 @@ func apiMetadataTriageHandler(w http.ResponseWriter, r *http.Request) {
 		forceUpdate:  true}
 	tm := shared.NewTriageMetadata(ctx, botClient, user.GitHubHandle, user.GitHubEmail, fetcher)
 
-	gac := shared.NewGitAccessControl(ctx, ds, botClient, *token)
+	gac, err := shared.NewGitHubAccessControl(ctx, ds, botClient, user, *token)
+	if err != nil {
+		http.Error(w, "Unable to create GitHub OAuth client: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
 	handleMetadataTriage(ctx, gac, tm, w, r)
 }
 
@@ -101,25 +105,14 @@ func handleMetadataTriage(ctx context.Context, gac shared.GitHubAccessControl, t
 		return
 	}
 
-	code, err := gac.IsValidAccessToken()
-	if err != nil {
-		http.Error(w, "Failed to validate user token:"+err.Error(), http.StatusInternalServerError)
-		return
-	}
-
-	if code != http.StatusOK {
-		http.Error(w, "User token invalid; please log in again.", http.StatusUnauthorized)
-		return
-	}
-
-	code, err = gac.IsValidWPTMember()
+	valid, err := gac.IsValidWPTMember()
 	if err != nil {
 		http.Error(w, "Failed to validate web-platform-tests membership: "+err.Error(), http.StatusInternalServerError)
 		return
 	}
 
-	if code != http.StatusOK {
-		http.Error(w, "Logged-in user must be a member of the web-platform-tests GitHub organization. To join, please contact wpt.fyi team members.", http.StatusBadRequest)
+	if !valid {
+		http.Error(w, "Logged-in user must be a member of the web-platform-tests GitHub organization. To join, please contact wpt.fyi team members.", http.StatusForbidden)
 		return
 	}
 
--- a/api/metadata_handler.go
+++ b/api/metadata_handler.go
@@ -51,21 +51,22 @@ func apiMetadataTriageHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	wptfyiBotClient, err := shared.NewGitHubClientFromToken(ctx, "github-wpt-fyi-bot-token")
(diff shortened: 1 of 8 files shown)


api/metadata_handler.go
```diff
@@ -51,21 +51,22 @@ func apiMetadataTriageHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	wptfyiBotClient, err := shared.NewGitHubClientFromToken(ctx, "github-wpt-fyi-bot-token")
+	botToken, err := shared.GetSecret(ds, "github-wpt-fyi-bot-token")
 	if err != nil {
 		http.Error(w, "Unable to get GitHub client for wpt-fyi-bot: "+err.Error(), http.StatusInternalServerError)
```
KyleJu, line 56: Two things here
- maybe "Unable to get the GitHub OAuth token for wpt-fyi-bot:"?
- I know this should be asked early - is it safe to return a http.Error? It might be leaking information related to the secret?

shared/github_oauth.go
```diff
@@ -30,15 +31,22 @@ type User struct {
 
 // GitHubAccessControl encapsulates implementation details of access control for the wpt-metadata repository.
 type GitHubAccessControl interface {
-	IsValidAccessToken() (int, error)
-	IsValidWPTMember() (int, error)
+	// IsValid* functions also verify the access token with GitHub.
```
KyleJu, line 34:
```suggestion
	// IsValidWPTMember also verifies the access token with GitHub.
```

shared/github_oauth.go
```diff
@@ -129,47 +132,50 @@ func NewGitHubOAuth(ctx context.Context) (GitHubOAuth, error) {
...
-	tp := github.BasicAuthTransport{
-		Username: clientID,
-		Password: secret,
+	if !valid {
+		return false, errors.New("Invalid access token")
 	}
-
-	oauthAppClient := github.NewClient(tp.Client())
-	_, res, err := oauthAppClient.Authorizations.Check(gaci.ctx, clientID, gaci.token)
+	_, res, err := gaci.botClient.Organizations.GetOrgMembership(gaci.ctx, gaci.user.GitHubHandle, "web-platform-tests")
```
KyleJu, line 152: Note to myself: double check this API

shared/github_oauth.go
Hexcles: Sorry I forgot to mention that there's a second `IsValidAdmin` coming in the next PR

api/metadata_handler.go
Hexcles: Both of these are addressed by #2167

shared/github_oauth.go
```diff
@@ -106,16 +114,11 @@ func (g *githubOAuthImpl) GetGitHubUser(client *github.Client) (*github.User, er
...
-		log.Errorf("Failed to get github-oauth-client-id secret: %s", err.Error())
-		return &githubOAuthImpl{}, err
-	}
 
-	secret, err := GetSecret(store, "github-oauth-client-secret")
+	clientID, secret, err := getOAuthClientIDSecret(store)
 	if err != nil {
-		log.Errorf("Failed to get github-oauth-client-secret: %s", err.Error())
-		return &githubOAuthImpl{}, err
+		log.Errorf("Failed to get github-oauth-client-{id,secret}: %e", err)
```
KyleJu, line 120:
```suggestion
		log.Errorf("Failed to get github-oauth-client-{id,secret}: %s", err.Error())
```
%e is used for scientific notation?

Hexcles: Ahh thanks so,ejpw I thought it was for error...

KyleJu (approved):
Sorry for the late reply. LGTM

KyleJu:
oops I wasn't able to triage in the instance, the error message is `
500: Failed to validate web-platform-tests membership: GET https://api.github.com/orgs/web-platform-tests/memberships/KyleJu: 404 Not Found []`

KyleJu:
> oops I wasn't able to triage in the staging instance; the error message is ` 500: Failed to validate web-platform-tests membership: GET https://api.github.com/orgs/web-platform-tests/memberships/KyleJu: 404 Not Found []`. To reproduce it, you can go to staging.wpt.fyi, open chrome dev and go to cookies. You can copy and paste the`session` cookie and its secret to this staging instance

Was able to reproduce it locally using my personal token. I suspect `GetOrgMembership` requires extra permission to check non-authenticated user. The correct API here should be [IsMember](https://pkg.go.dev/github.com/google/go-github/github#OrganizationsService.IsMember) and worked when I tested it locally.

shared/github_oauth.go
KyleJu, line 152:
```suggestion
	 IsMember, res, err := gaci.botClient.Organizations. IsMember(gaci.ctx, "web-platform-tests", gaci.user.GitHubHandle, )
```

Hexcles:
Thanks for spotting this! This looks like a permission issue to me; I suspect the authenticated user (i.e. the bot account in our case) needs to be an admin of the org as I can use my personal token with the same OAuth scopes on this API. I'll change it to `IsMember`.

Hexcles:
Alright it works :) https://github.com/web-platform-tests/wpt-metadata/pull/412

PR #2162, opened 2020-09-21
Migrate appengine.NewContext to request.Context()

KyleJu:
A part of #1747,  AppEngine Context migration.

### DIFF
--- a/api/interop_medium_test.go
+++ b/api/interop_medium_test.go
@@ -3,7 +3,6 @@
 package api
 
 import (
-	"context"
 	"encoding/json"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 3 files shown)


KyleJu:
Note that (1) In 70f6ac6, I tried using a background context for `interop_medium_test.go` (because I was under the impression that background context should be used in test), but that test is an integration test that requires a derived HTTP request Context. (2) [search_test.go for wpt-interop](https://github.com/web-platform-tests/wpt.fyi/blob/master/webdriver/search_test.go#L23) timed out in one of the GitHub check runs. I reran it and it passed.

Hexcles (approved):
LGTM

PR #2164, opened 2020-09-21
Fixup: restore Content-Type in CloudTasks

Hexcles:
Follow up to #2153 and part of #1747 .

TODO (in a future PR): figure out a way to test this locally. The new SDK no longer provides a local emulator for TaskQueue.

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -273,7 +273,11 @@ func (a appEngineAPIImpl) ScheduleTask(queueName, taskName, target string, param
 				AppEngineHttpRequest: &taskspb.AppEngineHttpRequest{
 					HttpMethod:  taskspb.HttpMethod_POST,
 					RelativeUri: target,
-					Body:        []byte(params.Encode()),
+					// In appengine.taskqueue, The default for POST task was
+					// application/x-www-form-urlencoded, but the new SDK
+					// defaults to application/octet-stream.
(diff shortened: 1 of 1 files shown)


shared/appengine.go
```diff
@@ -273,7 +273,11 @@ func (a appEngineAPIImpl) ScheduleTask(queueName, taskName, target string, param
 				AppEngineHttpRequest: &taskspb.AppEngineHttpRequest{
 					HttpMethod:  taskspb.HttpMethod_POST,
 					RelativeUri: target,
-					Body:        []byte(params.Encode()),
+					// In appengine.taskqueue, The default for POST task was
+					// application/x-www-form-urlencoded, but the new SDK
+					// defaults to application/octet-stream.
+					Headers: map[string]string{"Content-Tye": "application/x-www-form-urlencoded"},
```
KyleJu, line 279:
```suggestion
					Headers: map[string]string{"Content-Type": "application/x-www-form-urlencoded"},
```

Hexcles: Oops... Thanks for catching this :)

KyleJu (approved):
Normally the header format is `map[string][]string`. Kind of weird that its `map[string]string` here. LGTM

shared/appengine.go
KyleJu: np!

PR #2166, opened 2020-09-21
[webapp] Hide GitHub login button in some pages

Hexcles:
## Review Information

Go to Recent Runs, Insights, Processor & About pages. The GitHub login button shouldn't show up.

### DIFF
--- a/webapp/components/wpt-header.js
+++ b/webapp/components/wpt-header.js
@@ -67,7 +67,7 @@ class WPTHeader extends WPTFlags(PolymerElement) {
           <img src="/static/logo.svg" alt="wpt.fyi logo">
           <a href="/">web-platform-tests dashboard</a>
         </h1>
-        <template is="dom-if" if="[[githubLogin]]">
+        <template is="dom-if" if="[[showLogin]]">
           <github-login user="[[user]]" is-triage-mode="{{isTriageMode}}"></github-login>
         </template>
       </div>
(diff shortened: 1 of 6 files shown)


stephenmcgruer:
> The GitHub login button shouldn't show up.

Why? I know it isn't useful on those pages, but isn't it indicating a site-wide status? (Being logged in)

Hexcles:
It is a side-wide status, but somewhat confusing on those pages (especially the help text which mentions the triage mode).

Another minor reason is that fixing the template rendering is *slightly* more annoying (there's actually another bug that login always redirects back to the homepage instead of the previous page), but I don't want to take that into consideration and let's do the right thing here -- do you think it's more intuitive to always show the login button instead?

KyleJu:
I think it is more consistent to show the side-wide status, however

1. When a user is logged in, the triage mode button should be disabled in the Recent Runs, Insights, Processor, and About tabs.
2. Like Rob mentioned, when a user logs in or logs out, they should be redirected to the page there were on (I forgot to set the [return URL here](https://github.com/web-platform-tests/wpt.fyi/blob/b8b8106cd7320a71f628214803cf583bde53f231/webapp/login.go#L39))
3. Another bug where when a logged in user switches to one of the Recent Runs, Insights, Processor, and About tabs, it shows the login button on the header.

I am up for merging this CL because the three things mentioned above confuse users. Number 1 and 3 can be fixed by tweaking the `noLogin` in this change as well

stephenmcgruer:
> do you think it's more intuitive to always show the login button instead?

I do. My general expectation of a site is that if I can or am logged in, that information is visible in a header or sidebar (one of them, consistently) on every page I go to.  Otherwise it is unclear to me (a) if I am logged in or (b) how I log in, if not. 

I think the triage mode slider shouldn't appear unless it is usable, and yes we should fix the bug with redirect after login and the bug (3) that Kyle mentions. But I'd rather we fix those 'properly' rather than break the site in a different way (imo!) to address them.

(This is all my opinion, though I happen to think it's the correct one ;))

Hexcles:
I'm convinced! Stay tuned for a proper fix :)

PR #2172, opened 2020-09-24
[Draft] G112 log migration

KyleJu:
Draft


shared/logger.go
```diff
@@ -54,37 +54,26 @@ func NewNilLogger() Logger {
...
-
-// newGAELogger returns a Google App Engine Standard Environment logger bound to
-// the given context.
-func newGAELogger(ctx context.Context) Logger {
-	return gaeLogger{ctx}
-}
-
 // NewAppEngineContext creates a new Google App Engine Standard-based
 // context bound to an http.Request.
 func NewAppEngineContext(r *http.Request) context.Context {
```
KyleJu, line 59: Have to [close the client](https://pkg.go.dev/cloud.google.com/go/logging#hdr-Closing_your_Client)

KyleJu:
not logging the severity on app engine - might be an old issue

KyleJu:
With the `e755d9d`, the severity level for Standard App Engine is shown properly under the log_name `stdout`.  Now I will examine the structured logging setup - correlate entries in the app log with the request log (it is clearly not)

Hexcles (commented):
https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870
> Note that constructing Logger per request (to have trace_id in CommonLabels) leads to a rapid goroutine (and memory) leaks, since Client "remembers" all Loggers until it itself is closed.  ... So either requests must reuse Loggers, or they must not share Clients. 

According to this, we probably don't want to create a new Logger everytime. Might as well store the logger in Clients and reuse it.

KyleJu:
> [googleapis/google-cloud-go#720 (comment)](https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870)
> 
> > Note that constructing Logger per request (to have trace_id in CommonLabels) leads to a rapid goroutine (and memory) leaks, since Client "remembers" all Loggers until it itself is closed.  ... So either requests must reuse Loggers, or they must not share Clients.
> 
> According to this, we probably don't want to create a new Logger everytime. Might as well store the logger in Clients and reuse it.

Thanks! It helps a lot. I think this explains the timeout issues in the tests.

KyleJu:
With 55420a7, it can be built locally. Now looking at the nil context issue in staging instance.

shared/logger.go
```diff
@@ -54,37 +53,16 @@ func NewNilLogger() Logger {
...
 	ctx := r.Context()
-	return withLogger(ctx, newGAELogger(ctx))
+	ctx, err := newAppEngineFlexContext(r, os.Getenv("GOOGLE_CLOUD_PROJECT"))
+	if err != nil {
+		return r.Context()
+	}
+
+	return ctx
```
KyleJu, line 65: not backward compatible with newGAELogger

shared/appengine.go
```diff
@@ -42,18 +46,38 @@ func (c *clientsImpl) Init(ctx context.Context) (err error) {
 		return nil
 	}
 	c.cloudtasks, err = cloudtasks.NewClient(ctx)
-	return err
+	if err != nil {
+		return err
+	}
+
+	c.gclog, err = gclog.NewClient(ctx, runtimeIdentity.AppID)
```
KyleJu, line 53: Project ID is different between flex and standard

KyleJu:
with 56d2e32, all checks pass but searchcache is returning 500 in staging. I think flex and app engine have a different way of getting project ID. Going forward, I will revert changes in searchcache and experimenting with logger setup in standard first. Will move the initialization of logger client into the main() function.

In terms of structured logging, there needs to be parent logger wrapping each http request; [see here](https://github.com/googleapis/google-cloud-go/issues/448#issuecomment-352132523). Similar to `HandleWithGoogleCloudLogging`, we can wrap all the APIs in routes.go

shared/appengine.go
Hexcles: Weird, `GOOGLE_CLOUD_PROJECT` should be available on Flex, too, as well as `GAE_*`: https://cloud.google.com/appengine/docs/flexible/go/migrating

Hexcles:
Why is searchcache returning 500? Is there any error detail perhaps in the logs?

shared/appengine.go
KyleJu: Well I was guessing that it could be the cause for searchcache failure. It seems to [retrieve/verify project ID from metadata](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/cache/service/main.go#L115). Might be the same underlyingly?

KyleJu:
> Why is searchcache returning 500? Is there any error detail perhaps in the logs?

Actually it is 502 and here is what I get from standard log. Still trying to dig the searchcache log.
"Error from request: POST https://g112-log-migration-dot-searchcache-dot-wptdashboard-staging.uk.r.appspot.com/api/search/cache: STATUS 502: <html>
<head><title>502 Bad Gateway</title></head>
<body>
<center><h1>502 Bad Gateway</h1></center>
<hr><center>nginx</center>
</body>
</html>
"

KyleJu:
> > Why is searchcache returning 500? Is there any error detail perhaps in the logs?
> 
> Actually it is 502 and here is what I get from standard log. Still trying to dig the searchcache log.
> 
> `Error from request: POST https://g112-log-migration-dot-searchcache-dot-wptdashboard-staging.uk.r.appspot.com/api/search/cache: STATUS 502:
> 
> <title>502 Bad Gateway</title>
> # 502 Bad Gateway
> nginx `

and I see nothing in the searchcache log..

Hexcles:
Hmm it seems to work now: https://g112-log-migration-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=master&label=experimental&product=chrome&aligned&q=status%3Afail

Did you redeploy the instance? If not, I wonder if the symptom (a bunch of 422 followed by 502) suggests that we are not reserving enough memory for serving new requests (i.e. for a while all memory was used for loading & deserializing runs).

KyleJu:
> Hmm it seems to work now: https://g112-log-migration-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=master&label=experimental&product=chrome&aligned&q=status%3Afail
> 
> Did you redeploy the instance? If not, I wonder if the symptom (a bunch of 422 followed by 502) suggests that we are not reserving enough memory for serving new requests (i.e. for a while all memory was used for loading & deserializing runs).

Nope. `status:fail` seems to work but `chrome:fail` still returns a combination of 422 and 502 

https://g112-log-migration-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=master&label=experimental&product=chrome&aligned&q=chrome%3Afail

Hexcles:
`chrome:fail` also works now.

It's normal for the searchcache to take a while loading the test runs, during when it'd return 422. The 502 is the problem here, which suggests that the searchcache wasn't able to respond at all (perhaps spending all the CPU and/or memory loading runs). And eventually when the run is loaded completely, searchcache can serve the specific request correctly.

PR #2173, opened 2020-09-25
Use GitHub to authenticate admin access to api/*

Hexcles:
## Description

This is a follow-up to #2175 (which is also a part of #1747 ), finishing up the remaining migration from appengine.user to GitHub OAuth in api/* (namely the results receiver and its admin upload page).

## Review Information

The tests should be fairly self-explanatory. For manual testing:

* Try https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com/admin/results/upload ; You should see permission denied.
* Login on https://staging.wpt.fyi and copy the session cookies to https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com . Then visit https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com/admin/flags again ; you should see again permission denied.
* I'll add your GitHub handle to the admin table. Now you should see the admin upload page. Try uploading a random JSON (not necessarily a wptrunner output). You should see a task successfully scheduled (it won't create a run as the JSON is invalid).

PR #2175, opened 2020-09-25
Use GitHub to manage access to admin/* pages

Hexcles:
Extracted from #2173 (the first half)

#2173 will be rebased to contain only the api/* part after this lands. 

## Review Information

* Try https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com/admin/flags ; You should see permission denied.
* Login on https://staging.wpt.fyi and copy the session cookies to https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com . Then visit https://admin-via-github-dot-wptdashboard-staging.uk.r.appspot.com/admin/flags again ; you should see again permission denied.
* I'll add your GitHub handle to the admin table. Now you should see the admin page. Note that the actual upload function won't work until #2173 lands.

## Changes

It'd be easier to review by commits. Please skip all `*_mock.go` files.

### DIFF
--- a/shared/github_oauth.go
+++ b/shared/github_oauth.go
@@ -34,6 +34,7 @@ type User struct {
 type GitHubAccessControl interface {
 	// IsValid* functions also verify the access token with GitHub.
 	IsValidWPTMember() (bool, error)
+	IsValidAdmin() (bool, error)
 }
 
 type githubAccessControlImpl struct {
@@ -154,7 +155,26 @@ func (gaci githubAccessControlImpl) IsValidWPTMember() (bool, error) {
 	return isMember, err
 }
 
-// NewGitHubAccessControl returns a GitHubAccessControl for checking the permission of a logged-in GitHub user.
+func (gaci githubAccessControlImpl) IsValidAdmin() (bool, error) {
+	valid, err := gaci.isValidAccessToken()
+	if err != nil {
+		return false, err
+	}
+	if !valid {
+		return false, errors.New("Invalid access token")
+	}
+	key := gaci.ds.NewNameKey("Admin", gaci.user.GitHubHandle)
+	var dst struct{}
+	if err := gaci.ds.Get(key, &dst); err == ErrNoSuchEntity {
+		return false, nil
+	} else if err != nil {
+		return false, err
+	}
+	return true, nil
+}
+
+// NewGitHubAccessControl returns a GitHubAccessControl for checking the
+// permission of a logged-in GitHub user.
 func NewGitHubAccessControl(ctx context.Context, ds Datastore, botClient *github.Client, user *User, token string) (GitHubAccessControl, error) {
 	clientID, secret, err := getOAuthClientIDSecret(ds)
 	if err != nil {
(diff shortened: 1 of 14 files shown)


shared/github_oauth.go
```diff
@@ -153,7 +155,24 @@ func (gaci githubAccessControlImpl) IsValidWPTMember() (bool, error) {
...
+	if err != nil {
+		return false, err
+	}
+	if !valid {
+		return false, errors.New("Invalid access token")
+	}
+	key := gaci.ds.NewNameKey("Admin", gaci.user.GitHubHandle)
+	var dst struct{}
+	if err := gaci.ds.Get(key, &dst); err != nil {
+		return false, err
```
KyleJu, line 169: I am getting `Error checking admin` after copied the session cookie, but I am expecting `Admin only` response. What does `gaci.ds.Get(key, &dst);` returns if my name is not in the datastore?

webapp/admin_handler.go
```diff
@@ -15,38 +15,86 @@ import (
...
+		http.Error(w, "Log in from the homepage first", http.StatusUnauthorized)
+		return false
+	}
+	admin, err := acl.IsValidAdmin()
+	if err != nil {
+		log.Errorf("Error checking admin: %s", err.Error())
+		http.Error(w, "Error checking admin", http.StatusInternalServerError)
+		return false
+	}
+	if !admin {
```
KyleJu, line 29: This will never hit -  there is no case that `err == nil` but `admin == false`

shared/github_oauth.go
```diff
@@ -174,20 +193,33 @@ func NewGitHubAccessControl(ctx context.Context, ds Datastore, botClient *github
...
+func NewGitHubAccessControlFromRequest(aeAPI AppEngineAPI, ds Datastore, r *http.Request) (GitHubAccessControl, error) {
+	ctx := aeAPI.Context()
+	botClient, err := aeAPI.GetGitHubClient()
 	if err != nil {
-		log.Errorf("Failed to get secure-cookie-hashkey secret: %s", err.Error())
 		return nil, err
 	}
+	user, token := GetUserFromCookie(ctx, ds, r)
+	if user == nil {
+		return nil, nil
```
KyleJu, line 207: This should return an `error`

webapp/admin_handler.go
```diff
@@ -15,38 +15,86 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+func checkAdmin(acl shared.GitHubAccessControl, log shared.Logger, w http.ResponseWriter) bool {
+	if acl == nil {
```
KyleJu, line 19: Realized this is the `nil` check for the empty/expiry cookie case. I think it makes more sense to return a http error earlier. If you do, you can remove this check here; if not, I am always good with it.

shared/github_oauth.go
Hexcles: Hmm, I'm trying to reserve errors for unexpected stuff (my response to your comment above is in the same vein). In this case, I think missing (valid) cookies is expected (user not logged in) so it shouldn't be an error. WDYT?

webapp/admin_handler.go
Hexcles: You mean setting `http.Error` in `NewGitHubAccessControlFromRequest`? That'd require passing `http.ResponseWriter` into it. I also think setting `http.Error` in a shared function that's meant to be reusable is a bit weird.

webapp/admin_handler.go
Hexcles: You're right; I'm fixing `IsValidAdmin`.

webapp/admin_handler.go
KyleJu: I meant that case where `acl, err := shared.NewGitHubAccessControlFromRequest(a, ds, r)` return `(nil,nil)` when the cookie is not set. It is the only case that `acl` being passed in here is nil. I think your [comment](https://github.com/web-platform-tests/wpt.fyi/pull/2175/files#r495510239) answered my question here - you are reserving the `err` for unexpected stuff.

shared/github_oauth.go
KyleJu: Yea that makes sense. Also see [my comment below](https://github.com/web-platform-tests/wpt.fyi/pull/2175/files#r495681167)

KyleJu (approved):
Tested with the staging instance in this PR and LGTM

PR #2176, opened 2020-09-25
Revert "Add atom for filtering by test-file contents (#1216)"

Hexcles:
This reverts commit 14c45844f0ee96efb359a30b4361e584472b191c, removing
the search-by-content feature which never launched* and whose underlying
AppEngine API is being deprecated (#1747 ).

* It was a PoC. To launch it, we'd also need a cronjob to update the full-text search index.

Also remove the remote_api from the prod configuration; after this revert, we now only use remote_api locally for testing.

This is almost a pure revert (only one small merge conflict). To review, maybe try some structured queries in the staging instance to do some smoke tests.

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -56,16 +56,6 @@ func (tnp TestNamePattern) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return tnp
 }
 
-// FileContentsQuery is a query atom that matches test contents to a query string.
-type FileContentsQuery struct {
-	Query string
-}
-
-// BindToRuns for FileContentsQuery is a no-op; it is independent of test runs.
-func (fcq FileContentsQuery) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
-	return fcq
-}
-
 // SubtestNamePattern is a query atom that matches subtest names to a pattern string.
 type SubtestNamePattern struct {
 	Subtest string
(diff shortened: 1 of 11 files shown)


api/query/cache/index/filter.go
```diff
@@ -236,7 +159,7 @@ func (tnp SubtestNamePattern) Filter(t TestID) bool {
 }
 
 // Filter interprets a TestPath as a filter function over TestIDs.
-func (tp *TestPath) Filter(t TestID) bool {
```
stephenmcgruer, line 239: Can you explain why the various `*` and `&` changes in this file were first made (if you know), and if we need to or should revert them?

api/query/cache/poll/poll.go
```diff
@@ -67,7 +67,7 @@ func KeepRunsUpdated(store shared.Datastore, logger shared.Logger, interval time
 			next := errs[1:]
 			for i := range next {
 				if errs[i] != nil && next[i] == nil {
-					logger.Errorf("Ingested run after skipping %d runs; ingest run attempt errors: %v", i, errs)
```
stephenmcgruer, line 70: We should keep this,  I think :)

stephenmcgruer (approved):
LGMT % comments

api/query/cache/index/filter.go
Hexcles: The reverted change introduced the first mutation on a bound query (see `fcq.searchResults`), which required the receiver type to be a pointer, so `newFilter` had to return `&FileContentsQuery` to match the `filter` interface, but returning a pointer only in some cases can be confusing and error-prone in the future, so `newFilter` was changed to always return struct pointers. 

I don't have strong preference over either (these are tiny structs) but I'd prefer a clean revert here.

api/query/atoms.go
```diff
@@ -979,109 +948,120 @@ func MetadataQualityFromString(quality string) (MetadataQuality, error) {
 func unmarshalQ(b []byte) (AbstractQuery, error) {
 	{
 		var tnp TestNamePattern
-		if err := json.Unmarshal(b, &tnp); err == nil {
```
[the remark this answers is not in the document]
Hexcles: Sure, let me add a follow-up commit to clean it back again; I'll keep the revert pure.

api/query/cache/poll/poll.go
```diff
@@ -67,7 +67,7 @@ func KeepRunsUpdated(store shared.Datastore, logger shared.Logger, interval time
...
+					logger.Errorf("Ingested run after skipping %d runs; ingest run attempt errors: %v", errs)
```
Hexcles, line 70:
```suggestion
					logger.Errorf("Ingested run after skipping %d runs; ingest run attempt errors: %v", i, errs)
```

api/query/cache/index/filter.go
stephenmcgruer: Ok, revert sounds good to me in that case. Thanks!

PR #2190, opened 2020-10-07
[Go 1.12] Correct the Google Cloud logging setup for searchcache

KyleJu:
A part of https://github.com/web-platform-tests/wpt.fyi/issues/1747, this PR modifies searchcache logging to the correct google cloud logging setup.

With this change, searcache logging will show structured logging and severity in the appengine. A few behavioral changes are introduced:

- When fail to initiate gclog, the program is [terminated](https://github.com/web-platform-tests/wpt.fyi/pull/2190/files#diff-29135f505c736f13d37da653c9755a6a17118cd4363bee5998e54d5a04cb615cR168), rather than [continue without a logger](https://github.com/web-platform-tests/wpt.fyi/pull/2190/files#diff-63407b1c55c2275cab0b2de74003827d3e5a4fdaa4c3208dcce9e5fb91b580bcL150)
- HTTP response status codes are not set for gclog, and it might also affect Stackdriver alerting; see the [discussions](https://github.com/web-platform-tests/wpt.fyi/pull/2190#discussion_r502648715). This issue can be sort of bypassed by filtering the severity level of parent logs in appengine.
- In process of testing, the staging instance occasionally returns 502 errors; see the [comment here](https://github.com/web-platform-tests/wpt.fyi/pull/2190#pullrequestreview-505048774).

## Test
 To view this log, go to [appengine](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gae_app%22%20resource.labels.module_id%3D%22searchcache%22%20resource.labels.version_id%3D%22go112-log-draft%22%0AlogName%3D%22projects%2Fwptdashboard-staging%2Flogs%2Frequest_log%22;timeRange=PT1H?serviceId=searchcache&project=wptdashboard-staging) and search log_name `request_log`.

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -14,7 +14,6 @@ import (
 	"syscall"
 	"time"
 
-	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/datastore"
 	gclog "cloud.google.com/go/logging"
 	"github.com/sirupsen/logrus"
@@ -29,7 +28,7 @@ import (
 
 var (
 	port                   = flag.Int("port", 8080, "Port to listen on")
-	projectID              = flag.String("project_id", "", "Google Cloud Platform project ID, if different from ID detected from metadata service")
+	projectID              = flag.String("project_id", "", "Google Cloud Platform project ID, if different from ID detected from environment")
 	gcpCredentialsFile     = flag.String("gcp_credentials_file", "", "Path to Google Cloud Platform credentials file, if necessary")
 	numShards              = flag.Int("num_shards", runtime.NumCPU(), "Number of shards for parallelizing query execution")
 	monitorInterval        = flag.Duration("monitor_interval", time.Second*5, "Polling interval for memory usage monitor")
@@ -107,17 +106,17 @@ func init() {
 
 	maxRunsPerRequestMsg = fmt.Sprintf("Too many runs specified; maximum is %d.", *maxRunsPerRequest)
 
-	autoProjectID, err := metadata.ProjectID()
-	if err != nil {
-		logrus.Warningf("Failed to get project ID from metadata service")
+	autoProjectID := os.Getenv("GOOGLE_CLOUD_PROJECT")
+	if autoProjectID == "" {
+		logrus.Warningf("Failed to get project ID from environment")
 	} else {
 		if *projectID == "" {
-			logrus.Infof(`Using project ID from metadata service: %s`, autoProjectID)
+			logrus.Infof("Using project ID from environment: %s", autoProjectID)
 			*projectID = autoProjectID
 		} else if *projectID != autoProjectID {
-			logrus.Warningf(`Using project ID from flag: "%s" even though metadata service reports project ID of "%s"`, *projectID, autoProjectID)
+			logrus.Warningf("Using project ID from flag: %s, even though environment reports project ID: %s", *projectID, autoProjectID)
 		} else {
-			logrus.Infof(`Using project ID: %s`, *projectID)
+			logrus.Infof("Using project ID: %s", *projectID)
 		}
 	}
 
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -167,6 +167,7 @@ func main() {
 	}
 	defer gclogClient.Close()
 
+	// Reuse loggers to prevent leaking goroutines: https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870
 	childLogger := gclogClient.Logger("request_log_entries", gclog.CommonResource(&monitoredResource))
 	parentLogger := gclogClient.Logger("request_log", gclog.CommonResource(&monitoredResource))
 
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -16,6 +16,7 @@ import (
 
 	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/datastore"
+	gclog "cloud.google.com/go/logging"
 	"github.com/sirupsen/logrus"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/backfill"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
@@ -158,12 +159,22 @@ func main() {
 	var netClient = &http.Client{
 		Timeout: time.Second * 5,
 	}
+
+	// Initializes Logger
+	var gcClient *gclog.Client
+
+	gcClient, err = gclog.NewClient(context.Background(), *projectID)
+	if err != nil {
+		logrus.Fatalf("Failed to initiate gclog: %v", err)
+	}
+	defer gcClient.Close()
+
 	// Polls Metadata update every 10 minutes.
 	go poll.KeepMetadataUpdated(netClient, logger, time.Minute*10)
 
 	http.HandleFunc("/_ah/liveness_check", livenessCheckHandler)
 	http.HandleFunc("/_ah/readiness_check", readinessCheckHandler)
-	http.HandleFunc("/api/search/cache", shared.HandleWithGoogleCloudLogging(searchHandler, *projectID, &monitoredResource))
+	http.HandleFunc("/api/search/cache", shared.HandleWithGoogleCloudLogging(searchHandler, gcClient, *projectID, &monitoredResource))
 	logrus.Infof("Listening on port %d", *port)
 	logrus.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
 }
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -160,17 +160,17 @@ func main() {
 		Timeout: time.Second * 5,
 	}
 
-	// Initializes Logger
-	var gcClient *gclog.Client
+	// Initializes Logger.
+	var gclogClient *gclog.Client
 
-	gcClient, err = gclog.NewClient(context.Background(), *projectID)
+	gclogClient, err = gclog.NewClient(context.Background(), *projectID)
 	if err != nil {
-		logrus.Fatalf("Failed to initiate gclog: %v", err)
+		logrus.Fatalf("Failed to initiate gclog Client: %v", err)
 	}
-	defer gcClient.Close()
+	defer gclogClient.Close()
 
-	childLogger := gcClient.Logger("request_log_entries", gclog.CommonResource(&monitoredResource))
-	parentLogger := gcClient.Logger("request_log", gclog.CommonResource(&monitoredResource))
+	childLogger := gclogClient.Logger("request_log_entries", gclog.CommonResource(&monitoredResource))
+	parentLogger := gclogClient.Logger("request_log", gclog.CommonResource(&monitoredResource))
 
 	// Polls Metadata update every 10 minutes.
 	go poll.KeepMetadataUpdated(netClient, logger, time.Minute*10)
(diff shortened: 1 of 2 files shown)


KyleJu:
with d4978bf,  the setup looks correct
good

KyleJu:
compared to staging logging for the same `chrome:fail` query
bad

KyleJu:
> with [d4978bf](https://github.com/web-platform-tests/wpt.fyi/commit/d4978bf1763965fa980b2cde27c6576318adab84), the setup looks correct
> good

@Hexcles How does this logging look to you?

shared/logger.go
```diff
@@ -143,14 +135,31 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
...
 			h(w, r)
 			return
 		}
 		h(w, r.WithContext(ctx))
+		parentLogger := gcClient.Logger("request", gclog.CommonResource(commonResource))
+		e := gclog.Entry{
+			Trace:    traceID,
+			Severity: gclog.Info,
+			HTTPRequest: &gclog.HTTPRequest{
+				Request: r,
```
Hexcles, line 156: Let's try setting the latency here (by getting a timestamp at the beginning of the function and subtracting it from the current timestamp). Looks like this might be the missing piece to group logs correctly.

Hexcles (commented):
Thinking more about the 502 errors, I wonder if we are seeing a memory leak similar to https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870 . We might want to try reusing the loggers.

shared/logger.go
KyleJu: Right, it is mentioned in [the issue you found](https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-320780311)

KyleJu:
Searchcache still encounters 502 errors but the logging is nested properly. To view this log, go to [appengine](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gae_app%22%20resource.labels.module_id%3D%22searchcache%22%20resource.labels.version_id%3D%22go112-log-draft%22%0AlogName%3D%2528%22projects%2Fwptdashboard-staging%2Flogs%2Frequest%22%2529;timeRange=PT1H?serviceId=searchcache&project=wptdashboard-staging) and search log_name `reqest`.

work

shared/logger.go
```diff
@@ -143,14 +134,35 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
...
 			h(w, r)
 			return
 		}
 		h(w, r.WithContext(ctx))
+
+		end := time.Now()
+		e := gclog.Entry{
+			Timestamp: end,
+			Trace:     traceID,
+			Severity:  gclog.Info,
```
KyleJu, line 156: Note that Severity here has the same problem as statuscode, since it should be `max(severity of all children logs)`

KyleJu: e.g. https://source.chromium.org/chromium/infra/infra/+/master:go/src/go.chromium.org/luci/gae/impl/cloud/request.go;l=174

shared/logger.go
```diff
@@ -143,14 +135,31 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
...
 			return
 		}
 		h(w, r.WithContext(ctx))
+		parentLogger := gcClient.Logger("request", gclog.CommonResource(commonResource))
+		e := gclog.Entry{
+			Trace:    traceID,
+			Severity: gclog.Info,
+			HTTPRequest: &gclog.HTTPRequest{
+				Request: r,
+				Status:  200,
```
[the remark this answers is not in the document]
KyleJu: There is one issue and one question if set it to 0 

- Without a status code, it is **difficult** to screen failed requests in the log. But we can bypass it by filtering the severity level and the failed requests are nested under the filtered logs
- How would it affect error reporting and monitoring in Stackdriver? I assume alerts are trigged by the number of non-200 responses?

shared/logger.go
KyleJu: I can fix this by keeping track of max severity in the gcLogger, since it is attached to the context

shared/logger.go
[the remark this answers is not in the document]
KyleJu: I also have a hacky solution in mind - storing it in the `context.WithValue`... although it is not recommended in the documentation 

`Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.`

shared/logger.go
KyleJu: we can store it in the `gcLogger` using a map that maps traceID to severity. `gcLogger` is created per request and hopefully there is no race condition

shared/logger.go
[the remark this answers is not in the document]
Hexcles: Yeah, not setting status code is definitely not great and indeed might affect our alerting (but only after we apply the new logging setup to the default webapp), but it's still better than setting the wrong status code unconditionally. My point is, if it turns out to be difficult to implement without major refactoring, we should probably work on it in a separate PR and leave it as 0 (unset) in this PR.

Let's leave this open for now and I'll also try to find a way to get the status code. Putting this on context does seem a bit borderline.

api/query/cache/service/main.go
```diff
@@ -158,12 +159,25 @@ func main() {
...
+	// Initializes Logger
+	var gcClient *gclog.Client
+
+	gcClient, err = gclog.NewClient(context.Background(), *projectID)
+	if err != nil {
+		logrus.Fatalf("Failed to initiate gclog: %v", err)
+	}
+	defer gcClient.Close()
+
+	childrenLogger := gcClient.Logger("request_logs", gclog.CommonResource(&monitoredResource))
```
Hexcles, line 172: Nit: probably better to call it `childLogger`. Also, let's use `request_log_entries` for consistency with the current setup (note the current `childLogger` logs to `request_log_entries`).
```suggestion
	childLogger := gcClient.Logger("request_log_entries", gclog.CommonResource(&monitoredResource))
```

shared/logger.go
```diff
@@ -118,20 +118,11 @@ func (gcl *gcLogger) Errorf(format string, params ...interface{}) {
 
 // newAppEngineFlexContext creates a new Google App Engine Flex-based
 // context, with a Google Cloud logger client bound to an http.Request.
-func newAppEngineFlexContext(r *http.Request, project string, commonResource *mrpb.MonitoredResource) (ctx context.Context, err error) {
+func newAppEngineFlexContext(r *http.Request, gcClient *gclog.Client, traceID string, childrenLogger *gclog.Logger) (ctx context.Context, err error) {
```
Hexcles, line 121: Does this still need `gcClient`?

shared/logger.go
```diff
@@ -143,14 +134,35 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
 // commonResource is an optional override to the monitored resource details appended to each log.
 // e.g. in the Flex environment, it pays to override this value to type gae_app, to ensure finding
 // logs is consistent between services.
-func HandleWithGoogleCloudLogging(h http.HandlerFunc, project string, commonResource *mrpb.MonitoredResource) http.HandlerFunc {
+func HandleWithGoogleCloudLogging(h http.HandlerFunc, gcClient *gclog.Client, project string, childrenLogger, parentLogger *gclog.Logger) http.HandlerFunc {
```
Hexcles, line 137: same here for `gcClient`

shared/logger.go
Hexcles: Yeah that sounds good.

shared/logger.go
```diff
@@ -143,14 +134,35 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
...
 		h(w, r.WithContext(ctx))
+
+		end := time.Now()
+		e := gclog.Entry{
+			Timestamp: end,
+			Trace:     traceID,
+			Severity:  gclog.Info,
+			HTTPRequest: &gclog.HTTPRequest{
+				Request: r,
+				Status:  0,
```
Hexcles, line 159:
```suggestion
				// TODO: set Status based on w
```
0 is the default. Let's not set it again and leave a comment instead.

api/query/cache/service/main.go
```diff
@@ -158,12 +159,25 @@ func main() {
...
+	parentLogger := gcClient.Logger("request", gclog.CommonResource(&monitoredResource))
```
Hexcles, line 173:
```suggestion
	parentLogger := gcClient.Logger("request_log", gclog.CommonResource(&monitoredResource))
```
(trying to follow the naming convention of App Engine standard here)

Hexcles (commented):
Awesome! The logs look correct now!

api/query/cache/service/main.go
```diff
@@ -158,12 +159,25 @@ func main() {
 	var netClient = &http.Client{
 		Timeout: time.Second * 5,
 	}
+
+	// Initializes Logger.
+	var gclogClient *gclog.Client
+
+	gclogClient, err = gclog.NewClient(context.Background(), *projectID)
```
Hexcles, line 166: Would this work?
```suggestion
	gclogClient, err := gclog.NewClient(context.Background(), *projectID)
```

shared/logger.go
```diff
@@ -90,9 +90,15 @@ func NewAppEngineContext(r *http.Request) context.Context {
 type gcLogger struct {
 	childLogger *gclog.Logger
 	traceID     string
+	maxSeverity gclog.Severity
 }
 
 func (gcl *gcLogger) log(severity gclog.Severity, format string, params ...interface{}) {
+	// The underlying type of gclog.Severity is int, https://pkg.go.dev/cloud.google.com/go/logging#Severity.
```
Hexcles, line 97: The more important piece of doc is about the order:
```suggestion
	// "Severity levels are ordered, with numerically smaller levels treated as less severe than numerically larger levels".
	// https://pkg.go.dev/cloud.google.com/go/logging#Severity.
```

api/query/cache/service/main.go
```diff
@@ -158,12 +159,25 @@ func main() {
...
+	if err != nil {
+		logrus.Fatalf("Failed to initiate gclog Client: %v", err)
+	}
+	defer gclogClient.Close()
+
```
Hexcles, line 171:
```suggestion

	// Reuse loggers to prevent leaking goroutines: https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870
```

Hexcles (approved):
LGTM % nits

api/query/cache/service/main.go
KyleJu: Yup it will work. I just thought that declaring `gclogClient` would make it more clear on what is being deferred.

shared/logger.go
KyleJu: Thanks!

KyleJu:
@Hexcles I reran `go_test (pull_request)` but it couldn't seem to pass. Could you admin merge?

shared/logger.go
```diff
@@ -143,14 +154,35 @@ func newAppEngineFlexContext(r *http.Request, project string, commonResource *mr
...
+		h(gcResponseWriter{gcLogger: &gcl, w: w}, r.WithContext(withLogger(r.Context(), &gcl)))
+
+		end := time.Now()
+		e := gclog.Entry{
+			Timestamp: end,
+			Trace:     traceID,
+			Severity:  gcl.maxSeverity,
+			HTTPRequest: &gclog.HTTPRequest{
+				Request: r,
+				Status:  gcl.statusCode,
```
Hexcles, line 180: @KyleJu FYI I've fixed this by wrapping the `ResponseWriter`.

KyleJu: Thanks for the fix! The same fix would work for https://github.com/web-platform-tests/wpt.fyi/pull/2208. I was just thinking if there could be any problem when HttpResponder and context shared the same instance of a Logger but nothing came up. Very clean fix nevertheless!

Issue #2189, opened 2020-10-06
Inconsistent location IDs in prod


Hexcles:
For some reason, AppEngine in our prod project runs in the `us-central` region, which doesn't exist on https://cloud.google.com/about/locations#americas . However, Cloud Tasks runs in the correct `us-central1` region. This inconsistency causes `Failed to schedule task: rpc error: code = InvalidArgument desc = Location 'us-central' is not a valid location. Use ListLocations to list valid locations.` when trying to schedule tasks using the new Cloud Tasks SDK (#2153).

Hexcles:
Rolling back #2188 for now. We've lost about a day's worth of test runs on prod.

PR #2200, opened 2020-10-13
Actually use the correct location when scheduling tasks

Hexcles:
Follow-up to #2189 where I forgot to actually use the new variable.

Sorry about the mistake. We really should have a test for this. Short of a local emulator for Cloud Tasks, WDYT about introspecting the proto?

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -241,39 +241,7 @@ func (a appEngineAPIImpl) ScheduleTask(queueName, taskName, target string, param
 		panic("Clients.cloudtasks is nil")
 	}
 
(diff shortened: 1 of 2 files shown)


stephenmcgruer (approved):
> Follow-up to #2189 where I forgot to actually use the new variable.

Ahaha! This is why we have code review, so this comes down 100% on your code reviewer imo, which was me :D.

> Sorry about the mistake. We really should have a test for this. Short of a local emulator for Cloud Tasks, WDYT about introspecting the proto?

Yes, this would be good. Probably a bit of a mess of introducing even more API shims so we can inject mocks, but this logic got complex enough that it's worth testing. Here or follow-up, your choice.

PR #2207, opened 2020-10-18
Refactor some API handlers to stop using memcache directly

Hexcles:
This is part of #1747 and a pure refactor without any behaviour changes.

Refactor `api/manifest.go` (and its test) and `api/metadata_cache.go` to use the Memcache interfaces from `shared/cache.go` in preparation for switching over to Cloud Redis.


api/manifest.go
```diff
@@ -41,57 +40,96 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
-		// "latest" caches which SHA is latest; Return the manifest for that SHA.
-		if sha == "latest" {
-			return getManifest(ctx, manifestAPI, string(cached.Value), paths)
+	if shared.IsLatest(sha) {
+		// Attempt to find the "latest" SHA in cache.
+		if latestSHA, err := readByKey(latestMC, "latest"); err != nil {
+			log.Debugf("Latest SHA not found in cache: %v", err)
+		} else {
+			// Found! Now delegate to get manifest for that specific SHA.
+			return getManifest(ctx, manifestAPI, string(latestSHA), paths)
```
KyleJu: I guess it is safe to assume cached != nil?

api/manifest.go
```diff
@@ -41,57 +40,96 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
 
-	if err != nil && err != memcache.ErrCacheMiss {
-		return "", nil, err
-	} else if cached != nil {
-		// "latest" caches which SHA is latest; Return the manifest for that SHA.
-		if sha == "latest" {
-			return getManifest(ctx, manifestAPI, string(cached.Value), paths)
+	if shared.IsLatest(sha) {
+		// Attempt to find the "latest" SHA in cache.
+		if latestSHA, err := readByKey(latestMC, "latest"); err != nil {
```
KyleJu, line 51: `mc`

api/manifest.go
```diff
@@ -41,57 +40,96 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
 
 func getManifest(ctx context.Context, manifestAPI manifest.API, sha string, paths []string) (string, []byte, error) {
-	if sha == "" {
-		sha = "latest"
-	}
-	fetchedSHA := sha
+	log := shared.GetLogger(ctx)
+	mc := shared.NewMemcacheReadWritable(ctx, time.Hour*48)
+	// Shorter expiry for latest SHA, to keep it current.
+	latestMC := shared.NewMemcacheReadWritable(ctx, time.Minute*5)
```
KyleJu: This can be moved to line 85. see [my comment](https://github.com/web-platform-tests/wpt.fyi/pull/2207/commits/5843e88e16d7a5d2c770257fd0c5b6eaf1140ec6#r508210968)

api/manifest.go
```diff
@@ -40,58 +38,96 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil && err != memcache.ErrCacheMiss {
-		return "", nil, err
-	} else if cached != nil {
-		// "latest" caches which SHA is latest; Return the manifest for that SHA.
-		if sha == "latest" {
-			return getManifest(ctx, manifestAPI, string(cached.Value), paths)
+	if shared.IsLatest(sha) {
+		// Attempt to find the "latest" SHA in cache.
+		if latestSHA, err := readByKey(latestMC, "latest"); err != nil {
+			log.Debugf("Latest SHA not found in cache: %v", err)
```
KyleJu, line 52: Just double check: this line would have `return "", nil, err` in the original code. With your changes, manifest will be fetched regardless?

KyleJu:
@Hexcles I reviewed commit by commit and the comments are misaligned. All my comments so far are actually for commit one. Sorry about the inconvenience!

api/manifest.go
```diff
@@ -40,58 +38,96 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
 	var body []byte
-	cached, err := memcache.Get(ctx, manifestCacheKey(sha))
 
-	if err != nil && err != memcache.ErrCacheMiss {
-		return "", nil, err
-	} else if cached != nil {
-		// "latest" caches which SHA is latest; Return the manifest for that SHA.
-		if sha == "latest" {
-			return getManifest(ctx, manifestAPI, string(cached.Value), paths)
+	if shared.IsLatest(sha) {
```
KyleJu, line 49: Food for thought:  rewriting it [this way](https://play.golang.org/p/geYvFjHPvA4) can get rid of the nested `else`

Hexcles: That looks better. Thanks!

api/manifest.go
```diff
@@ -40,58 +38,98 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
-	memcache.Set(ctx, item)
 
-	// Shorter expiry for latest SHA, to keep it current.
+	// Write latest SHA to cache, if needed.
 	if shared.IsLatest(sha) {
-		latestSHAItem := &memcache.Item{
-			Key:        manifestCacheKey("latest"),
-			Value:      []byte(fetchedSHA),
-			Expiration: time.Minute * 5,
+		if err := writeByKey(latestMC, "latest", []byte(fetchedSHA)); err != nil {
```
KyleJu, line 85: Without knowing too much about manifest, would there be an issue if this step fails, the `latest` sha in the cache is outdated but `getManifest` returns the correct sha? e.g. [line 48](https://github.com/web-platform-tests/wpt.fyi/pull/2207/files#diff-fe0077ecdbb4f52d4a445846f85fd781e87d4a66d862b3778ceb4c503ee3cf67R48) returns on the incorrect `latest` sha?

api/manifest.go
```diff
@@ -40,58 +38,98 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
+	writer, err := writable.NewWriteCloser(ikey)
+	if err != nil {
+		return err
+	}
+	n, err := writer.Write(body)
+	if err != nil {
+		return err
+	}
+	if n != len(body) {
+		return fmt.Errorf("incomplete write; expected %d bytes but %d written", len(body), n)
```
KyleJu, line 132:
```suggestion
		return fmt.Errorf("Incomplete write; expected %d bytes but %d written", len(body), n)
```

Hexcles: Logs are capitalized, but errors are not. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

api/manifest.go
Hexcles: That's possible. Without an atomic write (e.g. combining "latest" and the manifest content into a single item), there's going to be some race condition and/or inconsistency one way or another. The original implementation also didn't check for errors here. I think the risk is tolerable given the short expiration for "latest".

api/manifest.go
```diff
@@ -40,58 +38,98 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
-			Value:      []byte(fetchedSHA),
-			Expiration: time.Minute * 5,
+		if err := writeByKey(latestMC, "latest", []byte(fetchedSHA)); err != nil {
+			log.Errorf("Error writing latest SHA to cache: %v", err)
 		}
-		memcache.Set(ctx, latestSHAItem)
 	}
 
+	// Decompress the manifest and filter it by paths if needed.
 	gzReader, err := gzip.NewReader(bytes.NewReader(body))
```
KyleJu, line 91: I think this needs to be closed; see [doc](https://golang.org/pkg/compress/gzip/#NewReader) and [an example](https://github.com/web-platform-tests/wpt.fyi/blob/master/shared/metadata_util.go#L74)

KyleJu, line 91: I think this needs to be closed; see [doc](https://golang.org/pkg/compress/gzip/#NewReader) and [an example](https://github.com/web-platform-tests/wpt.fyi/blob/master/shared/metadata_util.go#L74)

api/manifest.go
KyleJu: Aha thank you. I will keep that in mind

api/manifest.go
Hexcles: Good point! Thanks.

api/manifest.go
```diff
@@ -40,58 +38,98 @@ func apiManifestHandler(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		return fetchedSHA, nil, err
 	}
-	body, err = ioutil.ReadAll(gzReader)
-	return fetchedSHA, body, err
+	unzipped, err := ioutil.ReadAll(gzReader)
```
Hexcles, line 95:
```suggestion
	defer gzReader.Close()
	unzipped, err := ioutil.ReadAll(gzReader)
```

api/manifest.go
KyleJu: Right that is good point on the short expiration

KyleJu (approved):

Hexcles:
Tested manually using https://memcache-dot-wptdashboard-staging.uk.r.appspot.com/api/metadata?product=chrome by checking AppEngine logs.

PR #2208, opened 2020-10-19
[Go 1.12] Set up Google Cloud Logging in standard AppEngine

KyleJu:
A part of #1747, this PR migrates the appengine logging to Google Cloud Logging in standard AppEngine

With this change,  minor behavioral changes are introduced:

- In dev_appserver, logrus is attached to  HTTP incoming request contexts.

## Test
[See logs here](https://pantheon.corp.google.com/logs/query;query=resource.type%3D%22gae_app%22%20resource.labels.module_id%3D%22default%22%20resource.labels.version_id%3D%22gclog-standard%22%0AlogName%3D%22projects%2Fwptdashboard-staging%2Flogs%2Frequest_log%22;timeRange=PT1H?serviceId=default&project=wptdashboard-staging)

### DIFF
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -15,7 +15,6 @@ import (
 	"time"
 
 	"cloud.google.com/go/datastore"
-	gclog "cloud.google.com/go/logging"
 	"github.com/sirupsen/logrus"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/backfill"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
@@ -159,23 +158,18 @@ func main() {
 		Timeout: time.Second * 5,
 	}
 
-	// Initializes Logger.
-	gclogClient, err := gclog.NewClient(context.Background(), *projectID)
-	if err != nil {
-		logrus.Fatalf("Failed to initiate gclog Client: %v", err)
+	// Initializes clients.
+	if err = shared.Clients.Init(context.Background()); err != nil {
+		logrus.Fatalf("Failed to initialize Google Cloud clients: %v", err)
 	}
-	defer gclogClient.Close()
-
-	// Reuse loggers to prevent leaking goroutines: https://github.com/googleapis/google-cloud-go/issues/720#issuecomment-346199870
-	childLogger := gclogClient.Logger("request_log_entries", gclog.CommonResource(&monitoredResource))
-	parentLogger := gclogClient.Logger("request_log", gclog.CommonResource(&monitoredResource))
+	defer shared.Clients.Close()
 
 	// Polls Metadata update every 10 minutes.
 	go poll.KeepMetadataUpdated(netClient, logger, time.Minute*10)
 
 	http.HandleFunc("/_ah/liveness_check", livenessCheckHandler)
 	http.HandleFunc("/_ah/readiness_check", readinessCheckHandler)
-	http.HandleFunc("/api/search/cache", shared.HandleWithGCLFlex(searchHandler, *projectID, childLogger, parentLogger))
+	http.HandleFunc("/api/search/cache", shared.HandleWithLogging(searchHandler))
 	logrus.Infof("Listening on port %d", *port)
 	logrus.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
 }
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -175,7 +175,7 @@ func main() {
 
 	http.HandleFunc("/_ah/liveness_check", livenessCheckHandler)
 	http.HandleFunc("/_ah/readiness_check", readinessCheckHandler)
-	http.HandleFunc("/api/search/cache", shared.HandleWithGoogleCloudLogging(searchHandler, *projectID, childLogger, parentLogger))
+	http.HandleFunc("/api/search/cache", shared.HandleWithGCLFlex(searchHandler, *projectID, childLogger, parentLogger))
 	logrus.Infof("Listening on port %d", *port)
 	logrus.Fatal(http.ListenAndServe(fmt.Sprintf(":%d", *port), nil))
 }
--- a/api/query/cache/service/main.go
+++ b/api/query/cache/service/main.go
@@ -22,12 +22,11 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/poll"
 	"github.com/web-platform-tests/wpt.fyi/shared"
 	"google.golang.org/api/option"
-	mrpb "google.golang.org/genproto/googleapis/api/monitoredres"
 )
 
 var (
 	port                   = flag.Int("port", 8080, "Port to listen on")
-	projectID              = flag.String("project_id", "", "Google Cloud Platform project ID, if different from ID detected from environment")
+	projectID              = flag.String("project_id", "", "Google Cloud Platform project ID, used for connecting to Datastore")
 	gcpCredentialsFile     = flag.String("gcp_credentials_file", "", "Path to Google Cloud Platform credentials file, if necessary")
 	numShards              = flag.Int("num_shards", runtime.NumCPU(), "Number of shards for parallelizing query execution")
 	monitorInterval        = flag.Duration("monitor_interval", time.Second*5, "Polling interval for memory usage monitor")
@@ -45,8 +44,6 @@ var (
 	idx index.Index
 	mon monitor.Monitor
 )
-
-var monitoredResource mrpb.MonitoredResource
 
 func livenessCheckHandler(w http.ResponseWriter, r *http.Request) {
 	w.Write([]byte("Alive"))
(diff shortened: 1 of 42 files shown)


shared/appengine.go
```diff
@@ -39,18 +45,52 @@ func (c *clientsImpl) Init(ctx context.Context) (err error) {
...
+	if err != nil {
+		return err
+	}
+
+	monitoredResource := mrpb.MonitoredResource{
+		Type: "gae_app",
+		Labels: map[string]string{
+			"project_id": os.Getenv("GOOGLE_CLOUD_PROJECT"),
+			// https://cloud.google.com/appengine/docs/standard/go/runtime#environment_variables
+			"module_id":  os.Getenv("GAE_SERVICE"),
```
KyleJu, line 63: use ` runtimeIdentity` instead of os

KyleJu:
Setup looks correct and surprisingly the status code is shown correctly as well

Screen Shot 2020-09-28 at 6 16 05 PM
Screen Shot 2020-10-19 at 4 26 34 PM

KyleJu:
See another example `api/user`; the response code is 401

Screen Shot 2020-10-19 at 4 25 59 PM

shared/logger.go
```diff
@@ -54,45 +54,62 @@ func NewNilLogger() Logger {
...
-// newGAELogger returns a Google App Engine Standard Environment logger bound to
-// the given context.
-func newGAELogger(ctx context.Context) Logger {
-	return gaeLogger{ctx}
+	}
 }
 
 // NewAppEngineContext creates a new Google App Engine Standard-based
 // context bound to an http.Request.
 func NewAppEngineContext(r *http.Request) context.Context {
```
KyleJu, line 98: Remove the method and replace all instances with `r.Context()`

KyleJu: Double check if it affects mocks

webapp/web/main.go
```diff
@@ -28,6 +28,7 @@ func init() {
 }
 
 func main() {
+	// TODO: The issue here is if cloudtasks initialized successfully but gclog doesn't, cloudtasks still needs to be closed.
```
KyleJu, line 31: Address this TODO

shared/logger.go
```diff
@@ -54,45 +54,62 @@ func NewNilLogger() Logger {
...
-}
-
-func (l gaeLogger) Debugf(format string, args ...interface{}) {
-	gaelog.Debugf(l.ctx, format, args...)
-}
+// HandleWithStandardGCL - TODO: Refactor
+func HandleWithStandardGCL(h http.HandlerFunc, project string, childLogger, parentLogger *gclog.Logger) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if project == "" || childLogger == nil || parentLogger == nil {
+			h(w, r)
```
KyleJu, line 61: This case pollutes the log with `Context without logger`; [see](https://github.com/web-platform-tests/wpt.fyi/blob/42dd9f9bc9e9e34f1cfe346395e2981e9a93c373/shared/logger.go#L162)

shared/logger.go
KyleJu: Assign `NewNilLogger` in dev_appserver

shared/logger.go
KyleJu: Will attach logrus in dev_appserver

shared/logger.go
```diff
@@ -54,45 +54,62 @@ func NewNilLogger() Logger {
...
 }
 
-type gaeLogger struct {
-	ctx context.Context
-}
-
-func (l gaeLogger) Debugf(format string, args ...interface{}) {
-	gaelog.Debugf(l.ctx, format, args...)
-}
+// HandleWithStandardGCL - TODO: Refactor
```
KyleJu, line 57: Fix this TODO

shared/logger.go
```diff
@@ -54,45 +54,62 @@ func NewNilLogger() Logger {
...
 }
 
 type gcLogger struct {
 	childLogger *gclog.Logger
 	traceID     string
+	maxSeverity gclog.Severity
 }
 
 func (gcl *gcLogger) log(severity gclog.Severity, format string, params ...interface{}) {
+	if int(severity) > int(gcl.maxSeverity) {
```
KyleJu, line 103: TODO: this should be gone after merge

KyleJu:
Structured logging seems incorrect, as it only show the same children logs

Screen Shot 2020-10-19 at 8 08 20 PM

shared/logger.go
```diff
@@ -54,45 +54,56 @@ func NewNilLogger() Logger {
...
+			traceID = fmt.Sprintf("projects/%s/traces/%s", project, traceID)
+		}
 
-func (l gaeLogger) Warningf(format string, args ...interface{}) {
-	gaelog.Warningf(l.ctx, format, args...)
-}
+		gcl := gcLogger{
+			childLogger: childLogger,
+			traceID:     traceID,
+			maxSeverity: gclog.Default,
```
KyleJu, line 75: maybe change the default severity to Info, consistent with the appengine default level

KyleJu:
There is a race condition between the construction of parent/childLogger vs AddRoute, where the former is called in [main()](https://github.com/web-platform-tests/wpt.fyi/blob/9fe7d7ea2f431ed79690a1e7775ccb7cc6a17831/webapp/web/main.go#L30) and the latter is called in [init()](https://github.com/web-platform-tests/wpt.fyi/blob/9fe7d7ea2f431ed79690a1e7775ccb7cc6a17831/webapp/web/main.go#L18)

Hexcles:
All `init`s should happen before `main`. The execution order among multiple `init`s is less clear (either undefined or shouldn't be relied on anyway).

KyleJu:
> All `init`s should happen before `main`. The execution order among multiple `init`s is less clear (either undefined or shouldn't be relied on anyway).

I guess race condition is not exact the right word but the problem is here - Logger is used before they are initialized; used in init() but constructed in main()

KyleJu:
I noticed the same `{}` in [searchcache](https://github.com/web-platform-tests/wpt.fyi/pull/2190) but besides this issue, the setup looks correct
Screen Shot 2020-10-21 at 10 22 08 AM (2)

KyleJu:
@Hexcles while I am trying to figure out [{} issue](https://github.com/web-platform-tests/wpt.fyi/pull/2208#issuecomment-713616190), this code is ready for review. Notice that I will send a follow-up CL to merge the https://github.com/web-platform-tests/wpt.fyi/pull/2190 changes.

It is easier to review commit by commit.

KyleJu:
@Hexcles Ready for review

Hexcles (approved):
LGTM % nits

I've pushed a commit directly to the branch to address some minor nits. PTAL.

api/query/cache/service/main.go
```diff
@@ -159,23 +158,18 @@ func main() {
 		Timeout: time.Second * 5,
 	}
 
-	// Initializes Logger.
-	gclogClient, err := gclog.NewClient(context.Background(), *projectID)
-	if err != nil {
-		logrus.Fatalf("Failed to initiate gclog Client: %v", err)
+	// Initializes clients.
+	if err = shared.Clients.Init(context.Background()); err != nil {
```
KyleJu, line 162: Two small things:

- There is a case where  *projectID !=  os.Getenv("GOOGLE_CLOUD_PROJECT"); [see here](https://github.com/web-platform-tests/wpt.fyi/pull/2208/commits/74e90b3816bca2b1b030d4ef997207e66d6bfe17#diff-29135f505c736f13d37da653c9755a6a17118cd4363bee5998e54d5a04cb615cL117)

- As a result of this change, cloudtasks will also be initialized, which isn't used anywhere in flex. It looks like the library is available in the [Flex environment](https://cloud.google.com/tasks/docs/migrating), while Task Queues is only available in Standard. (The initialization is redundant but it is safe to do so)

I will also remove the initialization `monitoredResource` in the next commit

shared/appengine.go
```diff
@@ -75,17 +76,16 @@ func (c *clientsImpl) Close() {
 	if c.cloudtasks != nil {
 		err := c.cloudtasks.Close()
 		if err != nil {
-			logrus.Warningf("Error closing cloudtasks: %s", err.Error())
+			log.Printf("Error closing cloudtasks: %s", err.Error())
```
KyleJu, line 79: Is there a reason to use log vs logrus here?

api/query/cache/service/main.go
Hexcles: > There is a case where *projectID != os.Getenv("GOOGLE_CLOUD_PROJECT"); see here

Great point. I decided to clarify in the help message that `--project_id` is only used for connecting to Datastore, which is useful when debugging searchcache locally. I don't think we'd ever want sending local logs to GCP. And I've removed `monitoredResource`.

> As a result of this change, cloudtasks will also be initialized, which isn't used anywhere in flex. It looks like the library is available in the Flex environment, while Task Queues is only available in Standard. (The initialization is redundant but it is safe to do so)

I think this is not a big concern. We are just holding onto a small client that's never used. When we have more (and more expensive) clients in `shared.Clients`, we'd consider conditionally initializing them.

shared/appengine.go
Hexcles: This is on the way out (exiting the program) and we can only log to stderr, so the logging level and formatting provided by logrus are less useful here (it'd end up as plaintext in stderr either way).

KyleJu:
> LGTM % nits
> 
> I've pushed a commit directly to the branch to address some minor nits. PTAL.

Thanks for the nit commit! I will merge after double checking the staging :)

Hexcles:
SGTM

Let's squash-merge this PR.

PR #2222, opened 2020-10-24
Add support for searching all products' bug trackers

foolip:
## Description
I was doing a bit of triage and noticed search WebKit's Bugzilla isn't supported, so I added support for that and the others too.

## Review Information
This code is completely untested by me, done in the web editor. Are there tests for this code that I should update?

## Changes
Meta: I think we should remove this section from the template, what would one ever write here?

## Requirements
Nope.


webapp/components/wpt-amend-metadata.js
```diff
@@ -256,19 +256,34 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
     return link;
   }
 
-  hasHref(product) {
-    return product === 'chrome';
+  hasSearchURL(product) {
+    return product === 'chrome' || product === 'edge' || product === 'firefox' ||
```
KyleJu, line 260: Looks like we will support search for all browsers?

KyleJu: If that is the case, perhaps we can remove the `if="[[hasSearchURL(node.product)]]"`

KyleJu:
@foolip Thanks for the fix! tests can be added [here](https://github.com/web-platform-tests/wpt.fyi/blob/968fac884daab5434f166a1a18294d28a47d4926/webapp/components/test/wpt-amend-metadata.html#L130). Looks like there are minor formatting issues (caused lint to fail)

KyleJu:
@stephenmcgruer I cleaned it up and added tests. Ready for a look

stephenmcgruer (approved):
LGTM, though couldn't test as I've forgotten again how to login to deployed instances. Will check that staging seems right once it reaches there.

stephenmcgruer:
Dammit, no, I believe that broke bug searching entirely based on staging.wpt.fyi.

I'll fire up wpt.fyi locally and see if I can figure out why.

KyleJu:
Looking...

stephenmcgruer:
Hah, the test name and the product are reversed in the call to `getSearchURL`

KyleJu:
> Hah, the test name and the product are reversed in the call to `getSearchURL`

my bad. I clearly fixed it but when I did pull from origin, I think it got overwritten

PR #2226, opened 2020-10-26
[Go 1.12] Refactor the direct use of /appengine/datastore library with datastore_ae

KyleJu:
This is part of #1747 and a pure refactor without any behaviour changes.

Refactor all direct use of .../appengine/datastore library with datastore_ae. Implement the missing `NewIncompleteKey` and `PutMulti` APIs.

### DIFF
--- a/api/interop.go
+++ b/api/interop.go
@@ -11,9 +11,8 @@ import (
 	"net/http"
 
 	mapset "github.com/deckarep/golang-set"
-	"github.com/web-platform-tests/wpt.fyi/shared/metrics"
 	"github.com/web-platform-tests/wpt.fyi/shared"
-	"google.golang.org/appengine/datastore"
+	"github.com/web-platform-tests/wpt.fyi/shared/metrics"
 )
 
 // interopHandler handles the view of test results broken down by the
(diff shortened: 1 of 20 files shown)


shared/datastore_ae.go
```diff
@@ -92,6 +96,20 @@ func (d aeDatastore) Put(key Key, src interface{}) (Key, error) {
...
+func (d aeDatastore) PutMulti(keys []Key, src interface{}) ([]Key, error) {
+	cast := make([]*datastore.Key, len(keys))
+	for i := range keys {
+		cast[i] = keys[i].(*datastore.Key)
+	}
+
+	cast, err := datastore.PutMulti(d.ctx, cast, src)
+	newKeys := make([]Key, len(cast))
+	for i := range cast {
+		newKeys[i] = cast[i]
```
Hexcles, line 108: Nit: let's not reuse the variable to avoid confusion.

```suggestion
	srcKeys, err := datastore.PutMulti(d.ctx, cast, src)
	newKeys := make([]Key, len(srcKeys))
	for i := range srcKeys {
		newKeys[i] = srcKeys[i]
```

api/receiver/api_medium_test.go
```diff
@@ -194,7 +194,8 @@ func TestAddTestRun(t *testing.T) {
 	assert.Equal(t, int64(123456), key.IntID())
 
 	var testRun2 shared.TestRun
-	datastore.Get(ctx, datastore.NewKey(ctx, key.Kind(), "", key.IntID(), nil), &testRun2)
+	store := shared.NewAppEngineDatastore(ctx, false)
+	store.Get(store.NewIDKey(key.Kind(), key.IntID()), &testRun2)
```
Hexcles, line 198: Can this be
```suggestion
	store.Get(key, &testRun2)
```
?

Hexcles (approved):
LGTM % nits

api/interop.go
```diff
@@ -127,7 +126,7 @@ func loadFallbackInteropRun(ctx context.Context, filters shared.TestRunFilter) (
 	found := false
 	for {
 		_, err := it.Next(&interop)
-		if err == datastore.Done {
+		if err == store.Done().(error) {
```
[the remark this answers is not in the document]
KyleJu: Interesting. When comparing interface types and non-interface types, `They are equal if t's dynamic type is identical to X and t's dynamic value is equal to x.`

KyleJu:
Tested util/populate_dev_data.go and the local instance runs successfully. Submitting the changes

PR #2239, opened 2020-11-03
Encapsulate MultiError from datastore SDKs

Hexcles:
Return shared.MultiError from both AppEngine and Cloud
datastore.GetMulti, including unification of ErrNoSuchEntity.

Part of #1747

### DIFF
--- a/shared/datastore_ae.go
+++ b/shared/datastore_ae.go
@@ -7,6 +7,7 @@ package shared
 import (
 	"context"
 
+	"google.golang.org/appengine"
 	"google.golang.org/appengine/datastore"
 )
 
@@ -89,7 +90,19 @@ func (d aeDatastore) GetMulti(keys []Key, dst interface{}) error {
 	for i := range keys {
 		cast[i] = keys[i].(*datastore.Key)
 	}
-	return datastore.GetMulti(d.ctx, cast, dst)
+	err := datastore.GetMulti(d.ctx, cast, dst)
+	if multiError, ok := err.(appengine.MultiError); ok {
+		errors := make([]error, len(multiError))
+		for i, err := range multiError {
+			if err == datastore.ErrNoSuchEntity {
+				errors[i] = ErrNoSuchEntity
+			} else {
+				errors[i] = err
+			}
+		}
+		return NewMultiError(errors, "datastore.GetMulti")
+	}
+	return err
 }
 
 func (d aeDatastore) Put(key Key, src interface{}) (Key, error) {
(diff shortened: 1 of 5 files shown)


shared/datastore_ae.go
```diff
@@ -89,7 +90,19 @@ func (d aeDatastore) GetMulti(keys []Key, dst interface{}) error {
 	for i := range keys {
 		cast[i] = keys[i].(*datastore.Key)
 	}
-	return datastore.GetMulti(d.ctx, cast, dst)
+	err := datastore.GetMulti(d.ctx, cast, dst)
+	if multiError, ok := err.(appengine.MultiError); ok {
```
KyleJu, line 94: Optional nit: I think if we change (`datastore.ErrNoSuchEntity`)[https://github.com/web-platform-tests/wpt.fyi/blob/4ae33813438d237bade774b6426979c6e5f4dc79/shared/datastore.go#L18) to `errors.New("datastore: no such entity")` (also see [here](https://pkg.go.dev/cloud.google.com/go/datastore#pkg-variables), we can return NewMultiError(multiError, "datastore.GetMulti") right away.

KyleJu (approved):

shared/datastore_ae.go
Hexcles: To make sure I understand correctly, you are suggesting a different error message (that's the same as the original one from Datastore SDK) to avoid the for loop that replaces `ErrNoSuchEntity` with our own? My intention here is to make sure the error returned can be compared against the variable exported from `shared` regardless of which SDK is used under the hood. Two errors with the same error message aren't equal.

KyleJu: Yea that is my intention. I thought two errors (interface) will be equal if their dynamic types and values are equal, which is still [true](https://play.golang.org/), but I think error is an exception according to the [documentation](https://play.golang.org/)

Hexcles: Did you paste a wrong link to documentation?

KyleJu: Oops sorry [here](https://golang.org/pkg/errors/#New)

"New returns an error that formats as the given text. Each call to New returns a distinct error value even if the text is identical"

Hexcles: Right, it's not much of an exception but more like a design decision.

See the source code: https://golang.org/src/errors/errors.go
`errors.New` returns a pointer to a new struct, so now the "value" of the error interface is an address, which is different from another error created by `errors.New` even if the pointed structs have the same content (error string).

PR #2241, opened 2020-11-03
Migrate from AppEngine memcache to Memorystore (Redis)

Hexcles:
Part of #1747 . I followed https://cloud.google.com/appengine/docs/standard/go/using-memorystore#setup_redis_db for the setup (note that this needs to be repeated on prod before the next prod release; **update**: I've completed this).

This completely migrates us off of the legacy memcache in AppEngine to Cloud Memorystore (Redis). With this change, each instance of webapp creates a pool of Redis connections on startup for API handlers to use if it is running in the cloud; when running locally, it'd skip memcache altogether.

Tested manually (by verifying the logs and inspecting the Redis instance). Unfortunately we have to drop a test for now as it requires a local Redis, which we don't have.

### DIFF
--- a/shared/cache.go
+++ b/shared/cache.go
@@ -18,15 +18,17 @@ import (
 	"net/http"
 	"time"
 
-	"google.golang.org/appengine/memcache"
+	"github.com/gomodule/redigo/redis"
 )
 
 var (
 	errNewReadCloserExpectedString        = errors.New("NewReadCloser(arg) expected arg string")
 	errMemcacheWriteCloserWriteAfterClose = errors.New("memcacheWriteCloser: Write() after Close()")
+	errMemcacheInvalidResponseType        = errors.New("memcache: type received from GET is not []byte")
 	errByteCachedStoreExpectedByteSlice   = errors.New("contextualized byte CachedStore expected []byte output arg")
 	errDatastoreObjectStoreExpectedInt64  = errors.New("datastore ObjectStore expected int64 ID")
 	errCacheMiss                          = errors.New("cache miss")
+	errNoRedis                            = errors.New("not connected to redis")
 )
 
 // Readable is a provider interface for an io.ReadCloser.
@@ -152,9 +154,9 @@ type memcacheReadWritable struct {
 }
 
 type memcacheWriteCloser struct {
-	memcacheReadWritable
+	rw         memcacheReadWritable
+	conn       redis.Conn
 	key        string
-	expiry     time.Duration
 	b          bytes.Buffer
 	hasWritten bool
 	isClosed   bool
@@ -165,14 +167,24 @@ func (mc memcacheReadWritable) NewReadCloser(iKey interface{}) (io.ReadCloser, e
 	if !ok {
 		return nil, errNewReadCloserExpectedString
 	}
-
-	item, err := memcache.Get(mc.ctx, key)
-	if err == memcache.ErrCacheMiss {
-		return nil, errCacheMiss
-	} else if err != nil {
-		return nil, err
+	if Clients.redisPool == nil {
+		return nil, errNoRedis
 	}
-	return ioutil.NopCloser(bytes.NewReader(item.Value)), nil
+	conn := Clients.redisPool.Get()
+	defer conn.Close()
+
+	// https://redis.io/commands/get
+	result, err := conn.Do("GET", key)
+	if err != nil {
+		return nil, err
+	} else if result == nil {
+		return nil, errCacheMiss
+	}
+	b, ok := result.([]byte)
+	if !ok {
+		return nil, errMemcacheInvalidResponseType
+	}
+	return ioutil.NopCloser(bytes.NewReader(b)), nil
 }
 
 func (mc memcacheReadWritable) NewWriteCloser(iKey interface{}) (io.WriteCloser, error) {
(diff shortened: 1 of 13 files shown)


Hexcles:
FTR, there were some permission issues when deploying the service on Travis. I ended up adding two permissions to the Travis deployer service account: `vpcaccess.connectors.use` & `compute.globalOperations.get`

webapp/web/app.yaml
```diff
@@ -10,6 +10,14 @@ inbound_services:
 
 default_expiration: "1d"
 
+vpc_access_connector:
+  name: projects/wptdashboard-staging/locations/us-east4/connectors/appengine-connector
+
+env_variables:
+  REDISHOST: '10.171.142.203'
```
KyleJu, line 17: Double quota for consistency?

KyleJu: Same below

webapp/web/app.yaml
```diff
@@ -10,6 +10,14 @@ inbound_services:
 
 default_expiration: "1d"
 
+vpc_access_connector:
```
KyleJu, line 13: This is more of a personal question: a VPC network is not needed anywhere else in this project? e.g. datastore?

webapp/web/app.yaml
```diff
@@ -10,6 +10,14 @@ inbound_services:
...
+  name: projects/wptdashboard-staging/locations/us-east4/connectors/appengine-connector
+
+env_variables:
+  REDISHOST: '10.171.142.203'
+  REDISPORT: '6379'
+
+
```
KyleJu, line 20: extra space?

shared/cache.go
```diff
@@ -165,23 +167,36 @@ func (mc memcacheReadWritable) NewReadCloser(iKey interface{}) (io.ReadCloser, e
...
 func (mc memcacheReadWritable) NewWriteCloser(iKey interface{}) (io.WriteCloser, error) {
 	key, ok := iKey.(string)
 	if !ok {
 		return nil, errNewReadCloserExpectedString
 	}
-
-	return &memcacheWriteCloser{mc, key, mc.expiry, bytes.Buffer{}, false, false}, nil
+	var conn redis.Conn
+	if Clients.redisPool != nil {
+		conn = Clients.redisPool.Get()
```
KyleJu, line 197: I don't see any immediate benefit between creating `conn` here vs in `Close()` but if I have to choose, I will put it in `Close()`. It allows `conn` to be created and closed immediately

webapp/web/app.yaml
```diff
@@ -10,6 +10,14 @@ inbound_services:
 
 default_expiration: "1d"
 
+vpc_access_connector:
+  name: projects/wptdashboard-staging/locations/us-east4/connectors/appengine-connector
+
+env_variables:
+  REDISHOST: '10.171.142.203'
+  REDISPORT: '6379'
+
```
Hexcles, line 19:
```suggestion
  REDISPORT: "6379"
```

webapp/web/app.yaml
Hexcles:
```suggestion
```

webapp/web/app.yaml
Hexcles: No. It *seems* to me that calling a Google Cloud API generally doesn't require the VPC network, but connecting directly to an internal IP from Serverless (including App Engine) does.

Also there's actually a problem here. The prod project has a different name for its VPC access connector... Let me see how I can fix that

webapp/web/app.dev.yaml
```diff
@@ -12,7 +12,6 @@ inbound_services:
 - warmup
 
 default_expiration: "1d"
```
Hexcles, line 14:
```suggestion
default_expiration: "1d"

```

shared/cache.go
Hexcles: Yeah I agree. I was going to write to Redis in `Write` but decided not to eventually. Let me fix it up.

webapp/web/app.yaml
KyleJu: Yea I was gonna mention that the prod and staging share the same `env_variables` below while vpc_access_connector is different from prod and staging (both names and regions). 

Also I assume there shouldn't an issue if prod and staging share the same redis? e.g. caching to the same key name? Does it mean that they share 1GB together?

Hexcles: So the two projects are completely separate. They in fact have their own internal network. And the Redis instances just happen to share the same IP (on two different private networks) between them.

The VPC connector is less fortunate, as it has the project and location ID coded in as you said.

util/deploy.sh
```diff
@@ -37,6 +37,7 @@ done
 if [[ "${APP_PATH}" == ""  ]]; then fatal "app path not specified."; fi
 case "${APP_PATH}" in
   "webapp/web" | \
+  "webapp/web/app.staging.yaml" | \
```
KyleJu, line 40: My understanding in deployment in travis is rather vague but it looks like `APP_PATH` is changed as a result. Would it affect places like [travis-deploy-staging.sh](https://github.com/web-platform-tests/wpt.fyi/blob/master/util/travis-deploy-staging.sh#L24)?

Hexcles: Good catch! Fixed.

shared/cache.go
```diff
@@ -194,15 +202,15 @@ func (mw *memcacheWriteCloser) Write(p []byte) (n int, err error) {
 
 func (mw *memcacheWriteCloser) Close() error {
 	mw.isClosed = true
-	if !mw.hasWritten {
+	if Clients.redisPool == nil || !mw.hasWritten {
```
KyleJu, line 205: maybe return `errNoRedis` here

KyleJu (approved):
LGTM

KyleJu:
> FTR, there were some permission issues when deploying the service on Travis. I ended up adding two permissions to the Travis deployer service account: `vpcaccess.connectors.use` & `compute.globalOperations.get`

Could you briefly explain the cause of permission issue?

shared/cache.go
Hexcles: I think it'd be a bit redundant with the same error on the reader side -- I don't think we currently have anywhere that writes directly to memcache without trying to read from it first; we'd end up having two "not connected to Redis" logs every request if Redis isn't available (e.g. when running locally). WDYT?

Hexcles:
These two new permissions are required to use the VPC access connector during deployment.

shared/cache.go
KyleJu: Yea that's fine!

PR #2244, opened 2020-11-06
Tweak local testing setup in preparation for go1.12 switchover

Hexcles:
Part of #1747 

1. Enable Cloud Datastore emulator when running tests (this will unblock us from switching to Cloud Datastore SDK)
2. Attach a nil logger to local dev_appserver (this suppresses a large number of "logs will be lost" warnings when running tests)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -191,7 +191,6 @@ git: apt-get-git
 psmisc: apt-get-psmisc
 python3: apt-get-python3.7
 python: apt-get-python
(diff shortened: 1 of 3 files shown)


KyleJu:
@stephenmcgruer A bit caught up with finishing up my design doc. @stephenmcgruer would you mind reviewing this one?

stephenmcgruer (approved):

PR #2252, opened 2020-11-11
Migrate from App Engine Datastore to Cloud Datastore

Hexcles:
This is a part of #1747 .

This is a big change. Please review by commits and read the commit messages.

Only the first three commits affect production, which are relatively simple.

The remaining changes are fixing the test setup (specifically for medium and large tests which require the local dev_appserver).


KyleJu:
@stephenmcgruer I will review this one :)

shared/sharedtest/util.go
```diff
@@ -25,28 +28,104 @@ type Instance interface {
...
+
+	os.Setenv("DATASTORE_PROJECT_ID", project)
+	os.Setenv("DATASTORE_EMULATOR_HOST", i.hostPort)
+	return nil
+}
+
+func (i aeInstance) stop() error {
+	// Do not kill, terminate or interrupt the emulator process; its subprocesses will keep running.
+	postShutdown := func() {
+		res, err := http.PostForm(fmt.Sprintf("http://%s/shutdown", i.hostPort), nil)
```
KyleJu, line 99: what is this for?

shared/sharedtest/util.go
```diff
@@ -25,28 +28,104 @@ type Instance interface {
...
+		if err == nil {
+			res.Body.Close()
+		}
 	}
-	return req.WithContext(ctxWithNilLogger(req.Context())), err
+
+	stopped := make(chan error)
+	go func() {
```
KyleJu, line 106: IIUC, this code will keep running `postShutdown()` until gcd command is exited? could you add a bit explanations

shared/sharedtest/util.go
```diff
@@ -25,28 +28,104 @@ type Instance interface {
...
+	}()
+	select {
+	case <-started:
+		break
+	case <-time.After(time.Second * 10):
+		i.stop()
+		return errors.New("timed out starting Datastore emulator")
+	}
+
+	os.Setenv("DATASTORE_PROJECT_ID", project)
```
KyleJu, line 91: This seems to be a `gcloud beta emulators datastore env-init` step before this

shared/sharedtest/util.go
Hexcles: Shutting down the Datastore emulator.

The emulator starts a bunch of subprocesses and it's hard to kill them all (sending SIGINT/SIGTERM to the main process leaves the subprocesses hanging around), and then I found https://github.com/googleapis/google-cloud-go/issues/224#issuecomment-218327626

shared/sharedtest/util.go
```diff
@@ -25,28 +28,104 @@ type Instance interface {
...
+	os.Setenv("DATASTORE_EMULATOR_HOST", i.hostPort)
+	return nil
+}
+
+func (i aeInstance) stop() error {
+	// Do not kill, terminate or interrupt the emulator process; its subprocesses will keep running.
```
Hexcles, line 97:
```suggestion
	// Do not kill, terminate or interrupt the emulator process; its subprocesses will keep running.
	// https://github.com/googleapis/google-cloud-go/issues/224#issuecomment-218327626
```

shared/sharedtest/util.go
Hexcles: You're correct that these two `Setenv` are equivalent to `env-init`, but we can't use `env-init` here.

`env-init` simply prints out the shell commands to set these two env vars. And one needs to execute them in the shell, e.g. `$(gcloud beta emulators datastore env-init)`, but we are not in a shell here, and starting a new shell to execute these two commands won't affect the current process (and its future subprocesses).

shared/sharedtest/util.go
Hexcles: It will run `postShutdown`, wait for the process to exit on itself (cleanly), and if that times out (after a second) run `postShutdown` again. Did you mean adding some comments here?

shared/sharedtest/util.go
KyleJu: Thanks for the comment!

shared/sharedtest/util.go
KyleJu: Nah that's good!

KyleJu:
Let me go through the PR once again

util/populate_dev_data.go
```diff
@@ -40,12 +40,17 @@ func main() {
...
+	if *project != "" {
+		os.Setenv("DATASTORE_PROJECT_ID", *project)
+	}
+	if *datastoreHost != "" {
+		os.Setenv("DATASTORE_EMULATOR_HOST", *datastoreHost)
 	}
 
-	log.Printf("Adding dev data to host %s...", *localRemoteAPIHost)
+	ctx := context.Background()
+	shared.Clients.Init(ctx)
```
KyleJu, line 51: Missing shared.Clients.Close()

Hexcles:
```suggestion
	shared.Clients.Init(ctx)
	defer shared.Clients.Close()
```

util/populate_dev_data.go
```diff
@@ -40,12 +40,17 @@ func main() {
 	log.SetFlags(log.LstdFlags | log.Lshortfile)
 	flag.Parse()
 
-	ctx, err := getRemoteAPIContext()
-	if err != nil {
-		log.Fatal(err)
+	if *project != "" {
+		os.Setenv("DATASTORE_PROJECT_ID", *project)
```
KyleJu, line 44: where is it being used?

Hexcles: This and `DATASTORE_EMULATOR_HOST` are later used by the datastore client (in `NewClient`).

KyleJu: oh right! I got confused for a moment

KyleJu (approved):
LGTM. IIUC, as of this change, most of tests are using Cloud Datastore emulators directly, except large webdriver tests and puppeteer tests, where dev_appserver.py manages the emulator for us. You also mentioned that you will "migrate it off" later. what do you mean by that? (As in we will use the emulator directly right)

Hexcles:
Yeah, that's all correct. A follow-up PR is coming your way soon. Thanks!

Issue #2224, opened 2020-10-26
Occasional "transport is closing" error when scheduling Cloud Tasks


Hexcles:
`Failed to schedule task: rpc error: code = Unavailable desc = transport is closing`

I created a new alert for 5xx errors from `/api/results/upload`, which brought my attention to this error. It happens occasionally (a couple of times per week). It's not clear whether it was caused by #2153 (we might have missed it before due to the lack of the alert), but we should have some retry mechanism for this critical API nonetheless.

Tentatively low-prio due to the low frequency

stephenmcgruer:
What is this impact of this bug - do we drop test results?

Hexcles:
Yes, and that's bad, but with a low chance of happening, and the vast majority of API calls being PR results, the expectation for losing a full run is low.

* At least that's what I was going to say... Looking at the logs more closely, I realized I got the filtering wrong and it actually happens ~once per day, which is a lot more concerning. Continue looking...

PR #2253, opened 2020-11-12
Prevent a potential race condition when closing clients

Hexcles:
One theory for https://github.com/googleapis/google-cloud-go/issues/3205
(and https://github.com/web-platform-tests/wpt.fyi/issues/2224)
is that the instance might receive a request when it is shutting down
(and closing the clients).

To discover and/or prevent this scenario, we
* Add a log message at the beginning of shared.Clients.Close()
* Set the clients to nil before closing them (this would cause a panic
  if a client is later used)

### DIFF
--- a/shared/appengine.go
+++ b/shared/appengine.go
@@ -86,7 +86,11 @@ func (c *clientsImpl) Init(ctx context.Context) (err error) {
 // Close closes all clients in Clients. It must be called once and only once
 // before the server exits. Do not use AppEngineAPI afterwards.
 func (c *clientsImpl) Close() {
-	log.Printf("Closing clients")
+	log.Println("Closing clients")
+	// In the code below, we set clients to nil before closing them. This would
+	// cause a panic if we use a client that's being closed, which should never
+	// happen but we are not sure how exactly App Engine manages instances.
+
 	if c.cloudtasks != nil {
 		client := c.cloudtasks
 		c.cloudtasks = nil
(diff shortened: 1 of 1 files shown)


shared/appengine.go
```diff
@@ -86,27 +86,31 @@ func (c *clientsImpl) Init(ctx context.Context) (err error) {
 // Close closes all clients in Clients. It must be called once and only once
 // before the server exits. Do not use AppEngineAPI afterwards.
 func (c *clientsImpl) Close() {
+	log.Printf("Closing clients")
```
stephenmcgruer, line 89: Nit; could be `Print` or `Println` since you don't have format arguments, I think? (Both appear equivalent for `log`, which always adds a newline)

shared/appengine.go
```diff
@@ -86,27 +86,31 @@ func (c *clientsImpl) Init(ctx context.Context) (err error) {
...
 	if c.cloudtasks != nil {
-		if err := c.cloudtasks.Close(); err != nil {
+		client := c.cloudtasks
```
stephenmcgruer, line 91: May be worth a comment saying why we're doing this very weird thing. (And ditto below)

stephenmcgruer (approved):

shared/appengine.go
Hexcles:
```suggestion
	log.Println("Closing clients")
```

shared/appengine.go
Hexcles: Added a comment at the top of the function.

stephenmcgruer (approved):
Still LGTM

PR #2257, opened 2020-11-18
Completely remove service worker

Hexcles:
Service worker was first introduced in #979 to cache components as a
mitigation against our way too many unpacked JavaScript files. It was
later disabled in #1324 due to caching bugs, but an "uninstaller" was
left in place to uninstall any installed service workers.

It has been more than a year since #1324 landed and whoever had our
buggy service worker installed originally must have uninstalled it
(either via our uninstaller or by the browser) by now, so let's remove
this code.


webapp/components/wpt-flags.js
```diff
@@ -57,7 +57,6 @@ Object.defineProperty(wpt, 'ServerSideFeatures', {
       'paginationTokens',
       'pendingChecks',
       'runsByPRNumber',
-      'serviceWorker',
```
KyleJu, line 60: Probably need to delete this in Datastore on staging and prod

KyleJu (approved):
LGTM

webapp/components/wpt-flags.js
Hexcles: Yup, I'll do that!

PR #2265, opened 2020-11-19
Switch AppEngine Standard to Go 1.12

### LINKED ISSUE #1747: Migrate Go AppEngine to use cloud SDK and eventually to golang 1.12

Hexcles:
This fixes #1747 (finally... or I hope?).

Please review by commits. The "core" part of the change is really the first and the fifth commit (the "flag day" change). The three commits in between are again fixing test setup in the new world.


Hexcles:
(@KyleJu feel free to defer to Stephen)

KyleJu:
> (@KyleJu feel free to defer to Stephen)

I will review this one :). I will get back to you by Monday

webapp/web/main.go
```diff
@@ -33,5 +34,12 @@ func main() {
...
 	defer shared.Clients.Close()
-	appengine.Main()
+
+	// This behaves differently in prod and locally:
+	// * Prod: the provided app.yaml is not used; it simply starts the
+	//   DefaultServerMux on $PORT (defaults to 8080).
+	// * Local: in addition to the prod behaviour, it also starts some
+	//   static handlers according to app.yaml, which effectively replaces
+	//   dev_appserver.py.
+	nicehttp.Serve("webapp/web/app.staging.yaml", nil)
```
KyleJu, line 44: can we just not run it in prod (nested it under `if isDevAppserver()`)?

Makefile
```diff
@@ -44,7 +44,10 @@ python_test: python3 tox
 go_build: git mockgen packr2
 	make webapp_node_modules_prod
 	go generate ./...
+	# Check all packages without producing any output.
 	go build ./...
+	# Build the webapp.
+	go build ./webapp/web
```
KyleJu, line 50: why is this being run separately? it wasn't built prior to this change?

webapp/web/main.go
Hexcles: We could. That was in fact my first implementation, but later I realized that `nicehttp.Serve` does basically exactly the same thing line by line except that the condition is flipped (it has its `IsProdAppEngine`). Do you prefer the explicitness?

Makefile
Hexcles: No, when given multiple packages (e.g. `./...`), `go build` does not store the output, so we'd need to have a separate invocation. This is built for local dev (as well as large/webdriver tests). Essentially this binary is the entrypoint to our server both on prod and locally now.

webapp/web/main.go
KyleJu: Right I did notice the prod check in `nicehttp.Serve`, but why do they[start a localhost:8080](https://github.com/samthor/nicehttp/blob/0bb46c93be00919c226a1bf06908e455f31f62d2/gae.go#L161) in prod?

Hexcles: That's in line with https://cloud.google.com/appengine/docs/standard/go/go-differences#writing_a_main_package

KyleJu: I see

webdriver/webapp_server.go
```diff
@@ -126,146 +104,19 @@ func NewWebserver() (s AppServer, err error) {
...
-	return nil
-}
-
 func addStaticData(i *devAppServerInstance) (err error) {
 	cmd := exec.Command(
 		"go",
 		"run",
 		"../util/populate_dev_data.go",
-		fmt.Sprintf("--project=%s", i.project),
-		fmt.Sprintf("--datastore_host=127.0.0.1:%d", i.gcdPort),
```
KyleJu, line 261: Don't we still need this?

Hexcles: No, `sharedtest.NewAEInstance` sets the env vars which will be picked up by `datastore.NewClient`.

https://github.com/web-platform-tests/wpt.fyi/blob/4720ce1316df824a536e0cf933c7ddfc82f64f55/shared/sharedtest/util.go#L91-L92

webdriver/appserver.js
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  _awaitReady(process) {
+    return new Promise(resolve => {
+      const logDevAppserver = debug('wpt.fyi:appserver');
+      process.stderr.on('data', buffer => {
+        logDevAppserver(buffer.toString());
+      });
+
+      // TODO(Hexcles): do we still need to check if the server is up?
```
KyleJu, line 56: Reminder here in case you forget

Hexcles: Thanks. I'll leave this TODO as is for now. The server starts almost instantaneously.

KyleJu (approved):
LGTM

webdriver/webapp_server.go
KyleJu: yea you are right

Hexcles:
Thanks for reviewing this big change, Kyle!

PR #2142, opened 2020-09-09
[web-components-test] Pin Chrome to 84 to workaround hang

stephenmcgruer:
This is a hopefully temporary workaround for https://github.com/web-platform-tests/wpt.fyi/issues/2128,
to stop making unrelated PRs red and avoid missing real regressions.

PR #2266, opened 2020-11-19
Unpin Chrome but disable script streaming

Hexcles:
This still does **not** address #2142 but seems like a better workaround to me.


stephenmcgruer (approved):
Hrm, ok. But we may need to pin again if the flag is removed at some point.

Hexcles:
Long shot: now that we've had a new base Docker image, let's give it another try (without the special flag)

Hexcles:
And of course that did not work.

PR #2267, opened 2020-11-19
Unify Dockerfiles for local dev and GitHub Actions

Hexcles:
## Description

Use the same `Dockerfile` for local dev and GitHub Actions. `Dockerfile` for GitHub Actions was used as the base and tweaked a bit for local dev, while the original `Dockerfile` in the repo root was discarded.

In addition to simplifying things, this also has the added benefit of automatically building and updating our Docker image using the existing workflow on GitHub Actions (in other words, we'd no longer need to run `docker build .` locally).

## Review Information

Tested locally and on GitHub Actions: https://github.com/web-platform-tests/wpt.fyi/actions/runs/373292356


stephenmcgruer (approved):

Hexcles:
This also has the side benefit of reducing local setup time from 2~5min to <1min as you'd no longer need to build the Docker image locally.

PR #2273, opened 2020-11-24
Pull image from Docker Hub on Travis

Hexcles:
instead of building from scratch.

This should speed up our Travis builds by 2x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -78,12 +78,12 @@ jobs:
       - MAKE_TEST_TARGET=cleanup_staging_versions
 
 before_install: |
-  export DOCKER_IMAGE=wptd-dev
-  export DOCKER_INSTANCE=wptd-dev-${RANDOM}
+  export DOCKER_IMAGE="webplatformtests/wpt.fyi:latest"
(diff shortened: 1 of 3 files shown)


KyleJu (approved):

.travis.yml
```diff
@@ -78,12 +78,12 @@ jobs:
       - MAKE_TEST_TARGET=cleanup_staging_versions
 
 before_install: |
-  export DOCKER_IMAGE=wptd-dev
-  export DOCKER_INSTANCE=wptd-dev-${RANDOM}
+  export DOCKER_IMAGE="webplatformtests/wpt.fyi:latest"
```
KyleJu, line 81: what if we change the Dockerfile in the PR?

Hexcles: Right, that'd be the same as
https://github.com/web-platform-tests/wpt.fyi/blob/2e806310e2c7e3d48de85f028c1f9cc52e7d27c9/.github/actions/make-in-docker/action.yml#L9

You'd need to change both to point to the branch instead of "latest". The branch tag is built by https://github.com/web-platform-tests/wpt.fyi/blob/master/.github/workflows/docker-update.yml.

Issue #2272, opened 2020-11-24
[UI] Query builder (edit) not working when runs are populated from run_ids or max-count


KyleJu:
Query builder isn't able to delete or add the correct browsers  when runs are populated from `run_ids` or `max-count`.

Go to https://wpt.fyi/results/css/css-images?run_id=483120004&run_id=505970001&run_id=488890002 and click on edit. There are four browsers to be deleted but only three are shown in the test result. Similarly, adding a product doesn't change the test result view.

The `products` isn't updated from `testRuns` results. Also See [discussions](https://github.com/web-platform-tests/wpt.fyi/pull/2007#issuecomment-646239787).

PR #2278, opened 2020-11-25
[UI] Hide EDIT when URL params use run_id or max_count

KyleJu:
A temporary fix for #2272. This change hides EDIT button when the URL params use `run_id` or `max_count`. 

To properly fix it, we first need to figure out how a browser can be added when a user queries by `run_id` (e.g. fetching the latest run_id of that browser). I think it is better to hide it for now.

### DIFF
--- a/webapp/components/test-runs-query.js
+++ b/webapp/components/test-runs-query.js
@@ -266,7 +266,7 @@ const TestRunsQuery = (superClass, opt_queryCompute) => class extends QueryBuild
 
   computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master, runIds) {
     if (runIds && runIds.length) {
-      return 'Showing ${runIds.length} run(s) based on query parameters';
+      return `Showing ${runIds.length} run(s) based on query parameters`;
     }
     const latest = this.computeIsLatest(shas) ? 'the latest ' : '';
     const branch = master ? 'master ' : '';
--- a/webapp/components/test-runs-query.js
+++ b/webapp/components/test-runs-query.js
@@ -266,7 +266,7 @@ const TestRunsQuery = (superClass, opt_queryCompute) => class extends QueryBuild
 
   computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master, runIds) {
     if (runIds && runIds.length) {
-      return 'Showing based on query parameters';
+      return 'Showing ${runIds.length} run(s) based on query parameters';
     }
     const latest = this.computeIsLatest(shas) ? 'the latest ' : '';
     const branch = master ? 'master ' : '';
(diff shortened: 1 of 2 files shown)


Hexcles:
1. We should also change the "from the latest ... default product set" sentence, and ideally add a link to go back to the default product set (which simply drops the query param from URL)
2. We should hide the EDIT button when it's disabled.

image

KyleJu:
@Hexcles Ready for another look. I am working around <a> and use `onclick` to force a reload. I have tried setting `href="/[[page]][[path]]"` but it is being treated as a navigate without any reload. `href=""` just so that it looks like a hyperlink

webapp/views/wpt-app.js
```diff
@@ -112,13 +112,16 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
       <template is="dom-if" if="[[resultsTotalsRangeMessage]]">
         <info-banner>
           [[resultsTotalsRangeMessage]]
+          <template is="dom-if" if="[[!editable]]">
+            <a href="" onclick="[[loadDefaultProductURL]]"> (switch to the default product set instead)</a>
```
Hexcles, line 116:
```suggestion
            <a href="javascript:window.location.search='';"> (switch to the default product set instead)</a>
```

You're fighting against `app-location` here (see the [docs](https://www.webcomponents.org/element/@polymer/app-route/elements/app-location#integrating-with-other-routing-code:~:text=Integrating%20with%20other%20routing%20code)). You don't need to have an onclick listener for this though.

Hexcles: And if you take this suggestion, remember to delete `loadDefaultProductURL` as well.

webapp/components/test-runs-query.js
```diff
@@ -264,7 +264,10 @@ const TestRunsQuery = (superClass, opt_queryCompute) => class extends QueryBuild
     return batchUpdate;
   }
 
-  computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master) {
+  computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master, runIds) {
+    if (runIds && runIds.length) {
+      return 'Showing based on query parameters';
```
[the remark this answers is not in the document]
KyleJu: Do you mean ${runIds}? This suggestion will look like Showing 39166 tests from 3 based on query parameters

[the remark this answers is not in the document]
Hexcles: Sorry I missed "runs"

webapp/components/test-runs-query.js
```diff
@@ -264,7 +264,10 @@ const TestRunsQuery = (superClass, opt_queryCompute) => class extends QueryBuild
     return batchUpdate;
   }
 
-  computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master) {
+  computeResultsRangeMessage(shas, productSpecs, from, to, maxCount, labels, master, runIds) {
+    if (runIds && runIds.length) {
+      return 'Showing ${runIds.length} run(s) based on query parameters';
```
KyleJu, line 269:
```suggestion
      return `Showing ${runIds.length} run(s) based on query parameters`;
```

Hexcles (approved):
LGTM

PR #2283, opened 2020-12-01
Fix name of query parameter in /api/search payload

Hexcles:
Just "q" has no effect, and is not what the wpt.fyi frontend uses.


Hexcles:
Sorry I didn't notice you used the wrong base branch :) @foolip

KyleJu (approved):

foolip:
Should we just delete the master branch? I think I created the PR from the web interface starting from a URL with "master" in it, which should redirect if the branch is deleted.

Hexcles:
@foolip yup I've just deleted the master branch, woohoo!

PR #2285, opened 2020-12-02
Document run_ids as non-optional for /api/search

foolip:
https://github.com/web-platform-tests/wpt.fyi/blob/f8d9beebe98a4d254d1ae0d0283c4fe7cef3196e/api/query/atoms.go#L437

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -430,11 +430,10 @@ Search for test results over some set of test runs.
 
 __Parameters__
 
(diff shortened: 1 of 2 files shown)


foolip:
I think was supposed to work at some point, but as far as I can tell it doesn't. The frontend still first calls /api/runs and then /api/search.

Hexcles (approved):
I agree we should fix the doc rather than trying to make it work at this point.

We can file a separate issue on API ergonomics.

PR #2293, opened 2020-12-04
Update api/search GET/POST methods in doc

KyleJu:
Update api/search GET/POST methods in doc


Hexcles:
Oh thank you! I finally realized what I missed a few days ago. Let me revert some recent doc changes first.

foolip:
Oh wow, so the GET variant is the useful one that I didn't know existed. Why doesn't the wpt.fyi front end use this, given that the actual runs are also returned?

Hexcles:
@foolip it does **not** support structured queries. `q` simply filters by *test names*.

foolip:
Oh, yeah that's not what I want at all :)

Hexcles (approved):

KyleJu:
@Hexcles Thanks Robert!

PR #2295, opened 2020-12-04
Cleanup helpers and timeouts WebDriver tests

Hexcles:
The main motivation is to extend the timeout everywhere as we started to see timeouts more often since the Go 1.12 migration (which is presumably caused by the slower Datastore emulator), but I took the chance to do some other cleanup as well.

### DIFF
--- a/webdriver/webdriver.go
+++ b/webdriver/webdriver.go
@@ -14,6 +14,7 @@ import (
 	"github.com/tebeka/selenium"
 )
 
+// Flags
 var (
 	debug            = flag.Bool("debug", false, "Turn on debug logging")
 	browser          = flag.String("browser", "firefox", "Which browser to run the tests with")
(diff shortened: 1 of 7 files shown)


webdriver/webdriver.go
```diff
@@ -9,17 +9,26 @@ import (
...
 	debug            = flag.Bool("debug", false, "Turn on debug logging")
 	browser          = flag.String("browser", "firefox", "Which browser to run the tests with")
 	startFrameBuffer = flag.Bool("frame_buffer", frameBufferDefault(), "Whether to use a frame buffer")
 )
 
+const (
+	// LongTimeout is the timeout for waiting the full page to load, with
+	// data coming from Datastore. You may not need this if you only need
+	// to wait for the initial Polymer rendering.
+	LongTimeout = time.Second * 30
```
KyleJu, line 29:
```suggestion
	longTimeout = time.Second * 30
```
A local constant

KyleJu: And change the case in the test if you decide to make this change

KyleJu (approved):
LGTM with a comment

webdriver/webdriver.go
Hexcles: Although this is only used within this package (i.e. technically no need to be exported), it's used widely in multiple files and there's a convention in this package to export this kind of helpers (see the other helpers in this package).

The main reason for this change was mentioned in the PR description: "to extend the timeout everywhere as we started to see timeouts more often since the Go 1.12 migration (which is presumably caused by the slower Datastore emulator)".

PR #2312, opened 2020-12-15
[BSF] Backend for BSF on wpt.fyi milestone one

KyleJu:
The backend for [BSF on wpt.fyi](https://docs.google.com/document/d/1YsWyovoje7rvW_u-kaCJAWvFJ2QjafvyVqWKtmafFHw/edit#) milestone one.

- Rather than POST requests, /api/bsf uses GET requests for milestone one, without request payload


shared/fetch_bsf.go
```diff
@@ -0,0 +1,130 @@
...
+		return BSFData{}
+	}
+
+	// The lateset revision should be the last row at the 0th index.
+	response.LastUpdateRevision = data[len(data)-1][0]
+	response.Data = data
+	return response
+}
+
+// FetchBSF encapsulates the Fetch() method for testing.
```
KyleJu, line 93:
```suggestion
// FetchBSF encapsulates the Fetch(isExperimental bool) method for testing.
```

api/README.md
```diff
@@ -653,3 +654,62 @@ This endpoint returns the URL of a PR that is created in the wpt-metadata repo.
...
+__`to`__ : (Optional) RFC3339 timestamp, for which to include BSF data that occured before the given time exclusively.
+
+__`experimental`__ : A boolean to return BSF data for experimental or stable runs. Defaults to false.
+
+__JSON Response__
+
+`fields` corresponds to the fields (columns) in the `data` table.
+
+`data` returns BSF data in chronological order.
+```json
```
Hexcles, line 679: Wrap this in `<details>`?

api/README.md
```diff
@@ -653,3 +654,62 @@ This endpoint returns the URL of a PR that is created in the wpt-metadata repo.
...
+__Parameters__
+
+__`from`__ : (Optional) RFC3339 timestamp, for which to include BSF data that occured after the given time inclusively.
+
+__`to`__ : (Optional) RFC3339 timestamp, for which to include BSF data that occured before the given time exclusively.
+
+__`experimental`__ : A boolean to return BSF data for experimental or stable runs. Defaults to false.
+
+__JSON Response__
+
```
Hexcles, line 675: Add a summary "the response has three top-level fields:" and make a list. Also mention `lastUpdateRevision`.

api/bsf_handler.go
```diff
@@ -0,0 +1,60 @@
...
+func apiBSFHandler(w http.ResponseWriter, r *http.Request) {
+	handleBSF(w, r, shared.NewFetchBSF())
+}
+
+func handleBSF(w http.ResponseWriter, r *http.Request, fetcher shared.FetchBSF) {
+	q := r.URL.Query()
+	isExperimental := false
+	val, _ := shared.ParseBooleanParam(q, "experimental")
+	// If the experimental parameter is missing or present with no value, set
+	// isExperimental to the default value, false.
```
Hexcles, line 25: This comment is slightly confusing as it's not what the code actually does (though with the same effect). Maybe just drop it.

api/bsf_handler.go
```diff
@@ -0,0 +1,60 @@
...
+	if val != nil {
+		isExperimental = *val
+	}
+
+	lines, err := fetcher.Fetch(isExperimental)
```
Hexcles, line 30: Maybe move this after we parse the timestamps to avoid wasting resource on bad requests.

api/bsf_handler.go
```diff
@@ -0,0 +1,60 @@
...
+	bsfData := shared.FilterandExtractBSFData(lines, from, to)
+	marshalled, err := json.Marshal(bsfData)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	_, err = w.Write(marshalled)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
```
Hexcles, line 57: It is indeed a good practice to check all errors but this one is a bit unnecessary as we can't actually change the status code of our response after we've started writing the body (`200 OK` will have already been sent to the client by default). There's really little we can do if we fail to write (most likely due to connection errors) so we usually just ignore this error.

shared/fetch_bsf.go
```diff
@@ -0,0 +1,130 @@
...
+	"time"
+)
+
+const (
+	// experimentalBSFURL is the GitHub URL for fetching the experimental BSF data
+	// for Chrome, Firefox and Safari.
+	experimentalBSFURL = "https://raw.githubusercontent.com/Ecosystem-Infra/wpt-results-analysis/gh-pages/data/experimental-browser-specific-failures.csv"
+	// stableBSFURL is the GitHub URL for fetching the stable BSF data
+	// for Chrome, Firefox and Safari.
+	stableBSFURL = "https://raw.githubusercontent.com/Ecosystem-Infra/wpt-results-analysis/gh-pages/data/stable-browser-specific-failures.csv"
```
Hexcles, line 22: Is there a rate limit for `raw.githubusercontent.com`? Shall we consider caching the responses?

api/bsf_handler.go
KyleJu: Good point

shared/fetch_bsf.go
KyleJu: I think it falls under user-to-server requests for GitHub APIs, which should be 5000 calls/hour. But agreed, caching at the request level should be good for now

api/bsf_handler.go
```diff
@@ -12,26 +12,27 @@ import (
...
+// BSFHandler is an http.Handler for the /api/bsf endpoint.
+type BSFHandler struct {
+	fetcher shared.FetchBSF
+}
+
 // apiBSFHandler fetches browser-specific failure data based on the URL params.
 func apiBSFHandler(w http.ResponseWriter, r *http.Request) {
-	handleBSF(w, r, shared.NewFetchBSF())
+	ctx := r.Context()
+	// Serve cached with 5 minute expiry. Delegate to BSFHandler on cache miss.
```
Hexcles, line 23: Given the frequency of the cron job (weekly), I think it's safe to set the expiry to 1 hour.

api/bsf_handler.go
```diff
@@ -12,26 +12,27 @@ import (
...
+	shared.NewCachingHandler(
+		r.Context(),
```
Hexcles, line 25: Use `ctx` instead.

Hexcles (approved):
LGTM % nits

Issue #2287, opened 2020-12-02
De-emphasize "Interoperability" on the front-end


Hexcles:
Currently there is quite a lot of clutter above the fold of wpt.fyi:
![image](https://user-images.githubusercontent.com/1436812/100924302-6ea20d80-34ae-11eb-9af3-8a42d419fca2.png)
Among which the tab bar is very rarely used despite its footprint. Very few people look at the interop tab.

In the long term we might want to surface the interop metrics in a different place (e.g. some kind of team dashboard, along with browser-specific failures), or get rid of it altogether, but before that, we can de-emphasize it on the UI, e.g. move "interop" into the (much smaller) top nav bar and remove the tab switcher.

PR #2325, opened 2020-12-29
[UI] Move the interop tab to headers

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2287

### DIFF
--- a/webapp/components/wpt-header.js
+++ b/webapp/components/wpt-header.js
@@ -75,9 +75,9 @@ class WPTHeader extends WPTFlags(PolymerElement) {
       <nav>
         <!-- TODO: handle onclick with wpt-results.navigate if available -->
         <a href="/">Latest Run</a>
-        <a href="/interop[[path]]?[[query]]">Interoperability</a>
         <a href="/runs">Recent Runs</a>
         <a href="/insights">Insights</a>
+        <a href="/interop[[path]]?[[query]]">Interoperability</a>
         <template is="dom-if" if="[[processorTab]]">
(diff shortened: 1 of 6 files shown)


KyleJu:
@stephenmcgruer - I also experimented with lazy loading the interop view but encountered WebDriver tests timeout issues. With lazy loading, I saw a slight speedup on the loading (0.5~ second)

webapp/components/wpt-header.js
```diff
@@ -75,6 +75,7 @@ class WPTHeader extends WPTFlags(PolymerElement) {
       <nav>
         <!-- TODO: handle onclick with wpt-results.navigate if available -->
         <a href="/">Latest Run</a>
+        <a href="/interop[[path]]?[[query]]">Interoperability</a>
```
stephenmcgruer, line 78: Let's move this to between 'Insights' and 'Processor'.

(Ultimately I think we'll want to shuffle things around, but for now this seems a bit more coherent than Latest / Interop / Recent / Insights / Processor. Plus we know people rarely use this view anyway.)

stephenmcgruer (approved):
This is awesome! It looks much cleaner, and it sort of gives me the idea that we could one day use the 'interoperability' page for things like BSFs/etc... (and then make it the default page, and then make 'latest results' not really a thing, and... ahem, getting ahead of myself)

KyleJu:
`suites_medium_test.go` encountered a timeout erro `timed out starting Datastore emulator`. Retrying checks

PR #2327, opened 2020-12-31
[BSF] Set up Google Analytics for hover and hide actions

KyleJu:
Set up Google Analytics for hover and hide actions:

- `hover` tracks the duration spent on the general area of the BSF graph, including the banner
- `hide` tracks the number of times users hide/open the BSF graph

For the setup, see https://developers.google.com/analytics/devguides/collection/analyticsjs/events.

Also work around a [known issue](https://stackoverflow.com/questions/17244996/why-do-the-mouseenter-mouseleave-events-fire-when-entering-leaving-child-element) where the children svg (in this case Google Chart) interferes with parent's `onmouseenter` event. Use `isInteracting` to keep track of the state of interaction with the BSF chart.

Note that unlike the origin design, we can track clicks on stable/experiment through the /api/bsf endpoint usage and zoom-in/zoom-out are covered by `hover` actions.

## Test
See [Google Analytics](https://analytics.google.com/analytics/web/#/report/content-event-overview/a89387706w132586808p136568610/_u.date00=20201229&_u.date01=20201230/)

### DIFF
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -233,6 +233,17 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
     };
     this.handleCollapse = () => {
       this.isBSFCollapsed = !this.isBSFCollapsed;
+      // Record hide/open actions on the BSF graph. Currently, we only
+      // show it on the homepage.
+      if ('ga' in window) {
+        window.ga('send', {
+          hitType: 'event',
+          eventCategory: 'bsf',
+          eventAction: 'hide',
+          eventLabel: this.path,
+          eventValue: this.isBSFCollapsed ? 1 : -1
+        });
+      }
       this.setLocalStorageFlag(this.isBSFCollapsed, 'isBSFCollapsed');
     };
     this.submitQuery = this.handleSubmitQuery.bind(this);
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -246,7 +246,7 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
         window.ga('send', {
           hitType: 'event',
           eventCategory: 'bsf',
-          eventAction: 'hide',
+          eventAction: 'visibility change',
           eventLabel: this.path,
           eventValue: this.isBSFCollapsed ? 1 : 0
         });
@@ -254,12 +254,18 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
       this.setLocalStorageFlag(this.isBSFCollapsed, 'isBSFCollapsed');
     };
     this.enterBSF = () => {
+      // The use of isInteracting is a workaround for a known issue,
+      // https://stackoverflow.com/questions/17244996/why-do-the-mouseenter-mouseleave-events-fire-when-entering-leaving-child-element;
+      // when users interact with the BSF chart itself, enterBSF is triggered unexpectedly.
+      // In that case, isInteracting is set to true to avoid resetting bsfStartTime.
       if (this.isInteracting) {
         return;
       }
       this.bsfStartTime = new Date();
     };
     this.exitBSF = () => {
+      // Similarly, when users interact with the BSF chart, isInteracting is set to
+      // true to avoid sending analytics prematurely in exitBSF.
       if (this.isInteracting || !this.bsfStartTime) {
         return;
       }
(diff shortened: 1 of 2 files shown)


KyleJu:
@stephenmcgruer my workaround using `isInteracting` makes the interaction with google chart somewhat janky. I am not able to click the BSF chart once in a while. Let me know if you encounter the same experience

KyleJu:
> @stephenmcgruer my workaround using `isInteracting` makes the interaction with google chart somewhat janky. I am not able to click the BSF chart once in a while. Let me know if you encounter the same experience

Actually I can reproduce it on wpt.fyi staging.  It seems to be an issue If you click really fast on the chart

webapp/views/wpt-app.js
```diff
@@ -235,8 +242,45 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
...
+          hitType: 'event',
+          eventCategory: 'bsf',
+          eventAction: 'hide',
+          eventLabel: this.path,
+          eventValue: this.isBSFCollapsed ? 1 : 0
+        });
+      }
       this.setLocalStorageFlag(this.isBSFCollapsed, 'isBSFCollapsed');
     };
+    this.enterBSF = () => {
```
stephenmcgruer, line 258: What exactly are we trying to measure here? It seems like we are measuring just the time spent in the general BSF area, not the chart specifically, and as such I don't get the purpose of the `isInteracting` variable.

webapp/views/wpt-app.js
```diff
@@ -235,8 +242,45 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
...
+      if (this.isInteracting) {
+        return;
+      }
+      this.bsfStartTime = new Date();
+    };
+    this.exitBSF = () => {
+      if (this.isInteracting || !this.bsfStartTime) {
```
stephenmcgruer, line 265: `this.bsfStartTime` is never cleared, so after the first call to `exitBSF` it will always be non-null and won't trigger this early return (which I presume is meant to guard against `exitBSF` firing with an equivalent `enterBSF`?)

stephenmcgruer (changes requested):

webapp/views/wpt-app.js
```diff
@@ -235,8 +242,45 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
     };
     this.handleCollapse = () => {
       this.isBSFCollapsed = !this.isBSFCollapsed;
+      // Record hide/open actions on the BSF graph. Currently, we only
+      // show it on the homepage.
+      if ('ga' in window) {
+        window.ga('send', {
+          hitType: 'event',
+          eventCategory: 'bsf',
+          eventAction: 'hide',
```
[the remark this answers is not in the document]
KyleJu: I try to represent eventValue: 1 as hide and 0 as show. How about eventAction: ` hide/show`?

webapp/views/wpt-app.js
KyleJu: Correct, I want to measure the time spent in the general area because it would count the scenario when users look at the BSF graph without any interaction or only switch between stable and experimental.

`isInteracting` is a workaround for [this issue](https://stackoverflow.com/questions/17244996/why-do-the-mouseenter-mouseleave-events-fire-when-entering-leaving-child-element). where users interact with Google Chart, `enterBSF ` is triggered. I use `isInteracting` to avoid resetting `bsfStartTime`

webapp/views/wpt-app.js
KyleJu: Yea it is somewhat redundant. I put it here just in case weird things happen. Happy to remove it if it confuses you

webapp/views/wpt-app.js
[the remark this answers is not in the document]
stephenmcgruer: Sure, or `visibility change` or something like that :)

webapp/views/wpt-app.js
stephenmcgruer: Oh this is super interesting! I was originally going to say I couldn't reproduce this, but then found that if I mouse quickly over the actual lines on the graph (maybe hitting the legend popups?) I can cause multiple `enterBSF`s to trigger - though without an `exitBSF` in the middle! E.g.

image

So I guess this LGTM, but please document `isInteracting`, it's very unclear why it exists!

webapp/views/wpt-app.js
stephenmcgruer: This is fine, but we should probably clear `bsfStartTime` in `exitBSF` after using it then.

KyleJu:
@stephenmcgruer ready for another look :)

stephenmcgruer (approved):

PR #2331, opened 2021-01-05
Unify metadata cache

KyleJu:
Use redis in flex, https://cloud.google.com/memorystore/docs/redis/connect-redis-instance-flex


KyleJu:
Judging from the Pathenon log, the cache is pulled from Redis correctly. However the staging instance doesn't work well, as it keeps returning 502; it might be a deployment issue

PR #2334, opened 2021-01-06
[Triage Metadata] Unskip metadataMap from seenProductURLs

### LINKED ISSUE #2333: [Triage Metadata] Triaged subtests are not shown
In https://github.com/web-platform-tests/wpt-metadata/pull/735, the newly-triaged subtest `'[foo=''bar'' s] <div foo="bar"> with querySelector in standards mode'` already exists but [wpt.fyi](https://wpt.fyi/results/css/selectors/attribute-selectors/attribute-case/semantics.html?label=experimental&label=master&aligned) failed to show any icons. In fact, none of subtests are shown in [css/selectors/attribute-selectors/attribute-case/META.yml](https://github.com/web-platform-tests/wpt-metadata/blob/master/css/selectors/attribute-selectors/attribute-case/META.yml)

KyleJu:
Fix  #2333. Don't skip `metadataMap` from `seenProductURLs` . 

`metadataMap` stores metadata for all triaged tests and is used to show inline icons. `displayedMetadata` displays the metadata on the bottom of test results view and only show metadata on the test-file level. 

## Test
See [here](https://fix-subtest-missing-dot-wptdashboard-staging.uk.r.appspot.com/results/css/selectors/attribute-selectors/attribute-case/semantics.html?label=experimental&label=master&aligned)

### DIFF
--- a/webapp/components/wpt-metadata.js
+++ b/webapp/components/wpt-metadata.js
@@ -203,11 +203,9 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
     for (const test of Object.keys(metadata).filter(k => this.shouldShowMetadata(k, path, testResultSet))) {
       const seenProductURLs = new Set();
       for (const link of metadata[test]) {
-        const serializedProductURL = link.product.trim() + '_' + link.url.trim();
-        if (link.url === '' || seenProductURLs.has(serializedProductURL)) {
+        if (link.url === '') {
           continue;
         }
-        seenProductURLs.add(serializedProductURL);
         const urlHref = this.getUrlHref(link.url);
         const subtestMap = {};
         if ('results' in link) {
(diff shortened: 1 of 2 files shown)


stephenmcgruer (commented):
Can we add a unittest for this, that would have caught the original bug?

webapp/components/wpt-metadata.js
```diff
@@ -203,11 +203,9 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
     for (const test of Object.keys(metadata).filter(k => this.shouldShowMetadata(k, path, testResultSet))) {
```
stephenmcgruer, line 203: It would be useful to document this method, either just before the loop or adding method-level documentation. Something like:

```
    // This loop constructs both the metadataMap, which is used to show inline
    // bug icons in the test results, and displayedMetdata, which is the list of
    // metadata links shown at the bottom of the page.
    let metadataMap = {};
    ....
```

webapp/components/wpt-metadata.js
```diff
@@ -229,6 +227,11 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
         } else {
           metadataMap[metadataMapKey] = Object.assign(metadataMap[metadataMapKey], subtestMap);
         }
+        const serializedProductURL = link.product.trim() + '_' + link.url.trim();
```
stephenmcgruer, line 230: Again, worthwhile explaining why we only check this now.

```
    // Avoid showing duplicate bug links in the list of metadata shown at the bottom of the page.
    const serializedProductURL ...
```

stephenmcgruer (approved):

PR #2335, opened 2021-01-07
[UI] Remove iron-pages and stop loading interop data

KyleJu:
`iron-pages` loads `ready()` in `wpt-interop` and `wpt-results` by default.  Remove loadData() in the `ready()` of  interop to improve the results page load time

### DIFF
--- a/webapp/views/wpt-interop.js
+++ b/webapp/views/wpt-interop.js
@@ -32,7 +32,7 @@ const interopQueryCompute =
   'interopQueryParams(shas, aligned, master, labels, productSpecs, to, from, maxCount, offset, search)';
 
 class WPTInterop extends WPTColors(WPTFlags(LoadingState(PathInfo(
-  TestRunsQueryLoader(TestRunsUIQuery(PolymerElement, interopQueryCompute)))))) {
+    TestRunsQueryLoader(TestRunsUIQuery(PolymerElement, interopQueryCompute)))))) {
   static get template() {
     return html`
   <style>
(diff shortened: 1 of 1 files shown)


KyleJu:
@stephenmcgruer would you know where the timeout is set? I am getting `  Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.` for [this interop test](https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/components/test/wpt-interop.html#L80). I failed to see where it sets the 10000ms

stephenmcgruer:
> @stephenmcgruer would you know where the timeout is set? I am getting Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. for this interop test. I failed to see where it sets the 10000ms

Summarizing our discussion over VC:

1. This is because the test is now broken; the logic change in this PR makes calling `queryChanged` mandatory and the test doesn't do that (it relies on `ready()`).
2. Requiring `queryChanged` is ok but a bit messy; the better path would be to refactor wpt-app.js to not load both the subpages when only one of them is now used at a time.
3. The performance benefit isn't huge (and somewhat unstable), so this is possibly not worth doing.

KyleJu:
> > @stephenmcgruer would you know where the timeout is set? I am getting Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. for this interop test. I failed to see where it sets the 10000ms
> 
> Summarizing our discussion over VC:
> 
> 1. This is because the test is now broken; the logic change in this PR makes calling `queryChanged` mandatory and the test doesn't do that (it relies on `ready()`).
> 2. Requiring `queryChanged` is ok but a bit messy; the better path would be to refactor wpt-app.js to not load both the subpages when only one of them is now used at a time.
> 3. The performance benefit isn't huge (and somewhat unstable), so this is possibly not worth doing.

Just to chime in on the point 3, the interop data is also likely cached in prod.

PR #2347, opened 2021-01-15
[Triage Metadata] wrong order of getSearchURL params

KyleJu:
A follow-up fix for https://github.com/web-platform-tests/wpt.fyi/pull/2222#issuecomment-761100542 as it is breaking the staging instance


stephenmcgruer (approved):

KyleJu:
Screen Shot 2021-01-15 at 1 40 29 PM

Tested on staging and it should be good now

PR #2352, opened 2021-01-20
[taskcluster] Fetch only 30 check-runs per page from GitHub

stephenmcgruer:
In https://github.com/web-platform-tests/wpt/issues/27243, we are seeing
500 errors from GitHub when we try to request 100 check-runs per page.
As a workaround whilst they investigate, GitHub have suggested
requesting fewer check-runs at a time.

### DIFF
--- a/api/taskcluster/webhook.go
+++ b/api/taskcluster/webhook.go
@@ -439,15 +439,18 @@ func (api apiImpl) ListCheckRuns(owner string, repo string, checkSuiteID int64) 
 	var runs []*github.CheckRun
 	options := github.ListCheckRunsOptions{
 		ListOptions: github.ListOptions{
-			// 100 is the maximum allowed items per page; see
-			// https://developer.github.com/v3/guides/traversing-with-pagination/#changing-the-number-of-items-received
-			PerPage: 100,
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
(I'm a little worried about a 4x increase in API load here, but we can monitor for such errors as well and revert/increase the page size if needed.)

LukeZielinski (approved):
Maybe a newb question - bit this should hit staging shortly after landing? So we can check for the effect there before releasing on the live version?

stephenmcgruer:
> Maybe a newb question - bit this should hit staging shortly after landing? So we can check for the effect there before releasing on the live version?

Not a newb question at all! Yes, staging is automatically deployed from HEAD on a fairly quick basis. You can check the version at the bottom of https://staging.wpt.fyi/about

Looking at https://staging.wpt.fyi/runs?label=master&label=stable&from=2021-01-13T00%3A00&product=chrome&product=firefox&product=safari it shows the same problem as prod, so yeah we should be able to monitor for that improving.

stephenmcgruer:
Note that `api/checks/runs.go` appears to also have a similar query to this one, so we may run into problems there too, but I wanted to start with the place where we definitely know we're struggling and see if we can get past that.

PR #2366, opened 2021-01-28
[results-processor] Make conflicts easier to debug

stephenmcgruer:
Currently conflicts reported by the processor report the field that
conflicted but not the values that it saw, e.g.:

"Conflicting 'browser_version' found in the merged report"

This change adds the values, to better help debugging what conflict
occurred. Note that we bail on the first report that fails, so it is
possible that we could have (e.g.) 5 reports with 5 different
'browser_version's, and we would only report the first two after
this change.

### DIFF
--- a/results-processor/wptreport_test.py
+++ b/results-processor/wptreport_test.py
@@ -168,8 +168,8 @@ class WPTReportTest(unittest.TestCase):
                 },
             }, f)
         with open(tmp_path, 'rb') as f:
-            with self.assertRaisesRegex(ConflictingDataError,
-                                        "product, browser_version"):
+            reg = r"product: \[chrome, firefox\], browser_version: \[70, 59\]"
+            with self.assertRaisesRegex(ConflictingDataError, reg):
                 r.load_json(f)
 
         # Fields without conflict should be preserved.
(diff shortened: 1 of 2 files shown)


results-processor/wptreport_test.py
```diff
@@ -168,8 +168,8 @@ def test_load_json_multiple_chunks_conflicting_data(self):
                 },
             }, f)
         with open(tmp_path, 'rb') as f:
-            with self.assertRaisesRegex(ConflictingDataError,
-                                        "product, browser_version"):
+            reg = r"product: \[chrome, firefox\], browser_version: \[70, 59\]"
```
KyleJu, line 171: Could you explain what prompts this change and why `browser_version`: is `[70, 59]`? Wouldn't browser versions change over time?

stephenmcgruer: > what prompts this change

I debugged a missing run case the other day, which just turned out to be a minor update of Safari Stable on Azure Pipelines having only reached some machines (2/5 of the machines were using Safari 14.0, the other 3 were using Safari 14.0.2). However to discover that it was a minor update I had to go to Azure Pipelines, download the report JSON, and grep through the files myself - all for information that wpt.fyi already knows

> why browser_version: is [70, 59]?

Browser version being `[70, 59]` is just what the test is using as the conflicting values. (And was already using).

> Wouldn't browser versions change over time?

This conflict detection is within one report. Browser versions should not change across shards of one report, otherwise the data from the run is junk (some tests running with version X, some with version Y...).

KyleJu (approved):

PR #2386, opened 2021-02-05
[BSF] BSF graph captions

KyleJu:


webapp/components/wpt-bsf.js
```diff
@@ -61,6 +61,7 @@ class WPTBSF extends LoadingState(PolymerElement) {
       </style>
       <div class="bsf">
         <div class="left">
+          <h5>Browser-specific failures are the number of WPT tests which fail in exactly one browser. This graph shows the BSF scores for Chrome, Firefox and Safari over the last two years.</h5>
```
stephenmcgruer, line 64: I don't think `over the last two years` is particularly useful information. Let's drop that last part (and just end on "This graph shows the BSF scores for Chrome, Firefox and Safari.").

stephenmcgruer (approved):
LGTM % comment.

I should probably do a fuller write-up at some point and then we can link to that, but for now this is an improvement I think.

webapp/components/wpt-bsf.js
KyleJu, line 64:
```suggestion
          <h5>Browser-specific failures are the number of WPT tests which fail in exactly one browser. This graph shows the BSF scores for Chrome, Firefox and Safari.</h5>
```

PR #2433, opened 2021-03-04
Remove obsolete 'experimentalAlignedExceptEdge' admin flag

stephenmcgruer:
This flag was used to fetch edge runs back when they were not tagged
with the experimental label. It is no longer useful, and is also
misleading (the name implies it might remove Edge from the default
query, which it does not do). As such, remove it.

This also makes util/populate_dev_data.go closer to what is shipped on
wpt.fyi

See https://github.com/web-platform-tests/wpt.fyi/pull/2431#issuecomment-790748838
and the following comment


KyleJu (approved):

stephenmcgruer:
(I believe this is a real failure; can reproduce locally and it goes away when I revert. Will dig in.)

stephenmcgruer:
@KyleJu I've made non-trivial changes, please re-review. Thanks!

KyleJu:
LGTM. Thanks Stephen!

Issue #2421, opened 2021-02-24
[Triage] Support searching for test-level triage


stephenmcgruer:
https://github.com/web-platform-tests/wpt.fyi/pull/2407 added support for test-level triage, not connected to a specific product. This currently cannot be searched for; the metadata has no `product` and so our search atom of `triaged:[product]` doesn't make any sense.

We should add support for something like `triaged:test-level` or `triaged:test-issue` to allow searching for these cases.

PR #2440, opened 2021-03-07
[triage] Add support for searching for test-level triage

stephenmcgruer:
This PR both adds the ability to search for test-level metadata (that is, metadata
that has no product associated with it), and stops test-level metadata showing up
for specific product searches.

To achieve this, we:

1. Expand the `triagedExp` atom on the front-end to also accept `test-issue` and
    to convert it to an empty product.
2. Change `shared.GetMetadataResponse` to invert its handling of empty-product
    metadata from matching any product to matching no products.

This does mean that for a request like: `"query":{"exists":[{"triaged":"test-issue"}]}` we
create N identical `query.Triaged` objects, one for each test-run in the query. This is
inefficient (because `triaged:test-issue` should not be bound to multiple runs), but is an
artifact of how `AbstractExists` works and is identical to the behavior for `link` today.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/2421

To test: open deployed instance, try searching for `triaged:test-issue` (should get a
handful of results), and for `triaged:chrome` (should get slightly less results than on
wpt.fyi). You can also try hitting `/api/metadata?product=chrome` on the deployed
instance and note that it no longer has entries for the test-level triaged tests.

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -250,14 +250,16 @@ func (l AbstractLessThan) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return LessThan{c}
 }
 
-// AbstractLink is represents the root of a link query, which matches Metadata URLs
-// to a pattern string; it is independent of test runs.
+// AbstractLink is represents the root of a link query, which matches Metadata
+// URLs to a pattern string.
 type AbstractLink struct {
 	Pattern         string
 	metadataFetcher shared.MetadataFetcher
 }
 
-// BindToRuns for AbstractLink is a no-op; it is independent of test runs
+// BindToRuns for AbstractLink fetches metadata for either test-level issues or
+// issues associated with the given runs. It does not filter the metadata by
+// the pattern yet.
 func (l AbstractLink) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	if l.metadataFetcher == nil {
 		l.metadataFetcher = searchcacheMetadataFetcher{}
@@ -279,8 +281,8 @@ type AbstractTriaged struct {
 	metadataFetcher shared.MetadataFetcher
 }
 
-// BindToRuns for AbstractTriaged binds each run of the AbstractTriaged ProductSpec
-// to a triaged object.
+// BindToRuns for AbstractTriaged binds each run matching the AbstractTriaged
+// ProductSpec to a triaged object.
 func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	cq := make([]ConcreteQuery, 0)
 
@@ -289,16 +291,21 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	}
 	for _, run := range runs {
 		if t.Product == nil || t.Product.Matches(run) {
+			// We only want to fetch metadata for this specific run (or for no runs, if
+			// the search is for test-level issues).
 			includeTestLevel := false
-			metadataRuns := runs
+			metadataRuns := []shared.TestRun{run}
+
 			// Product being nil means that we want test-level issues.
 			if (t.Product == nil) {
 				includeTestLevel = true
-				metadataRuns = make([]shared.TestRun, 0)
+				metadataRuns = []shared.TestRun{}
 			}
 			metadata, _ := shared.GetMetadataResponse(metadataRuns, includeTestLevel, logrus.StandardLogger(), t.metadataFetcher)
 			metadataMap := shared.PrepareLinkFilter(metadata)
-			cq = append(cq, Triaged{run.ID, metadataMap})
+			if len(metadataMap) > 0 {
+				cq = append(cq, Triaged{run.ID, metadataMap})
+			}
 		}
 	}
 
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -297,7 +297,7 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 			metadataRuns := []shared.TestRun{run}
 
 			// Product being nil means that we want test-level issues.
-			if (t.Product == nil) {
+			if t.Product == nil {
 				includeTestLevel = true
 				metadataRuns = []shared.TestRun{}
 			}
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -262,7 +262,8 @@ func (l AbstractLink) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	if l.metadataFetcher == nil {
 		l.metadataFetcher = searchcacheMetadataFetcher{}
 	}
-	metadata, _ := shared.GetMetadataResponse(runs, logrus.StandardLogger(), l.metadataFetcher)
+	includeTestLevel := true
+	metadata, _ := shared.GetMetadataResponse(runs, includeTestLevel, logrus.StandardLogger(), l.metadataFetcher)
 	metadataMap := shared.PrepareLinkFilter(metadata)
 
 	return Link{
@@ -288,7 +289,14 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	}
 	for _, run := range runs {
 		if t.Product == nil || t.Product.Matches(run) {
-			metadata, _ := shared.GetMetadataResponse(runs, logrus.StandardLogger(), t.metadataFetcher)
+			includeTestLevel := false
+			metadataRuns := runs
+			// Product being nil means that we want test-level issues.
+			if (t.Product == nil) {
+				includeTestLevel = true
+				metadataRuns = make([]shared.TestRun, 0)
+			}
+			metadata, _ := shared.GetMetadataResponse(metadataRuns, includeTestLevel, logrus.StandardLogger(), t.metadataFetcher)
 			metadataMap := shared.PrepareLinkFilter(metadata)
 			cq = append(cq, Triaged{run.ID, metadataMap})
 		}
--- a/api/query/atoms_test.go
+++ b/api/query/atoms_test.go
@@ -890,10 +890,9 @@ func TestStructuredQuery_bindLink(t *testing.T) {
 		},
 	}
 
-	// AbstractLink should include test-level issues since it's about
-	// searching for links across all runs. It does not include the
-	// Chromium link because there is no run for Chromium (and thus no
-	// reason to include it, as the frontend won't show it).
+	// AbstractLink should bind test-level issues too as the pattern might match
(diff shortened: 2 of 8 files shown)


stephenmcgruer:
(Currently doesn't work; searching `triaged:test-issue` returns all tests which have **any** triage metadata, not specifically the test-level one. Probably need to change the `t.Product == nil` handling in `api/query/atoms.go` to filter the metadata map accordingly).

stephenmcgruer:
@KyleJu - PTAL. I do not like my approach of passing an empty array to `shared.GetMetadataResponse`, would love your thoughts there. Apart from that, I would also like to reduce the inefficiency of creating multiple `query.Triaged` objects, but that seems fundamentally tied to how `exists` works :(

api/query/atoms.go
```diff
@@ -257,7 +257,13 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	}
 	for _, run := range runs {
```
KyleJu, line 258: Like you mentioned, this `triaged:test-issue` creates multiple `query.Triaged` objects, but in fact it is independent of `runs` information. In the [filter](https://github.com/web-platform-tests/wpt.fyi/blob/4aff68a58e7a7c90cef004eaf350a06a21f71993/api/query/cache/index/filter.go#L330), all it filters out is the testname. It means that if this `test-issue` exists in one run, it will exist in all runs. I would do something like this

```suggestion
        if t.Product == nil && len(runs) != 0 {
            metadata, _ := shared.GetMetadataResponse(make([]shared.TestRun, 0), logrus.StandardLogger(), t.metadataFetcher)
	    metadataMap := shared.PrepareLinkFilter(metadata)
             return Or{Triaged{runs[0].ID, metadataMap}}
       }
	for _, run := range runs {
```

let me know if it makes sense

shared/metadata.go
```diff
@@ -111,30 +135,20 @@ func constructMetadataResponse(productSpecs ProductSpecs, metadata map[string]Me
...
-								},
-							}
-						}
-						if _, ok := res[fullTestName]; !ok {
-							res[fullTestName] = MetadataLinks{output}
-						} else {
-							res[fullTestName] = append(res[fullTestName], output)
-						}
+					// Matches on browser type if a version is not specified.
+					if link.Product.MatchesProductSpec(productSpec) {
```
KyleJu, line 150: Without `|| link.Product.BrowserName == "" `, aren't we skipping all the metadata for `test-issue` in `/api/metadata`? But looks like it is still working on [staging](https://smcgruer-test-issue-se-dot-wptdashboard-staging.uk.r.appspot.com/results/css/css-align?label=master&label=experimental&aligned)?

KyleJu:
> @KyleJu - PTAL. I do not like my approach of passing an empty array to `shared.GetMetadataResponse`, would love your thoughts there.

Why don't we create a separate function that retrieves `test-issue` metadata only? something like `constructTestIssueMetadataResponse( metadata map[string]Metadata) MetadataResults`. You can call this function separately with my [suggestion](https://github.com/web-platform-tests/wpt.fyi/pull/2440#discussion_r590926421)

> Apart from that, I would also like to reduce the inefficiency of creating multiple `query.Triaged` objects, but that seems fundamentally tied to how `exists` works :(
[Commented above](https://github.com/web-platform-tests/wpt.fyi/pull/2440#discussion_r590926421)

You mentioned that `/api/metadata?product=chrome on the deployed
instance and note that it no longer has entries for the test-level triaged tests.` Like [my comment](https://github.com/web-platform-tests/wpt.fyi/pull/2440#discussion_r590944210), how come this change doesn't break wpt-metadata?

shared/metadata.go
stephenmcgruer: Ok, I fixed this by adding an `includeTestLevel` flag, but I could also see adding `test-issue` as an acceptable `product/products` option for `/api/metadata` only. Lmk what you think.

KyleJu:
oops missed it. Looking at it now

KyleJu (commented):
let me think about the TODOs in this PR

api/query/atoms_test.go
```diff
@@ -871,10 +890,123 @@ func TestStructuredQuery_bindLink(t *testing.T) {
 		},
 	}
 
+	// AbstractLink should include test-level issues since it's about
+	// searching for links across all runs.
+	// TODO: Shouldn't this also pick up the chromium link too (a.html),
```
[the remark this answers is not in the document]
KyleJu: In other words, `link` searches metadata for products displayed on UI, otherwise `TestRuns` should include Chrome as well

api/query/atoms_test.go
```diff
@@ -871,10 +890,123 @@ func TestStructuredQuery_bindLink(t *testing.T) {
...
+		},
+	}
+	assert.Equal(t, expect, q.BindToRuns(runs...))
+
+	// This case doesn't match, so we should get back False.
+	q = AbstractTriaged{
+		Product:         &safari,
+		metadataFetcher: mockFetcher,
+	}
+	// TODO: This should be False{}, but we pass the entire 'runs' array to
```
KyleJu, line 950: Right, but like you said these `runs` in the runtime is an array of length 1. So

shared/metadata_test.go
```diff
@@ -209,13 +209,15 @@ func TestConstructMetadataResponse_OneMatchingBrowserVersion(t *testing.T) {
 		},
 	}
 
-	MetadataResults := constructMetadataResponse(productSpecs, metadataMap)
+	MetadataResults := constructMetadataResponse(productSpecs, false,metadataMap)
```
KyleJu, line 212:
```suggestion
	MetadataResults := constructMetadataResponse(productSpecs, false, metadataMap)
```

shared/metadata_test.go
```diff
@@ -109,7 +109,7 @@ func TestConstructMetadataResponse_OneLink(t *testing.T) {
 		},
 	}
 
-	MetadataResults := constructMetadataResponse(productSpecs, metadataMap)
+	MetadataResults := constructMetadataResponse(productSpecs, false, metadataMap)
```
KyleJu, line 112: I think we are losing some test coverages because in fact `includeTestLevel` is `true` most of the time, as most of the usage is from /api/metadata. Could you add test-level metadata to some of these tests?

api/query/atoms.go
```diff
@@ -257,7 +258,14 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
...
 		if t.Product == nil || t.Product.Matches(run) {
-			metadata, _ := shared.GetMetadataResponse(runs, logrus.StandardLogger(), t.metadataFetcher)
+			includeTestLevel := false
+			metadataRuns := runs
+			// Product being nil means that we want test-level issues.
+			if (t.Product == nil) {
+				includeTestLevel = true
+				metadataRuns = make([]shared.TestRun, 0)
```
KyleJu, line 266:
```suggestion
```
We don't need this anymore as test-level metadata will [break before the loop](https://github.com/web-platform-tests/wpt.fyi/pull/2440/files#diff-48d7909c1ce4d6c5e32eb0144ec673308fee32ab1f56f6b265010b988934e5ebR144)

api/query/atoms_test.go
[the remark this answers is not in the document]
stephenmcgruer: That makes sense. I have updated the documentation in `atoms.go` to make that clearer.

api/query/atoms_test.go
stephenmcgruer: Saying that this 'should be fine' is relying on the implementation detail of `AbstractExists` performing a disjunction over runs. If we want to bake that into `AbstractTriage` (i.e. assume that it is **only** ever called with 1 run), we should add code to error out of `AbstractTriage::BindToRuns` if runs has length >1.

I've done a quick audit, and I think all root atoms only pass single runs into their query atoms. So we could make that change to assume runs is length 1 - WDYT?

api/query/atoms.go
stephenmcgruer: I think we still need this; that break is only for a single result, for a single link in the metadata. For example, if you had:

```
links:
    - url: http://example.org
      results:
        - test: foo.html
    - product: chrome
      url: https://crbug.com/1234
      results:
        - test: bar.html
```

And you searched `triaged:test-issue`, the loop in `constructMetadataResponse` (assuming includeTestLevel=true and runs = [chrome, firefox, safari]) is:

```
1. for i := range data.Links
    a. link := data.Links[0] = (url: http://example.org, product: nil, results: [ test: foo.html ])
    b. for result := range link.Results
        i. result = (test: foo.html)
        ii. link.Product.BrowserName == "", and includeTestLevel=true, so add link and break
    [ next iteration of outer loop ]
    a. link := data.Links[1] = (url: http://crbug.com/1234, product: Chrome, results: [ test: bar.html ])
    b. for result := range link.Results
        i. result = (test: bar.html)
        ii. link.Product.BrowserName != ""
        iii. for productSpec := range [chrome, firefox, safari]
            1. productSpec = chrome
            2. if link.Product matches chrome, success, so **add link and break**
```

So we end up adding bar.html even though we only wanted test-level issues!

shared/metadata_test.go
stephenmcgruer: (Ack, need to think about this a bit)

api/query/atoms_test.go
KyleJu: Yea that is a good idea

api/query/atoms.go
KyleJu: Yea you are right. I was confused a little bit. Now we can return 

- all metadata by passing in `runs` and true for `includeTestLevel`
- metadata without test-level by passing in `runs` and false for `includeTestLevel`
- test-level only by passing empty runs and true for `includeTestLevel`

KyleJu (approved):

api/query/atoms.go
```diff
@@ -250,19 +250,22 @@ func (l AbstractLessThan) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return LessThan{c}
 }
 
-// AbstractLink is represents the root of a link query, which matches Metadata URLs
-// to a pattern string; it is independent of test runs.
+// AbstractLink is represents the root of a link query, which matches Metadata
```
KyleJu, line 253:
```suggestion
// AbstractLink is the root of a link query, which matches Metadata
```

KyleJu:
There were issues with GitHub but LGTM. I will let you follow up on [the test comment](https://github.com/web-platform-tests/wpt.fyi/pull/2440#discussion_r593228289)

api/query/atoms_test.go
KyleJu: +1 for the fix and I agree. The search logic in searchcache is quite difficult, and it would require some rethinking to refactor

PR #2449, opened 2021-03-09
Remove deprecated experimentalByDefault and experimentalAligned flags

stephenmcgruer:
Both of these flags have been default-on in wpt.fyi for many years, and
we are exceedingly unlikely to change them to be false at any point.
Removing them clarifies the intent of the code - and has also revealed
an existing inconsistency in our codebase (see Default() in
shared/test_run_filter.go).

### DIFF
--- a/shared/test_run_filter.go
+++ b/shared/test_run_filter.go
@@ -94,6 +94,10 @@ func (filter TestRunFilter) IsDefaultQuery() bool {
 // OrDefault returns the current filter, or, if it is a default query, returns
 // the query used by default in wpt.fyi.
 func (filter TestRunFilter) OrDefault() TestRunFilter {
+	// TODO(smcgruer): OrAlignedStableRuns is not the default query in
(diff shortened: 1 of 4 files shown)


KyleJu (approved):
Note that these flags need to be deleted from [Prod Datastore](https://pantheon.corp.google.com/datastore/entities;kind=Flag;ns=__$DEFAULT$__/query/kind?project=wptdashboard) after the upcoming release and [staging Datastore](https://pantheon.corp.google.com/datastore/entities;kind=Flag;ns=__$DEFAULT$__/query/kind?project=wptdashboard-staging) after this PR is merged

stephenmcgruer:
> Note that these flags need to be deleted from Prod Datastore after the upcoming release and staging Datastore after this PR is merged

A good reminder, thanks! I've scheduled myself a task to do so next week. I'll also audit the other datastore flags at that time.

Issue #2450, opened 2021-03-09
[Travis] Lighthouse errors on the main branch


KyleJu:
Starting from https://github.com/web-platform-tests/wpt.fyi/commit/1b8bd43c1e4c90445b28b27f6f5b259a4658ceff, Travis CI on the main branch started seeing [lighthouse errors](https://travis-ci.com/github/web-platform-tests/wpt.fyi/jobs/489479094):

Running Lighthouse 3 time(s) on https://staging.wpt.fyi/results/?label=experimental&label=master&aligned
Run #1...failed!
Error: Lighthouse failed with exit code 1
    at ChildProcess.child.on.code (/home/user/wpt.fyi/webapp/node_modules/@lhci/cli/src/collect/node-runner.js:119:21)
    at ChildProcess.emit (events.js:198:13)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:248:12)

This error likely happened before [this PR](https://github.com/web-platform-tests/wpt.fyi/commit/1b8bd43c1e4c90445b28b27f6f5b259a4658ceff) because Travis had been failing on `Stage 1: deploy main` steps for a long time

KyleJu:
The last time when Travis CI was green completely was on [this commit](https://github.com/web-platform-tests/wpt.fyi/commit/a4602edec0d22be096759f9f3ee7f2b575758361)

KyleJu:
I have found similar errors here - https://github.com/GoogleChrome/lighthouse-ci/issues/398

KyleJu:
Digging into the log more, `NO_FCP` indicates that the error is related to wpt.fyi page loading issue. 

"runtimeError": {
    "code": "NO_FCP",
    "message": "Something went wrong with recording the trace over your page load. Please run Lighthouse again. (NO_FCP)"
  },

Running Lighthouse through devtools still works as expected, although it has a very low performance score. Given that we will only maintain wpt.fyi in the near future, I will disable Lighthouse tests and keep an eye on the accessibility score through devtools

KyleJu:
Travis is green now

PR #2452, opened 2021-03-09
[Travis] Remove lighthouse from CI

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2450. 

See [the origin setup](https://github.com/web-platform-tests/wpt.fyi/pull/1767)


webapp/package-lock.json
```diff
@@ -765,252 +765,6 @@
         "@polymer/polymer": "^3.0.0"
```
KyleJu, line 765: Changes in this file is the result of running `npm install`

PR #2463, opened 2021-03-09
[Pyup] Change pyup update frequency to monthly

KyleJu:
Schedule monthly pyup update, instead of weekly

See `schedules` in https://pyup.io/docs/bot/config/


stephenmcgruer (approved):
Good start; we can see how this frequency feels and then adjust more if necessary :)

PR #2464, opened 2021-03-10
[Renovate] Schedule renovate update every Monday

KyleJu:
Switch renovate to batch updates every Monday. Currently Renovate sends updates whenever there is one available.

See the [schedule](https://docs.renovatebot.com/presets-schedule/#scheduleearlymondays) syntax

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,7 @@
 {
+  "schedule": [
+    "before 3am on Monday"
+  ],
   "extends": [
     "config:base",
     ":automergeMinor",


stephenmcgruer (approved):
Can I clarify; when I read "on demand" I usually interpret it as "someone has to make a request for this to happen". However here I think by "on demand" you mean "renovate will schedule an update instantly whenever there is one".

If that's correct, I would suggest rewording the PR description a little to clarify that we are switching to batch updates (which I believe is fairly common terminology), rather than as-available updates.

KyleJu:
> Can I clarify; when I read "on demand" I usually interpret it as "someone has to make a request for this to happen". However here I think by "on demand" you mean "renovate will schedule an update instantly whenever there is one".
> 
> If that's correct, I would suggest rewording the PR description a little to clarify that we are switching to batch updates (which I believe is fairly common terminology), rather than as-available updates.

Updated the description!

PR #2468, opened 2021-03-10
Remove deprecated interopScoreColumn flag

stephenmcgruer:
This flag was never turned on, and the feature it hid was never
completed. We've since moved on from this view almost entirely, so
remove the dead code.


stephenmcgruer:
Low priority cleanup

KyleJu (approved):

PR #2469, opened 2021-03-10
Remove the never-launched manifest frontend code

stephenmcgruer:
The use of the manifest in the frontend never shipped, and the flags are
off in prod and staging. This code is non-trivial and removing it makes
the code easier to understand.

The /api/manifest endpoint is not removed because it is still [used by
Bikeshed](https://github.com/tabatkins/bikeshed/blob/e797b9548c0bd140c9b1500809869b09f50f7d99/bikeshed/update/updateWpt.py)
and the [GitHub checks in wpt-metadata](https://github.com/web-platform-tests/wpt-metadata/blob/557b49eb329670395d90744b084869e848b268ac/test_paths_test.go#L30)

Updates #56

### DIFF
--- a/webapp/components/wpt-flags.js
+++ b/webapp/components/wpt-flags.js
@@ -40,7 +40,6 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'colorHomepage',
       'diffFromAPI',
       'displayMetadata',
-      'fetchManifestForTestList',
       'githubCommitLinks',
       'githubLogin',
       'interopScoreColumn',
@@ -48,12 +47,8 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'processorTab',
       'queryBuilder',
       'queryBuilderSHA',
-      'reftestAnalyzerMockScreenshots',
-      'reftestIframes',
       'searchCacheInterop',
       'showBSF',
-      'showTestType',
-      'showTestRefURL',
       'structuredQueries',
       'triageMetadataUI',
       'webPlatformTestsLive',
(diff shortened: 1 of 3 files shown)


stephenmcgruer:
Low priority and needs a careful look at it, but I think this is a nice cleanup of a chunk of code.

webapp/components/wpt-flags.js
```diff
@@ -40,20 +40,15 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
...
       'interopScoreColumn',
       'permalinks',
       'processorTab',
       'queryBuilder',
       'queryBuilderSHA',
-      'reftestAnalyzerMockScreenshots',
-      'reftestIframes',
       'searchCacheInterop',
       'showBSF',
-      'showTestType',
```
KyleJu, line 55: This is true on [staging](https://pantheon.corp.google.com/datastore/entities;kind=Flag;ns=__$DEFAULT$__/query/kind?project=wptdashboard-staging)

KyleJu: so is `showTestRefURL`

stephenmcgruer: Yes, however I'm fairly sure they have zero effect - the actual usage of them relies on `this.manifest` being filled, but the top-level flag being off means thats always null. I'd love if you could help me double-check that though!

KyleJu: Yep good call! `fetchManifestForTestList` is never used -> `this.manifest` is always null -> `testType` is always ''. I set a few break points at the removed code in this PR and they are never triggered.

KyleJu (approved):

KyleJu:
Note that /api/manifest is also used by [GitHub checks in wpt-metadata](https://github.com/web-platform-tests/wpt-metadata/blob/557b49eb329670395d90744b084869e848b268ac/test_paths_test.go#L30)

PR #2470, opened 2021-03-11
[compat2021] Add summary scores component

stephenmcgruer:
This is mostly a copy of the existing code from
https://ecosystem-infra.github.io/wpt-results-analysis/compat, but with
the necessary modifications for the polymer ecosystem (e.g. using data
binding, etc). A few TODOs are left for cases where using more polymer
features may be useful.

### DIFF
--- a/webapp/components/compat-2021.js
+++ b/webapp/components/compat-2021.js
@@ -4,7 +4,10 @@
  * found in the LICENSE file.
  */
 
+import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
 import { html, PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
+
+const GITHUB_URL_PREFIX = 'https://raw.githubusercontent.com/Ecosystem-Infra/wpt-results-analysis/gh-pages';
 
 // Compat2021 is a custom element that holds the overall compat-2021 dashboard.
 // The dashboard breaks down into top-level summary scores, a small description,
@@ -13,7 +16,7 @@ class Compat2021 extends PolymerElement {
   static get template() {
     return html`
       <h1>Compat 2021 Dashboard</h1>
-      <p>TODO: Summary scores</p>
+      <compat-2021-summary stable="[[stable]]"></compat-2021-summary>
       <p>
         These scores represent how well browser engines are doing on the 2021
         Compat Focus Areas, as measured by wpt.fyi test results. Each feature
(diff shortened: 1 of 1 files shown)


webapp/components/compat-2021.js
```diff
@@ -34,6 +39,218 @@ class Compat2021 extends PolymerElement {
...
+const STABLE_TITLES = [
+  'Chrome/Edge Stable',
+  'Firefox Stable',
+  'Safari Stable',
+];
+
+const EXPERIMENTAL_TITLES = [
+  'Chrome/Edge Dev',
+  'Firefox Nightly',
+  'Safari Tech Preview',
```
foolip, line 67:
```suggestion
  'Safari Technology Preview',
```

Or maybe "Safari Preview" or  "Safari TP" if it needs to be short.

webapp/components/compat-2021.js
```diff
@@ -34,6 +39,218 @@ class Compat2021 extends PolymerElement {
...
+
+class Compat2021Summary extends PolymerElement {
+  static get template() {
+    return html`
+      <style>
+        #summaryContainer {
+          padding-top: 1em;
+          display: flex;
+          justify-content: center;
+          gap: 30px;
```
foolip, line 78: :)

webapp/components/compat-2021.js
```diff
@@ -34,6 +39,218 @@ class Compat2021 extends PolymerElement {
...
+    const label = stable ? 'stable' : 'experimental';
+    const url = `${GITHUB_URL_PREFIX}/data/compat2021/summary-${label}.csv`;
+    const csvResp = await fetch(url);
+    if (!csvResp.ok) {
+      throw new Error(`Fetching chart csv data failed: ${csvResp.status}`);
+    }
+    const csvText = await csvResp.text();
+    const csvLines = csvText.split('\n');
+    csvLines.shift();  // We don't need the CSV header.
+    csvLines.pop();  // Trailing empty line.
```
foolip, line 206:
```suggestion
    const csvLines = csvText.split('\n').filter(l => l);
    csvLines.shift();  // We don't need the CSV header.
```

That seems less brittle in case the trailing newline is removed.

foolip (approved):

webapp/components/compat-2021.js
stephenmcgruer: Went for "Safari Preview" for now as the full string was too long. We can always change it later.

PR #2471, opened 2021-03-11
[compat2021] Add chart component

stephenmcgruer:
This is mostly a copy of the existing code from
https://ecosystem-infra.github.io/wpt-results-analysis/compat, but with
the necessary modifications for the polymer ecosystem (e.g. using data
binding, etc). A few TODOs are left for cases where using more polymer
features may be useful.

### DIFF
--- a/webapp/components/compat-2021.js
+++ b/webapp/components/compat-2021.js
@@ -4,6 +4,7 @@
  * found in the LICENSE file.
  */
 
+import {load} from '../node_modules/@google-web-components/google-chart/google-chart-loader.js';
 import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
 import { html, PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
 
@@ -31,7 +32,7 @@ class Compat2021 extends PolymerElement {
           experimental features enabled.
         </template>
       </p>
-      <p>TODO: Individual feature graph</p>
+      <compat-2021-feature-chart stable="[[stable]]"></compat-2021-feature-chart>
       <p>TODO: Test results table</p>
 `;
   }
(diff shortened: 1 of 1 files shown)


webapp/components/compat-2021.js
```diff
@@ -254,3 +255,228 @@ class Compat2021Summary extends PolymerElement {
...
+          margin: 0 auto;
+        }
+      </style>
+
+      <h3>Individual Features</h3>
+
+      <!-- TODO: replace with paper-dropdown-menu -->
+      <div>
+        <label for="featureSelect">Focus area:</label>
+        <select id="featureSelect">
```
foolip, line 275: A problem here is there's no way to go back to the overall view once you've chosen a subview. Can you add an explicitly "everything" option which is selected by default?

webapp/components/compat-2021.js
```diff
@@ -254,3 +255,228 @@ class Compat2021Summary extends PolymerElement {
...
+          },
+          tooltip: {
+            trigger: 'both',
+          },
+          hAxis: {
+            title: 'Date',
+            format: 'MMM-YYYY',
+          },
+          vAxis: {
+            title: 'Percentage of passing tests',
```
foolip, line 310:
```suggestion
            title: 'Percentage of tests passing',
```

Right?

webapp/components/compat-2021.js
```diff
@@ -254,3 +255,228 @@ class Compat2021Summary extends PolymerElement {
...
+    const div = this.$.failuresChart;
+    const label = stable ? 'stable' : 'experimental';
+    const url = `${GITHUB_URL_PREFIX}/data/compat2021/${feature}-${label}.csv`;
+
+    const csvResp = await fetch(url);
+    if (!csvResp.ok) {
+      throw new Error(`Fetching chart csv data failed: ${csvResp.status}`);
+    }
+    const csvText = await csvResp.text();
+    const csvLines = csvText.split('\n');
```
foolip, line 375: This code looks very familiar. Will there be more, or just one dupe? If it's trivial to factor out a `fetchCSV()` that might be an improvement even with only two call sites.

webapp/components/compat-2021.js
stephenmcgruer: I'm not clear what the purpose of that would be? Currently, as requested by Chris, the overall view is just blank (https://ecosystem-infra.github.io/wpt-results-analysis/compat).

However typing that, it becomes instantly clear that the no-feature view should probably be a graph of the overall score over time -_-. Does that SGTY? If so, it will take time to put it together so I'd suggest we keep this behavior for now.

foolip: Aha, so I had imagined that the overall big numbers were also updating, and that there's no way to go back.

An overall graph would perhaps be nice, but yeah let's leave this for now.

foolip (approved):

PR #2472, opened 2021-03-11
[compat2021] Add stable checkbox and update URL as feature/stable change

stephenmcgruer:
This also brings in most of Una's changes from
https://github.com/Ecosystem-Infra/wpt-results-analysis/pull/38 and
https://github.com/Ecosystem-Infra/wpt-results-analysis/pull/42

Includes a few TODOs for places where we could be using polymer
elements/etc.

### DIFF
--- a/webapp/components/compat-2021.js
+++ b/webapp/components/compat-2021.js
@@ -32,8 +32,18 @@ class Compat2021 extends PolymerElement {
           experimental features enabled.
         </template>
       </p>
-      <compat-2021-feature-chart stable="[[stable]]"></compat-2021-feature-chart>
-      <p>TODO: Test results table</p>
+
+      <!-- TODO: Use a paper-toggle instead -->
+      <div>
+        <label for="stableVersionCheckbox">Show results from Stable browser releases:</label>
+        <input id="stableVersionCheckbox" type="checkbox" />
+      </div>
+
+      <compat-2021-feature-chart stable="[[stable]]"
+                                 feature="{{feature}}">
+      </compat-2021-feature-chart>
+
+      <!-- TODO: Test results table -->
 `;
   }
 
@@ -44,7 +54,14 @@ class Compat2021 extends PolymerElement {
   static get properties() {
     return {
       stable: Boolean,
+      feature: String,
     };
+  }
+
+  static get observers() {
+    return [
+      'updateUrlParams(stable, feature)',
+    ];
   }
 
   ready() {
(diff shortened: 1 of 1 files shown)


webapp/components/compat-2021.js
```diff
@@ -32,8 +32,18 @@ class Compat2021 extends PolymerElement {
           experimental features enabled.
         </template>
       </p>
-      <compat-2021-feature-chart stable="[[stable]]"></compat-2021-feature-chart>
-      <p>TODO: Test results table</p>
+
+      <!-- TODO: Use a paper-toggle instead -->
+      <div>
+        <label for="stableVersionCheckbox">Show results from Stable browser releases:</label>
+        <input id="stableVersionCheckbox" type="checkbox" />
```
foolip, line 39: Can we use the toggle component, like Robert suggested?

stephenmcgruer: Are you thinking of a toggle that looks like:

```
stable (   (*)) experimental
[click]
stable ((*)   ) experimental
```

I.e. has text on either side? Or are you thinking of something different?

One different approach we could also take is the experimental/stable buttons you find on the BSF graph on https://wpt.fyi, where they're separate buttons but they can't both be selected.

foolip: I guess a toggle indeed isn't great, it's better if one can understand the current state, not just that it's *not* stable, if that were the name of the toggle.

The buttons on the BSF graphs seem OK. They take up a lot of horizontal space, but maybe that's OK/

foolip:
There are conflicts to be resolved here now, I guess after Una's fixes.

stephenmcgruer:
> There are conflicts to be resolved here now, I guess after Una's fixes.

Una's fixes were to https://ecosystem-infra.github.io/wpt-results-analysis/compat, not the wpt.fyi version. That said, they do still need to be incorporated and I still need to follow through on https://github.com/web-platform-tests/wpt.fyi/pull/2472#discussion_r593202256

I tend to follow the GitHub UI - when it's ready for re-review I press the 're-request review' button :).

stephenmcgruer:
@foolip PTAL, I've switched to BSF-style buttons and pulled in a bunch of Una's fixes. Only thing I haven't gotten working yet is her 700px max-width (wpt.fyi itself has a max-width of 1200px)

webapp/components/compat-2021.js
```diff
@@ -16,6 +17,45 @@ const GITHUB_URL_PREFIX = 'https://raw.githubusercontent.com/Ecosystem-Infra/wpt
...
+          font-size: 18px;
+          display: flex;
+          justify-content: center;
+          flex-direction: column;
+        }
+
+        .unselected {
+          background-color: white;
+        }
+        .selected{
```
foolip, line 47:
```suggestion
        .selected {
```

webapp/components/compat-2021.js
```diff
@@ -44,14 +113,73 @@ class Compat2021 extends PolymerElement {
...
+  }
+
   ready() {
     super.ready();
 
     const params = (new URL(document.location)).searchParams;
     this.stable = params.get('stable') !== null;
+    this.feature = params.get('feature');
+
+    // The default behavior of the page (when loaded with no params) is to not
```
foolip, line 133: This was added earlier and just moved now, but I don't understand the comment. Setting `value` has the same effect regardless isn't it? Or is this in reference to some work that isn't being done which the reader would suspect is needed but actually isn't?

webapp/components/compat-2021.js
```diff
@@ -44,14 +113,73 @@ class Compat2021 extends PolymerElement {
...
+
+    const params = [];
+    if (feature) {
+      params.push(`feature=${feature}`);
+    }
+    if (stable) {
+      params.push('stable');
+    }
+
+    // We always append a '?' at the very least, as passing empty-string to
```
foolip, line 156: If we use the full URL, or at least prepend `/compat2021`, can we avoid appending the `?`?

I think it's just the empty string that's special cased.

foolip (approved):
LGTM % nits

webapp/components/compat-2021.js
stephenmcgruer: Ah, excellent. `location.pathname` works, thanks!

PR #2477, opened 2021-03-15
[Triage Metadata] Implement Redis commands and /api/metadata/pending

KyleJu:
Implement Redis commands and the /api/metadata/pending endpoint. The purpose of this PR is to allow instant reflection of the triage icons. For more information, see the [design doc](https://docs.google.com/document/d/1tK8PiR-wpQEOEHQl92rqLmcbYWYsAnI1y-rPIf-s6hY/edit#heading=h.9cn1ocwm764n)

Redis commands - see [documentations](https://pkg.go.dev/github.com/gomodule/redigo/redis#hdr-Executing_Commands) and [an example](https://github.com/web-platform-tests/wpt.fyi/blob/main/shared/cache.go#L174)

### Test
Tested in [prod](https://github.com/web-platform-tests/wpt.fyi/pull/2477#issuecomment-799841986) using [dummy data](https://github.com/web-platform-tests/wpt.fyi/pull/2477/commits/df76f345a1b52f2b3709bf13f65b451837bdc123)


KyleJu:
` go_test (pull_request) ` failed two runs in a row. It might interfere with Redis emulators or other test setup

KyleJu:
Tested using dummy data
Screen Shot 2021-03-15 at 8 02 22 PM

stephenmcgruer (commented):
One high-level question I have (that I should have realized when reviewing the design doc, sorry): 

Should we be storing the pending PR information in datastore instead? Redis is intended for caching of data that does not need to survive (as all data can be evicted at any time, I believe). That usually means that the data is permanently backed by another source, but in this case there is no retrievable backing source? (Unless we plan to fetch outstanding PRs from the github repo, and try to reverse-engineer the diffs back into MetadataLink objects?)

The current implementation could be ok if we're making the assumption that this is a best-effort API, and that some pending PRs may not be shown in it. If we are, we should be explicit about that in the documentation.

KyleJu:
> One high-level question I have (that I should have realized when reviewing the design doc, sorry):
> 
> Should we be storing the pending PR information in datastore instead? Redis is intended for caching of data that does not need to survive (as all data can be evicted at any time, I believe). 

Per our offline discussions, the metadata information we stored is not intended to survive for a long time. For the cache set, there is no expiry time and we will remove them once its corresponding PR is merged or closed. For the temporary metadata stored in `NewJSONObjectCache`, I am planning to set a one-week TTL because (at least I hope) all pending PRs will be addressed within a week.

>That usually means that the data is permanently backed by another source, but in this case there is no retrievable backing source? (Unless we plan to fetch outstanding PRs from the github repo, and try to reverse-engineer the diffs back into MetadataLink objects?)

Even in the case of data corruption or Reids going down, the information is not permanently lost because they will be reflected in the /api/metadata endpoints once pending PRs are addressed. But yes, they might be temporarily lost as long as the PRs are stuck. But whats the chance of Redis going down? It should be unlikely because Redis has [one replica](https://cloud.google.com/memorystore/docs/redis/high-availability), in case of deployment, upgrade or maintenance. Also judging from what we see from Patheon, we didn't see any particular point where Redis' usage went to 0.

api/metadata_handler_test.go
```diff
@@ -310,3 +311,102 @@ func getMetadataTestData() map[string][]byte {
...
+	handlePendingMetadata(ctx, mockCache, mockSet, w, r)
+
+	assert.Equal(t, http.StatusOK, w.Code)
+
+	res := w.Body.String()
+	json.Unmarshal([]byte(res), &actual)
+	assert.Equal(t, expected, actual)
+}
+
+func TestPendingMetadataHandler_Fail(t *testing.T) {
```
stephenmcgruer, line 397: Nit; we don't have coverage for "the set has the key but the json cache doesn't".

shared/cache.go
```diff
@@ -431,3 +433,66 @@ func FlushCache() error {
 	_, err := conn.Do("FLUSHALL")
 	return err
 }
+
+// RedisSet is an interface for an redisSetReadWritable,
+// which performs Add/Remove/GetAll operations via the App Engine Redis API.
+type RedisSet interface {
+	// Add inserts value to the set stored at key; ignored if value is
+	// already a member of this set.
```
stephenmcgruer, line 441:
```suggestion
	// already a member of the set.
```

shared/cache.go
```diff
@@ -431,3 +433,66 @@ func FlushCache() error {
...
+	Add(key string, value string) error
+	// Remove removes value to the set stored at key; ignored if value is
+	// not a member of this set.
```
stephenmcgruer, line 444:
```suggestion
	// not a member of the set.
```

shared/cache.go
```diff
@@ -431,3 +433,66 @@ func FlushCache() error {
...
 	return err
 }
+
+// RedisSet is an interface for an redisSetReadWritable,
+// which performs Add/Remove/GetAll operations via the App Engine Redis API.
+type RedisSet interface {
+	// Add inserts value to the set stored at key; ignored if value is
+	// already a member of this set.
+	Add(key string, value string) error
+	// Remove removes value to the set stored at key; ignored if value is
```
stephenmcgruer, line 443:
```suggestion
	// Remove removes value from the set stored at key; ignored if value is
```

shared/cache.go
```diff
@@ -431,3 +433,66 @@ func FlushCache() error {
...
+	}
+	conn := Clients.redisPool.Get()
+	defer conn.Close()
+
+	// https://redis.io/commands/smembers
+	value, err := redis.Strings(conn.Do("SMEMBERS", key))
+	if err != nil {
+		return nil, err
+	} else if value == nil {
+		return nil, errCacheMiss
```
stephenmcgruer, line 494: The documentation above says that we return `[]string` if the key is not present, but this looks like we return an error instead?

stephenmcgruer (approved):
LGTM modulo comments

shared/cache.go
KyleJu: You are right. Well technically `value` can be never nil in that case. I will remove this else if

PR #2478, opened 2021-03-15
[compat2021] Link to the compat2021 page from the site header

stephenmcgruer:


stephenmcgruer:
image

foolip (approved):
✨LGTM✨

This feels like a somewhat silly way to bring attention to it, but also it will work, so let's sparkle!

PR #2486, opened 2021-03-16
Fix alignment of test results page

stephenmcgruer:
In cf18d4857908847293741223559193b4755a019c, we removed the unused
manifest code. However we overlooked that there was an invisible
zero-width table column in existence for this unused feature, and when
it was removed the 'path' cell stretched too far.

Fix this by making the path column single-wide, not double.


stephenmcgruer:
Deployed instance looks identical to prod to me now.

KyleJu (approved):

PR #2487, opened 2021-03-18
Rename all Memcache to Redis

KyleJu:
Since go 1.14 migration, we have migrated from Memcache to Redis. Rename all occurrences of memcache to redis.


KyleJu:
Might be easier to review commit by commit

stephenmcgruer (approved):

PR #2488, opened 2021-03-19
[compat2021] Add a default summary-over-time view on the graph

stephenmcgruer:
The UI change here is not massive, but requires substantially
refactoring how the page works. We now load the unified CSV rather than
individual CSV files for each feature. To aid with this and provide a
layer of caching, a specific data management class is introduced
(generalized from just chart data as it will eventually be used for the
summary scores at the top as well).

In addition to the above, this PR also improves the browser changelog
tooltip action. Loading changelogs automatically is now supported for
Firefox stable, and paper-dialogs pop-ups are added for Firefox Nightly
and Safari.

### DIFF
--- a/webapp/components/compat-2021.js
+++ b/webapp/components/compat-2021.js
@@ -6,6 +6,8 @@
 
 import {load} from '../node_modules/@google-web-components/google-chart/google-chart-loader.js';
 import '../node_modules/@polymer/paper-button/paper-button.js';
+import '../node_modules/@polymer/paper-dialog/paper-dialog.js';
+import '../node_modules/@polymer/paper-input/paper-input.js';
 import '../node_modules/@polymer/polymer/lib/elements/dom-if.js';
 import { html, PolymerElement } from '../node_modules/@polymer/polymer/polymer-element.js';
 
@@ -551,8 +553,55 @@ class Compat2021FeatureChart extends PolymerElement {
           height: 350px;
           margin: 0 auto;
         }
+
+        paper-dialog {
+          max-width: 600px;
+        }
       </style>
       <div id="failuresChart" class="chart"></div>
+
+      <paper-dialog with-backdrop id="firefoxNightlyDialog">
+        <h2>Firefox Nightly Changelogs</h2>
+        <div>
+          Nightly builds of Firefox are all given the same sub-version,
+          <code>0a1</code>, so we cannot automatically determine the changelog.
+          To find the changelog of a specific Nightly release, locate the
+          corresponding SHAs on the
+          <a href="https://hg.mozilla.org/mozilla-central/firefoxreleases"
+             target="_blank">release page</a>, enter them below, and click "Go".
+          <paper-input id="firefoxNightlyDialogFrom" label="From SHA"></paper-input>
+          <paper-input id="firefoxNightlyDialogTo" label="To SHA"></paper-input>
+        </div>
+
+        <div class="buttons">
+          <paper-button dialog-dismiss>Cancel</paper-button>
+          <paper-button dialog-confirm on-click="onFirefoxNightlyDialogGo">Go</paper-button>
+        </div>
+      </paper-dialog>
+
+      <paper-dialog with-backdrop id="safariDialog">
+        <h2>Safari Changelogs</h2>
+        <template is="dom-if" if="[[stable]]">
+          <div>
+            Stable releases of Safari do not publish changelogs, but some insight
+            may be gained from the
+            <a href="https://developer.apple.com/documentation/safari-release-notes"
+               target="_blank">Release Notes</a>.
+          </div>
+        </template>
+        <template is="dom-if" if="[[!stable]]">
+          <div>
+            For Safari Technology Preview releases, release notes can be found on
+            the <a href="https://webkit.org/blog/" target="_blank">WebKit Blog</a>.
+            Each post usually contains a revision changelog link - look for the
+            text "This release covers WebKit revisions ...".
+          </div>
+        </template>
+
+        <div class="buttons">
+          <paper-button dialog-dismiss>Dismiss</paper-button>
+        </div>
+      </paper-dialog>
 `;
   }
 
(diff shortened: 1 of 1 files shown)


webapp/components/compat-2021.js
```diff
@@ -6,10 +6,152 @@
...
+];
+
+// Compat2021DataManager encapsulates the loading of the CSV data that backs
+// both the summary scores and graphs shown on the Compat 2021 dashboard. It
+// fetches the CSV data, processes it into sets of datatables, and then caches
+// those tables for later use by the dashboard.
+class Compat2021DataManager {
+  constructor() {
+    this._dataLoaded = load().then(() => {
+      return Promise.all([this._loadCsv(true), this._loadCsv(false)]);
```
foolip, line 31:
```suggestion
      return Promise.all([this._loadCsv('stable'), this._loadCsv('experimental')]);
```

Seems just as well since the first thing done with the boolean is to turn it into these labels.

But booleans are still used in other calls, if you think it's better to be consistent please reject.

webapp/components/compat-2021.js
```diff
@@ -6,10 +6,152 @@
...
+    });
+  }
+
+  // Fetches the datatable for the given feature and stable/experimental state.
+  // This will wait as needed for the underlying CSV data to be loaded and
+  // processed before returning the datatable.
+  async getDatatable(feature, stable) {
```
foolip, line 38:
```suggestion
  async getDataTable(feature, stable) {
```

Since these are `google.visualization.DataTable`, and also with the "atat" in this word it's easier to read like this.

webapp/components/compat-2021.js
```diff
@@ -6,10 +6,152 @@
...
+    const browserVersions = [[], [], []];
+
+    csvLines.forEach(line => {
+      // We control the CSV data source, so are quite lazy with parsing it.
+      //
+      // The format is:
+      //   date, [browser-version, browser-feature-a, browser-feature-b, ...]+
+      const csvValues = line.split(',');
+
+      // Javascript Date objects use 0-indexed months whilst the CSV is
```
foolip, line 97:
```suggestion
      // JavaScript Date objects use 0-indexed months whilst the CSV is
```

webapp/components/compat-2021.js
```diff
@@ -6,10 +6,152 @@
...
+      const date = new Date(dateParts[0], dateParts[1] - 1, dateParts[2]);
+
+      // Initialize a new row for each feature, with the date column set.
+      const newRows = new Map(features.map(feature => {
+        return [feature, [date]];
+      }));
+
+      // Now handle each of the browsers. For each there is a version column,
+      // then the scores for each of the five features.
+      for (let i = 1; i < csvValues.length; i += 6) {
```
foolip, line 109: This loop works, but if there's a way to instead loop over `browserVersions` the right number of times it could end up more readable. I wouldn't know without trying it, though, so feel free to ignore if you don't think so. Not important.

webapp/components/compat-2021.js
```diff
@@ -400,54 +553,63 @@ class Compat2021FeatureChart extends PolymerElement {
...
+        <h2>Firefox Nightly Changelogs</h2>
+        <div>
+          Nightly builds of Firefox are all given the same sub-version,
+          <code>0a1</code>, so we cannot automatically determine the changelog.
+          To find the changelog of a specific Nightly release, locate the
+          corresponding SHAs on the
+          <a href="https://hg.mozilla.org/mozilla-central/firefoxreleases"
+             target="_blank">release page</a>, enter them below, and click "Go".
+          <paper-input id="firefoxNightlyDialogFrom" label="From SHA"></paper-input>
+          <paper-input id="firefoxNightlyDialogTo" label="To SHA"></paper-input>
```
foolip, line 573:
```suggestion
          corresponding revisions on the
          <a href="https://hg.mozilla.org/mozilla-central/firefoxreleases"
             target="_blank">release page</a>, enter them below, and click "Go".
          <paper-input id="firefoxNightlyDialogFrom" label="From revision"></paper-input>
          <paper-input id="firefoxNightlyDialogTo" label="To revision"></paper-input>
```

Matching the terminology of the linked page.

webapp/components/compat-2021.js
```diff
@@ -400,54 +553,63 @@ class Compat2021FeatureChart extends PolymerElement {
...
+            <a href="https://developer.apple.com/documentation/safari-release-notes"
+               target="_blank">Release Notes</a>.
+          </div>
+        </template>
+        <template is="dom-if" if="[[!stable]]">
+          <div>
+            For Safari Technology Preview releases, release notes can be found on
+            the <a href="https://webkit.org/blog/" target="_blank">WebKit Blog</a>.
+            Each post usually contains a revision changelog link - look for the
+            text "This release covers WebKit revisions ...".
```
foolip, line 597: Attention to detail here :)

webapp/components/compat-2021.js
```diff
@@ -461,124 +623,135 @@ class Compat2021FeatureChart extends PolymerElement {
...
         while (row > 0 && lastVersion === version) {
           row -= 1;
           lastVersion = browserVersions[browserIdx][row];
         }
-        // TODO: If row == -1, we've failed, but we should grey out the
-        // option instead in that case.
-        window.open(this.getChromeDiffUrl(lastVersion, version));
+        // TODO: If row == -1 here then we've failed.
+
+        if (browserIdx === 0) {
```
foolip, line 663: Can we look up the browser name using `browserIdx` and switch on that?

webapp/components/compat-2021.js
```diff
@@ -461,124 +623,135 @@ class Compat2021FeatureChart extends PolymerElement {
...
+    // The version numbers are reported as XX.Y.Z, but pushlog wants
+    // 'FIREFOX_XX_Y_Z_RELEASE'.
+    const fromParts = fromVersion.split('.');
+    const fromRelease = `FIREFOX_${fromParts.join('_')}_RELEASE`;
+    const toParts = toVersion.split('.');
+    const toRelease = `FIREFOX_${toParts.join('_')}_RELEASE`;
+    return `https://hg.mozilla.org/mozilla-unified/pushloghtml?fromchange=${fromRelease}&tochange=${toRelease}`;
+  }
+
+  onFirefoxNightlyDialogGo() {
```
foolip, line 701: `onFirefoxNightlyDialogGo` look like pretty unusual naming to me. `openFirefoxNightlyDialog`? Or what does the "go" bit mean?

foolip (approved):
LGTM % nits

webapp/components/compat-2021.js
stephenmcgruer: Yeah, there are better ways to do this but I think this is ok for now (no time!).

webapp/components/compat-2021.js
stephenmcgruer: As with elsewhere we can and probably should, but this works for now and I don't want to spend more time refactoring this. Sorry!

webapp/components/compat-2021.js
stephenmcgruer: Oh, the 'Go' button is in the pop-up dialog and this is the handler for it. I'll add the word `Button` or something to clarify a bit.

stephenmcgruer:
I'll give Chris and Robert a few hours to raise any related feature requests then merge this.

PR #2492, opened 2021-03-22
[Triage Metadata] Cache pending metadata in a triage

KyleJu:
Cache a PR number and its pending metadata to Redis in a triage. If cache fails, the request still returns 200.

The TTL of pending metadata in Redis is 7 days. If a PR is still stuck in the wpt-metadata repo after 7 days, the PR should be removed from the pending metadata cache set. The pending metadata information is temporarily lost until the stuck PR is addressed.

PR #2493, opened 2021-03-22
[Triage Metadata] Clean orphaned pending metadata in searchcache polling

KyleJu:
Set up Redis in searchcache and remove no longer PRs and their pending metadata from Redis every 10 minutes.

[Instructions to set up Redis](https://cloud.google.com/memorystore/docs/redis/connect-redis-instance-flex). It should share the same Redis memorystore as webapp. 

## Test
Checked the log in staging using https://github.com/web-platform-tests/wpt-metadata/pull/1068
Screen Shot 2021-03-23 at 12 00 35 AM

### DIFF
--- a/api/query/cache/poll/poll.go
+++ b/api/query/cache/poll/poll.go
@@ -5,9 +5,12 @@
 package poll
 
 import (
+	"context"
 	"net/http"
+	"strconv"
 	"time"
 
+	"github.com/google/go-github/v33/github"
 	"github.com/web-platform-tests/wpt.fyi/api/query"
 	"github.com/web-platform-tests/wpt.fyi/api/query/cache/index"
 	"github.com/web-platform-tests/wpt.fyi/shared"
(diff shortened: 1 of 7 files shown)


KyleJu:
@stephenmcgruer nothing weird in the log so feel free to take a look. I will add some dummy data through https://github.com/web-platform-tests/wpt.fyi/pull/2492

stephenmcgruer:
> @stephenmcgruer nothing weird in the log so feel free to take a look. I will add some dummy data through #2492

Sorry,is this ready for review? (Or do you want a pre-review of something specific?) I wasn't sure as it's still marked draft.

KyleJu:
> > @stephenmcgruer nothing weird in the log so feel free to take a look. I will add some dummy data through #2492
> 
> Sorry,is this ready for review? (Or do you want a pre-review of something specific?) I wasn't sure as it's still marked draft.

Oops sorry about the confusion. I want to do quick staging testing using  #2492. Other than it's ready for a review. I will remove the `draft` once the testing is done

KyleJu:
Turns out 'logger.Errorf` doesn't really work in `logrus`. Logging started showing up after `logger.Infof`

Screen Shot 2021-03-23 at 12 00 35 AM

KyleJu:
@stephenmcgruer Ready for review

api/query/cache/poll/poll.go
```diff
@@ -84,20 +87,77 @@ func wait(start time.Time, total time.Duration) {
 	}
 }
 
-// KeepMetadataUpdated implements updates to metadataMapCached via simple polling every
+// MetadataPollingService performs metadata-related services via simple polling every
 // interval duration.
-func KeepMetadataUpdated(client *http.Client, logger shared.Logger, interval time.Duration) {
-	logger.Infof("Metadata update via polling started")
+func MetadataPollingService(ctx context.Context, logger shared.Logger, interval time.Duration) {
```
stephenmcgruer, line 92: Nit; usually a method name that has some result should be a verb-sentence. This is a noun currently, so its not really clear what happens with the 'MetadataPollingService'.

Perhaps: `StartMetadataPollingService`?

api/query/cache/poll/poll.go
```diff
@@ -84,20 +87,77 @@ func wait(start time.Time, total time.Duration) {
...
+	logger.Infof("Metadata polling service started")
```
stephenmcgruer, line 93: Super-nit: this log is technically inaccurate as the service hasn't started yet, so it could cause confusion if something goes wrong below. I.e. you'd get the log lines:

```
Metadata polling service started
Unable to get GitHub client: foo bar
```

So if you read just the log lines, you might think the service has started :). Probably worth making it near-future tense (`starting`).

api/query/cache/poll/poll.go
```diff
@@ -84,20 +87,77 @@ func wait(start time.Time, total time.Duration) {
...
+
+	if metadataCache != nil {
+		query.MetadataMapCached = metadataCache
+	}
+}
+
+// cleanOrphanedPendingMetadata cleans and removes orphaned pending metadata in Redis.
+func cleanOrphanedPendingMetadata(ctx context.Context, ghClient *github.Client, cacheSet shared.RedisSet, logger shared.Logger) {
+	logger.Infof("Running cleanOrphanedPendingMetadata...")
+	if ghClient == nil {
```
stephenmcgruer, line 125: Optional: perhaps we should skip calling the function in the outer-loop, and require the ghClient be non-null?

stephenmcgruer (approved):
Optional: consider testing `cleanOrphanedPendingMetadata`, as it's logic is non-trivial. (Could also test `keepMetadataUpdated` but that's more trivial logic)

stephenmcgruer:
> Turns out logger.Errorf does not really work in logrus. Logging started showing up after changing logger.Errorf to logger.Infof

Missed his before. This is a bit of a yikes, do we know why this doesn't work?!

Note that you have some `Errorf` left in the PR I think, you'll need to change those then right?

KyleJu:
> > Optional: consider testing cleanOrphanedPendingMetadata, as it's logic is non-trivial. 

I will send out a follow-up PR for tests

> Turns out logger.Errorf does not really work in logrus. Logging started showing up after changing logger.Errorf to logger.Infof
> 
> Missed his before. This is a bit of a yikes, do we know why this doesn't work?!
> 
> Note that you have some `Errorf` left in the PR I think, you'll need to change those then right?

It is related to the implementation of `logrus.StandardLogger()` because the GCP logging isn't set up at this point yet. I will dig into it. Also yup I replaced the last `Errorf`

KyleJu:
Verified /api/metadata/pending endpoint after closing https://github.com/web-platform-tests/wpt-metadata/pull/1068. It works correctly

Issue #2494, opened 2021-03-22
[compat2021] responsive layout for narrow screens


foolip:
https://wpt.fyi/compat2021?feature=summary&embedded when embedded on wpt.fyi in an iframe gets horizontal scrollbars for narrow screens, the breakpoint probably being 800px judging by the width of the graph component.

For wpt.fyi as a whole it has never seemed important to make it work well on small screens, but for this dashboard it would be useful. I would not say it's super important, however.

stephenmcgruer:
FYI, the embedded form has nothing to do with the rest of wpt.fyi; it entirely (afaik) removes all CSS/HTML that is used for wpt.fyi

The problem here is the graph, as noted in the OP. Specifically the fact that Google Charts width is set at draw time (even if one uses `width: 100%` in the chart options), and will not resize dynamically with the page. To fix this, I believe we could:

1. Set https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/components/compat-2021.js#L709 to `"100%"` rather than `800`, and style the parent element accordingly.
2. Add a resize handler to redraw the chart when the page size changes.

We could do just the first one if we only wanted an initial lower-sized load to work and don't care about supporting resize.

PR #2495, opened 2021-03-22
[compat2021] Make the feature graph responsive to narrow layouts

stephenmcgruer:
This PR also:

1. Sets a max-width of 700px for the page (based on Una's old PR)
2. Centers the page.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/2494

### DIFF
--- a/webapp/components/compat-2021.js
+++ b/webapp/components/compat-2021.js
@@ -160,6 +160,10 @@ class Compat2021 extends PolymerElement {
     return html`
       <style>
         :host {
+          display: block;
+          max-width: 700px;
+          /* Override wpt.fyi's automatically injected common.css */
+          margin: 0 auto !important;
           font-family: system-ui, sans-serif;
           line-height: 1.5;
         }
@@ -551,6 +555,8 @@ class Compat2021FeatureChart extends PolymerElement {
              with the JavaScript defined height. */
           height: 350px;
           margin: 0 auto;
+          display: flex;
+          justify-content: center;
         }
 
         paper-dialog {
@@ -622,6 +628,14 @@ class Compat2021FeatureChart extends PolymerElement {
     return 'compat-2021-feature-chart';
   }
 
+  ready() {
+    super.ready();
+
+    window.addEventListener('resize', () => {
+      this.updateChart(this.feature, this.stable);
+    });
+  }
+
   async updateChart(feature, stable) {
     // Our observer may be called before the feature is set, so debounce that.
     if (!feature) {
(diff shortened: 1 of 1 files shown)


stephenmcgruer:
Feel free to play around with https://smcgruer-responsive-gr-dot-wptdashboard-staging.uk.r.appspot.com/compat2021?feature=summary and https://smcgruer-responsive-gr-dot-wptdashboard-staging.uk.r.appspot.com/compat2021?embedded&feature=summary to see how it responds and looks. Maybe the chart is a little small at max-width 700px?

foolip:
With the graph scaling down like vector graphics (/me checks source, yes it's SVG) the text does end up a bit small. If we could slightly increase the text size to look more consistent at 700px and just allow it to be smaller on even narrower screens, I think that'd be OK.

webapp/components/compat-2021.js
```diff
@@ -750,6 +760,30 @@ class Compat2021FeatureChart extends PolymerElement {
       };
     }
 
+    // We draw the chart in two ways, depending on the viewport width. In
+    // 'full' mode the legend is on the right and we limit the chart size to
+    // 700px wide. In 'mobile' mode the legend is on the top and we use all the
+    // space we can get for the chart.
+    //
+    // Google Charts is not responsive, so once drawn the settings are static
```
foolip, line 768: But can one ask it to redo the layout? There's already a resize event listener, and https://stackoverflow.com/questions/41319720/make-google-chart-responsive suggest calling a `drawChart()` which recreates the whole chart. So I guess that's one heavyhanded option.

stephenmcgruer: I don't quite follow what you're asking here. The only way to redo the layout is to `drawChart()`, which is exactly what the resize event listener I added in this PR does. Did this comment mislead you in some way? (Note that it is not near the resize listener when reading the code normally; obviously in GitHub code review view the code is collapsed).

stephenmcgruer:
> With the graph scaling down like vector graphics (/me checks source, yes it's SVG) the text does end up a bit small. If we could slightly increase the text size to look more consistent at 700px and just allow it to be smaller on even narrower screens, I think that'd be OK.

Not sure I interpreted this request correctly, but increased the font size slightly to 14px. Any larger and it overflowed on my screen in a non-pretty way.

foolip (approved):

webapp/components/compat-2021.js
foolip: Yes, I took the comment to mean Google Charts is not responsive (in its layout), and nothing is done to work around that. While it seems like it could be worked around by doing something in a resize listener.

Issue #2280, opened 2020-11-26
Change the default branch to "main"


Hexcles:
https://tools.ietf.org/id/draft-knodel-terminology-02.html

I'm going to do this now that we have relatively few PRs open.

Note: this issue is about the repo of wpt.fyi itself.

Hexcles:
Locally, I've done
```bash
git branch -c master main
git checkout main
git push -u origin main
```
which gives us https://github.com/web-platform-tests/wpt.fyi/tree/main

Hexcles:
Fixed up everything I can find in the code: https://github.com/web-platform-tests/wpt.fyi/commit/f8d9beebe98a4d254d1ae0d0283c4fe7cef3196e

Hexcles:
Looks like both GitHub Actions and Travis went well: https://github.com/web-platform-tests/wpt.fyi/runs/1460905946

I've changed the default branch to "main" but will keep the "master" around for a short while. The previous branch rule for master has been modified to apply to main. I also added a new branch rule to master to prevent push except from admins.

@web-platform-tests/wpt-fyi please run the following command locally. Thanks!
```bash
git branch -m master main
git checkout main
git fetch origin
git branch -u origin/main
```

Hexcles:
Updated the base branch of all open PRs.

I'm closing this issue now. Feel free to reopen if you encounter any issue.

Hexcles:
Update: the `master` branch has been deleted. All existing URLs (or git refs) to the `master` branch should be automatically redirected to the new `main` branch by GitHub.

foolip:
I can confirm the redirects work in the most common case:
https://github.com/web-platform-tests/wpt.fyi/blob/master/CODE_OF_CONDUCT.md

It doesn't work in all cases though:
https://github.com/web-platform-tests/wpt.fyi/commits/master

Hexcles:
> https://github.com/web-platform-tests/wpt.fyi/commits/master

Do we have this link somewhere?

foolip:
Not that I know of, I just tested two kinds of URLs I often end up at. I suspect some specs link to URLs like this for "recent changes" but that's not relevant for wpt.fyi.

PR #2497, opened 2021-03-24
Update three links to point to main branch instead of master, removing the redirect

Schweinepriester:
## Description

On <https://wpt.fyi/results/> I've noticed when clicking on "[`view the search documentation`](https://github.com/web-platform-tests/wpt.fyi/blob/master/api/query/README.md)" you'll be taken to the documentation, but a banner from GitHub distracts you, telling you've been redirected to the default branch. This PR removes the redirect by directly linking to the main branch (which is the default). I've also changed two other occurrences. The redirect was acknowledged in https://github.com/web-platform-tests/wpt.fyi/issues/2280#issuecomment-737269724, but I thought it would be nicer without it.

There are more links containing master linking to other repos. If you want me to I can check and change those too.

## Review Information

Not applicable, code is run as before, no change.

## Changes

Not really applicable. While the files differ widely, it's only a change to each link.

## Requirements

Not applicable.

### DIFF
--- a/api/receiver/client/client.go
+++ b/api/receiver/client/client.go
@@ -50,7 +50,7 @@ func (c client) CreateRun(
 	resultURLs []string,
 	screenshotURLs []string,
 	labels []string) error {
-	// https://github.com/web-platform-tests/wpt.fyi/blob/master/api/README.md#url-payload
+	// https://github.com/web-platform-tests/wpt.fyi/blob/main/api/README.md#url-payload
 	payload := make(url.Values)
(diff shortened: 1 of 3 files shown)


KyleJu:
@Schweinepriester Thanks for the changes! And yes it would be great if you can change [the rest](https://github.com/web-platform-tests/wpt.fyi/search?q=%22blob%2Fmaster%22&type=) as well!

stephenmcgruer (approved):

Schweinepriester:
Thanks, will do the follow up shortly :)

Issue #2202, opened 2020-10-14
Triaging doesn't take effect until minutes after and no way to know when/if done


foolip:
I've been triaging in https://wpt.fyi/results/css/css-flexbox?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=safari%3Afail%20none%28triaged%3Asafari%29 today and found that I was slowed down a lot by not knowing what I had already done, or whether what I had done had taken effect or not.

Triaging opens a pull request which is automatically merged, but even after that, furiously reloading the page doesn't seem to always bring in the new data. Eventually, after perhaps 10 minutes, it's always settled, however.

What I'd like to happen:
- Triage a bug
- Show a spinner or something to indicate that saving is in progress (no link to the PR necessary IMHO)
- Once done, automatically reload to filter out the tests I just triaged.

foolip:
I see one can also get into an in-between state, where the bugs icons are shown, and yet those tests aren't filtered out. Screenshot:
![image](https://user-images.githubusercontent.com/498917/96052354-a95bd200-0e7d-11eb-8653-540c5fc9c1c2.png)

My query was `safari:fail none(triaged:safari)` which should have filtered out the red cells with a bug icon in the Safari column.

PR #2498, opened 2021-03-25
[Triage Metadata] Add pending metadata to UI

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2202. Plumb /api/metadata/pending to the UI. 

`/api/metadata/pending` is invoked in constructor() or when users triage new metadata, while `/api/metadata` is invoked only when `products` changes

## Test
Copy and paste session cookies from staging.wpt.fyi and triage a metadata.  The icon(s) should show up instantly

### DIFF
--- a/webapp/components/test/wpt-metadata.html
+++ b/webapp/components/test/wpt-metadata.html
@@ -51,8 +51,8 @@ suite('<wpt-metadata>', () => {
       appFixture.pendingMetadata = {
         '/bar.html': [
           {
-            url: 'bug1',
-            product: 'chrome'
+            url: 'bug2',
+            product: 'safari'
           }
         ]
       };
--- a/webapp/components/test/wpt-metadata.html
+++ b/webapp/components/test/wpt-metadata.html
@@ -25,12 +25,14 @@ suite('<wpt-metadata>', () => {
   suite('compute display metadata', () => {
     test('null case', () => {
       appFixture.path = null;
-      appFixture.metadata = {};
+      appFixture.pendingMetadata = {};
+      appFixture.mergedMetadata = {};
       assert.equal(appFixture.displayedMetadata, undefined);
     });
     test('simple case', () => {
       appFixture.searchResults = [{ test: '/foo.html' }];
-      appFixture.metadata = {
+      appFixture.pendingMetadata = {};
+      appFixture.mergedMetadata = {
         '/foo.html': [
           {
             url: 'bug1',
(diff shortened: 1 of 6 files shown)


webapp/components/wpt-amend-metadata.js
```diff
@@ -361,6 +365,9 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
       .then(text => {
         this.prLink = text;
         this.prText = 'Created PR: ' + text;
+        // Create an observeable change to trigger /api/metadata/pending
+        // reload.
+        this.triageObserver = !this.triageObserver;
```
stephenmcgruer, line 370: If what we're trying to do is inform some ancestor of ours that an event has occurred (metadata being triaged), we should use dispatchEvent with a `computed=true, bubbles=true` event. Our ancestor can then have an event listener that will receive the events, and not have to hack this two-way binding all the way down!

See:

https://polymer-library.polymer-project.org/3.0/docs/devguide/events#custom-events
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/dispatchEvent

webapp/components/wpt-metadata.js
```diff
@@ -170,13 +184,39 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
...
-        this.metadata = metadata;
+      window.fetch(url).then(r => r.json()).then(mergedMetadata => {
+        this.mergedMetadata = mergedMetadata;
       })
     );
   }
 
-  computTestResultSet(searchResults) {
+  // loadPendingMetadata is called when wpt-metadata.js is initialized
+  // through constructor() or when users triage new metadata, unlike loadMergedMetadata().
```
stephenmcgruer, line 194: This feels like it could cause triaged metadata UI to disappear when in a triaging session? E.g.:

```
1. Triage test foo.html. A pending PR is created, it is added to /api/metadata/pending, and it shows up in the UI
2. Continuing triaging. In the background, the created PR is merged because its good.
3. The searchcache periodic sync runs, and cleans up the pending PR metadata
4. The user triages another test, bar.html. The metadata icon for foo.html disappears ?!
```

Is this correct? If so, should we be thinking about refreshing `/api/metadata` as well?

stephenmcgruer (commented):
Two big comments, so starting with those and then will look at lower-level stuff once we've resolved them.

webapp/components/wpt-metadata.js
KyleJu: For step 4, When user triages another test, webapp will force fetch the latest copy of the wpt-metadata repo, thus icons will remain. However, if user refreshes the page, there is a time period where icons will disappear. Refreshing `/api/metadata` won't help because the in-memory copy could still be old. Let me think about it more...

KyleJu: To solve this issue, I will only remove a pending PR from /api/pending/metadata 10 minutes after it is merged

KyleJu: Will send out a fix in a separate PR

KyleJu:
@foolip Ready for review

webapp/components/wpt-metadata.js
KyleJu: https://github.com/web-platform-tests/wpt.fyi/pull/2515

webapp/components/test/wpt-metadata.html
```diff
@@ -44,9 +46,68 @@
...
+          {
+            url: 'bug1',
+            product: 'chrome'
+          }
+        ]
+      };
+      appFixture.mergedMetadata = {
+        '/foo.html': [
+          {
+            url: 'bug1',
```
foolip, line 62: Should a different url and product be used here, to rule out the test passing if pending and merged are mixed up?

webapp/components/wpt-metadata.js
```diff
@@ -170,13 +184,39 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
...
+  loadPendingMetadata() {
+    const url = new URL('/api/metadata/pending', window.location);
```
foolip, line 196: This appears to only be for turning a relative URL into an absolute one, but `fetch(url)` already resolves relative URLs. Does just `fetch('/api/metadata/pending')` work?

webapp/views/wpt-results.js
```diff
@@ -365,6 +367,16 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     ];
   }
 
+  connectedCallback() {
+    super.connectedCallback();
+    this.addEventListener('triagemetadata', this.handleReloadPendingMetadata.bind(this));
+  }
+
+  disconnectedCallback() {
+    this.removeEventListener('triagemetadata', this.handleReloadPendingMetadata.bind(this));
```
foolip, line 376: This `removeEventListener` call won't work, because calling `this.handleReloadPendingMetadata.bind(this)` twice will return different functions, and `removeEventListener` needs exactly the same arguments as `addEventListener` to work.

If removing the event listener is important, you'll need to store the callback (`this.handleReloadPendingMetadata.bind(this)`) somewhere are pass it in here.

foolip (approved):
LGTM % nits. I haven't tried if the UI works differently, because I think I'd need to make real metadata changes for that?

webapp/components/wpt-metadata.js
KyleJu: You are right. `fetch will load the path relative to the url you have opened if you don't begin with / or specify absolute url.` However, all other instances of `fetch` in this code base uses absolute path. Lets keep it that way for consistency.

webapp/views/wpt-results.js
KyleJu: Good catch!

KyleJu:
> LGTM % nits. I haven't tried if the UI works differently, because I think I'd need to make real metadata changes for that?

I tested on staging :)

foolip:
@KyleJu doesn't staging still use the real wpt-metadata repo? That's why I didn't want to try it.

KyleJu:
> @KyleJu doesn't staging still use the real wpt-metadata repo? That's why I didn't want to try it.

Yes I tested it with real data and deleted it after

PR #2500, opened 2021-03-26
[compat2021] Rename browsers to browser engines

stephenmcgruer:


foolip:
As I think I mentioned somewhere else, an issue or a meeting, I don't think we should do this renaming unless another browser vendor insists on it. https://wpt.fyi/compat2021?feature=summary is clearer about what we're testing for the most part, and if differences between Chrome and Edge surface, we could fix it by either addressing the differences, or worst case combining the results of Chrome and Edge runs.

Issue #2215, opened 2020-10-22
[Triage Metadata] Verify the validity of a triage metadata URL on the frontend


KyleJu:
When users click on `triage` button, there should be lint check that verifies the validity of triage metadata URLs and prompt users if they are invalid on the spot. Also See https://github.com/web-platform-tests/wpt-metadata/issues/591.

foolip:
Would this mean fixing up the URLs automatically when possible?

KyleJu:
I think I will prompt users when a URL is invalid. For example, showing a text that says `Your URL is invalid` and user isn't able to traige

foolip:
Alright, so maybe we can also try to prepend 'https://' and checking if that's a valid URL, so that things like "crbug.com/1234" still work.

PR #2538, opened 2021-05-18
[Triage Metadata] Validate URL prefix

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2215. This change will check if an URL starts `http(s)://`. 

This will also help avoid manually normalizing URLs, e.g. https://github.com/web-platform-tests/wpt-metadata/pull/777

### DIFF
--- a/webapp/components/wpt-amend-metadata.js
+++ b/webapp/components/wpt-amend-metadata.js
@@ -163,7 +163,7 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement))) 
             <div class="metadata-entry">
               <img class="browser" src="[[displayMetadataLogo(node.product)]]">
               :
-              <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
+              <paper-input label="Bug URL" value="{{node.url}}" auto-validate pattern="^((http|https):\/\/)[\w.-]+" error-message="Missing http(s):// prefix" autofocus></paper-input>
             </div>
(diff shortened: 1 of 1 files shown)


KyleJu:
@foolip I am reinforcing the `https://` to determine a valid URL. Alternatively, I could use `new URL()` and see if it throws `TypeError` to determine the validity

foolip:
@KyleJu I think the ideal behavior here would be:
- Trim leading and trailing whitespace, call the result `url`
- If `new URL(url)` doesn't throw, it's all good.
- If it did throw, try `new URL('https://' + url)`. If that doesn't throw, prepend 'https://'
- If it still can't be parsed as a URL, treat it as invalid, asking the user to fix.

foolip:
@KyleJu got a notification about this. Is it clear what next steps are here, does my suggestion in https://github.com/web-platform-tests/wpt.fyi/pull/2538#issuecomment-843576746 make sense?

KyleJu:
> @KyleJu got a notification about this. Is it clear what next steps are here, does my suggestion in [#2538 (comment)](https://github.com/web-platform-tests/wpt.fyi/pull/2538#issuecomment-843576746) make sense?

Yea it makes sense. I was trying to test https://github.com/web-platform-tests/wpt.fyi/issues/2560 on this. It should create a new deployment when I close and open but it didn't. Something isn't working

foolip:
@KyleJu did you mean to close this? This was never landed, right?

Issue #2560, opened 2021-06-10
Staging deployment missing


KyleJu:
For the recent PRs, I don't see any staging deployment. The deployment on staging [wpt.fyi](https://staging.wpt.fyi/about) seems to be behind as well.

e.g. https://github.com/web-platform-tests/wpt.fyi/pull/2557

KyleJu:
@foolip I will look into it but tag you in case you have experience with it before

foolip:
I haven't tinkered with the staging deployment code, but I have set up CI for deploying to AppEngine elsewhere so if you're stumped I could take a look.

First off, did the deploy step run in CI at all? If yes, how did it fail?

KyleJu:
Tested in #2561. Deploy step doesn't run in CI at all

KyleJu:
On the https://travis-ci.com/github/web-platform-tests/wpt.fyi/requests, I see `
Could not authorize build request for web-platform-tests/wpt.fyi.` for most of the missing deployment

KyleJu:
@foolip [“Could not authorize build request”, usually means that the account’s subscription expired or that it ran out of trial builds](https://docs.travis-ci.com/user/common-build-problems/). Not sure how the billing is setup?

foolip:
The linked billing account seems fine. Have you checked logs in GCP to see if they say something interesting.

Another idea is maybe our service account keys need updating, do they really have infinite lifetime?

KyleJu:
Checked GCP but didn't see anything interesting. There is no log for the missing deployment.

How do we update the service account keys? Do I have access to it.

Also I see the pricing change article here - https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing. it looks like we have to choose a pricing plan?

foolip:
Oh wait, I misunderstood. It's Travis CI saying that build can't be started, not a step within that has GCP saying the GCP deploy build can't be started.

I think we should migrate from Travis CI to GitHub Actions to sidestep the whole problem.

KyleJu:
[In latest GitHub actions deployment}(https://github.com/web-platform-tests/wpt.fyi/runs/2948394952), Webapp is deployed successfully and it confirmed the fix for https://github.com/web-platform-tests/wpt.fyi/issues/2411. The only issue with webapp is the version name. It should incorporate WPT SHA instead of [refs-heads-main](https://pantheon.corp.google.com/appengine/versions?project=wptdashboard-staging&serviceId=default)

[...]

PR #2566, opened 2021-06-14
fix: deno canary is experimental

lucacasonato:


KyleJu (approved):
LGTM. I don't think we can verify it on staging since the deployment is broken. Lets merge it once https://github.com/web-platform-tests/wpt.fyi/issues/2560 is fixed

PR #2567, opened 2021-06-15
Migrate from Travis CI to GitHub Actions

foolip:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2560. 

To verify, compare it to [travis build](https://github.com/web-platform-tests/wpt.fyi/runs/2652653294)


foolip:
@KyleJu if you want to build on this, here's an example from another repo of having a JSON file with secrets stored as a variable in GitHub Actions:
https://github.com/foolip/mdn-bcd-collector/blob/f19ffe3e18bef66576d34bbeff8bd0f44c103adb/.github/workflows/push.yml#L38-L40

KyleJu:
@foolip I think I am almost there in the latest run but im stuck on
`openssl aes-256-cbc -K $encrypted_2fd045226a67_key -iv $encrypted_2fd045226a67_iv -in client-secret.json.enc -out client-secret.json -d`

It looks like I need `$encrypted_2fd045226a67_key` and `$encrypted_2fd045226a67_iv` as secrets in GitHub actions. I cannot access them in [Travis](https://travis-ci.com/github/web-platform-tests/wpt.fyi/settings). They are used for 

`gcloud auth activate-service-account --key-file /home/user/wpt.fyi/client-secret.json`

Any idea about this encryption key and auth token?

foolip:
We won't be able to keep the existing `client-secret.json.enc`, instead we'll need to generate a new one, put that in a `CLIENT_SECRETS_JSON` secret, and do `echo "$CLIENT_SECRETS_JSON" > client-secret.json` as one of the steps.

foolip:
It's used here: `gcloud auth activate-service-account --key-file $(WPTD_PATH)client-secret.json`

foolip:
I created a new service account (github-actions@wptdashboard-staging.iam.gserviceaccount.com) with the same role as the existing one (Travis AppEngine Deployment) and put it in `GCLOUD_KEY_FILE_JSON`.

KyleJu:
cd26a07 Looks good with the PR branch deployment. I will incorporate the main branch now

KyleJu:
@foolip I think I got the deployment for PR to work, although I have some questions with the TODO in the file:
 - `DOCKER_INSTANCE` should be `"wptd-dev-${RANDOM}"`, where ${RANDOM} is an expression. I cannot get it to work in GitHub actions
 - `docker exec -t "${DOCKER_INSTANCE}" make go_build;` runs before every deployment in travis. Is that necessary?
 - Do we need to set `set -e`?
 - `puppeteer_chrome_test` step has been failing for a long time. Should we move it?

foolip:
@KyleJu, sweet, almost there!

> `DOCKER_INSTANCE` should be `"wptd-dev-${RANDOM}"`, where ${RANDOM} is an expression. I cannot get it to work in GitHub actions

I think`${RANDOM}` is https://tldp.org/LDP/abs/html/randomvar.html, a feature of Bash. That code has to be evaluated by shell, can you try `shell: bash` as in https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions?

> `docker exec -t "${DOCKER_INSTANCE}" make go_build;` runs before every deployment in travis. Is that necessary?

I think this depends on what exactly the `gcloud app deploy` step does, whether or not it will use [Cloud Build](https://cloud.google.com/build/docs). From GCP logs it looks like Cloud Build is involved, and that means that build steps in the CI steps (GitHub Actions) _could_ be unnecessary if they're also done when building the Docker image.

But I can't see where `make go_build` happens if not there. If you remove that step, does the deploy still succeed?

> Do we need to set `set -e`?

That's for failing early on errors, and I'd use that only in a shell script that invokes multiple command. I wouldn't bother in a single short step inlined into the CI config.

> `puppeteer_chrome_test` step has been failing for a long time. Should we remove it?

At least it shouldn't be running in CI if it's not expected to pass. What does it test, and is it important to avoid regressions in that thing?

KyleJu:
> @KyleJu, sweet, almost there!
> 
> > `DOCKER_INSTANCE` should be `"wptd-dev-${RANDOM}"`, where ${RANDOM} is an expression. I cannot get it to work in GitHub actions
> 
> I think`${RANDOM}` is https://tldp.org/LDP/abs/html/randomvar.html, a feature of Bash. That code has to be evaluated by shell, can you try `shell: bash` as in https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions?
> 
I got it to work :)

> > `docker exec -t "${DOCKER_INSTANCE}" make go_build;` runs before every deployment in travis. Is that necessary?
> 
> I think this depends on what exactly the `gcloud app deploy` step does, whether or not it will use [Cloud Build](https://cloud.google.com/build/docs). From GCP logs it looks like Cloud Build is involved, and that means that build steps in the CI steps (GitHub Actions) _could_ be unnecessary if they're also done when building the Docker image.
> 
> But I can't see where `make go_build` happens if not there. If you remove that step, does the deploy still succeed?

By each deployment, I meant that `docker exec -t "${DOCKER_INSTANCE}" make go_build;` is run three times for `build webapp`, `build processor' and `build searchcahe`, but I think it should only run once in the setup. Looking at [a Travis example](https://travis-ci.com/github/web-platform-tests/wpt.fyi/jobs/505768056#L345), the only difference between each installation is a random `DOCKER_INSTANCE`. I don't see any reason why we have to run it more than one time but I don't have enough knowledge on it.

> 
> > Do we need to set `set -e`?
> 
> That's for failing early on errors, and I'd use that only in a shell script that invokes multiple command. I wouldn't bother in a single short step inlined into the CI config.
> 
> > `puppeteer_chrome_test` step has been failing for a long time. Should we remove it?
> 
> At least it shouldn't be running in CI if it's not expected to pass. What does it test, and is it important to avoid regressions in that thing?

It is not expected to pass and all tests have been failing due to infra errors. I will remove them for now

KyleJu:
I did remove a few unnecessary steps but it looks like we have a huge speedup? For comparable builds, it takes [18 minutes on travis](https://github.com/web-platform-tests/wpt.fyi/runs/2652653294) but it takes 2 minutes here. The speedup is from removing redundant installation steps and `puppeteer_chrome_test`

KyleJu:
I see errors after switching `push` on `github-actions` branch to `pull_request`
`
fatal: ambiguous argument 'refs/pull/2567/merge..HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
`

KyleJu:
> I see errors after switching `push` on `github-actions` branch to `pull_request`
> 
> `fatal: ambiguous argument 'refs/pull/2567/merge..HEAD': unknown revision or path not in the working tree. Use '--' to separate paths from revisions, like this: 'git <command> [<revision>...] -- [<file>...]'`

It is failing on `git diff --name-only ${GITHUB_REF}..HEAD | egrep -v "${EXCLUSIONS}" | egrep "${APP_DEPS_REGEX}" ||` to detect if there are changes that should be deployed. `GITHUB_REF` is `refs/heads/github-actions` in this case. @foolip Isn't `refs/heads/github-actions` HEAD already? Or it points to `HEAD~` in the GitHub action?

Also I am unsure why `${GITHUB_REF}` was 'refs/pull/2567/merge`. Let me dig around

foolip:
Perhaps we can use https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to trigger deploy only on certain paths and avoid the problem entirely?

foolip:
As for the error, the orginal setup seems to have depended on how Travis fetches the commit to test, and that the `refs/heads/pull/1234/merge` ref is fetched under that name. With GitHub Actions the same merge commit will be fetched, but maybe it's not named. One could instead compare `HEAD` and `HEAD^1` to see what's changed in the PR.

foolip:
(`refs/pull/2567/merge` is the name given to merge commit generated by GitHub merging the PR branch into the current `main` branch.)

KyleJu:
> As for the error, the original setup seems to have depended on how Travis fetches the commit to test, and that the `refs/pull/2567/merge` ref is fetched under that name. With GitHub Actions the same merge commit will be fetched, but maybe it's not named. One could instead compare `HEAD` and `HEAD^1` to see what's changed in the PR.

After some experiments, I finally resolved this issue by increasing the fetch-depth to 2. It defaults to fetching only one commit by actions/checkout@v2. `HEAD` and `HEAD^1` work very well in this case as well.

KyleJu:
> Perhaps we can use https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onpushpull_requestpaths to trigger deploy only on certain paths and avoid the problem entirely?

I think it makes more sense that we only create deployment with a PR, rather a `push` to any branches

KyleJu:
@foolip This is ready for review :)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+        with:
+          fetch-depth: 2
+
+      - name: secrets
+        run: echo "$GCLOUD_KEY_FILE_JSON" > client-secret.json
+        env:
+         GCLOUD_KEY_FILE_JSON: ${{ secrets.GCLOUD_KEY_FILE_JSON }}
+
+      - name: Set DOCKER_INSTANCE
+        run: echo "DOCKER_INSTANCE=wptd-dev-$(echo $RANDOM)" >> $GITHUB_ENV
```
foolip, line 25: Interesting, I guess `$(echo $RANDOM)` creates a subshell where just `$RANDOM` does not.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: |
+          docker pull "${DOCKER_IMAGE}"
+          bash ./util/docker-dev/run.sh -d -q
+
+      - name: installation
+        run: docker exec -t "${DOCKER_INSTANCE}" make go_build;
+
+      # Set -f for main branch.
+      - name: set deployment flag
+        if: ${{ github.ref == 'refs/heads/main' }}
```
foolip, line 37: What will `github.ref` be for a pull request from a fork with the changes on the `main` branch? If it is also `refs/heads/main` then there's a potential problem here. The fork shouldn't have access to secrets, but just going down this path is wrong.

A simple way to avoid this would be to also check `github.event_name`.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: echo "FORCE_DEPLOYMENT=-f" >> $GITHUB_ENV
+
+      - name: build webapp
```
foolip, line 40:
```suggestion
      - name: deploy webapp
```

I would suggest this edit below as well. I first thought this would only build some stuff locally in CI, not deploy anything.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,55 @@
...
+        run: ./util/deploy-staging.sh "${FORCE_DEPLOYMENT}" webapp/web/app.staging.yaml
+
+      - name: Run go_large_test in main
```
foolip, line 43: Should these tests really be run as part of the deploy job, such that if they fail the webapp deployment (above) happens but the rest (below) does not?

Perhaps this can be split into its own workflow and only run on PRs? Alternatively, if passing certain tests is important before deploying, it would be good to make that more explicit by running the tests first.

foolip (commented):
I can't approve this since I started the PR, but everything LGTM % nits now. Feel free to land after addressing whichever of my last comments you think are worth addressing.

.github/workflows/deploy.yml
KyleJu: Good catch. I will check `${{ github.event_name == 'push' && github.ref == 'refs/heads/main' }}` instead

.github/workflows/deploy.yml
KyleJu: `go_large_test` appears to be the webdriver integration tests that need staging deployment. In the [origin travis setup](https://github.com/web-platform-tests/wpt.fyi/blob/207813b3ed18bae81068934caa478daffd782d36/.travis.yml#L61), they only run in the deployment of main branch. Maybe it is too costly to run in PRs?

KyleJu (approved):

KyleJu:
Ready to merge. Let's see how it does on the main branch

PR #2582, opened 2021-07-03
fix: render missing tests gray

### LINKED ISSUE #2581: "0 / 0" cells should not be red
In https://twitter.com/dunxen/status/1410558940656570372, @duncandean noted in reaction to this image:
image

> I suppose 0/0 should be all green or some undefined colour Face with tears of joy

I agree this it's weird to treat 0 / 0 as a failure (red) when actually it's missing. It used to be that this only happened when there was some problem since all browsers should run all tests, but with Deno results it's going to be much more common since only a subset of WPT is run.

It can happen for browsers too, however, like here:
https://wpt.fyi/results/accessibility/crashtests?run_id=5684579232907264&run_id=5683320908152832&run_id=5735626798268416&run_id=5742689167343616

Drilling into one of these "0 / 0" cells usually (always?) leads to a gray "MISSING" cell.

I think we should hoist this gray "MISSING" and show that anywhere there's currently going to be a red "0 / 0" cell.

lucacasonato:
Before:
image

After:
image

Closes #2581

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -724,12 +724,16 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
 
       return `delta ${delta > 0 ? 'positive' : 'negative'}`;
     } else {
-      // Non-diff case: total=0 -> 'none'; path='/' -> 'top';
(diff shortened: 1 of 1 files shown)


KyleJu:
Thanks for the fix1

KyleJu (approved):

Issue #2585, opened 2021-07-07
[GitHub actions] Could not read json file /home/user/wpt.fyi/client-secret.json


KyleJu:
Staging Deployment failed in https://github.com/web-platform-tests/wpt.fyi/pull/2582 due to following errors:

ERROR: (gcloud.auth.activate-service-account) Could not read json file /home/user/wpt.fyi/client-secret.json: Expecting value: line 2 column 1 (char 1)
make: *** [Makefile:238: gcloud_login] Error 1
Error: Process completed with exit code 2.

This is the first occurrence since the migration. This issue should be prioritized if it happens more often

lucacasonato:
@KyleJu I think it happened because the staging run tried to run on my PR, even though I don't have access to the secrets in this repo.

KyleJu:
Aha you are right. fork shouldn't be able to access to the secret. @foolip I guess this is expected?

foolip:
@KyleJu that's expected, yes, but it's related to a comment I made on another PR. We should never go down the deploy code path in this case, some extra condition is needed to avoid that.

PR #2587, opened 2021-07-09
[GitHub Actions] Check if PRs are from a fork

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/2585. Do not run the job if a PR is from a fork

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,7 +6,7 @@ on:
   pull_request:
 jobs:
   deploy-staging:
-    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+    if: ${{ github.repository == 'web-platform-tests/wpt.fyi' }}
     name: Deploy staging.wpt.fyi
     runs-on: ubuntu-latest
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -6,6 +6,7 @@ on:
   pull_request:
 jobs:
   deploy-staging:
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
```
foolip, line 9: Will this also work when this workflow is run because something was pushed to the `main` branch? I suspect it won't. If it should, then something that hardcodes the official repo like https://github.com/web-platform-tests/wpt/pull/29195/files might be needed.

KyleJu: Good catch!

KyleJu:
#2595 #2594 are the two examples that developers tend to fork the repo. It is kind of troublesome since we have no better way to verify UI.

KyleJu:
@foolip friendly ping :)

foolip (approved):

PR #2594, opened 2021-07-13
Link to the GitHub source for the current commit, in addition to master

### LINKED ISSUE #2593: "View source on GitHub" doesn't link to the current PR or commit
The "view source on GitHub" link always links to the current test in the `master` branch of [web-platform-tests/wpt](https://github.com/web-platform-tests/wpt), even when the test results being shown are from a pull request or from a commit that was selected explicitly with the `sha` search parameter. Having a link for the current commit as well as one for the master branch might also help reduce confusion when one of the browsers in the default product set hasn't had runs for some time.

andreubotella:
## Description
Changes the "View source on GitHub" link to have two links instead: one to the master branch, as before, and one that links directly to the current commit. This is useful because the test results shown might not be from the master branch: they could be from a pull request, from a historical commit, or they could be from the last commit that has had runs from all selected browsers, which might not be the current HEAD of the master branch.

Closes #2593.

## Review Information
TODO

## Changes
TODO

## Requirements
TODO

### DIFF
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -92,17 +92,20 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
           <div class="links">
             <ul>
               <li>
-                <a href\$="https://github.com/web-platform-tests/wpt/blob/master[[path]]" target="_blank">View source on GitHub</a></li>
+                View source on GitHub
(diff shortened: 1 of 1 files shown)


KyleJu:
Thanks for fixing the indentation as well :)

KyleJu (approved):

KyleJu:
The change should be available on https://staging.wpt.fyi/results/ shortly and available in prod in the next release

PR #2641, opened 2021-09-20
Fix link to "Meta qualities" in structured search docs

foolip:


KyleJu (approved):
LGTM. The deployment failure is caused by an auth issue

PR #2668, opened 2021-10-28
[compat2021] Support use_webkitgtk query parameter

stephenmcgruer:
When set, this parameter switches the compat2021 page to use WebKitGTK
results rather than Safari. This allows showing up-to-date results for
WebKit even in the absence of newer STP versions being available on
wpt.fyi.


foolip:
@KyleJu will there be a new prod release soon including this change?

KyleJu:
> @KyleJu will there be a new prod release soon including this change?

I will do a release next Monday :)

foolip:
Thanks Kyle!

Issue #1491, opened 2019-09-17
Add support for arbitrary labeling of tests / results.


lukebjerring:
This represents part of the design for the wpt-metadata repository.
https://docs.google.com/document/d/1oWYVkc2ztANCGUxwNVTQHlWV32zq6Ifq9jkkbYNbSAg/edit#heading=h.mv91jownsl1e

- Add a format for appending labels to wpt-metadata META.yml files.
- Add UI for displaying the labels, and for appending new labels

lukebjerring:
This relates to #1489, in that there are meta flags in the contents of test files that are a different way of denominating the part of a spec that a test applies to (not just the spec as a whole).

This feature is blocking a feature request from the CSSWG (surfacing the flags inside a test as labels of the tests that we can filter the results by). In theory, once we can label arbitrarily, we can also write a bot that scrapes the flags metadata inside a test file and updates the labels in wpt-metadata.

KyleJu:
This feature has been implemented

PR #2678, opened 2021-11-09
Add support for arbitrary labeling of tests  

KyleJu:
Add test labelling support for https://github.com/web-platform-tests/wpt.fyi/issues/1491.

- Add an optional `label` entry in Metadata `Link`.
- Add the ability to search triaged tests based on its `labels`. The label exists at the test-level (that is, metadata
that has no product associated with it), regardless of browsers. One test can associate with multiple labels.

### How to test
On the [staging instance](https://metadata-label-dot-wptdashboard-staging.uk.r.appspot.com/results/css/css-align/gaps?q=testlabel%3Aexample-label1&run_id=5312968510144512), search `label:example-label1`, which should show a dummy data `gap-animation-001.html`

### Follow-up
Support multiple labels

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -344,7 +344,7 @@ regardless of browsers.
 
     {"testlabel": label}
 
-Where label is a string.
+Where label is a string and case-insensitive.
 
  E.g.
 
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -337,6 +337,21 @@ Search triaged Chrome tests -
 
     chrome:pass and triaged:chrome
 
+#### testlabel
+
+`testlabel` query atoms perform a search for tests that have a matching metadata label,
+regardless of browsers.
+
+    {"testlabel": label}
+
+Where label is a string.
+
+ E.g.
+
+Search triaged tests with a label interop2020 -
+
+    testlabel:interop2020
+
 #### is
 
 `is` query atoms perform a search for tests that possess some meta quality.
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -337,12 +337,12 @@ Search triaged Chrome tests -
 
     chrome:pass and triaged:chrome
 
-#### testlabel
+#### label
 
-`testlabel` query atoms perform a search for tests that have a matching metadata label,
+`label` query atoms perform a search for tests that have a matching metadata label,
 regardless of browsers.
 
-    {"testlabel": label}
+    {"label": label}
 
 Where label is a string and case-insensitive.
 
@@ -350,7 +350,7 @@ Where label is a string and case-insensitive.
 
 Search triaged tests with a label interop-2022:
 
-    testlabel:interop-2022
+    label:interop-2022
 
 #### is
 
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -350,7 +350,7 @@ Where label is a string.
 
 Search triaged tests with a label interop-2022:
 
-    testlabel:interop2020
+    testlabel:interop-2022
 
 #### is
 
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -348,7 +348,7 @@ Where label is a string.
 
  E.g.
 
-Search triaged tests with a label interop2020 -
+Search triaged tests with a label interop-2022:
 
     testlabel:interop2020
 
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -316,8 +316,8 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return Or{cq}
 }
 
-// AbstractTestLabel is represents the root of a testlabel query, which matches Metadata
-// labels to a string.
+// AbstractTestLabel represents the root of a testlabel query, which matches test-level metadata
+// labels to a searched label.
 type AbstractTestLabel struct {
 	Label           string
 	metadataFetcher shared.MetadataFetcher
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -316,6 +316,30 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	return Or{cq}
 }
 
+// AbstractTestLabel is represents the root of a testlabel query, which matches Metadata
+// labels to a string.
+type AbstractTestLabel struct {
+	Label           string
+	metadataFetcher shared.MetadataFetcher
+}
+
+// BindToRuns for AbstractTestLabel fetches test-level metadata; it is independent of test runs.
+func (t AbstractTestLabel) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
+	if t.metadataFetcher == nil {
+		t.metadataFetcher = searchcacheMetadataFetcher{}
+	}
+
+	includeTestLevel := true
+	// Passing []shared.TestRun{} means that we want test-level issues.
+	metadata, _ := shared.GetMetadataResponse([]shared.TestRun{}, includeTestLevel, logrus.StandardLogger(), t.metadataFetcher)
+	metadataMap := shared.PrepareTestLabelFilter(metadata)
+
+	return TestLabel{
+		Label:    t.Label,
+		Metadata: metadataMap,
+	}
+}
+
 // MetadataQuality represents the root of an "is" query, which asserts known
 // metadata qualities to the results
 type MetadataQuality int
@@ -906,6 +930,26 @@ func (l *AbstractLink) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// UnmarshalJSON for AbstractTestLabel attempts to interpret a query atom as
+// {"testlabel":<label string>}.
+func (t *AbstractTestLabel) UnmarshalJSON(b []byte) error {
+	var data map[string]*json.RawMessage
+	if err := json.Unmarshal(b, &data); err != nil {
+		return err
+	}
+	labelMsg, ok := data["testlabel"]
+	if !ok {
+		return errors.New(`Missing testlabel pattern property: "testlabel"`)
+	}
+	var label string
+	if err := json.Unmarshal(*labelMsg, &label); err != nil {
+		return errors.New(`Missing testlabel pattern property "label" is not a string`)
+	}
+
+	t.Label = label
+	return nil
+}
+
 // UnmarshalJSON for AbstractTriaged attempts to interpret a query atom as
 // {"triaged":<browser name>}.
 func (t *AbstractTriaged) UnmarshalJSON(b []byte) error {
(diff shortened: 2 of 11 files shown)


KyleJu:
Staging instance - https://metadata-label-dot-wptdashboard-staging.uk.r.appspot.com]

foolip:
On the staging instance, I see a link to http://example.org. Is that unrelated metadata, or is it because of the `example-label1` label?

api/query/README.md
```diff
@@ -337,6 +337,21 @@ Search triaged Chrome tests -
 
     chrome:pass and triaged:chrome
 
+#### testlabel
```
foolip, line 340: Would using just "label" as the query atom run into some problem? Is this to avoid ambiguity with the product labels?

api/query/README.md
```diff
@@ -337,6 +337,21 @@ Search triaged Chrome tests -
...
+`testlabel` query atoms perform a search for tests that have a matching metadata label,
+regardless of browsers.
+
+    {"testlabel": label}
+
+Where label is a string.
+
+ E.g.
+
+Search triaged tests with a label interop2020 -
```
foolip, line 351: Let's go with the label I think we'll use as the example:

```suggestion
Search triaged tests with a label interop-2022:
```

api/query/README.md
```diff
@@ -337,6 +337,21 @@ Search triaged Chrome tests -
...
+
+    testlabel:interop2020
```
foolip, line 353:
```suggestion
    testlabel:interop-2022
```

api/query/atoms.go
```diff
@@ -316,6 +316,30 @@ func (t AbstractTriaged) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
...
+}
+
+// BindToRuns for AbstractTestLabel fetches test-level metadata; it is independent of test runs.
+func (t AbstractTestLabel) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
+	if t.metadataFetcher == nil {
+		t.metadataFetcher = searchcacheMetadataFetcher{}
+	}
+
+	includeTestLevel := true
+	// Passing []shared.TestRun{} means that we want test-level issues.
```
foolip, line 333: Thanks for the comment!

api/query/atoms.go
```diff
@@ -906,6 +930,26 @@ func (l *AbstractLink) UnmarshalJSON(b []byte) error {
...
+	if err := json.Unmarshal(b, &data); err != nil {
+		return err
+	}
+	labelMsg, ok := data["testlabel"]
+	if !ok {
+		return errors.New(`Missing testlabel pattern property: "testlabel"`)
+	}
+	var label string
+	if err := json.Unmarshal(*labelMsg, &label); err != nil {
+		return errors.New(`Missing testlabel pattern property "label" is not a string`)
```
foolip, line 946: This says "label" and not "testlabel", is that intentional?

api/query/atoms_test.go
```diff
@@ -25,76 +25,76 @@ func TestStructuredQuery_empty(t *testing.T) {
 func TestStructuredQuery_missingRunIDs(t *testing.T) {
 	var rq RunQuery
 	err := json.Unmarshal([]byte(`{
-		"query": {
-			"pattern": "/2dcontext/"
-		}
-	}`), &rq)
+        "query": {
```
foolip, line 28: Did the indention need to change here, or could that be done separately?

api/query/atoms_test.go
```diff
@@ -462,16 +462,51 @@ func TestStructuredQuery_triaged(t *testing.T) {
 		}}, rq)
 }
 
+func TestStructuredQuery_testlabel(t *testing.T) {
+	var rq RunQuery
+	err := json.Unmarshal([]byte(`{
+        "run_ids": [0, 1, 2],
+        "query": {
+            "exists": [{
+                "testlabel": "interop1"
```
foolip, line 471: Can you also add a test where there are multiple labels one some tests, and a search for tests that have both labels (`&`) and another for tests that have either label (`|`)?

api/query/cache/index/filter.go
```diff
@@ -302,6 +309,27 @@ func (tr Triaged) Filter(t TestID) bool {
...
+		return false
+	}
+
+	labels, ok := tl.metadata[name]
+	if !ok {
+		return false
+	}
+
+	for _, label := range labels {
+		if strings.ToLower(label) == strings.ToLower(tl.label) {
```
foolip, line 325: Can you note that labels are case-insensitive in the docs? Or do labels need to be case-insensitive, does that follow a precedent of other query atoms?

api/query/cache/index/index_filter_test.go
```diff
@@ -627,6 +627,60 @@ func TestBindExecute_TriagedWildcards(t *testing.T) {
...
+			},
+		},
+	})
+	metadata := map[string][]string{
+		"/foo/bar/b.html": {"random"},
+		matchingTestName:  {"interop1", "INTEROP2"},
+		"/d/e/f":          {""},
+	}
+
+	link := query.Or{Args: []query.ConcreteQuery{query.TestLabel{Label: "interop2", Metadata: metadata}}}
```
foolip, line 661: Is the `query.Or` needed here? There's just a single condition inside, right?

How about tests for "and" and "or" with multiple labels? (Also asked earlier, I'm not sure where it would be best to test that, but at one of the layers seems sufficient.)

shared/metadata_test.go
```diff
@@ -321,3 +330,48 @@ func TestGetMetadataFilePath(t *testing.T) {
...
+	subtestName := "Something should happen"
+	fail := TestStatusFail
+	metadataResults := map[string]MetadataLinks{
+		"/foo/bar/a.html": []MetadataLink{
+			{
+				URL: "https://bug.com/item",
+				Results: []MetadataTestResult{{
+					SubtestName: &subtestName,
+					Status:      &fail,
+					Label:       &label,
```
foolip, line 345: I'm not too sure what this test is about, but can subtests really have an associated label? I guess it comes for free by copying the link metadata, but wouldn't it also mean that all subtests will be filtered out when searching for a label, because none on the subtests have that label?

api/query/README.md
KyleJu: Correct. I wanted to distinguish between metadata labels and labels in [/api/runs](https://github.com/web-platform-tests/wpt.fyi/tree/main/api#apiruns). What do you think?

api/query/atoms.go
KyleJu: Yup it is intentional. It is different from checking if `labelMsg, ok := data["testlabel"]` key words exist. The `label` here is the label of a metadata in string format.

api/query/atoms_test.go
KyleJu: Let me separate them

shared/metadata_test.go
KyleJu: That is correct. Technically this format won't exist because label is at the test level. Let me change that

api/query/atoms_test.go
KyleJu: I added a case for `&` abstract query. I realized that the code is yet to support multiple labels. so I will send up a follow-up PR for it

api/query/cache/index/filter.go
KyleJu: Added in the doc. I thought it would make more sense to make it case-insensitive in case of cap issues.

api/query/cache/index/index_filter_test.go
KyleJu: `query.Or` is not needed and I removed here. I will include the `and` and `or` test in the follow-up PR. But yes, this is the place to test that

shared/metadata_test.go
KyleJu: Subtests cannot be associated with a test-level test but this test is for `TestPrepareLinkFilter`.  This is a drive-by change. See `TestPrepareTestLabelFilter` below

KyleJu:
> example-label1

Well it is because `URL` is a non-optional field in metadata. So at the moment, if we want to label a test, we also need a `URL`. Do you think we should change that?

shared/metadata_test.go
KyleJu: Actually you are right. Fixed it. Labels should be at the test-level

foolip:
> > example-label1
> 
> I added [`http://example.org`](https://github.com/web-platform-tests/wpt-metadata/pull/2172/files). It is because `URL` is a non-optional field in metadata. So at the moment, if we want to label a test, we also need a `URL`. Do you think we should change that?

I think so, yes. We don't want a link *and* a label, we want a link *or* a label. Maybe the code that validates metadata could check that we have exactly one of them?

api/query/cache/index/filter.go
foolip: I searched around a bit for other uses of `ToLower` or `ToUpper` to do case-insensitive compares, and found some:
https://github.com/web-platform-tests/wpt.fyi/blob/207813b3ed18bae81068934caa478daffd782d36/api/query/cache/index/filter.go#L166-L169
https://github.com/web-platform-tests/wpt.fyi/blob/207813b3ed18bae81068934caa478daffd782d36/api/query/search.go#L186
https://github.com/web-platform-tests/wpt.fyi/blob/207813b3ed18bae81068934caa478daffd782d36/api/query/atoms.go#L586

So there's some precedent for case-insensitive, so let's go with that! Thanks for documenting it!

api/query/README.md
foolip: Would the two kinds of labels really go in the same place? In the user-visible URLs at least, run labels are query string parameters, but test labels are inside the `q` parameter, and that's true of all search cache bits.

What would the collision look like in `/api/runs`? If it can't be avoided there, could it still be `label` in the user-facing query language?

KyleJu: There is no collision in the query - one is in the query para and one is in `q`, but it can be confusing when `labels` mean different things. It is like the use of `metadata`

api/query/atoms.go
KyleJu: Yup you are right. The original wording was wrong because nothing was `missing` there. Thanks for the correction!

api/query/cache/index/filter.go
KyleJu: Thanks for double checking :)

KyleJu:
> we want a link _or_ a label.

That makes sense. Let me put `a link or a label` and `multiple labels` into the follow-up PR because the metadata format needs to be tweaked

api/query/README.md
KyleJu: I think from the perspective of searchcache, `label` implies labels for tests because we are searching the `label` of some tests under some `runs`. We are not searching for `runs`

KyleJu: I renamed the search expression to `label` but kept `testlabel` on the backend. I think it will be more clear internally in the code

foolip (approved):

foolip:
> > we want a link _or_ a label.
> 
> That makes sense. Let me put `a link or a label` and `multiple labels` into the follow-up PR because the metadata format needs to be tweaked

Sounds good! I'll land this so that it's deployed to staging and we can try a label in the real metadata repo.

PR #2687, opened 2021-11-19
Tweak Metadata label field and support multiple labels

KyleJu:
Tweak Metadata label format per https://github.com/web-platform-tests/wpt-metadata/pull/2186. A follow-up for https://github.com/web-platform-tests/wpt.fyi/pull/2678. This PR includes
- Move metadata `label` field to a level above in the `link`
- Support multiple labels per test
- Add tests for `&` and '|` testlabel search logic for searchcache

### DIFF
--- a/api/query/cache/index/index_filter_test.go
+++ b/api/query/cache/index/index_filter_test.go
@@ -658,6 +658,7 @@ func TestBindExecute_QueryAndTestLabel(t *testing.T) {
 		"/d/e/f":          {""},
 	}
 
+	// It is equivalent to searching "label:interop1 & label:interop2".
 	testlabel := query.And{[]query.ConcreteQuery{query.TestLabel{Label: "interop2", Metadata: metadata}, query.TestLabel{Label: "interop1", Metadata: metadata}}}
 	plan, err := idx.Bind(runs, testlabel)
 	assert.Nil(t, err)
@@ -712,6 +713,7 @@ func TestBindExecute_QueryOrTestLabel(t *testing.T) {
 		"/d/e/f":          {"interop1"},
 	}
 
+	// It is equivalent to searching "label:interop1 | label:interop2".
 	testlabel := query.Or{[]query.ConcreteQuery{query.TestLabel{Label: "interop2", Metadata: metadata}, query.TestLabel{Label: "interop1", Metadata: metadata}}}
 	plan, err := idx.Bind(runs, testlabel)
 	assert.Nil(t, err)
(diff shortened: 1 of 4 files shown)


api/query/cache/index/index_filter_test.go
```diff
@@ -627,6 +627,126 @@ func TestBindExecute_TriagedWildcards(t *testing.T) {
...
+			},
+		},
+	})
+	metadata := map[string][]string{
+		"/foo/bar/b.html": {"random"},
+		matchingTestName:  {"interop1", "INTEROP2"},
+		"/d/e/f":          {""},
+	}
+
+	testlabel := query.And{[]query.ConcreteQuery{query.TestLabel{Label: "interop2", Metadata: metadata}, query.TestLabel{Label: "interop1", Metadata: metadata}}}
```
foolip, line 661: Can you add a comment here explaining what the query is in more plain text, or using the search cache syntax? I think this is `label:interop1 & label:interop2`. (I assume the order doesn't matter.)

api/query/cache/index/index_filter_test.go
```diff
@@ -627,6 +627,126 @@ func TestBindExecute_TriagedWildcards(t *testing.T) {
...
+			},
+		},
+	})
+	metadata := map[string][]string{
+		"/foo/bar/b.html": {"random"},
+		matchingTestName:  {"INTEROP2"},
+		"/d/e/f":          {"interop1"},
+	}
+
+	testlabel := query.Or{[]query.ConcreteQuery{query.TestLabel{Label: "interop2", Metadata: metadata}, query.TestLabel{Label: "interop1", Metadata: metadata}}}
```
foolip, line 715: Document this as `label:interop1  | label:interop2`?

shared/metadata_test.go
```diff
@@ -32,6 +31,7 @@ links:
       status: FAIL
     - test: c.html
   - url: https://github-issue.com/1234
+    label: labelB
```
foolip, line 34: Here there's both `url` and `label`. I guess that makes sense when it applies to a single test, but is it also possible to split these into two items? If it is possible, I wonder if it wouldn't be easier overall to *only* allow the split form, in other words disallowing having both `url` and `label` at the same time?

foolip: Per discussion with @KyleJu this is just internal to the wpt.fyi representation, in wpt-metadata it will be allowed to have either `url` or `label`, or both. And we'll have lints to catch the situation where there's redundant entries. In that case, allowing both forms is fine.

foolip (approved):
LGTM with a few comments added.

PR #2694, opened 2021-11-24
Remove break for constructing test-level metadata

KyleJu:
Fixed a existing bug when constructing test-level metadata. The issue is surfaced in https://github.com/web-platform-tests/wpt-metadata/pull/2196#issuecomment-977133256.


KyleJu:
I will create a separate PR since the history is polluted

PR #2700, opened 2021-11-30
Add triage UI for label

KyleJu:
Add Triage UI for #1491

### Tested here
https://github.com/web-platform-tests/wpt-metadata/pull/2251

### DIFF
--- a/shared/triage_metadata_test.go
+++ b/shared/triage_metadata_test.go
@@ -1,4 +1,6 @@
+//go:build small
 // +build small
+
 // Copyright 2019 The WPT Dashboard Project. All rights reserved.
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file.
@@ -248,6 +250,94 @@ links:
 	assert.Equal(t, "chrome", actual.Links[2].Product.BrowserName)
 	assert.Equal(t, "foo1", actual.Links[2].URL)
 	assert.Equal(t, "a.html", actual.Links[2].Results[0].TestPath)
+}
+
+func TestAddToFiles_AddNewMetadataLink_Label(t *testing.T) {
+	var amendment MetadataResults
+	json.Unmarshal([]byte(`{
+		"/foo/foo1/abc.html": [
+			{
+				"label": "interop"
+			}
+		]
+	}`), &amendment)
+
+	var path = "foo/foo1"
+	var fileMap = make(map[string]Metadata)
+	fileInBytes := []byte(`
+links:
+  - product: chrome
+    url: foo
+    results:
+    - test: b.html
+  - product: firefox-2
+    url: https://bug.com/item
+    results:
+    - test: b.html
+      subtest: Something should happen
+      status: FAIL
+    - test: c.html
+`)
+	var file Metadata
+	yaml.Unmarshal(fileInBytes, &file)
+	fileMap[path] = file
+
+	actualMap := addToFiles(amendment, fileMap, NewNilLogger())
+
+	assert.Equal(t, 1, len(actualMap))
+	actualInBytes, ok := actualMap["foo/foo1"]
+	assert.True(t, ok)
+
+	var actual Metadata
+	yaml.Unmarshal(actualInBytes, &actual)
+	assert.Equal(t, 3, len(actual.Links))
+	assert.Equal(t, "chrome", actual.Links[0].Product.BrowserName)
+	assert.Equal(t, "foo", actual.Links[0].URL)
+	assert.Equal(t, 1, len(actual.Links[0].Results))
+	assert.Equal(t, "b.html", actual.Links[0].Results[0].TestPath)
+	assert.Equal(t, "firefox", actual.Links[1].Product.BrowserName)
+	assert.Equal(t, "https://bug.com/item", actual.Links[1].URL)
+	assert.Equal(t, "", actual.Links[2].Product.String())
+	assert.Equal(t, "", actual.Links[2].URL)
+	assert.Equal(t, "interop", actual.Links[2].Label)
+}
+
+func TestAddToFiles_AddNewMetadataResults_Label(t *testing.T) {
+	var amendment MetadataResults
+	json.Unmarshal([]byte(`{
+		"/foo/foo1/abc.html": [
+			{
+				"label": "interop"
+			}
+		]
+	}`), &amendment)
+
+	var path = "foo/foo1"
+	var fileMap = make(map[string]Metadata)
+	fileInBytes := []byte(`
+links:
+  - label: interop
+    results:
+    - test: b.html
+`)
+	var file Metadata
+	yaml.Unmarshal(fileInBytes, &file)
+	fileMap[path] = file
+
+	actualMap := addToFiles(amendment, fileMap, NewNilLogger())
+
+	assert.Equal(t, 1, len(actualMap))
+	actualInBytes, ok := actualMap["foo/foo1"]
+	assert.True(t, ok)
+
+	var actual Metadata
+	yaml.Unmarshal(actualInBytes, &actual)
+	assert.Equal(t, 1, len(actual.Links))
+	assert.Equal(t, "", actual.Links[0].Product.BrowserName)
+	assert.Equal(t, "interop", actual.Links[0].Label)
+	assert.Equal(t, 2, len(actual.Links[0].Results))
+	assert.Equal(t, "b.html", actual.Links[0].Results[0].TestPath)
+	assert.Equal(t, "abc.html", actual.Links[0].Results[1].TestPath)
 }
 
 func TestAddToFiles_AddNewMetadataLink_Asterisk(t *testing.T) {
--- a/shared/triage_metadata_test.go
+++ b/shared/triage_metadata_test.go
@@ -116,7 +116,7 @@ links:
     url: https://external.com/item
     results:
     - test: a.html
-  - product: firefox-2
+  - product: firefox
     url: https://bug.com/item
     results:
     - test: b.html
@@ -166,7 +166,7 @@ links:
     url: foo
     results:
     - test: b.html
-  - product: firefox-2
+  - product: firefox
     url: https://bug.com/item
     results:
     - test: b.html
@@ -220,7 +220,7 @@ links:
     url: foo
     results:
     - test: b.html
-  - product: firefox-2
+  - product: firefox
     url: https://bug.com/item
     results:
     - test: b.html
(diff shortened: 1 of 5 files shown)


KyleJu:
The UI layout looks like:
Screen Shot 2021-12-01 at 4 54 55 PM

foolip:
@KyleJu I haven't reviewed the code yet, but it would be great with some validation on submit, so that trying to submit with neither URL nor label shows a meaningful error message. (Currently the triage API will return 500 instead.)

shared/triage_metadata_test.go
```diff
@@ -250,6 +252,94 @@ links:
...
+
+	var path = "foo/foo1"
+	var fileMap = make(map[string]Metadata)
+	fileInBytes := []byte(`
+links:
+  - product: chrome
+    url: foo
+    results:
+    - test: b.html
+  - product: firefox-2
```
foolip, line 273: Is the trailing "-2" here significant? I guess it means Firefox version 2, and in the expectations below the product is just "firefox", but I guess this is already covered by other tests?

shared/triage_metadata_test.go
```diff
@@ -250,6 +252,94 @@ links:
...
+    url: https://bug.com/item
+    results:
+    - test: b.html
+      subtest: Something should happen
```
foolip, line 277: Sounds like good life motto!

webapp/components/test/wpt-amend-metadata.html
```diff
@@ -94,6 +94,19 @@
 
       assert.deepEqual(appFixture.getTriagedMetadataMap(displayedMetadata), expected);
     });
+    test('getTriagedMetadataMap(displayedMetadata) for a label triage', () => {
+      appFixture.path = '/abc';
+      const displayedMetadata = [
+        { product: '', url: '', tests: ['testA.html'], label: 'interop' },
+      ];
+      // When there is no product, we have to remove the product from the
```
foolip, line 102: Useful comment, thanks!

webapp/components/wpt-amend-metadata.js
```diff
@@ -164,6 +164,9 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
               <img class="browser" src="[[displayMetadataLogo(node.product)]]">
               :
               <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
+              <template is="dom-if" if="[[!node.product]]">
+                <paper-input label="Label" value="{{node.label}}"></paper-input>
```
foolip, line 168: In the screenshot there's a lowercase "label" here. Is that from something on this line, what does `{{node.label}}` mean? In any case, capitalizing it like "Label" just like "Bug URL" would be nice.

webapp/components/wpt-amend-metadata.js
```diff
@@ -164,6 +164,9 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
               <img class="browser" src="[[displayMetadataLogo(node.product)]]">
               :
               <paper-input label="Bug URL" value="{{node.url}}" autofocus></paper-input>
+              <template is="dom-if" if="[[!node.product]]">
```
foolip, line 167: Here, perhaps the bug and label can go on separate lines?

Linking to some documentation about bug and label triage here would be helpful too.

webapp/components/wpt-amend-metadata.js
```diff
@@ -250,18 +253,24 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
       }
     } else {
       for (const entry of displayedMetadata) {
-        if (entry.url === '') {
+        if (entry.url === '' && (!entry.label || entry.label === '')) {
```
foolip, line 256: This can be simplified, because if `entry.label === ''`, then `!entry.label` is true and `entry.label === ''` will never be evaluated:

```suggestion
        if (entry.url === '' && !entry.label) {
```

And maybe a comment around here about why `url` and `label` are treated differently would be helpful.

webapp/components/wpt-amend-metadata.js
```diff
@@ -335,7 +344,11 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
     }
 
     for (const key in browserMap) {
-      this.displayedMetadata.push({ product: key, url: '', tests: browserMap[key] });
+      let node = { product: key, url: '', tests: browserMap[key] };
+      if (key === '') {
+        node['label'] = '';
```
foolip, line 349: Can you add a comment about this case? Why do we need to set the label to the empty string if there is no product?

shared/triage_metadata_test.go
KyleJu: I copied and pasted from old tests. Removed them all :)

webapp/components/wpt-amend-metadata.js
KyleJu: Yea I already changed it. The lowercase label is from `label="label"` and `{{node.label}}` is the object that stores user input of `label`

webapp/components/wpt-amend-metadata.js
KyleJu: I do feel like the UI looks cleaner with `bug` and `label` on the same line but let me add some documentation on the top

webapp/components/wpt-amend-metadata.js
KyleJu: Added. Basically we will set a `label` field when it is a test-level triage (product is empty)

KyleJu:
> @KyleJu I haven't reviewed the code yet, but it would be great with some validation on submit, so that trying to submit with neither URL nor label shows a meaningful error message. (Currently the triage API will return 500 instead.)

The new UI for this case
Screen Shot 2021-12-07 at 6 27 24 PM

webapp/components/wpt-amend-metadata.js
```diff
@@ -335,22 +346,43 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
     }
 
     for (const key in browserMap) {
-      this.displayedMetadata.push({ product: key, url: '', tests: browserMap[key] });
+      let node = { product: key, url: '', tests: browserMap[key] };
+      // when key (product) is empty, we will set a label field becuase
```
foolip, line 350:
```suggestion
      // when key (product) is empty, we will set a label field because
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -335,22 +346,43 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
...
   handleTriage() {
     const url = new URL('/api/metadata/triage', window.location);
+    const toast = this.shadowRoot.querySelector('#show-pr');
+
+    const triagedMetadataMap = this.getTriagedMetadataMap(this.displayedMetadata);
+    if (Object.keys(triagedMetadataMap).length === 0) {
+      this.selectedMetadata = [];
+      let errMsg = '';
+      if (this.displayedMetadata.length > 0 && this.displayedMetadata[0].product === '') {
+        errMsg = 'Invalid triage: Bug URL and Label fields cannot both be empty.';
```
foolip, line 368:
```suggestion
        errMsg = 'Failed to triage: Bug URL and Label fields cannot both be empty.';
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -335,22 +346,43 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
...
+      } else {
+        errMsg = 'Invalid triage: Bug URLs cannot be empty.';
```
foolip, line 370:
```suggestion
        errMsg = 'Failed to triage: Bug URLs cannot be empty.';
```

webapp/components/wpt-amend-metadata.js
```diff
@@ -335,22 +346,43 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
...
 
   handleTriage() {
     const url = new URL('/api/metadata/triage', window.location);
+    const toast = this.shadowRoot.querySelector('#show-pr');
+
+    const triagedMetadataMap = this.getTriagedMetadataMap(this.displayedMetadata);
+    if (Object.keys(triagedMetadataMap).length === 0) {
+      this.selectedMetadata = [];
+      let errMsg = '';
+      if (this.displayedMetadata.length > 0 && this.displayedMetadata[0].product === '') {
```
foolip, line 367: How does this condition relate to the bug URL and label both being empty? It looks like it depends on how `/api/metadata/triage` will fail in this case.

Can the validation instead happen on the client side before submitting? That would also avoid losing the state if one has selected a number of tests and then accidentally clicks submit.

foolip:
@KyleJu there's one comment still about doing the validation on the client side.

I guess that would be a bit of work, so please push back if you think it's not the way to go.

webapp/components/wpt-amend-metadata.js
KyleJu: > How does this condition relate to the bug URL and label both being empty? It looks like it depends on how `/api/metadata/triage` will fail in this case.
Correct. They have a different error messages.

> Can the validation instead happen on the client side before submitting? That would also avoid losing the state if one has selected a number of tests and then accidentally clicks submit.

Yea I think disabling the `triage` button if empty is one way to go but I have to do a few experiments with the UI

KyleJu:
> @KyleJu there's one comment still about doing the validation on the client side.
> 
> I guess that would be a bit of work, so please push back if you think it's not the way to go.

Yea I do have an alternative in mind but it does require a bit work. Since the current PR is already improvement, what if we submit this PR and I will send a follow-up?

foolip:
> > @KyleJu there's one comment still about doing the validation on the client side.
> > I guess that would be a bit of work, so please push back if you think it's not the way to go.
> 
> Yea I do have an alternative in mind but it does require a bit work. Since the current PR is already an improvement, what if we submit this PR and I will send a follow-up?

Sure, let's land this and see it on staging. Can you file an issue about the follow-up you have in mind?

foolip (approved):

KyleJu:
> > > @KyleJu there's one comment still about doing the validation on the client side.
> > > I guess that would be a bit of work, so please push back if you think it's not the way to go.
> > 
> > 
> > Yea I do have an alternative in mind but it does require a bit work. Since the current PR is already an improvement, what if we submit this PR and I will send a follow-up?
> 
> Sure, let's land this and see it on staging. Can you file an issue about the follow-up you have in mind?

Filed https://github.com/web-platform-tests/wpt.fyi/issues/2713

PR #2744, opened 2022-02-01
2022 Design Updates

argyleink:
fixes [#45](https://github.com/web-platform-tests/interop-2022/issues/45)

space to hold work efforts to improve the UI and UX of the 2022 compat dashboard

Estimated due date <del>`2/20/2020`</del> `2/14/2020`

Todo:
- [x] get stable safari scores in
- [x] restore line chart scoring
- [x] fix mismatch between feature keys in the data model
- [x] score breakdown table data needs filled out
- [x] score breakdown table should update when stable/experimental toggle is changed
- [x] score breakdown table should provide links to tests like the focus area does (nice to have)
- [x] older tooltip code removed
- [x] more cross browser QA
- [x] stable/experimental toggle needs screen reader state
- [x] browser icons need labels
- [x] stable/experimental toggle isnt clear enough which is toggled?
- [x] reduced motion check for the countup number animation?
- [ ] screenreader support for a changing data table?


KyleJu:
The `web_components_test` failure is caused by a flaky test, not due to the changes in this PR

KyleJu:
Deployed service [default] to [https://2022-design-updates-dot-wptdashboard-staging.uk.r.appspot.com]

argyleink:
latest push updates site to look like this

<details>

<summary>Screenshot</summary>

0 0 0 0_8080_interop-2022_feature=position-sticky stable (1)

</details>

argyleink:
latest staged build available at https://2022-design-updates-dot-wptdashboard-staging.uk.r.appspot.com/interop-2022

jensimmons:
This is all looking really good, Adam. I came here to give two notes, but hey, you've already done what I was going to suggest.

FYI, the Aspect Ratio tests do not just include the `aspect-ratio` CSS property. The new way browsers calculate the initial aspect ratio of an image/video replaced element from the width and height attributes (improving first paint) is also included. So... maybe call it "Aspect Ratio" instead of "aspect-ratio", and consider linking to the new text in the HTML spec, not just the CSS `aspect-ratio` element.

I'm not sure what you are doing for some of the other things — like "text" is several items, from several different specs, including Font Feature support. In fact, it was called "Typography" for a while.

bkardell:
A few comments raised during the meeting...

1. We'd like to make it more obvious which view (stable or experimental) is selected, it is visually indicated currently, but unclear which is actually active.
2. We'd like to not rely exclusively on icons
3. We'd like to do an a11y pass before merging this - though we can improve on it with separate issues as we go there are some basic questions here about whether we can or should for example respect prefers reduced animations in the counter changes, a few tab issues, and whether the counter changes as implemented create undue noise or troubles for AT. @bkardell has offered to breakout with @argyleink on this.

argyleink:
> The focus areas table should have `font-variant-numeric: tabular-nums`.

I added it in a previous iteration, but it's a monospace font and had no effect. did i miss something tho?!

argyleink:
> 3. We'd like to do an a11y pass before merging this

for sure! what tools do you test with? which assistive tech do we intend to support (there's so many)?

i just did a macOS screen reader test, feels like there's plenty to fix, but it's mostly about the page not communicating enough about state change than anything else? to fix most the issues the page should probably remove polymer and behave more like a document than an app.. maybe a v3 thing?

jensimmons:
I didn't get a chance to say in our meeting today a huge thank you to Adam, Philip, Una, and everyone else who's been working so hard on the new dashboard. It's a massive leap forward in providing real information about the project, and telling the story of what this is.

I haven't given notes yet, because each time I sit down to do so, and reload the page to see if there are any changes, you've already done most of what came to mind. 

I do still have these thoughts (beyond what we already talked about today in the meeting):

1. I'm not convinced the "How is this score calculated?" toggle to hide/show the data table is the best way to go. Perhaps it would be better to simple always show that table. And if you want to have the graphs be closer to the top on initial load, then put the graphs above the data table. It just feels to me like too many people won't see the phrase of text, won't click to open it, and won't see the information. Having the data table be further down the page where you have to scroll to get to it is at least far more expected.  

2. Could you style the dropdown? I see you are just using the default browser style — but then in Safari, the drop down is very small. The text is too small, the area to click/tap is too small. Maybe at least make it bigger, if not style the whole thing.

3. It's sad there's still not an easy way to get the percent values to line up on the decimal point, but could you write 100.0% instead of 100% in the data table so that the numbers line up better.

4. Style the links in the table, so they don't have the UA style colors of blue & purple?

5. I like the logos in the table, too — but there do need to be real words for accessibility. You could use headlines (or whatever HTML), and then hide them visually using the hack to move it off the screen: https://webaim.org/techniques/css/invisiblecontent/

jensimmons:
Oh and:
6. Look at this on mobile. It almost works (despite the fact this whole site is not designed for mobile) — but the content is a bit small.

jensimmons:
>Do we even want to display the figures with decimal places? 
I agree that perhaps we could just get rid of the decimal tens for the percents in the data table, and just show the rounded down number. (Down? If that's how the math is working.)

foolip:
Funny, I did some extra work to avoid the trailing ".0" for 100%, but just dropping the decimal point in the tables works. The graph is still using percentages with a decimal point, and I think that's good as it makes the less jagged and less likely that the lines for two browsers will exactly overlap.

foolip:
I've pushed a change to group the score table. @argyleink I left a "Adam, help!" in there when I couldn't figure out spacing inside the table, can you take a look?

foolip:
@argyleink can you help look into what remains of https://github.com/web-platform-tests/wpt.fyi/pull/2744#issuecomment-1035207759 and https://github.com/web-platform-tests/wpt.fyi/pull/2744#issuecomment-1035272463? A non-exhaustive list:
- Adding labels, not just relying on browser icons
- Moving score table to below graph and having it always expanded (no `<details>` needed)
- Visual touch up of the table, avoid repeating icons, and more consistent spacing/borders before final score?
- Accessibility

jensimmons:
The link for the 2022 Focus area of Cascade Layers (in the table) does not go to the right search. It's pulling up anything with "layer". Instead of "label:interop-2022-cascade". Which has the unfortunate result of listing tests that are not part of this effort. 

Could you also make sure the scores are using the correct label, and not a general "layer" keyword? We discovered the table link bug in trying to understand why Safari's result for Cascade Layers is so much lower than other browsers. It seems perhaps the score is not connected to the right set of tests.

jensimmons:
I saw you just added labels. I know that _Safari Technology Preview_ is a long name, but we'd like to request the correct name is used, and not an abbreviation like "Safari Tech Preview".

jensimmons:
Oh, btw, we noticed that the logo for Chrome is not the fancy new logo. https://www.theverge.com/2022/2/5/22919398/chrome-logo-change-eight-years

argyleink:
> Oh, btw, we noticed that the logo for Chrome is not the fancy new logo. https://www.theverge.com/2022/2/5/22919398/chrome-logo-change-eight-years

Thanks! looks like we fetch them from a dependency, i've just submitted my updated logos to them here https://github.com/alrra/browser-logos/issues/206

webapp/package.json
```diff
@@ -7,11 +7,11 @@
     "url": "https://github.com/web-platform-tests/wpt.fyi.git"
   },
   "devDependencies": {
-    "@browser-logos/chrome": "1.0.18",
-    "@browser-logos/chrome-beta": "2.1.10",
-    "@browser-logos/chrome-canary": "1.1.13",
-    "@browser-logos/chrome-dev": "2.1.11",
-    "@browser-logos/chromium": "1.1.5",
+    "@browser-logos/chrome": "2.0.0",
+    "@browser-logos/chrome-beta": "2.0.0",
```
alrra, line 11: @argyleink You should keep the original version for `chrome-beta` and `chrome-dev` as that is the latest version (when they will be updated as the rest those packages will be v3).

---

Note: For what changed, see the package's `CHANGELOG.md` file (e.g.: for [Chrome](https://github.com/alrra/browser-logos/blob/bb02a4d6df6d0261e12a8681b4731ffbbd8770da/src/chrome/CHANGELOG.md#changelog)).

argyleink: ty! will update!

argyleink:
removed dark mode until this bug is resolved https://bugs.webkit.org/show_bug.cgi?id=226893

jensimmons:
A total nitpick.. When making a screenshot, I wanted the opening paragraph ("These scores represent how browser engines are doing in 15 focus areas and 3 joint investigation efforts.") to not have a widow, so I adjusted: `.prose { max-inline-size: 40ch; }` to be `42ch`.

foolip:
> A total nitpick.. When making a screenshot, I wanted the opening paragraph ("These scores represent how browser engines are doing in 15 focus areas and 3 joint investigation efforts.") to not have a widow, so I adjusted: `.prose { max-inline-size: 40ch; }` to be `42ch`.

Ah, good catch! I tweaked the wording to bring it down to two lines (in Chrome) earlier, but didn't notice that it still wraps in Safari. I've applied the `42ch` change and checked that it's also 2 lines in Firefox now.

foolip (approved):
Let's get this into staging!

Issue #2743, opened 2022-02-01
realpath dependency is not documented


foolip:
@argyleink ran into this trying to get a local setup working on macOS. We use `realpath` in some scripts and the failure mode when following the README is pretty confusing if `realpath` isn't present.

We could document this dependency, or remove the need for it with some workaround.

KyleJu:
Notes that the `realpath` issue was resolved by installing `coreutils` through brew but there should be a better fix

ranashayan836-lang:
https://github.com/web-platform-tests/wpt.fyi/pull/2748

PR #2748, opened 2022-02-04
Replace realpath with git rev-parse --show-toplevel

foolip:
`realpath` was used in three places to get an absolute path to the repo
directory. We can use `git rev-parse --show-toplevel` for this instead
and avoid the dependency.

WPT_PATH was never used anywhere, and therefore removed.

In the pre-push script, commands.sh is sourced, so that will define
WPTD_PATH, so the redefinition can just be dropped.

Fixes https://github.com/web-platform-tests/wpt.fyi/issues/2743.


foolip:
@KyleJu with these changes I was able to follow the instructions in the README and get a working local dev instance. @argyleink FYI.

argyleink (approved):

KyleJu (approved):

PR #2768, opened 2022-02-21
Link to Interop 2022 in header navigation

foolip:
At the same time, remove the link to the interop view, since this could
now easily be confused with Interop 2022. The view itself isn't removed,
but it hasn't been as valuable since we added the search cache and can
filter to different combinations of pass and fail directly.


foolip:
@KyleJu can you review after the fact?

KyleJu (commented):
LGTM

PR #2774, opened 2022-02-24
Unlink Interop 2022 from navigation and say it will launch soon

foolip:
This is in case we need to or want to a prod push, for people who find
their way to the dashboard to understand it's not live yet.

### DIFF
--- a/webapp/components/interop-2022.js
+++ b/webapp/components/interop-2022.js
@@ -467,6 +467,12 @@ class Interop2022 extends PolymerElement {
       </style>
       <h1>Interop 2022 Dashboard</h1>
       <p class="prose">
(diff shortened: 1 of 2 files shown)


zcorpan (approved):

webapp/components/interop-2022.js
```diff
@@ -466,6 +466,12 @@ class Interop2022 extends PolymerElement {
         // }
       </style>
       <h1>Interop 2022 Dashboard</h1>
+      <p class="prose">
```
jgraham, line 469: I might make this stand out a bit more, but I don't know what classes you have.

jgraham (approved):

KyleJu (approved):

PR #2783, opened 2022-02-28
Scale the Interop 2022 summary score to 90% in graph

foolip:
This centralizes the calculation into the place where it will be natural
to add in the investigation score.


argyleink (approved):
lgtm

PR #2838, opened 2022-04-14
Fix android_webview icon size

KyleJu:
Fix android_webview icon size


KyleJu:
Screen Shot 2022-04-14 at 1 56 43 PM

WeizhongX (approved):
Thanks! Finally we made it correct.

Issue #2825, opened 2022-03-25
Showing percentage (not test/subtest) summary numbers


foolip:
@mfreed7 had this suggestion/request in email:

> I wanted to quickly revive this part of the discussion on Interop2022 scoring. I wanted to point out that it would be *so* nice to have a view [like this one](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-forms) on WPT.fyi, where the numbers in the table are the actual scores used for Interop2022. So for example, in [this list](https://wpt.fyi/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-forms), the `css/` folder currently shows "412 / 476", meaning Chrome passes 412 **sub-tests** out of 476 **sub-tests**. It would be great if instead it showed "43.95 / 90", which is my best estimate of the actual Interop2022 weight for the css/ folder. There are 90 individual test files, of which we fail 46, and one in which we get 325/343 subtests right.
>
> I ask because as we're looking for ways to improve our score, the above is what matters.
>
> Do you think this change is feasible/easy? I was initially thinking that this could be a separate display "mode", but the more I think about it, the more this scoring system seems generally better at conveying the pass/fail state of things.

I 100% agree that this would be valuable, and had this in mind when writing the [scoring code](https://github.com/web-platform-tests/results-analysis) for Interop 2022. I believe the rules are simple enough to implement on the frontend. I didn't file an issue though...

This is very similar to https://github.com/web-platform-tests/wpt.fyi/issues/2290, where the request it to show test counts instead. Whether we end up with 2 or 3 different ways of summarizing scores, I'd like a configuration ⚙️ for that, as well as the ability to sort, tracked by https://github.com/web-platform-tests/wpt.fyi/issues/2289.

@KyleJu @past let's discuss this for Q2 planning, getting the request from @mfreed7 makes me even more certain it's important :)

foolip:
We will have to revisit https://github.com/web-platform-tests/wpt.fyi/issues/62 when deciding on how to compute a percentage for testharness.js tests. Our options will be:
1. Ignore harness status entirely, [like Interop 2022](https://github.com/web-platform-tests/results-analysis/blob/d435f2156203f73c6c9228b12553f09254c8d72f/interop-2022/main.js#L218-L220)
2. Score as 0 if there's a harness error
[...]

PR #2841, opened 2022-04-19
Display test totals information on subtest and directory views

### LINKED ISSUE #2667: show a row with totals
It would be useful if wpt.fyi directory listings showed a row either at the top or bottom with the totals.

For example, I've been focusing a lot on the [list of css-transforms tests](https://wpt.fyi/results/css/css-transforms?label=master&label=experimental).  But the piece of information that I don't see in that view is the information shown in the "css-transforms" line within the [list of all css tests](https://wpt.fyi/results/css?label=master&label=experimental) that has the totals for the css-transforms directory.  I'd like to see that line shown as a "Total" row at either the top or the bottom of the css-transforms list.

I think it would be useful to show this total line in the css-transforms directory listing rather than having to go up to the css directory listing to see it.

### LINKED ISSUE #2759: wpt.fyi fails to count subtests
For example, see [this](https://wpt.fyi/results/webaudio/the-audio-api/the-audiocontext-interface/audiocontext-suspend-resume.html?sha=fdb441a036fae9246b9f7348778a41e2d233f5c5&label=master&label=experimental&product=chrome&product=firefox&product=safari) result view.

The UI says:

> Showing 33 subtests from master test runs from revision fdb441a

It also shows 40 subtests below that.

DanielRyanSmith:
## Description
fixes #2759 and #2667

This is a fix to ensure that the number of subtests shown in the blue banner matches of single test views will accurately represent the number of rows displayed to the user. Additionally, a "Total" row is added to subtest and directory views to display the aggregation of pass/total values of all of the tests numbers in the respective column.

~~**NOTE:** Issue #62 is still a problem for the aggregate number of subtests shown in levels higher than the subtest view, so this total will not match higher-level views by 1 in most instances.~~
Edit: This is a separate issue and this implementation has been removed from this PR.

## Changes
- The top blue banner now correctly displays the number of subtests being displayed in the subtest view.
- A row displaying the fraction of passes/total for tests in a directory or subtests within a test is displayed.
  - If a set of subtests for a browser has no data (all "MISSING"), the total row will show `0/0` with a gray background.
  - Tests that have no subtests and only display the Test Harness status and duration will not display a totals row.

Some additional small changes have been made (and can be removed if they are not correct):
- An additional sort for rows shown based on the test path name has been removed, as it had the (seemingly unintended) effect of sorting test paths with capital letter titles first.

- "TIMEOUT" and "NOTRUN" test status results show with a red background, as they are being considered failing anyway.
- "OK" test status results show with a green background, as they are being considered passing anyway.

### Subtests view (before)
Screen Shot 2022-04-20 at 10 01 41 AM

### Subtests view (after)
Screen Shot 2022-04-20 at 10 01 49 AM

### Directory view (before)
Screen Shot 2022-04-20 at 2 39 46 PM

### Directory view (after)
Screen Shot 2022-04-20 at 2 39 39 PM

### Totals with 0 tests will display with a gray background
Screen Shot 2022-04-20 at 2 43 41 PM

### Totals displayed in the root test directory will display with a gray background
Screen Shot 2022-04-20 at 2 44 59 PM

### DIFF
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -359,8 +359,8 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     if (rows.length <= 2) {
       return [];
     }
-    // keep a total for each for each browser.
-    let totals = new Array(rows[2].results.length);
+    // Keep a total for each browser.
+    const totals = new Array(rows[2].results.length);
     for (let i = 0; i < totals.length; i++) {
       totals[i] = {passes: 0, total: 0};
     }
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -171,10 +171,10 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     </template>
     <template is="dom-if" if="[[shouldShowTotals(totals)]]">
       <tr class="totals-row">
-        <td class="sub-test-name"><code>Total</code></td>
-        <template is="dom-repeat" items="[[totals]]" as="totals">
-          <td class$="[[ totalsColorClass(totals.passes, totals.total) ]]">
-            <code>[[ totals.passes ]]/[[ totals.total ]]</code>
+        <td class="sub-test-name"><code><strong>Total</strong></code></td>
+        <template is="dom-repeat" items="[[totals]]" as="columnTotal">
+          <td class$="[[ totalsColorClass(columnTotal.passes, columnTotal.total) ]]">
+            <code>[[ columnTotal.passes ]]/[[ columnTotal.total ]]</code>
           </td>
         </template>
       </tr>
@@ -359,29 +359,33 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     if (rows.length <= 2) {
       return [];
     }
+
     // Keep a total for each browser.
-    const totals = new Array(rows[2].results.length);
+    const totals = new Array(rows[0].results.length);
     for (let i = 0; i < totals.length; i++) {
       totals[i] = {passes: 0, total: 0};
     }
 
     // Tally the number of passes and total tests.
-    for (let i = 2; i < rows.length; i++) {
-      rows[i].results.forEach((result, index) => {
-        if (result.status === 'PASS') {
-          totals[index].passes++;
-        }
-        // If the test status is missing, it's not counted toward the total.
-        if (result.status) {
-          totals[index].total++;
-        }
-      });
-    }
+    rows.forEach(row => {
+      // Don't count the Harness duration row.
+      if (row.name !== 'Duration') {
+        row.results.forEach((result, index) => {
+          if (result.status === 'PASS' || result.status === 'OK') {
+            totals[index].passes++;
+          }
+          // If the test status is missing, it's not counted toward the total.
+          if (result.status) {
+            totals[index].total++;
+          }
+        });
+      }
+    });
     return totals;
   }
 
   colorClass(status) {
-    if (['PASS'].includes(status)) {
+    if (['PASS', 'OK'].includes(status)) {
       return this.passRateClass(1, 1);
     } else if (['FAIL', 'ERROR', 'TIMEOUT', 'NOTRUN'].includes(status)) {
       return this.passRateClass(0, 1);
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -107,6 +107,10 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     width: -moz-max-content;
     width: max-content;
   }
+  .totals-row {
+    border-top: 8px solid white;
+    padding: 8px;
+  }
   .view-triage {
     margin-left: 30px;
   }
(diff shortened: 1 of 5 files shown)


KyleJu:
Minor font differences in `Total` between testfile view and subtest vie
Screen Shot 2022-05-04 at 4 22 18 PM
Screen Shot 2022-05-04 at 4 22 27 PM

KyleJu:
Still some test count differences. Might be worth filing a separate issue. 
Screen Shot 2022-05-04 at 4 29 35 PM (2)
Screen Shot 2022-05-04 at 4 30 19 PM (2)

webapp/components/test-file-results-table.js
```diff
@@ -165,7 +169,16 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
         </template>
       </tr>
     </template>
-
+    <template is="dom-if" if="[[shouldShowTotals(totals)]]">
+      <tr class="totals-row">
+        <td class="sub-test-name"><code>Total</code></td>
+        <template is="dom-repeat" items="[[totals]]" as="totals">
```
KyleJu, line 175: `<template is="dom-repeat" items="[[totals]]" as="totals">` should be `as=total`. Otherwise you are reassigning the existing `totals` variables here.

KyleJu: Or "columnTotal" as `total.toal` is confuisng

webapp/components/test-file-results-table.js
```diff
@@ -335,15 +352,52 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
...
 
+  computeTotals(rows) {
+    // The first two rows display TestHarness status and duration,
+    // so we don't need to count them. If only these rows exist,
+    // there is no need to show totals.
+    if (rows.length <= 2) {
+      return [];
+    }
+    // keep a total for each for each browser.
+    let totals = new Array(rows[2].results.length);
```
KyleJu, line 363:
```suggestion
    const totals = new Array(rows[2].results.length);
```

webapp/components/test-file-results-table.js
```diff
@@ -335,15 +352,52 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
...
   }
 
+  computeTotals(rows) {
+    // The first two rows display TestHarness status and duration,
+    // so we don't need to count them. If only these rows exist,
+    // there is no need to show totals.
+    if (rows.length <= 2) {
+      return [];
+    }
+    // keep a total for each for each browser.
```
KyleJu, line 362:
```suggestion
    // Keep a total for each for each browser.
```

webapp/views/wpt-results.js
```diff
@@ -241,6 +246,20 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
               </tr>
             </template>
 
+            <template is="dom-if" if="[[ shouldDisplayTotals(displayedTotals, diffRun) ]]">
+              <tr class="totals-row">
+                <td>
+                  <code><strong>Total</strong></code>
+                </td>
+                <template is="dom-repeat" items="[[displayedTotals]]" as="totals">
```
KyleJu, line 254: ditto

webapp/components/test-file-results-table.js
DanielRyanSmith: That's great insight - I did not realize this was happening! I've change it as you suggested

webapp/views/wpt-results.js
DanielRyanSmith: 👌

DanielRyanSmith:
> Minor font differences in Total between testfile view and subtest view.

Good catch! forgot to `<strong>` one of them.

> Still some test count differences. Might be worth filing a separate issue. It shows 0/2 on the test file level but 0/1 on the subtest level.

This is true, and my original approach was to try and show this totals row as a true representation of the number of rows rather that did NOT count the TestHarness status. However, we do count the harness status as a separate subtest (at the moment), so this is a a separate issue. I changed the logic so that the total row should always match the number shown on the directory above. If/when we change the calculation, it can be part of a separate PR.

webapp/components/test-file-results-table.js
```diff
@@ -335,15 +352,56 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     return rows && rows.length && rows[0];
   }
 
+  computeTotals(rows) {
```
KyleJu, line 355: Here is my question for this bit, and perhaps I should have asked a bit early. Now we have removed the `Harness status` tweak, it appears that the total displayed on the test-level[1] is consistent with the subtest-level [2]

[1]
Screen Shot 2022-05-05 at 3 43 58 PM (2)

[2]
Screen Shot 2022-05-05 at 3 44 34 PM (2)

If that is the case, do we have to calculate them ourselves or simple pass it down when a user clicks on the test?

DanielRyanSmith: This is a very good question. I did a more thorough investigation when you mentioned this. The problem arises with the differences in search query results.
The numbers that are displayed in views above the single test level are plucked from the search results array of the entire test suite:
https://github.com/web-platform-tests/wpt.fyi/blob/5cadbbc2df6ac3557f17e6487c0d5f1a5ddd49f2/webapp/views/wpt-results.js#L627-L636
This array is typically huge and representative of the entire number of tests in WPT (right now it's usually around 46,000).
When we view a single test, we separately query for the subtest results that exist in that single test path:
https://github.com/web-platform-tests/wpt.fyi/blob/5cadbbc2df6ac3557f17e6487c0d5f1a5ddd49f2/webapp/components/test-file-results.js#L103-L107
This individual test query doesn't supply this aggregate count of passes and totals.
So then, why don't we just pass down the passes and totals that exist in the test suite search query? The problem is that the results are in the form of this very large array, so we would have to traverse the array to find the correct individual test result data to get these totals. Due to this, I thought it might be better to just take the individual test query data and check the subtest array length in a slightly modified way.

It might be worthwhile to change this on the server side at some point to include a subtest totals count when querying info on a specific test.

KyleJu: >> It might be worthwhile to change this on the server side at some point to include a subtest totals count when querying info on a specific test.

I agree with a fix on the server side. I do see a way to pass along test results without going through the very large array, similar to the way that `path` is being updated through a global variable, but messy. At the end of the day, your way is more accurate on a subtest level

The only thing I am worried is the potential discrepancy between the `total` on the test level and subtest level. We want to make sure that the data is consistent and has a single reliable source of truth. @foolip and I had the same discussions on https://github.com/web-platform-tests/wpt.fyi/issues/2825 but not in the scope of this PR

KyleJu (approved):

KyleJu:
Looking good!

Ms2ger:
I'm disappointed to discover this was added without wider discussion given that we'd explicitly decided not to add this misleading statistic.

PR #2842, opened 2022-04-20
chrome_android icon workaround

KyleJu:
A temporary workaround for the `chrome_android` icon.

### DIFF
--- a/webapp/components/product-info.js
+++ b/webapp/components/product-info.js
@@ -178,6 +178,11 @@ const ProductInfo = (superClass) => class extends superClass {
       if (channel) {
         name = `${name}-${channel}`;
       }
+      // TODO(kyle): A temporary workaround; remove this check when
+      // chrome_android is mapped to chrome on wptrunner.
+      if (name === 'chrome_android') {
+        name = 'chrome';
+      }
     }
     return `/static/${name}_64x64.png`;
   }


WeizhongX (approved):
Thanks for the change.

KyleJu:
Screen Shot 2022-04-20 at 1 48 27 PM (2)

webapp/components/product-info.js
```diff
@@ -178,6 +178,11 @@ const ProductInfo = (superClass) => class extends superClass {
       if (channel) {
         name = `${name}-${channel}`;
       }
+      // TODO(kyle): A temporary workaround; remove this check when
+      // chrome_android is mapped to chrome on wptrunner.
```
foolip, line 182: Is there an open issue for this?

WeizhongX: Looks like we need keep chrome_android as a product in wpt.fyi? We need something in the dropdown menu to pick chrome android, right? Do we have the 'platform' concept on wpt.fyi?

For wptrunner we need keep chrome_android as a different product. Running the test for chrome android is quite different to that of chrome. Maybe one thing we could do is to change the product when creating the wpt report. But is that similar to what we did here?

foolip: We do have a concept of platform, you can see little icons for Linux, macOS and Windows next to the browser icons. We have OSName and OSVersion columns in the TestRun datastore, so it must be possible to query by this and add it to the UI. @KyleJu do you think that sounds feasible?

KyleJu: I think the easiest solution is to keep `chrome_android` as a separate product and add the `Android` as a platform in the wpt report. So we can show the platform icons next to the browser icons like any other browsers on wpt.fyi. Filed a bug here https://github.com/web-platform-tests/wpt/issues/33794

PR #2843, opened 2022-04-21
Triage validate forms

DanielRyanSmith:
## Description
The triage submit button will be disabled until the user has filled out the required fields for submission.

## Changes
- The `Triage` button will be disabled until the user provides a bug URL (or either a bug URL or label) for each selected triage item.
- Tests that can be selected in triage mode will have their links disabled so that the user cannot accidentally change views when selecting triage items.

#### Triage button before changes (enabled with no inputs completed)
Screen Shot 2022-04-21 at 9 29 27 AM

#### Triage button after changes (disabled with no inputs completed)
Screen Shot 2022-04-21 at 9 28 39 AM

#### If the user can provide a URL or Label, only 1 is needed to enable triage submission.
Screen Shot 2022-04-21 at 11 03 21 AM

#### Test file links are not active during triage mode so the user can select the test item without navigating to another view.
Screen Shot 2022-04-21 at 11 14 01 AM


webapp/components/wpt-amend-metadata.js
```diff
@@ -356,6 +365,20 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
     }
   }
 
+  checkEnableTriageButton() {
```
KyleJu, line 368: The setup (used with `on-input`) with this method is rather inefficient. This function is called for every letter typed in any of the input field above. And each time, this function goes through `displayedMetadata`. Ideally this function should only called once for each input field.

DanielRyanSmith: I agree. I was having trouble finding an approach that can check if all fields are filled for a dynamic and unbounded number of input fields that have a state tracked by object pr. Reading about the [event objects model property](https://polymer-library.polymer-project.org/2.0/docs/devguide/templates) gave a better idea for a more succinct solution.

KyleJu: Per offline discussions: 

In terms of an efficient way to check it, I have two approaches in mind - you can use a use counter + onchange (https://www.w3schools.com/jsref/event_onchange.asp), and data bind the use counter to a function that checks if the use counter === the number of input fields.

The second one should be easier - instead of disabling triage button, we can always enable the triage button and check that all input fields are filled when clicks it. If they are not, we prompt user to fill them without closing the dialog. In this case, we don't have to check all input fields dynamically

webapp/components/wpt-amend-metadata.js
```diff
@@ -353,7 +367,38 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
...
+      this.fieldsFilled.numEmpty--;
+      this.fieldsFilled.filled[index] = true;
+    }
+    this.checkEnableTriageButton();
+  }
+
+  checkEnableTriageButton() {
+    // If all triage items have input, triage can be submitted.
+    this.triageSubmitDisabled = (this.fieldsFilled.numEmpty > 0
+      || this.displayedMetadata.length === 0);
```
KyleJu, line 401: `this.displayedMetadata.length === 0` isn't needed because user won't be able to enter the triage dialog if nothing is selected

webapp/components/wpt-amend-metadata.js
```diff
@@ -353,7 +367,38 @@ class AmendMetadata extends LoadingState(PathInfo(ProductInfo(PolymerElement)))
...
+    const label = this.displayedMetadata[index].label;
+
+    // Check if the input is empty.
+    if (url === '' && (label === '' || label === undefined)) {
+      // If the field was previously considered filled, it's now empty.
+      if (this.fieldsFilled.filled[index]) {
+        this.fieldsFilled.numEmpty++;
+      }
+      this.fieldsFilled.filled[index] = false;
+    } else if (!this.fieldsFilled.filled[index]) {
```
KyleJu, line 390: Optional nit: I would explicitly check this.fieldsFilled.filled[index] is false here because `!undefined` `!null` is also true in Javascript

KyleJu (approved):
LGTM % nits

webapp/components/wpt-amend-metadata.js
DanielRyanSmith: Makes sense - I added this 👌

PR #2853, opened 2022-05-02
Add a separate check for Chrome_Android

KyleJu:
Add a separate check for Chrome_Android. In some pages, it is looking for icons `chrome_android-dev_64x64` because it has channel


KyleJu:
Screen Shot 2022-05-02 at 2 51 56 PM (2)

DanielRyanSmith (approved):

PR #2854, opened 2022-05-04
[UI] Add the ability to sort test results

### LINKED ISSUE #2289: Sorting by summary numbers
I have found myself recently writing scripts to create spreadsheets from wpt.fyi data, which have pretty much exactly the same numbers as wpt.fyi itself, but gave me the ability to sort.

wpt.fyi is already like a large table, and a very common UX is being able to interact with column headers to filter and sort. Here's a very crude mockup of what that might look like:
image

P.S. Sorting by columns individually would be easy to understand, and probably sufficient. To sort by something more complicated like "number of tests that fail anywhere" would be nice, but perhaps that would be best done with a helper column showing that computed number, and then being able to filter by that column. So that feels like a separate proposal.

KyleJu:
Fix #2289. Add the ability to sort based on test names and test passes.

Screen Shot 2022-05-05 at 2 59 36 PM


KyleJu:
Add @foolip to review the UI changes and @DanielRyanSmith for the code itself

webapp/views/wpt-results.js
```diff
@@ -861,6 +879,41 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+        return 1;
+      }
+      return 0;
+    })
+
+    this.displayedNodes = sortedNodes;
+  }
+
+  sortTestResults() {
+    const [index] = arguments;
```
DanielRyanSmith, line 902: I thought of some possible changes/suggestions after testing the UI:

- Displaying which row is currently sorted by visually changing the sort arrow (slight color change?)
- (Reverse sorting) Keeping track of what row the data is sorted by and reversing that order if the user selects the row again would feel intuitive and could have additional use. Also changing the arrow to point upward in the case of reverse sorting as an obvious visual cue of what is happening.
- The sort row feels like it should have column labels of some sort, but those column names are essentially in the row above them. I'm not sure the best approach here, but it's something that felt odd when I first looked at it.

Nice work @KyleJu!

DanielRyanSmith:
I thought of some possible changes/suggestions after testing the UI:

- Displaying which row is currently sorted by visually changing the sort arrow (slight color change?)
- (Reverse sorting) Keeping track of what row the data is sorted by and reversing that order if the user selects the row again would feel intuitive and could have additional use. Also changing the arrow to point upward in the case of reverse sorting as an obvious visual cue of what is happening.
- The sort row feels like it should have column labels of some sort, but those column names are essentially in the row above them. I'm not sure the best approach here, but it's something that felt odd when I first looked at it.

Nice work @KyleJu!

webapp/views/wpt-results.js
```diff
@@ -861,6 +879,41 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     this.path = this.searchResults[(n + next) % n].test;
   }
 
+  sortTestname() {
+    if (!this.displayedNodes) {
+      return;
+    }
+    const sortedNodes = this.displayedNodes.slice();
+    sortedNodes.sort(function (a, b) {
+      if (a.path < b.path) {
```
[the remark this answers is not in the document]
KyleJu: good point!

webapp/views/wpt-results.js
```diff
@@ -861,6 +879,41 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+  sortTestResults() {
+    const [index] = arguments;
+    return () => {
+      if (!this.displayedNodes) {
+        return;
+      }
+
+      const sortedNodes = this.displayedNodes.slice();
+      sortedNodes.sort(function (a, b) {
+        return a.results[index].passes - b.results[index].passes;
```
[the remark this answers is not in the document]
KyleJu: https://github.com/web-platform-tests/wpt.fyi/pull/2854#issuecomment-1131828262. I think we decide to go with percentage since it will be the new display. The secondary sort is based on test names.

KyleJu:
I have implemented following suggested changes:

- Sorting by the fraction and a secondary sort by test names;
- clicking the arrow a second time reverses the sorting order

After some experiment with the UI, having a clickable arrow at the bottom right of the browser cell, per @foolip's suggestions, doesn't seem very obvious to users. I do like the idea of highlighting a cell after a sort. What do you think?

webapp/views/wpt-results.js
```diff
@@ -897,6 +929,92 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+    }
+
+    if (pathA > pathB) {
+      return 1;
+    }
+    return 0;
+  }
+
+  sortTestResults() {
+    const [index] = arguments;
```
DanielRyanSmith, line 971: Could `index` just be passed into the function as a param?

webapp/views/wpt-results.js
```diff
@@ -897,6 +929,92 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+    return () => {
+      if (!this.displayedNodes) {
+        return;
+      }
+
+      const sortedNodes = this.displayedNodes.slice();
+      const self = this;
+      sortedNodes.sort(function (a, b) {
+        if (self.sortRow[index]) {
```
DanielRyanSmith, line 980: I think `sortRow` could be represented by an integer here, being the index that is currently sorted on and compare it with the `index` var. The issue would be trying to switch to descending and ascending again when clicking the same column multiple times, but that should still work if you set this var to -1 whenever it was previously a match with the current index. It should remove the need to rebuild an array each time

DanielRyanSmith:
I'm looking at the staging environment and the switch between ascending and descending is working intuitively 😊  nice job!

DanielRyanSmith (changes requested):

webapp/views/wpt-results.js
KyleJu: I agree but `sortRow` array is a cleaner way to reset all other arrows. Every time sortRow is being reassigned, it triggers (through data binding) `<template is="dom-repeat" items="[[sortRow]]" as="sortItem">` to repopulate the arrow icons. I didn't see a good way to do that if `sortRow` is a number

webapp/views/wpt-results.js
```diff
@@ -897,6 +929,92 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+        }
+        // Both 0/0 cases; compare test names.
+        if (a.results[index].total === 0 && b.results[index].total === 0) {
+          return self.compareTestNameDefaultOrder(a, b);
+        }
+
+        // One of them is 0/0; compare passes;
+        if (a.results[index].total === 0 || b.results[index].total === 0) {
+          return a.results[index].passes - b.results[index].passes;
+        }
```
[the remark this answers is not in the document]
KyleJu: Good call!

webapp/views/wpt-results.js
KyleJu: Done

DanielRyanSmith (approved):
Thanks for addressing my comments!

foolip:
> After some experiment with the UI, having a clickable arrow at the bottom right of the browser cell, per @foolip's suggestions, doesn't seem very obvious to users. I do like the idea of highlighting a cell after a sort. What do you think?

If stuffing the arrows into the corner of the cells is messy, this is fine. Can be left as good starter bugs for someone in the future :)

One thing that would be great however is if the sorting is preserved on navigation into a subdirectory. Would that be messy?

Issue #1958, opened 2020-05-14
"Harness status" is shown for reftests


foolip:
This conditional is supposed to show "Test status" for reftests:

https://github.com/web-platform-tests/wpt.fyi/blob/0f000bc65bca1724dc453c63f6b20ff64fd7d5f7/webapp/components/test-file-results.js#L282

I'm sure that once worked, but it has since regressed. https://wpt.fyi/results/css/css-ui/appearance-auto-001.html now looks like this:
![image](https://user-images.githubusercontent.com/498917/81953434-f22a0900-9607-11ea-9869-26b275b67727.png)

Possibly the addition of duration is what broke this, if it's represented as a fake subtest or something.

foolip:
I noticed this again just now, because "Harness status" is incorrectly used as the subtest name when triaging:
https://github.com/web-platform-tests/wpt-metadata/pull/609/files?diff=split&w=1

@Hexcles @KyleJu

foolip:
I manually cleaned that up in https://github.com/web-platform-tests/wpt-metadata/pull/610, and confirmed there aren't any other instances of "Harness status" already in wpt-metadata.

DanielRyanSmith:
Has was fixed a while ago, and is currently not a problem in production 👍

PR #2858, opened 2022-05-09
Remove harness status aggregation and display percentages on Interop-202X scores

DanielRyanSmith:
## Description
Addresses #2825 and #1958

See [RFC #114](https://github.com/web-platform-tests/rfcs/pull/114) for detailed proposal information.

This change adds a new display for how test results are shown on wpt.fyi's Interop-2022 label view. Instead of displaying the flat number of tests that have passed over the test total, a percentage will display with this information. Additionally, a new scoring method has been implemented that no longer counts the Harness status toward the subtest count, and will mark the test as a failure if the test's status was not "OK" or "PASS".

[Here is a staging view](https://scoring-change-dot-wptdashboard-staging.uk.r.appspot.com/results/?run_id=4773272247861248&run_id=4872190042308608&run_id=5869271263477760&run_id=6279564925861888) of these changes in action. The summaries used on the runs in this link were generated with the new aggregation and display method.

[Here is another staging view](https://scoring-change-dot-wptdashboard-staging.uk.r.appspot.com/results/?run_id=4773272247861248&run_id=5255146506289152) displaying similar Chrome results. The **FIRST** run shown on the left is displaying results with the new aggregation method. This is a way to visualize the effect on the totals that will display from older vs. newer aggregations.

## Changes
- "Harness Status" will no longer be displayed on subtest views that do not use TestHarness and have been replaced with "Test Status".
- Interop-2022 results views will display more accurate aggregation that reflects how the scores are calculated.ry next to the directory name.
- Tests that experience harness errors will display a warning with harness error as title text.
- A new aggregation method has been added for creating and interpreting runs and run summary files. A Harness status for a test will no longer be counted toward the subtest total of a test. In addition, a test is marked at 0% passing if the Harness status is not "OK" (and a test's status is not "PASS").  The rationale here is that these error might stop further subtest failures from running, which can hide the scope of the problem. Marking as 0% makes it as visible as possible that something has gone wrong.
**NOTE**: This will result in an overall drop of passing percentages in these scenarios:
  - A test passes some subtests but has a non-"OK" Harness status. This passing percentage will drop to 0%.
  - A test has subtest failures but an "OK" Harness status. This status will not be counted toward the passing percentage. e.g. A test with 1 subtest that fails and an "OK" Harness status will be marked as 0% passing rather than 50%.
- Old summary files that were generated before this change will NOT take this new Harness status aggregation change into account, and so those results will display with the old aggregation method. (which will likely have an artificially inflated percentage when compared to newer summaries.

#### Tests with harness errors display warnings with title text next to results
Screen Shot 2022-06-21 at 10 45 08 AM

#### Interop-2022 results are viewed with an aggregation that is more indicative of actual interop-2022 scores
Screen Shot 2022-06-21 at 10 52 19 AM

#### Harness status will no longer count toward subtest totals
The run in the left column is aggregated using the new method, compared to the old totals displayed in the right column. This change more accurately represents the scope of test failures.
Screen Shot 2022-06-21 at 10 45 58 AM


DanielRyanSmith:
@KyleJu The deployment CI run has been having issues with resources it seems. Everything seems to successfully deploy except the new results processor. Maybe you have seen this problem before?

DanielRyanSmith:
I think it's probably best to separate this into two PRs - one with the visual changes to the UI and one implementing the scoring change. I'll separate them shortly.

api/query/query.go
```diff
@@ -14,9 +14,23 @@ import (
...
+	// Status represents the 1-2 character abbreviation for the status of the test.
+	Status string `json:"s"`
+	// Counts represents the subtest counts (passes and total).
+	Counts []int `json:"c"`
+}
 
+// summary is the golang type for the JSON format in pass/total summary files.
+// It has an old structure and a new structure - each which represent summary files
+// that match the old or new summary format.
+type summary struct {
```
KyleJu, line 28: In this case, `summary` is an unexported variable but with exported fields `Old` and `New`. Let us make it consistently and make it an exported variable

api/query/query.go
```diff
@@ -14,9 +14,23 @@ import (
...
+	// Old summary format. This will be null if the summary is using the new format.
+	Old map[string][]int
+	// New summary format that keeps status information.
+	New map[string]SummaryResult
```
KyleJu, line 32: I don't recommend using the naming of `New`. It is too close to the new in variable initialization

api/query/query.go
```diff
@@ -93,16 +107,25 @@ func (qh queryHandler) loadSummaries(testRuns shared.TestRuns) ([]summary, error
...
+				New: nil,
+			}
 			data, loadErr := qh.loadSummary(testRun)
 			if err == nil && loadErr != nil {
 				err = fmt.Errorf("Failed to load test run %v: %s", testRun.ID, loadErr.Error())
 				return
 			}
-			marshalErr := json.Unmarshal(data, &s)
+			// Try to unmarshal the json using the new aggregation structure.
+			marshalErr := json.Unmarshal(data, &s.New)
```
KyleJu, line 120: A thought for myself: summary files are big and this might increase the overheads of loading runs. Either add a flag somewhere to indict the old or new, or do a page load test

api/query/query_test.go
```diff
@@ -55,8 +56,14 @@ func TestLoadSummaries_success(t *testing.T) {
 		[]byte(`{"/x/y/z":[3,4]}`),
 	}
 	summaries := []summary{
-		map[string][]int{"/a/b/c": []int{1, 2}},
-		map[string][]int{"/x/y/z": []int{3, 4}},
+		{
+			Old: map[string][]int{"/a/b/c": {1, 2}},
```
KyleJu, line 60: Rather than modifying the existing one, we should add new tests to test the `New` format

api/query/search_test.go
```diff
@@ -39,14 +40,19 @@ func doTestIC(t *testing.T, p, q string) {
...
 	summaries := []summary{
-		map[string][]int{
-			"/a" + p + "c": []int{1, 2},
-			p + "c":        []int{9, 9},
-		},
-		map[string][]int{
-			"/z" + p + "c": []int{0, 8},
-			"/x/y/z":       []int{3, 4},
-			p + "c":        []int{5, 9},
+		{
```
KyleJu, line 43: Ditto: Add tests for the new format

api/query/search.go
```diff
@@ -187,21 +187,45 @@ func prepareSearchResponse(filters *shared.QueryFilter, testRuns []shared.TestRu
 	// Dedup visited file names via a map of results.
 	resMap := make(map[string]shared.SearchResult)
 	for i, s := range summaries {
```
KyleJu, line 189: Could you also double check what other queries could be affected? https://github.com/web-platform-tests/wpt.fyi/tree/main/api/query#readme. I will double check as well

webapp/components/test-file-results.js
```diff
@@ -194,7 +194,7 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
 
     // Set name for test-level status entry after subtests discovered.
     // Parameter is number of subtests.
-    resultsTable[0].name = this.statusName(resultsTable.length - 1);
+    resultsTable[0].name = this.statusName(resultsTable.length - 2);
```
KyleJu, line 197: Why is this change made?

KyleJu:
Errors from staging deployment (I will take a look):
```
ERROR: (gcloud.app.deploy) Error Response: [9] An internal error occurred while processing task /app-engine-flex/flex_await_healthy/flex_await_healthy>2022-05-26T20:59:07.274Z601.fm.2: 
No sufficient free disk space left for your App Engine Flexible application.
Please increase your VM instances disk size in the resource settings in the
app.yaml file for your deployment and retry.
See https://cloud.google.com/appengine/docs/flexible/python/reference/app-yaml#resource-settings
for how to set disk resource.
```

api/query/query.go
DanielRyanSmith: That's true, and I don't even think the export is required here (it was not previously). I'll  change them to no longer export.

api/query/query.go
DanielRyanSmith: That's a good idea - changed to `oldFormat` and `newFormat`

api/query/query.go
DanielRyanSmith: Yeah, I was reading up about using this approach to check if the summary is old or new, but I'm not certain it's the most efficient. I guess It might be worth doing some timing to figure out what kind of time this adds.

webapp/components/test-file-results.js
DanielRyanSmith: This is a separate small change that addresses [an old labeling issue](https://github.com/web-platform-tests/wpt.fyi/issues/1958).

api/query/query_test.go
DanielRyanSmith: Yes agreed, added a new test so that both the old and new formats are tested 👍

api/query/search_test.go
DanielRyanSmith: ✅

KyleJu:
> Errors from staging deployment (I will take a look):
> 
> ```
> ERROR: (gcloud.app.deploy) Error Response: [9] An internal error occurred while processing task /app-engine-flex/flex_await_healthy/flex_await_healthy>2022-05-26T20:59:07.274Z601.fm.2: 
> No sufficient free disk space left for your App Engine Flexible application.
> Please increase your VM instances disk size in the resource settings in the
> app.yaml file for your deployment and retry.
> See https://cloud.google.com/appengine/docs/flexible/python/reference/app-yaml#resource-settings
> for how to set disk resource.
> ```

OK I cleaned up some resources and got CI to build. I have proposed a solution for this issue in #2867.

davidsgrogan:
The description in this PR says it fixes https://github.com/web-platform-tests/wpt.fyi/issues/2825, but I don't think that's accurate. The request in 2825 is for "the numbers in the table are the actual scores used for Interop2022." AFAICT, this PR just shows the % in the cells but does not take Interop2022 into account.

foolip:
If combined with filtering by label, we get something that's similar:
https://scoring-change-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&q=label%3Ainterop-2022-forms

(If you see "Failed to fetch test runs" that's a resource issue on staging instances, you might have to imagine it working.)

But actually, this doesn't do a perfect job of making it clear where the biggest wins are. Starting at the top level and drilling down, the score is 0-100% at each level, so a few levels down it will still require some math to know how much the Interop 2022 score would increase by fixing all the tests in view.

The original suggestion of "43.95 / 90" in https://github.com/web-platform-tests/wpt.fyi/issues/2825 would give numbers that can be compared between subdirectories, since any 10 tests would contribute as much to the score. Although one still needs to compute `(90 - 43.95) / number_of_tests` to figure out the score improvement.

shared/models.go
```diff
@@ -480,6 +480,12 @@ type LegacySearchRunResult struct {
 	Passes int `json:"passes"`
 	// Total is the total number of test results for this run/file pair.
 	Total int `json:"total"`
+	// Status represents either the test status or harness status.
+	// This will be an empty string for old summaries.
+	Status string `json:"status"`
+	// NewAggProcess represents whether the summary was created with the old
+	// or new aggregation process.
+	NewAggProcess bool `json:"newAggProcess"`
```
jcscottiii, line 488: Quick question: do we anticipate changing the aggregation process again in the future? If so, do we want to change this to be `AggProcessVersion int`

webapp/views/wpt-results.js
```diff
@@ -855,6 +979,132 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+    return !node.isDir && status && !PASSING_STATUSES.includes(status)
+      && !node.results.every(testInfo => testInfo.singleSubtest);
+  }
+
+  getStatusDisplay(node, index) {
+    let status = node.results[index].status;
+    if (status in STATUS_ABBREVIATIONS) {
+      status = STATUS_ABBREVIATIONS[status];
+    }
+    return status;
```
jcscottiii, line 994: Quick question: In the future, is there a possibility where an additional abbreviation is added to the results processor STATUS_ABBREVIATIONS but has not been added to STATUS_ABBREVIATIONS here. Do we want to add an else conditional that defaults to something like `"Unknown Status: " + status`? Or could the browser cache this version of the javascript while the results processor has added a new abbreviation?

shared/models.go
DanielRyanSmith: The hope is not to change again, and to eventually remove the old format altogether, but I can see the value in adding a variable number of formats. I have an additional focus this quarter to change it so we don't use summaries with old formats, so hopefully we'll soon be back to 1 single format.

results-processor/wptreport.py
```diff
@@ -39,6 +39,20 @@
 # TODO(Hexcles): Remove this when the TC decision task is implemented.
 IGNORED_CONFLICTS = frozenset({'browser_build_id', 'browser_changeset'})
 
+# A map of abbreviations for test statuses. This will be used
+# to convert test statuses to smaller formats to store in summary files.
+STATUS_ABBREVIATIONS = {
```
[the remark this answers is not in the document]
DanielRyanSmith: That's a great idea! I'll update.

foolip (approved):

shared/models.go
DanielRyanSmith: > Does that mean that you will be retroactively correcting old entries to use the new format?

It still needs to be investigated what is the best approach for this, but the initial thought is to invalidate the old summaries and re-aggregate the run when requested using the new method. Regenerating the old summary files would take a lot of time and resources so it's not the ideal approach. See [this comment](https://github.com/web-platform-tests/rfcs/pull/114#discussion_r905634642) for a more in-depth description.

> This means that this variable is going to be temporary and go away once we get back to a single format? If so, I would just make an issue and add it as a TODO comment here

I think that's a good idea - adding a note that the old format should ideally not be used after future changes.

api/query/search_test.go
```diff
@@ -1,3 +1,4 @@
+//go:build small
```
jcscottiii, line 1: Could you remove this line and the same one added to api/query/search.go please? This syntax is only valid for go 1.17+ https://pkg.go.dev/go/build/constraint

DanielRyanSmith: Yes! This was automatically added by my editor - I'll remove it.

jcscottiii (approved):

PR #2866, opened 2022-05-25
Disable "Show History" button

DanielRyanSmith:
This feature has not properly worked for some time, and just serves to show curious users error messages. 😆  It will require a rework if there is a continued interest.


stephenmcgruer:
FWIW, the history button 'sort of' works today, if you refresh the page and try a few times (it seems to also help to wait a few seconds after the page loads):

image

And it can be very useful.

That said, it's definitely pretty broken as well, and so I cannot say that it's overall a positive feature atm 🤣

DanielRyanSmith:
Hey @stephenmcgruer, if this is a feature that is being derived some value out of, then perhaps it is worth keeping enabled for now. I was under the assumption that this was not working and no one was deriving utility from it. Perhaps handling the error message effectively is a better short-term approach?

webapp/views/wpt-results.js
```diff
@@ -288,14 +287,6 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
           <h3>
             History <span>(Experimental)</span>
           </h3>
-          <test-results-chart
```
KyleJu, line 291: Also remove `/webapp/components/test-results-chart.js`. This is the only place that uses `test-results-chart`

KyleJu (approved):
LGTM after removing unused files.

PR #2883, opened 2022-06-19
Interop 2022 - Fix page jumping from browser names

DanielRyanSmith:
[Staging link](https://interop-height-dot-wptdashboard-staging.uk.r.appspot.com/interop-2022)


foolip:
I've tried the staging instance in Chrome, Firefox and Safari and the text doesn't move around 👍

foolip (approved):

PR #2905, opened 2022-07-14
Enable custom runtime for webapp, upgrade webapp and searchcache to go1.18

### LINKED ISSUE #2652: Upgrade Go1.14 as of February 2021
WARNING: Go 1.14 is no longer supported by the Go community as of February 2021. We recommend you to upgrade to the latest version of Go runtime as soon as possible. For details on upgrading, see https://cloud.google.com/appengine/docs/standard/go/runtime.

jcscottiii:
[Staging Link](https://jcscottiii-go1-18-cust-dot-wptdashboard-staging.uk.r.appspot.com/)

## Description
Enable custom runtime for webapp, upgrade webapp and searchcache to go1.18

### Change app.yaml and app.staging.yaml to use custom runtime for webapp.
**Why**: The standard runtime for App Engine only supports up to [go 1.16](https://cloud.google.com/appengine/docs/standard/go/runtime). By moving to custom runtime, we can bring-our-own-container which allows us to control the go version. - There are some dependency updates by dependabot that cannot be upgraded because the latest version of those dependencies only use >=go1.17. We are unable to get the latest security updates because we cannot deploy with go1.17 for the webapp.

- Add Dockerfile (`webapp/web/Dockerfile`) for webapp that produces an image of ~40MB.
- Files affected: `webapp/web/Dockerfile`, `go.mod`, `go.sum`, `Makefile`
- More info about the : `Makefile` [changes](https://github.com/web-platform-tests/wpt.fyi/pull/2905/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R169-R174) that I learned personally: Followed the same special packaging used for searchcache (another go program in the repo). This is needed because the go files needed at build time need to be recopied to a single place. This is because gcloud will only take the files and directories in the location of the app.yaml and nothing from parent directories when deploying a custom runtime application. This special packaging will copy the whole repo again into a folder called wpt.fyi at the same level as app.yaml.

### Set go version for webapp and searchcache to 1.18
**Why**: Go supports up to N-2 versions. That means as of July 14, 2022, only go versions 1.18, 1.17 and 1.16 are still supported. In August go1.19 will come up and the window of supported versions shifts. By using 1.18, we will be on the latest. 

- Change the references in the Dockerfiles for webapp and searchcache to use 1.18.
- `docs/upgrading-go.md` - Added docs for future upgrades
- Run `go mod tidy` after upgrading go to 1.18. It does automatic changes to `go.mod` and `go.sum` 
- Files affected: `Dockerfile`, `api/query/cache/service/Dockerfile`, `go.mod`, `go.sum`, `docs/upgrading-go.md`

### Move the final image for searchcache to use [gcr.io/distroless/static-debian11](https://github.com/GoogleContainerTools/distroless/blob/main/base/README.md)
**Why**: The README lists which images to use. `gcr.io/distroless/base:latest` is the base image but not intended to be used. 

### Add go build flags so that the final image does not need C libraries
**Why**: With the distroless image, needed to add the CGO_ENABLED=0 flag. Without it, it would fail to run.

More details on using the distroless images [here](https://github.com/GoogleContainerTools/distroless#examples-with-docker)
- Files affected: `Makefile` and `api/query/cache/service/Dockerfile`

### Leverage nicehttp library to handle static serving
**Why**: The flexible environment (which support custom runtimes) does not support the `handlers` section of app.yaml for serving static content. Luckily, the nicehttp library that we already use can read the handlers section and do that.

The library has the ability to [serve static files](https://github.com/samthor/nicehttp/blob/554bd34ba7d447848631dfc195e96f126105d8aa/gae.go#L148-L153). It only turns on when not [deployed to app engine. It is determined by examining an environment variable.](https://github.com/samthor/nicehttp/blob/554bd34ba7d447848631dfc195e96f126105d8aa/gae.go#L21-L24). We can leverage this same functionality by reset that environment variable to empty at runtime.

More info:
[Documentation of serving static content in flexible environment (new way)](https://cloud.google.com/appengine/docs/flexible/go/serving-static-files#example_of_serving_static_files_with_your_app) vs [Documentation of serving static content in standard environment (old way)](https://cloud.google.com/appengine/docs/standard/go/serving-static-files#configuring_your_static_file_handlers)

Files affected: `webapp/web/app.staging.yaml`, `webapp/web/app.yaml`, `webapp/web/main.go`

### Other changes
- `.dockerignore` - Add a .dockerignore to help prevent redundant builds upon changing the Dockerfile
- `.gitignore` - Added the folders created when running `make package_service` for webapp and searchcache locally

Fixes #2652


KyleJu:
Solid attempt! I learnt a lot from this

Issue #2916, opened 2022-07-27
Migrate from Google Universal Analytics to Google Analytics 4


jcscottiii:
# Current state of the Analytics System

Currently, we leverage Universal Analytics which is due for [deprecation](https://support.google.com/analytics/answer/11583528) in favor of Google Analytics 4. There is a [guide](https://support.google.com/analytics/answer/9304153#add-tag&zippy=%2Cadd-your-tag-using-google-tag-manager%2Cadd-the-global-site-tag-directly-to-your-web-pages) of how to migrate to Google Analytics which involes making changes to the existing gtag.js script. However, our implemenation of Universal Analytics uses analytics.js. Analytics.js is a dated version of  and should migrate to gtag.js before migrating directly to Google Analytics. This [document](https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs) demonstrates the migration path and differences.

# Detailed Plan

- [x] 1. Migrate existing analytics.js script to gtag.js for Universal Analytics. #2917 
  - [x] Switch to using tracking code that uses gtag.js
  - [x] Change all custom events that use `window.ga` or `ga` to `window.gtag` or `gtag`
  - Notes:
    - No longer need to [explicitly](https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_pageviews_with_the_default_tracker) send page view events in gtag.js
    - Events have a different format [(1)](https://developers.google.com/analytics/devguides/migration/ua/analyticsjs-to-gtagjs#measure_events) [(2)](https://developers.google.com/analytics/devguides/collection/gtagjs/events#default-event-categories-and-labels)
    - Exceptions have a different [format](https://developers.google.com/analytics/devguides/collection/gtagjs/exceptions) in terms of the exception parameters.
- [x] 2. Add additional configs to send information to the new Google Analytics 4 properties while also sending data to the existing Universal Analytics property. #2917 
  - Might be advantageous to setup Google Tag Manager to manage environments automatically later on. #2919 
- [ ] 3. Once confimed that everything works, remove the old Universal Analytics tracking code and keep only the Google Analytics tracking. #2921 

cc: @past @DanielRyanSmith @KyleJu

jcscottiii:
After #2927 happened. It was confirmed that the Google Aanlytics 4 wptdashboard is now populating.

past:
[...]

PR #2917, opened 2022-07-27
Move from analytics.js to gtag.js

jcscottiii:
## Description
Part of #2916. Need to move to gtag.js. Instead of jumping straight to
gtag.js for Google Analytics 4, this work moves Universal Analytics to
use gtag.js. That we can test out the same functionality before point
the data to Google Analytics 4 without having to change the syntax.

A future PR will remove the Universal Analytics code.

## Changes
- Add Google Analytics 4 tracking codes for prod and non-prod while also keeping the Universal Analytics code.
- Remove the analytics.js script and place the gtag script from the Universal Analytics dashboard
- Move the exception event handling to the gtag.js. Syntax is a little different. [Docs](https://developers.google.com/analytics/devguides/collection/gtagjs/exceptions)
- Changed Page View events [syntax](https://developers.google.com/analytics/devguides/collection/gtagjs/pages#page_view_event) to work for gtag.js
- Changed other custom events [syntax](https://developers.google.com/tag-platform/gtagjs/reference#event) to work for gtag.js

### DIFF
--- a/webapp/components/self-navigator.js
+++ b/webapp/components/self-navigator.js
@@ -130,9 +130,11 @@ const SelfNavigation = (superClass) => class SelfNavigation extends PathInfo(sup
       .replace(/%3A00.000Z/g, '');
     window.history.pushState(params, '', url);
 
-    // Send Google Analytics pageview event
-    if ('ga' in window) {
-      window.ga('send', 'pageview', path);
+    // Send Google Analytics page_view event
+    if ('gtag' in window) {
+      window.gtag('event', 'page_view', {
+        page_path: path
+      });
     }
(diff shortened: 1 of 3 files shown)


jcscottiii:
See the events going to the new staging property:

image

Temporarily forced the events to go to the new prod property

image

For the old Universal Analytics property: Activity is still flowing but events are not registering. Will investigate
image

DanielRyanSmith (approved):
Looks great! Thanks for the detailed explanation and screenshots of things functioning. I'm not certain where we are on analytics for WebStatus, but this work might also be useful there.

PR #2920, opened 2022-07-28
Use searchcache results if summary file is old format

DanielRyanSmith:
Over changes:
- adds a check before summary files are used to determine if the summaries are in the new format (they have the `_v2` name suffix in their bucket URL).
- Relegates any queries that involve old summary files to the search cache, and will be processed at runtime if needed.
- Updates mock summary data used in test files to use the new summary format.

**Note**: The summary files in staging and production are currently being generated in the new summary file format and the `_v2` name suffix, so this change will have less impact needing to compute search results without summary files.

Decision process for handling queries:
Summary File Decision

## Major change explanations by file

`api/query/query.go`:
- New function `validateSummaryVersions` which checks each summary file URL associated with a test run and ensures it ends with the `-summary_v2.json.gz` suffix.
- Changes the struct `SummaryResult` to no longer hold an old format and new format for summary files, since we will no longer be using old format summary files.
- Rename `getRedisKey` function to `getSummaryFileRedisKey` to be more specific, as well as changing the existing redis key format.

`api/query/search.go`:
- Add a check to see if all summary files needed are in the new format. If not, treat the request as a "complex query" and use the search cache and to load the required data, or aggregate it as needed (this process does not use the summary files).
- Refactor some existing code to use Go best practices.
- Remove some logic to read old format summary files (non-complex queries), leaving only new format parsing.

`shared/util.go`
- Changed regex to require `_v2` suffix rather than allowing `_v2` as optional for summary files.

## Testing file changes

`api/query/query_test.go`
`api/query/search_test.go`
`shared/models_test.go`
`shared/test_run_query_medium_test.go`
`shared/util_test.go`
`webapp/components/test/fixtures/interop.json`
`util/populate_dev_data.go`
`webapp/components/test/fixtures/passrates.json`
`webapp/components/test/test-file-results.html`
`webapp/components/test/util/helpers.js`
- Change tests to only test using the new summary file format.

### A number of summary files exist in the repo to test against. These were previously in the old format and have been replaced with the new format.
`webapp/static/24278ab617/chrome[experimental]-summary_v2.json.gz`
`webapp/static/24278ab617/chrome[experimental].json`
`webapp/static/24278ab617/chrome[stable]-summary_v2.json.gz`
`webapp/static/24278ab617/chrome[stable].json`
`webapp/static/24278ab617/edge[experimental]-summary_v2.json.gz`
`webapp/static/24278ab617/edge[experimental].json`
`webapp/static/24278ab617/edge[stable]-summary_v2.json.gz`
`webapp/static/24278ab617/edge[stable].json`
`webapp/static/24278ab617/firefox[experimental]-summary_v2.json.gz`
`webapp/static/24278ab617/firefox[stable]-summary_v2.json.gz`
`webapp/static/24278ab617/firefox[stable]-summary_v2.json.gz`
`webapp/static/24278ab617/firefox[stable].json`
`webapp/static/24278ab617/safari[experimental]-summary_v2.json.gz`
`webapp/static/24278ab617/safari[experimental].json`
`webapp/static/24278ab617/safari[stable]-summary_v2.json.gz`
`webapp/static/24278ab617/safari[stable].json`

## Docs file changes
`api/README.md`
`docs/gcs.md`
- Add suffix `_v2` to summary file URL examples.

`webapp/components/test-file-results.js`
- Add a comment to appropriately explain the `resultsURL` method.


api/query/search.go
```diff
@@ -99,6 +99,25 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+		r2 := *r
+		r2url := *r.URL
+		r2.URL = &r2url
+		r2.Method = "GET"
+		runIDStrs := make([]string, 0, len(rq.RunIDs))
+		for _, id := range rq.RunIDs {
+			runIDStrs = append(runIDStrs, strconv.FormatInt(id, 10))
+		}
+		runIDsStr := strings.Join(runIDStrs, ",")
+		r2.URL.RawQuery = fmt.Sprintf("run_ids=%s&q=%s", url.QueryEscape(runIDsStr), url.QueryEscape(simpleQ.Pattern))
```
jcscottiii, line 111: r2.URL has a function called `[Query()](https://pkg.go.dev/net/url#URL.Query)`.
You should use that to get that existing query params.

Once you have that, you can use the [Add function](https://pkg.go.dev/net/url#Values.Add). And then re-encode it
```go
// do your loop
q.Add("run_ids",strconv.FormatInt(id, 10))

// any other query params

// finally add it back in
r2.URL.RawQuery = q.Encode()

```

api/query/search.go
```diff
@@ -99,6 +99,25 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
 			val, _ := shared.ParseBooleanParam(q, param)
 			isSimpleQ = isSimpleQ && (val == nil || !*val)
 		}
+		r2 := *r
```
jcscottiii, line 102: you should use request [clone()](https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/net/http/request.go;l=372)

api/query/query.go
```diff
@@ -61,6 +61,30 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return nil, err
+	}
+	versionFiles := make([]string, 0, len(testRuns))
+	for _, testRun := range testRuns {
+		summaryURL := shared.GetResultsURL(testRun, "")
+		url := summaryURL[:len(summaryURL)-16] + "_version.txt"
+		mkey := getRedisKey(testRun)
+		var data []byte
+		qh.dataSource.Get(mkey, url, &data)
```
jcscottiii, line 82: should there be an error check here? if the error doesn't need to be returned up, just log a warning

api/query/query.go
```diff
@@ -61,6 +61,30 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+	}
+	testRuns, filters, err := qh.getRunsAndFilters(filters)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
+		return nil, err
+	}
+	versionFiles := make([]string, 0, len(testRuns))
+	for _, testRun := range testRuns {
+		summaryURL := shared.GetResultsURL(testRun, "")
+		url := summaryURL[:len(summaryURL)-16] + "_version.txt"
```
jcscottiii, line 79: could you make 16 a constant and a comment about it please? i'm trying to learn about this offset.

api/query/query.go
```diff
@@ -61,6 +61,30 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+func (qh queryHandler) getVersionFiles(w http.ResponseWriter, r *http.Request) ([]string, error) {
+
+	filters, err := shared.ParseQueryFilterParams(r.URL.Query())
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return nil, err
+	}
+	testRuns, filters, err := qh.getRunsAndFilters(filters)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusNotFound)
```
jcscottiii, line 73: This is a style thing but it's usually safer to write to the `w http.ResponseWriter` in the initial handler function that receives the request. ResponseWriter should only be written to once in the request's life cycle. When a responseWriter is written to in a sub function, it makes it hard to know if responseWriter has already been written to. It's better to raise the error up and let the handler function write it. otherwise, you'll need to let other developers know not to write to responseWriter again.

DanielRyanSmith:
Thanks for taking a look James! I was hoping to view the changes deployed to a staging environment to see how they interact with a functional searchcache component. I usually add useful commenting and context for review when things feel more finalized, so I hope this wasn't too much of a mess to understand. 🙂

api/query/query.go
DanielRyanSmith: I see. So just to confirm I understand, it's best to pass back the error and handle the error writing in the calling function?

api/query/query.go
DanielRyanSmith: Sure, I'll update this to explain what is going on here. It's unintuitive (and probably not the best approach), but it definitely deserves an explanation 🙂

api/query/query.go
DanielRyanSmith: I'll add a debug log here - it's not unlikely that there should be an error fetching this file and it simply means that this is associated with an old summary file and we should disregard that file.

api/query/search.go
DanielRyanSmith: Would this be correct?
`r2 = r.Clone(r.Context())`

api/query/search.go
DanielRyanSmith: The issue here is that these run IDs live in `rq`, which is a `RunQuery` struct, and the run IDs aren't necessarily in the original query.
https://github.com/web-platform-tests/wpt.fyi/blob/8bf23a6f68d18acab002aa6a613fc5660afb0a85/api/query/atoms.go#L56-L62

That being said, I copied most of this code over from the another section of this function and I'm likely doing plenty wrong 😅 
https://github.com/web-platform-tests/wpt.fyi/blob/8bf23a6f68d18acab002aa6a613fc5660afb0a85/api/query/search.go#L149-L163

api/query/search.go
jcscottiii: That's right!

api/query/query.go
jcscottiii: Correct!

api/query/search.go
jcscottiii: i was thinking something like this:

```go
// Clone the existing request
r2 := r.Clone(r.Context())

// Extract the existing query parameters
q := r.URL.Query()

// Add specific query parameters that are missing [1/2]
for _, id := range rq.RunIDs {
  q.Add("run_ids", strconv.FormatInt(id, 10))
}

// Add specific query parameters that are missing [2/2]
q.Add("q", url.QueryEscape(simpleQ.Pattern))

// Encode into back into RawQuery
r2.URL.RawQuery = q.Encode()
```

api/query/query.go
```diff
@@ -61,28 +61,30 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
 	return &filters, testRuns, summaries, nil
 }
 
-func (qh queryHandler) getVersionFiles(w http.ResponseWriter, r *http.Request) ([]string, error) {
-
+func (qh queryHandler) validateSummaryVersions(w http.ResponseWriter, r *http.Request) (bool, error) {
```
jcscottiii, line 64: ResponseWriter is not needed anymore in the function.
You could change the signature to pass in only `query string`

api/query/query.go
```diff
@@ -61,6 +62,32 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+	for _, testRun := range testRuns {
+		summaryURL := shared.GetResultsURL(testRun, "")
+		// All summary URLs end with "-summary.json.gz". We remove this and
+		// add the correct suffix for the version file.
+		const suffixOffset = 16
+		url := summaryURL[:len(summaryURL)-suffixOffset] + "_version.txt"
+		mkey := getRedisVersionKey(testRun)
+		var data []byte
+		err := qh.dataSource.Get(mkey, url, &data)
+		if len(data) == 0 || err != nil {
```
jcscottiii, line 84: Since `len(data) == 0` is a condition where `validateSummaryVersions` needs to return, you should return an error. 
This function returns either: `false, err` or `true, nil`. By returning an error for this case, you can simplify the return type to just `error` instead of `(bool, error)`. (A bad alternative would be changing the name of this function to `areSummaryVersionsValidated` and return just a bool. But you would need to do a lot of logging since removing the error loses a lot of useful data.)

An example of splitting that out:
```go
if err !=nil {
    return err
} else if len(data) == 0 {
    return fmt.Errorf("custom error message: %s", mkey)
}
```

Callers of `validateSummaryVersions` can have a simple expectation that an error means something happened unexpectedly if an error is returned. If the caller needs to differentiate how to proceed after getting an error from `validateSummaryVersions`, it can always compare the errors.

jcscottiii (changes requested):

api/query/query.go
DanielRyanSmith: I see, that's true. I'll add an accurate error to this case.

api/query/query.go
DanielRyanSmith: True - I refactored it 👍

api/query/query.go
```diff
@@ -81,8 +81,10 @@ func (qh queryHandler) validateSummaryVersions(v url.Values) (bool, error) {
 		mkey := getRedisVersionKey(testRun)
 		var data []byte
 		err := qh.dataSource.Get(mkey, url, &data)
-		if len(data) == 0 || err != nil {
+		if err != nil {
 			return false, err
+		} else if len(data) == 0 {
+			return false, fmt.Errorf("No version found for : %v", mkey)
```
jcscottiii, line 87: since mkey is a string, you should use the %s [verb](https://pkg.go.dev/fmt#hdr-Printing) instead of the %v verb

api/query/search.go
```diff
@@ -99,51 +110,26 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
-		defer resp.Body.Close()
-		w.WriteHeader(resp.StatusCode)
-		_, err = io.Copy(w, resp.Body)
-		if err != nil {
-			logger.Errorf("Error forwarding response payload from search cache: %v", err)
-		}
-		return
+	// Use searchcache for a complex query or if old summary files exist.
+	if !isSimpleQ {
+		sh.useSearchcache(w, r, data, logger)
```
jcscottiii, line 132: This function writes to the response writer. You should return the appropriate error / success and write it to the response writer at this level.

DanielRyanSmith: Good idea! I'll refactor to do so.

api/query/search.go
```diff
@@ -99,68 +103,97 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
 	// handler.
-	r2 := *r
+	r2 := r.Clone(r.Context())
 	r2url := *r.URL
 	r2.URL = &r2url
 	r2.Method = "GET"
-	runIDStrs := make([]string, 0, len(rq.RunIDs))
-	for _, id := range rq.RunIDs {
-		runIDStrs = append(runIDStrs, strconv.FormatInt(id, 10))
+	r2.URL.RawQuery = fmt.Sprintf("run_ids=%s&q=%s",
```
jcscottiii, line 158: You should construct a new [`url.Values`](https://pkg.go.dev/net/url#Values), add the query params to it and use the [Encode](https://pkg.go.dev/net/url#Values.Encode) to safely build the query string

jcscottiii (changes requested):

api/query/search.go
```diff
@@ -99,68 +103,97 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+		url.QueryEscape(runIDsStr), url.QueryEscape(simpleQ.Pattern))
+
+	unstructuredSearchHandler{queryHandler: sh.queryHandler}.ServeHTTP(w, r2)
+}
+
+func (sh structuredSearchHandler) useSearchcache(w http.ResponseWriter, r *http.Request, data []byte, logger shared.Logger) (*http.Response, error) {
+	hostname := sh.api.GetServiceHostname("searchcache")
+	// TODO: This will not work when hostname is localhost (http scheme needed).
```
jcscottiii, line 166: I know this TODO is already here but could you log a github issue and add the github issue link/number here please?

api/query/query.go
```diff
@@ -61,6 +63,26 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+	testRuns, filters, err := qh.getRunsAndFilters(filters)
+	if err != nil {
+		return false, err
+	}
+
+	for _, testRun := range testRuns {
+		summaryURL := shared.GetResultsURL(testRun, "")
+		// All new summary URLs end with "-summary_v2.json.gz".
+		if !strings.HasSuffix(summaryURL, "-summary_v2.json.gz") {
+			return false, nil
```
KyleJu, line 80: Nit: Add a logging here for debugging purpose. Print out the file format

api/query/search.go
```diff
@@ -99,68 +103,97 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
-		req, err := http.NewRequest("POST", fwdURL.String(), bytes.NewBuffer(data))
+		r2 := r.Clone(r.Context())
+		r2.Method = "GET"
+		r2.URL.RawQuery = q.Encode()
+		summariesValid, err := sh.validateSummaryVersions(r2.URL.Query())
 		if err != nil {
-			logger.Errorf("Failed to create request to POST %s: %v", fwdURL.String(), err)
-			http.Error(w, "Error connecting to search API cache", http.StatusInternalServerError)
-			return
+			logger.Debugf("Error checking version files: %v", err)
```
KyleJu, line 119: If the file format is invalid, we should return `StatusInternalServerError` here

api/query/query.go
```diff
@@ -145,7 +167,7 @@ func (qh queryHandler) loadSummary(testRun shared.TestRun) ([]byte, error) {
 }
 
 func getRedisKey(testRun shared.TestRun) string {
```
KyleJu, line 169: nit: method name could be more descriptive, e.g. getSummaryFileRedisKey

api/query/query.go
```diff
@@ -145,7 +167,7 @@ func (qh queryHandler) loadSummary(testRun shared.TestRun) ([]byte, error) {
...
-	return "RESULTS_SUMMARY-" + strconv.FormatInt(testRun.ID, 10)
+	return "RESULTS_SUMMARY-v2-" + strconv.FormatInt(testRun.ID, 10)
```
KyleJu, line 170: Just a thought - what if we make the `_v2` consistent here, since `-summary_v2.json.gz`, e.g. `RESULTS_SUMMARY_v2-`

api/query/query.go
```diff
@@ -61,6 +63,26 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
 	return &filters, testRuns, summaries, nil
 }
 
+func (qh queryHandler) validateSummaryVersions(v url.Values) (bool, error) {
+	filters, err := shared.ParseQueryFilterParams(v)
+	if err != nil {
+		return false, err
+	}
+	testRuns, filters, err := qh.getRunsAndFilters(filters)
```
KyleJu, line 71:
```suggestion
	testRuns, _, err := qh.getRunsAndFilters(filters)
```

KyleJu (commented):
Good job! Just a few thoughts:

Overall it would be great if you could break down changes based on different services - one for searchcache and one for processor. (I understand the change to processor is rather minimal in this case). It easier to test/deploy/revert one service at a time.

On the design side, it seems to me that we have to validate new summary files for both webapp and searchcache? Does it mean that we have to direct all the traffic to searchcache first? or we only check it for some pages, like interop202x tests?

Alternatively, we can pack the changes in [api/query/search.go](https://github.com/web-platform-tests/wpt.fyi/pull/2920/files#diff-63d86010e4d9068ff8d8224425d39d41053c11e5e1b4eeb15c4d070249992be7) to an API endpoint, e.g. /has-summary-file that checks if a file exist. The reason being that webapp can scale and handle traffic much better than searchcache and much less resource intensive. Also by doing that, we can pretty much keep the searchcache logic intact, rather than change the `isSimpleQ` logic

api/query/query.go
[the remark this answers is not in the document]
DanielRyanSmith: We definitely will need to flush the cache prior to landing this change. I'm not sure of the protocol we have set up currently for cache data expiration, but this will certainly invalidate all of these old keys.

api/query/search.go
DanielRyanSmith: It's true, and I realized I'm already doing this when parsing `isSimpleQ` to check the summary file names, so I pulled it out the code block to make sure I'm not duplicating the effort and it works for both.

api/query/search.go
```diff
@@ -99,68 +103,97 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+	r2.URL.RawQuery = fmt.Sprintf("run_ids=%s&q=%s",
+		url.QueryEscape(runIDsStr), url.QueryEscape(simpleQ.Pattern))
+
+	unstructuredSearchHandler{queryHandler: sh.queryHandler}.ServeHTTP(w, r2)
+}
+
+func (sh structuredSearchHandler) useSearchcache(w http.ResponseWriter, r *http.Request, data []byte, logger shared.Logger) (*http.Response, error) {
+	hostname := sh.api.GetServiceHostname("searchcache")
+	// TODO: This will not work when hostname is localhost (http scheme needed).
+	fwdURL, _ := url.Parse(fmt.Sprintf("https://%s/api/search/cache", hostname))
```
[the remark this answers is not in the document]
DanielRyanSmith: Got it, added logging 👍

api/query/search.go
DanielRyanSmith: This is a comment I moved that existed previously, so I'm not sure of the context here, but I've created [an issue](https://github.com/web-platform-tests/wpt.fyi/issues/2941) and I'll link it in the comment.

api/query/query.go
DanielRyanSmith: Added 👍

api/query/search.go
DanielRyanSmith: I didn't do this just in case there's a corner case in which the summary files are not available for some reason but the search cache can still be used. Also, this was a poor error wording on my part from an old approach and I've changed it to be more clear.

api/query/query.go
DanielRyanSmith: name changed 👍

api/query/query.go
DanielRyanSmith: Seems like a good idea - changed

api/query/query.go
DanielRyanSmith: Added 👍

api/query/search.go
```diff
@@ -99,68 +103,97 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+		for _, id := range rq.RunIDs {
+			q.Add("run_id", strconv.FormatInt(id, 10))
+		}
 
-		client := sh.api.GetHTTPClientWithTimeout(time.Second * 15)
-		req, err := http.NewRequest("POST", fwdURL.String(), bytes.NewBuffer(data))
+		r2 := r.Clone(r.Context())
+		r2.Method = "GET"
+		r2.URL.RawQuery = q.Encode()
+		summariesValid, err := sh.validateSummaryVersions(r2.URL.Query())
```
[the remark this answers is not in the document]
DanielRyanSmith: I made a change to do this in a reusable way, since we also need this for the `unstructuredSearchHandler.ServeHTTP` call as well

DanielRyanSmith:
Thanks for taking a look @jcscottiii and @KyleJu 😊 
Sorry I took so long to respond!

> Overall it would be great if you could break down changes based on different services - one for searchcache and one for processor. (I understand the change to processor is rather minimal in this case). It easier to test/deploy/revert one service at a time.

I did a bit of this 

> On the design side, it seems to me that we have to validate new summary files for both webapp and searchcache? Does it mean that we have to direct all the traffic to searchcache first?

The summary files are not validated on webapp, but the summary URL is sliced and used to determine the URL for getting single test data results, because the single test data is prefaced with the same name as the summary (I've updated the method in `test-file-results.js` with a comment description to better explain what's happening there, as it wasn't obvious). For any test run query outside of single test views, the back end is sent a request and determines how to process the query.

When determining `isSimpleQ`, there is some additional work having to look at the summary file URLs and see if they have the correct suffix. However, this is relatively negligible and the possible strain comes from having to disregard the old summary files and aggregate the test results at query time, which is something that would need to happen even if we tried to determine the summary file viability on the webapp.
(I hope this makes sense - it's a convoluted explanation 😅 )

DanielRyanSmith:
@jcscottiii @KyleJu This PR is ready for review once again. I've tried to thoroughly explain the changes file-by-file in the description above. Hopefully that makes it a bit easier to understand the approach. Let me know if there's anything I can do to make this easier!

api/query/search.go
```diff
@@ -85,82 +85,112 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+	// Prepare logging.
+	ctx := sh.api.Context()
+	logger := shared.GetLogger(ctx)
+
 	var simpleQ TestNamePattern
+
+	r2 := r.Clone(r.Context())
+	r2url := *r.URL
+	r2.URL = &r2url
+	r2.Method = "GET"
```
jcscottiii, line 97: nit. use http.MethodGet

api/query/query.go
```diff
@@ -61,6 +57,27 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
...
+	}
+	testRuns, _, err := qh.getRunsAndFilters(filters)
+	if err != nil {
+		return false, err
+	}
+
+	for _, testRun := range testRuns {
+		summaryURL := shared.GetResultsURL(testRun, "")
+		// All new summary URLs end with "-summary_v2.json.gz".
+		if !strings.HasSuffix(summaryURL, "-summary_v2.json.gz") && !strings.HasSuffix(summaryURL, "-summary.json.gz") {
```
jcscottiii, line 73: In your change explanation, it says: `...ensures it ends with the -summary_v2.json.gz suffix`

`!strings.HasSuffix(summaryURL, "-summary_v2.json.gz") && !strings.HasSuffix(summaryURL, "-summary.json.gz")` will allow both v2 and v1.

Shouldn't v1 not be supported anymore?

api/query/query.go
```diff
@@ -61,6 +57,27 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
 	return &filters, testRuns, summaries, nil
 }
 
+func (qh queryHandler) validateSummaryVersions(v url.Values, logger shared.Logger) (bool, error) {
```
jcscottiii, line 60: You can return an error here instead of a boolean and error. This will simplify the signature of the method. For the most part every time you return false, you also return an error. The only major change would be at line 75. You should return a custom error.

More context around the change: From the caller's perspective, a method named "validate[SomeName]", it makes sense to return an error to indicate success or failure. But with a boolean and an error, there's an extra step the caller needs to do to ensure it really was validated.

api/query/search.go
```diff
@@ -85,82 +85,112 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
-		_, err = io.Copy(w, resp.Body)
-		if err != nil {
-			logger.Errorf("Error forwarding response payload from search cache: %v", err)
 		}
 		return
 	}
 
+	q = r.URL.Query()
+	q.Set("q", simpleQ.Pattern)
+	q.Set("run_ids", runIDsStr)
```
jcscottiii, line 150: I just realized above we use run_id for the query string parameter and here we use run_ids for the query string parameter. Is it possible to use only one? (Let me know if this might be a breaking change)

api/query/search.go
```diff
@@ -85,82 +85,112 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
+	logger := shared.GetLogger(ctx)
+
 	var simpleQ TestNamePattern
+
+	r2 := r.Clone(r.Context())
+	r2url := *r.URL
+	r2.URL = &r2url
+	r2.Method = "GET"
+	q := r.URL.Query()
+	q.Add("q", simpleQ.Pattern)
```
jcscottiii, line 99: nit: for the various query string parameter keys, we should declare constants

jcscottiii (changes requested):

api/query/query.go
DanielRyanSmith: Yes, you're correct. I'm glad you caught this, as it's a leftover from the previous change that allows both but writes as `_v2` 😅. I'll fix this and write a test so that this would be caught in the future.

api/query/query.go
DanielRyanSmith: I see, and I agree with you. I made the change to the best of my knowledge (hopefully it's understandable).

api/query/search.go
DanielRyanSmith: This is something that I've had issues with for earlier changes. I didn't personally write this logic, and so it's confusing when `run_id` and `run_ids` are used, and to my knowledge they are not necessarily interchangeable. It might be worth filing an issue for investigation later.

api/query/search.go
DanielRyanSmith: Yes, that would be useful, and might be best stored somewhere like `shared/params.go`. Maybe this is a possible code cleanup issue? I'm not sure the scope of this change, but it would probably be worth handling in all places.

api/query/search.go
DanielRyanSmith: Gotcha, changed 👍

api/query/query.go
```diff
@@ -61,6 +65,31 @@ func (qh queryHandler) processInput(w http.ResponseWriter, r *http.Request) (*sh
 	return &filters, testRuns, summaries, nil
 }
 
+func (qh queryHandler) validateSummaryVersions(v url.Values, logger shared.Logger) SummaryError {
```
jcscottiii, line 68: you can return type `error` instead of SummaryError

api/query/search.go
```diff
@@ -85,82 +85,112 @@ func (sh structuredSearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reque
...
-
-		client := sh.api.GetHTTPClientWithTimeout(time.Second * 15)
-		req, err := http.NewRequest("POST", fwdURL.String(), bytes.NewBuffer(data))
-		if err != nil {
-			logger.Errorf("Failed to create request to POST %s: %v", fwdURL.String(), err)
-			http.Error(w, "Error connecting to search API cache", http.StatusInternalServerError)
-			return
+		// Check old summary files. If any can't be found,
+		// use the searchcache to aggregate the runs.
+		summaryErr := sh.validateSummaryVersions(r2.URL.Query(), logger)
```
jcscottiii, line 125: this becomes:

```go
// Check old summary files. If any can't be found,
// use the searchcache to aggregate the runs.
err = sh.validateSummaryVersions(r2.URL.Query(), logger)
if err != nil {
   switch err {
   case errors.Is(err, ErrBadSummaryVersion):
      logger.Warnf("%s yields unsupported summary version. %s", r2.URL.Query().Encode(), err.Error()) // Maybe this should be Debugf?
      isSimpleQ = false
   default:
      logger.Debugf("Error checking summary file names: %v", err)
      // for all other errors, do we want to do nothing to simpleQ?
    }

}

```

jcscottiii (changes requested):
I added comments about the custom error part. I forgot to mention the ability to wrap errors and assert for errors earlier. Sorry about that.

Wrapping errors allows you to take an error and add more information to it. Then with `errors.Is`, it can unwrap the error and do the type assertion for you on your custom error.

More details about it:
- https://go.dev/blog/go1.13-errors
- Working with Error Wrapping [slides](https://s3.amazonaws.com/bizzabo.file.upload/JROQI70xSP24QPpSO26l_J%20Amsterdam%20-%20Working%20with%20Errors.pdf)


I have also created a go playground for you to test out if you want. https://go.dev/play/p/SYd3AUfKX1H

api/query/search.go
jcscottiii: You're right! Apologies. Added https://github.com/web-platform-tests/wpt.fyi/issues/2957

api/query/search.go
DanielRyanSmith: Originally, the method returned false if any error occurred, so we'll set `isSimpleQ` to false if any error is present.

Also, I was technically having issues here because `errors.Is()` returns a boolean, which is not consistent with comparing against the error, so I switched this to an if-else block.

DanielRyanSmith:
Thanks for the error handling help @jcscottiii 😊 maybe looks better now

jcscottiii (approved):

api/query/search.go
jcscottiii: you're right. in my example you would need to change the line `switch err` to `switch` (remove the variable to compare). Using an if-statement is completely fine too!

KyleJu (approved):
Looks good! LGTM

DanielRyanSmith:
Thanks for all the help @jcscottiii @KyleJu 😊

donatocayurin:
Thank you ❤️🙏

donatocayurin:
❤️🙏

PR #2939, opened 2022-08-16
Write new summaries as "_v2"

DanielRyanSmith:
Part of deprecating old summary formats. This change will now write new summary files with the `_v2` suffix, but will use any summary file, old or new. Additional checks will later be added to verify this suffix exists before using the file.

The plan is to ship this change and allow the new summaries to be written with the new filename, which will limit the need to delegate the work to the searchcache, as summary files are rarely referenced at later dates.

### DIFF
--- a/shared/util.go
+++ b/shared/util.go
@@ -144,7 +144,7 @@ func GetResultsURL(run TestRun, testFile string) (resultsURL string) {
 		// Assumes that result files are under a directory named SHA[0:10].
 		resultsBase := strings.SplitAfter(resultsURL, "/"+run.Revision)[0]
 		resultsPieces := strings.Split(resultsURL, "/")
-		re := regexp.MustCompile("(-summary)?\\.json\\.gz$")
+		re := regexp.MustCompile("(-summary(_v2)?)?\\.json\\.gz$")
 		product := re.ReplaceAllString(resultsPieces[len(resultsPieces)-1], "")
 		resultsURL = fmt.Sprintf("%s/%s/%s", resultsBase, product, testFile)
 	}
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -270,7 +270,8 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
 
   resultsURL(testRun, path) {
     path = this.encodeTestPath(path);
-    // This is relying on the assumption that result files end with '-summary.json.gz'.
+    // This is relying on the assumption that result
+    // files end with '-summary.json.gz' or '-summary_v2.json.gz'.
     let resultsSuffix = '-summary.json.gz';
     if (!testRun.results_url.contains('-summary')) {
       resultsSuffix = '-summary_v2.json.gz';
--- a/webapp/components/test-file-results.js
+++ b/webapp/components/test-file-results.js
@@ -273,7 +273,7 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
     // This is relying on the assumption that result
     // files end with '-summary.json.gz' or '-summary_v2.json.gz'.
     let resultsSuffix = '-summary.json.gz';
-    if (!testRun.results_url.contains('-summary')) {
+    if (!testRun.results_url.includes(resultsSuffix)) {
       resultsSuffix = '-summary_v2.json.gz';
     }
     const resultsBase = testRun.results_url.slice(0, testRun.results_url.lastIndexOf(resultsSuffix));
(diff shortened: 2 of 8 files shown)


webapp/components/test-file-results.js
```diff
@@ -270,8 +270,13 @@ class TestFileResults extends WPTFlags(LoadingState(PathInfo(
 
   resultsURL(testRun, path) {
     path = this.encodeTestPath(path);
-    // This is relying on the assumption that result files end with '-summary.json.gz'.
-    const resultsBase = testRun.results_url.slice(0, testRun.results_url.lastIndexOf('-summary.json.gz'));
+    // This is relying on the assumption that result
+    // files end with '-summary.json.gz' or '-summary_v2.json.gz'.
+    let resultsSuffix = '-summary.json.gz';
+    if (!testRun.results_url.contains('-summary')) {
```
KyleJu, line 276: Question about this part - if the `results_url` contains `-summary`, it is guaranteed to be the `v2` format? Is this a unique identifier for the `v2` format? if not, do we have existing files with the `-summary.json.gz` format and do we  have a backward compatibility?

jcscottiii, line 276: Won't the results_url always contain `-summary` for both the old format (`-summary.json.gz`) and the new format (`-summary_v2.json.gz`?

Also, what is the type for results_url?
javascript string has a method called `includes` but not `contains`
http://tpcg.io/_A9GPP0

DanielRyanSmith: Ah, this is a logic error on my part I forgot to replace, and this statement needs to check if the `results_url` contains `resultsSuffix`. Sorry fort the confusion!

The logic is supposed to be:
- Check for `summary.json.gz` substring in the url (signifies old format). Use this to find the index if it exists.
- If it does not exist in the url, it must be the new format. Instead use `summary_v2.json.gz`.

shared/util.go
```diff
@@ -144,7 +144,7 @@ func GetResultsURL(run TestRun, testFile string) (resultsURL string) {
 		// Assumes that result files are under a directory named SHA[0:10].
 		resultsBase := strings.SplitAfter(resultsURL, "/"+run.Revision)[0]
 		resultsPieces := strings.Split(resultsURL, "/")
-		re := regexp.MustCompile("(-summary)?\\.json\\.gz$")
+		re := regexp.MustCompile("(-summary(_v2)?)?\\.json\\.gz$")
```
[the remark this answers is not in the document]
DanielRyanSmith: Sure, that's a good idea! I'll add one.

webapp/components/test-file-results.js
DanielRyanSmith: I'm realizing I did not hit the code to test this as I expected lol, Kyle also mentioned an issue with this above. I'll get it fixed and make sure it's 100% working.

jcscottiii: To help test, you may want to append to this [list](https://github.com/web-platform-tests/wpt.fyi/blob/207813b3ed18bae81068934caa478daffd782d36/webapp/components/test/util/helpers.js#L18) or create a TEST_RUNS_DATA_V2.

If you do the latter, you'll want to create a copy of this [file](https://github.com/web-platform-tests/wpt.fyi/blob/8bf23a6f68d18acab002aa6a613fc5660afb0a85/webapp/components/test/test-file-results-table.html) but load up TEST_RUNS_DATA_V2
https://github.com/web-platform-tests/wpt.fyi/blob/8bf23a6f68d18acab002aa6a613fc5660afb0a85/webapp/components/test/test-file-results-table.html#L29

DanielRyanSmith: yes, I updated the `TEST_RUNS_DATA` you linked, and I edited some of the summary links to include a `_v2` suffix and left others with the old link. Since we're just validating that we can find the suffix as expected in the URLs, it should handle the test cases. Do you think this is sufficient? Or would it be best to update this further?

KyleJu (approved):
LGTM on my side. Please wait for James' LGTM to merge

webapp/components/test-file-results.js
jcscottiii: We should add a test suite to the [test-file-results.html](https://github.com/web-platform-tests/wpt.fyi/blob/3f36cb0ac4f036e73049547f08242dda9d9e5637/webapp/components/test/test-file-results.html#L94) that exercises [resultsURL](https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/components/test-file-results.js#L271).

Something like:
```js
    suite('resultsURL', () => {
      test('v1 url', () => {
        expect(
          tfr.resultsURL(
            TEST_RUNS_DATA[1],
            "bar"
            )).to.equal("foo")
      });
      test('v2 url', () => {
        expect(
          tfr.resultsURL(
            TEST_RUNS_DATA[0],
            "bar"
            )).to.equal("foo")
      });
    });
```

jcscottiii (approved):
LGTM

PR #2945, opened 2022-08-26
Move all dependency updates to dependabot

jcscottiii:
Remove renovate and use dependabot for go, python and node

1. Setup dependabot for go, python, node and docker. Make sure it does not auto merge.
2. Disable deployments for PRs made by dependabot 
  - Examples of why this is a good choice:
    - https://github.com/search?l=YAML&q=org%3Agoogle-github-actions+%27dependabot%5Bbot%5D%27&type=Code

---
Below are just misc. notes

Warnings

There are some differences between dependabot and renovate:
There are things renovate can do that dependabot can't do:
1. Run npm dedupe to keep the lock files pretty clean.
  [Notes](https://github.com/dependabot/dependabot-core/issues/2079)
2. Group updates together https://github.com/github/roadmap/issues/148

Extra work needed specifically for dependabot

1. Auto merging is possible but extra work needs to be done.
   [Notes](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#approve-a-pull-request)
  - I don't think auto merging is a good idea.
2. Need to create a GCLOUD_KEY_FILE_JSON secret specifically for dependabot to use. Make sure it has the necessary scopes and no more.

Misc notes

- For go dependencies - It does indeed run go mody tidy. (which is absolutely necessary) [Notes](https://github.blog/changelog/2020-10-19-dependabot-go-mod-tidy-and-vendor-support/)
- Dependabot has a docker updater too. We should use that.


.github/dependabot.yml
```diff
@@ -3,4 +3,38 @@ updates:
 - package-ecosystem: pip
   directory: results-processor/
   schedule:
-    interval: monthly
+    interval: weekly
+- package-ecosystem: gomod
+  directory: /
```
KyleJu, line 8: [Most examples](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-dependabotyml-file) I see add "" around the value, e.g. "/", "weekly"

.github/dependabot.yml
```diff
@@ -3,4 +3,38 @@ updates:
 - package-ecosystem: pip
   directory: results-processor/
   schedule:
-    interval: monthly
+    interval: weekly
+- package-ecosystem: gomod
```
KyleJu, line 7: nit: Add optional reviews, e.g.
    reviewers:
      - "kyleju"

KyleJu (approved):
lgtm with two comments

KyleJu:
> * For go dependencies - It does indeed run go mody tidy. (which is absolutely necessary) [Notes](https://github.blog/changelog/2020-10-19-dependabot-go-mod-tidy-and-vendor-support/)

Thanks for checking this! There was a time when Renovate did not run `go mod tidy` and it was an extra step to clean up manually every release

.github/dependabot.yml
jcscottiii: let's come back to this one. Maybe we should talk about this being on the task list for engineers on rotation.

.github/dependabot.yml
jcscottiii: Done!

PR #2954, opened 2022-09-06
Update Go dependencies

DanielRyanSmith:
Part of the Interop Tooling support rotation. Update wpt.fyi Go dependencies.


KyleJu (commented):
Were there notable changes from v42 to v47 in the changelog that may affect our APIs?

DanielRyanSmith:
According to the [release notes](https://github.com/google/go-github/releases):
- v47
  - Add ListOptions for listing user migrations
- v46
  - Support PagesSource as struct for update pages API
  - Change DependabotSecretsSelectedRepoIDs to []string
- v45
  - Support query parameter `visible_to_repository` in ListOrganizationRunnerGroups
- v44
  - Deprecate WebHookPayload
- v43
  - Add list options support for environments
  - GitHub Actions Billing returns total_paid_minutes_used as float64
  - Fix RequireCodeOwnerReviews and RequiredApprovingReviewCount fields with omitempty

I'm not sure the best way to test these, but none of them stuck out as something that would break our infra, so I ran the tests against the change and checked the staging environment to be functional. Is there anything else you would recommend we do to make sure nothing is breaking here?

KyleJu:
> 

Checking unit tests and staging is good enough indicator. I typically go through the changelog and keep an eye on breaking changes/depreciation. There were a few instances where I had to make small changes.

Thanks Daniel!

KyleJu (approved):

PR #2962, opened 2022-09-12
Align the node version between Docker and Makefile

KyleJu:


jcscottiii (approved):

DanielRyanSmith:
This change seems to have broken the local environment script. Running the docker setup locally no longer functioned recently, but reverting this fixed it. I'm not certain of the exact cause yet.

PR #1295, opened 2019-05-06
Migrate from self-navigator to @polymer/app-route

lukebjerring:
## Description
Rewrites the main `interop` and `wpt-results` components as "views", which are switched between using virtual navigation (drastically reducing load time when switching).

In addition, it drops our custom `self-navigator.js` code in favour of Polymer's `app-route` which handles virtual navigation better.

## Review Information
- Visit /admin/flags, enable the feature, visit homepage, change query, navigate back/forward etc.

PR #2963, opened 2022-09-12
Remove self-navigator in wpt-runs page. Delete self-navigator.js

jcscottiii:
Self-Navigator is no longer used. It was replaced by app-route in #1295. I think it was missed when removing it from other files. This removes its usage from the wpt-runs page

Way to test that navigation still works:
- Navigate to "Recent Runs"
- Click any link from that page. It should navigate successfully.

Context: Was trying to debug when the manual "page_view" event was being triggered in self-navigator.js. My breakpoint never triggered. Did research and saw that navigation was handled by app-route instead.


KyleJu (approved):

jcscottiii:
TestURL: https://clean-up-self-navigato-dot-wptdashboard-staging.uk.r.appspot.com/runs

I was wrong. There is still one usage of the SelfNavigator. Found it at the last minute.

On the runs page, if I:
- Click "Edit"
- Remove a Product
- Click Submit

The page does not reload anymore with the subset of products

It is because of the call to [`navigateToLocation`](https://github.com/web-platform-tests/wpt.fyi/blob/b2c1d3f82e80eda5cecc8609ee7ed4108a530f2d/webapp/components/wpt-runs.js#L441) which is in SelfNavigator

PR #2982, opened 2022-09-28
Investigation scores

DanielRyanSmith:
This change adds the ability to change/view the investigation scores designated on Interop Collaboration pages. The graph will also be updated to display the change in score over time.

### DIFF
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -208,12 +208,14 @@ class InteropDataManager {
         let start = info.scores_over_time.length - 1;
         let areaScore = 0;
         for (let i = start; i >= 0; i--) {
-          if (date < new Date(info.scores_over_time[i].date)) continue;
+          if (date < new Date(info.scores_over_time[i].date)) {
+            continue;
+          }
           areaScore = info.scores_over_time[i].score;
           break;
         }
         totalInvestigationScore += areaScore;
-      })
+      });
       summaryScore *= (1 - this.investigationWeight);
       totalInvestigationScore /= this.investigationScores.length;
       summaryScore += this.investigationWeight * totalInvestigationScore;
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -205,17 +205,14 @@ class InteropDataManager {
     if (this.investigationScores) {
       let totalInvestigationScore = 0;
       this.investigationScores.forEach(info => {
-        let start = info.scores_over_time.length - 1;
-        let areaScore = 0;
-        for (let i = start; i >= 0; i--) {
-          if (date < new Date(info.scores_over_time[i].date)) {
-            continue;
-          }
-          areaScore = info.scores_over_time[i].score;
-          break;
+        // Find the investigation score at the given date.
+        const entry = info.scores_over_time.findLast(
+          entry => date >= new Date(entry.date));
+        if (entry) {
+          totalInvestigationScore += entry.score;
         }
-        totalInvestigationScore += areaScore;
       });
+      // Add the investigation score and weight it as specified.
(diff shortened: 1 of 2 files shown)


webapp/components/interop.js
```diff
@@ -200,6 +201,28 @@ class InteropDataManager {
...
+      this.investigationScores.forEach(info => {
+        let start = info.scores_over_time.length - 1;
+        let areaScore = 0;
+        for (let i = start; i >= 0; i--) {
+          if (date < new Date(info.scores_over_time[i].date)) {
+            continue;
+          }
+          areaScore = info.scores_over_time[i].score;
+          break;
+        }
```
foolip, line 216: I think this could be simpler/clearer using `findLast()`. Untested suggestion:

```suggestion
        let entry = info.scores_over_time.findLast((entry) => {
          return date >= new Date(entry.date);
        });
        if (entry) {
          totalInvestigationScore += entry.score;
        }
```

DanielRyanSmith: Much more readable - I implemented it 👍

webapp/components/interop.js
```diff
@@ -172,10 +174,9 @@ class InteropDataManager {
 
         let summaryScore = testScore / numFocusAreas;
 
-        // TODO: get the investigation score at this date.
-        if (this.year === '2022') {
-          summaryScore = summaryScore * 0.9;
-        }
+        // Handle investigation scoring if applicable.
+        summaryScore = this.#addInvestigationScore(summaryScore, date);
```
foolip, line 178: The naming of `addInvestigationScore` and `summaryScore` bugs me a little bit, because this doesn't merely add to `summaryScore`, it also scales it by 0.9 (for 2022).

The problem is that we need the `investigationWeight` to know how to scale `summaryScore`. But I think it might be clearer like this:

```suggestion
        const [investigationScore, investigationWeight] = this.#getInvestigationScoreAndWeight(date);
```

And then compute a `summaryScore` here after the call. I think that'd make it easier to follow how the total score is computed from the individual parts.

webapp/components/interop.js
```diff
@@ -172,10 +174,9 @@ class InteropDataManager {
 
         let summaryScore = testScore / numFocusAreas;
```
foolip, line 175: This suggestion goes with the next one, if we'll first compute `testScore`, `investigationScore` and `investigationWeight`:

```suggestion
       testScore /= numFocusAreas;
```

webapp/components/interop.js
```diff
@@ -200,6 +201,25 @@ class InteropDataManager {
     }
   }
 
+  #addInvestigationScore(summaryScore, date) {
+    if (this.investigationScores) {
+      let totalInvestigationScore = 0;
+      this.investigationScores.forEach(info => {
```
foolip, line 207:
```suggestion
      for (const info of investigationScores) {
```

webapp/components/interop.js
```diff
@@ -172,10 +172,15 @@ class InteropDataManager {
...
         // Handle investigation scoring if applicable.
-        summaryScore = this.#addInvestigationScore(summaryScore, date);
+        const [investigationScore, investigationWeight] =
+          this.#getInvestigationScoreAndWeight(date);
+
+        // Factor in the the investigation score and weight as specified.
+        let summaryScore = testScore * (1 - investigationWeight);
+        summaryScore += investigationWeight * investigationScore;
 
         summaryScore = Math.floor(summaryScore);
```
foolip, line 185: This might be even simpler to understand as a single expression:

```suggestion
        const summaryScore = Math.floor(testScore * (1 - investigationWeight) +
                                        investigationScore * investigationWeight);
```

foolip (approved):

Issue #2985, opened 2022-09-30
web_components_test & go_chrome_test flakiness


KyleJu:
web_components_test & go_chrome_test are both experiencing flakiness and the issue lies in downloading the Chrome Stable build:

```
if [[ -z "$(which google-chrome)" ]]; then \
	ARCHIVE=google-chrome-stable_current_amd64.deb; \
	wget -q https://dl.google.com/linux/direct/${ARCHIVE}; \
	sudo dpkg --install ${ARCHIVE} 2>/dev/null || true; \
	sudo apt-get install --fix-broken -qqy; \
	sudo dpkg --install ${ARCHIVE} 2>/dev/null; \
fi
Selecting previously unselected package google-chrome-stable.
(Reading database ... 104859 files and directories currently installed.)
Preparing to unpack google-chrome-stable_current_amd64.deb ...
Unpacking google-chrome-stable ([106](https://github.com/web-platform-tests/wpt.fyi/actions/runs/3147582963/jobs/5146415715#step:3:107).0.5249.91-1) ...
Processing triggers for mime-support (3.62) ...
E: Failed to fetch http://deb.debian.org/debian/pool/main/libn/libnet-ssleay-perl/libnet-ssleay-perl_1.85-2+b1_amd64.deb  404  Not Found [IP: 146.75.106.132 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
(Reading database ... 104974 files and directories currently installed.)
Preparing to unpack google-chrome-stable_current_amd64.deb ...
Unpacking google-chrome-stable (106.0.5249.91-1) over (106.0.5249.91-1) ...
Processing triggers for mime-support (3.62) ...
make: *** [Makefile:[118](https://github.com/web-platform-tests/wpt.fyi/actions/runs/3147582963/jobs/5146415715#step:3:119): chrome] Error 1
```

DanielRyanSmith:
While investigating this issue, it might also be worth investigating a previously existent flaky test case that appears during browser testing. I'll look to see if I can find an example of this, but a flaky scenario that has existed for some time involves an error along the lines of `Error thrown outside of test function`. It is something that has occurred rarely for as some months, and is likely an async issue somewhere in the testing code.

past:
[...]

PR #1165, opened 2019-02-20
Fetch suitable chromedriver

lukebjerring:
## Description
Avoids selenium-standalone diverging from the latest version of Chrome.

PR #2987, opened 2022-10-01
Fix missing dependencies when Chrome install fails

past:
The error message in #2985 suggests to try passing --fix-missing, which although I am not certain it will help, I confirmed that it doesn't hurt. Code spelunking indicates that this bit was added in #1165 without a lot of rigor anyway.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,6 +118,7 @@ chrome: wget
 	if [[ -z "$$(which google-chrome)" ]]; then \
 		ARCHIVE=google-chrome-stable_current_amd64.deb; \
 		wget -q https://dl.google.com/linux/direct/$${ARCHIVE}; \
+		sudo apt-get update; \
 		sudo dpkg --install $${ARCHIVE} 2>/dev/null || true; \
 		sudo apt-get install --fix-broken --fix-missing -qqy; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -119,7 +119,7 @@ chrome: wget
 		ARCHIVE=google-chrome-stable_current_amd64.deb; \
 		wget -q https://dl.google.com/linux/direct/$${ARCHIVE}; \
 		sudo dpkg --install $${ARCHIVE} 2>/dev/null || true; \
```
jcscottiii, line 121:
```suggestion
		sudo apt-get update; \
		sudo dpkg --install $${ARCHIVE} 2>/dev/null || true; \
```

jcscottiii: I *think* that the index of package locations is stale. That is why it fails with the 404.

I tried adding `sudo apt-get update` and it seems to work.
https://github.com/web-platform-tests/wpt.fyi/pull/2992

past: I thought about running 'apt update', but I assumed it wouldn't help given that we essentially initialize a new docker container every time. Your run did succeed however, so maybe it's useful.

jcscottiii (approved):

PR #2994, opened 2022-10-03
Update major versions of Go dependencies

DanielRyanSmith:
Part of the Interop Tooling rotation.

```
go get -u -d ./...
go mod tidy
```


DanielRyanSmith:
@jcscottiii Do you know anything about this? Trying to update these dependencies causes `module requires Go 1.17` errors. Is this something we're just going to have to deal with until we update to Go 1.17, or is there something I should be doing to make sure we're only updating to versions compatible with 1.16?

jcscottiii:
>Is this something we're just going to have to deal with until we update to Go 1.17

Until App Engine supports Go 1.17 or later, we can't use those dependencies :( Hopefully, we can get access to this soon. You can close this until then

PR #2995, opened 2022-10-03
Fix import for Firefox and Safari

DanielRyanSmith:
The new changes to the Interop page did not function as intended on Firefox and Safari. This change fixes the JSON import causing the issue.

### DIFF
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -41,7 +41,7 @@ class InteropDataManager {
   async fetchYearData() {
     const resp = await fetch('../static/interop-data.json');
     const yearInfo = await resp.json();
-    return yearInfo
+    return yearInfo;
   }
 
   // Fetches the datatable for the given feature and stable/experimental state.
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -30,7 +30,7 @@ class InteropDataManager {
 
   async fetchYearData() {
     // prepare all year-specific info for reference.
-    const resp = await fetch('../static/interop-data.json');
+    const resp = await fetch('/static/interop-data.json');
     const paramsByYear = await resp.json();
 
     const yearInfo = paramsByYear[this.year];
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -20,13 +20,12 @@ class InteropDataManager {
   constructor(year) {
     this.year = year;
     // The data is loaded when the year data is obtained and the csv is loaded and parsed.
-    // The year data is needed for parsing the csv.
     this._dataLoaded = this.fetchYearData()
-    .then(async () => {
-      return load().then(() => {
-        return Promise.all([this._loadCsv('stable'), this._loadCsv('experimental')]);
+    // The year data is needed for parsing the csv.
+      .then(async () => {
+        await load();
+        return await Promise.all([this._loadCsv('stable'), this._loadCsv('experimental')]);
       });
-    });
   }
 
   async fetchYearData() {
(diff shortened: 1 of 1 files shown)


KyleJu:
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'replace')
    at InteropDataManager._loadCsv (interop.js:97:29)
    at interop.js:37:32
    at async InteropDataManager.getMostRecentScores (interop.js:61:5)
    at async InteropDashboard.ready (interop.js:645:32)

When I first load the page, it shows me the errors above. It can be reproduced from an incognito mode. There is a race condition between two asyn calls `fetchYearData` and `getDataTable`

webapp/components/interop.js
```diff
@@ -20,23 +19,31 @@ import PARAMS_BY_YEAR from '../static/interop-data.json' assert {type: 'json'};
...
-    this.summaryFeatureName = yearInfo.summary_feature_name;
-    this.csvURL = yearInfo.csv_url;
-    this.tableSections = yearInfo.table_sections;
-    this.prose = yearInfo.prose;
-    this.matrixURL = yearInfo.matrix_url;
-    this.issueURL = yearInfo.issue_url;
-    this.investigationScores = yearInfo.investigation_scores;
-    this.investigationWeight = yearInfo.investigation_weight;
-
+    this.fetchYearData(year)
```
KyleJu, line 22: Similar to the race condition between `fetchYearData` and `fetchYearData`, we don't want a race condition when `InteropDataManager`'s value is assigned to the view class and when this `fetchYearData` is fetched.  We should somehow update view after this promise succeeds, through observable values, or use `await`, e.g. await `this.dataManager.getDataTable(feature, stable)` to make it synchronous

DanielRyanSmith: I was not able to reproduce your error above, but I'm pretty sure I understand where it's coming from. All of these methods await `_dataLoaded`, which was set when both `_loadCsv` invocations were finished, but these invocations need to run after the year info is obtained. I'm not certain the best way to verify this since I wasn't able to reproduce, but I've added a change that handles each of these calls sequentially as intended.

DanielRyanSmith:
I was seeing a strange "deceptive link" warning on Chrome for the staging environment (possibly from an old implementation). For safety's sake, I created a draft PR with the same changes to generate a [separate staging url](https://url-testing-dot-wptdashboard-staging.uk.r.appspot.com/interop-2022) and it works on all browsers for me. Hopefully this is a sufficient fix 😅

webapp/components/interop.js
```diff
@@ -11,17 +11,29 @@ import '../node_modules/@polymer/paper-input/paper-input.js';
...
   constructor(year) {
-    // prepare all year-specific info for reference.
     this.year = year;
-    const yearInfo = PARAMS_BY_YEAR[year];
+    // The data is loaded when the year data is obtained and the csv is loaded and parsed.
+    this._dataLoaded = this.fetchYearData()
+    // The year data is needed for parsing the csv.
+      .then(async() => {
+        await load();
+        return await Promise.all([this._loadCsv('stable'), this._loadCsv('experimental')]);
```
KyleJu, line 27:
```suggestion
        return Promise.all([this._loadCsv('stable'), this._loadCsv('experimental')]);
```
Is `await` here needed?

KyleJu: IIUC, `this.fetchYearData()` is called synchronously at the time of `this._dataLoaded` creation, so we can guarantee that fetchYearData is fetched before use.

KyleJu (approved):
LGTM with a comment

foolip (approved):

webapp/components/interop.js
DanielRyanSmith: I think you're right, I've made the change 👍

PR #3001, opened 2022-10-14
Add a flag allowing interop view to be viewed everywhere

gsnedders:
## Description

This adds a flag which overrides the default logic for allowing the interop scores.

## Review Information

As I understand it, we pull defaults from somewhere server-side; it's not clear to me where actually controls this. As I understand, it sounds like 
people with admin accounts on wpt.fyi can change these defaults, but given they default to false this shouldn't need anything to be done there.

Note I haven't actually tested this (I don't have a working local setup), so someone who does should probably test this as part of the review.

### DIFF
--- a/webapp/components/test-runs-query.js
+++ b/webapp/components/test-runs-query.js
@@ -6,13 +6,14 @@
 import { pluralize } from './pluralize.js';
 import { Channels, DefaultProductSpecs, ProductInfo } from './product-info.js';
 import { QueryBuilder } from './results-navigation.js';
+import { WPTFlags } from './wpt-flags.js';
 
 const testRunsQueryComputer =
   'computeTestRunQueryParams(shas, aligned, master, labels, productSpecs, to, from, maxCount, offset, view)';
 
-const TestRunsQuery = (superClass, opt_queryCompute) => class extends QueryBuilder(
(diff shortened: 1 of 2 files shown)


KyleJu:
The deployment failed because this PR is from a fork. I am happy to deploy one manually after the review

DanielRyanSmith:
@jgraham Do you have opinions on this?

jgraham:
Yes, I continue to believe this is a bad idea. The Interop-202x tests are special in that we've got mutual agreement that they make sense to score in this way. We don't have that agreement for arbitary tests, so I don't think we should add an override like this.

gsnedders:
> Yes, I continue to believe this is a bad idea. The Interop-202x tests are special in that we've got mutual agreement that they make sense to score in this way. We don't have that agreement for arbitary tests, so I don't think we should add an override like this.

IMO, flags are sufficiently niche that this risk is very low, for the same reason as why having colours on the homepage behind a flag isn't harmful.

gsnedders:
Also, like, to be explicit, the current logic is very naïve and easy to work around: searching for `(label:interop-2022-color or not label:interop-2022-color)` will let you get interop view anywhere as is, and that's arguably more obvious than the `/flags` page nobody knows exists.

PR #3016, opened 2022-10-31
Use SecretManager to get sensitive uploader creds

jcscottiii:
Purpose:
- Currently, the secrets for uploaders are stored in Datastore. SecretManager is a better solution for storing secrets in Google Cloud

Changes:
- Add secret manager dependency to go.mod & go.sum
- Add secrets manager client to the singleton with the rest of the google cloud clients
- Switch GetUploader from using a datastore to a secrets manager
- Added a new "cloud" tag for tests. This tag will use actual cloud resources when there is no emulator. (SecretsManager does not have an emulator like the Datastore for the medium tests)

Other changes / pattern differences:
- The constructor function NewAppEngineSecretManager returns the concrete instance instead of the interface. Other functions, like NewAppEngineDatastore return the interface. The new function uses the pattern encouraged: "Accept interfaces, return structs". Should slowly look at doing this with the other structs / new structs
- To enable cloud tests, I switched the credentials in the Settings to a new service account that has access to only deploy App Engine apps and to read the single secret for the test. (The default App Engine service account is still used by the actual App Engine instances)
- Moved all the secrets from datastore to secret manager.


DanielRyanSmith (commented):
I'm going to let @KyleJu take a look at this, as you both are more familiar with this issue, but this is looking really good and thorough. Thanks for writing tests and taking care of this!

KyleJu:
> I'm going to let @KyleJu take a look at this, as you both are more familiar with this issue, but this is looking really good and thorough. Thanks for writing tests and taking care of this!

For Sure! I will get to it by the end of the day

shared/secret_manager_cloud_cloud_test.go
```diff
@@ -0,0 +1,37 @@
...
+	TestCloudSecretManagerGetSecret get the secrets on a real Google Cloud
+	Secret Manager because there is no emulator like Datastore.
+
+	If it needs to be setup again, run the following:
+	gcloud secrets create test-secret --replication-policy="automatic"
+	echo -n "test-secret-value" | gcloud secrets versions add test-secret --data-file=-
+	gcloud secrets add-iam-policy-binding test-secret --member='serviceAccount:github-cicd@wptdashboard-staging.iam.gserviceaccount.com' --role='roles/secretmanager.secretAccessor'
+*/
+func TestCloudSecretManagerGetSecret(t *testing.T) {
+	require.NotEmpty(t, runtimeIdentity.AppID, "Unable to find project ID")
```
KyleJu, line 23: Should a fake `runtimeIdentity.AppID` env be passed in the Makefile? Or this has to be the real wptdashboard-staging AppID?

KyleJu (approved):
LGTM with two questions

.github/workflows/ci.yml
```diff
@@ -47,3 +47,34 @@ jobs:
...
+  go_cloud_test:
+    # This job uses real Cloud resources.
+    # This means this CI job will have access to the service account.
+    # In that case, similar to deploy.yml, trust only pull requests that are
+    # made within web-platform-tests and exclude forks.
+    if: ${{ github.repository == 'web-platform-tests/wpt.fyi' }}
+    needs: [go_test, go_chrome_test, go_firefox_test]
+    runs-on: ubuntu-latest
+    env:
+      DOCKER_IMAGE: webplatformtests/wpt.fyi:latest
```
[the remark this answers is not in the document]
jcscottiii: it uses the same image used by deploy.yml
https://github.com/web-platform-tests/wpt.fyi/blob/adc496684c473b7cf1f45d3b40cf073b568e5678/.github/workflows/deploy.yml#L13

As long as the go_large_test still works in that workflow, this workflow will still work.
https://github.com/web-platform-tests/wpt.fyi/blob/adc496684c473b7cf1f45d3b40cf073b568e5678/.github/workflows/deploy.yml#L46

You bring up a good point about that image. If we upgrade the version of Go again but the docker image is not updated to use it, this workflow and the deploy workflow will break.

shared/secret_manager_cloud_cloud_test.go
jcscottiii: great question! it has to be the real wptdashboard-staging [AppID](https://github.com/web-platform-tests/wpt.fyi/pull/3016/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R94) because that project has the test secret manager value.

PR #3037, opened 2022-11-22
Start of the work to have a flexible scheme

### LINKED ISSUE #2941: Search cache testing will not work when hostname is localhost (http scheme needed)
(placeholder)

jcscottiii:
This will allow all services to be ran locally simultaneously

Fixes #2941


LeonardYam:
Hi, could I know what is the progress on this?

LeonardYam:
@jcscottiii Hi, thanks for replying! Admittedly, I am not that familiar with how the search-cache functions. Would changing all hard-coded schemes (as shown in #2941) be sufficient? And is there any test I could perform to check if the search-cache functions as intended?

BertsLifee (approved):

PR #3049, opened 2022-11-23
Ignore gcloud tests for dependabot & ignore certain npm packages

jcscottiii:
For security reasons, dependabot builds do not have access to the same secrets as a regular CI build. As a result, it cannot run gcloud tests properly. This patch ignores gcloud tests for dependabot builds.

Also, configure dependabot to ignore certain packages for /webapp. They were ignored in renovate previously. During the migration, they were only ignored for webdriver/ but not webapp/. This fixes that.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,6 +15,11 @@ updates:
   directory: "/webapp"
   schedule:
     interval: "weekly"
+  ignore:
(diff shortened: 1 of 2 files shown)


jcscottiii:
@KyleJu I didn't limit the number of PRs because I thought that the flood of dependabot PRs would be a one time thing this time. But after this one time, it should be manageable.

KyleJu (approved):

PR #3058, opened 2022-12-02
Interop 2022: update editing investigation score

foolip:
See https://github.com/web-platform-tests/interop-2022-editing/issues/10


foolip:
@DanielRyanSmith I think you already deployed, but if this can be deployed soonish too, that would be great :)

jensimmons (approved):

DanielRyanSmith (approved):

PR #3086, opened 2022-12-21
Interop 2022: update editing investigation score

foolip:
See https://github.com/web-platform-tests/interop-2022-editing/issues/10#issuecomment-1353156481


webapp/static/interop-data.json
```diff
@@ -96,7 +96,8 @@
                 "name": "Editing, contenteditable, and execCommand",
                 "scores_over_time": [
                     { "date": "2022-10-22", "score": 360 },
-                    { "date": "2022-11-25", "score": 460 }
+                    { "date": "2022-11-25", "score": 460 },
+                    { "date": "2022-12-31", "score": 520 }
```
foolip, line 100: @jgraham which date should we use here?

foolip: Answer in https://github.com/web-platform-tests/interop-2022-editing/issues/10#issuecomment-1361754258

jgraham (approved):

jcscottiii (approved):

PR #3125, opened 2023-01-28
Improve Interop table markup

nt1m:
- Each table is now a `<table>`, instead of having a giant `<table>` with multiple tables
- `<thead>` is now used for each table header
- `<tbody>` is now used for each table body
- `<tfoot>` is now used for each table footer

This should improve accessibility a bit.

### DIFF
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -411,7 +411,7 @@ class InteropDashboard extends PolymerElement {
         .score-table {
           width: 100%;
           border-collapse: collapse;
-          margin-top: 3em;
+          margin-top: 2.5em;
         }
 
         .score-table caption {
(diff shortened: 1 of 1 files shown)


webapp/components/interop.js
```diff
@@ -406,36 +406,41 @@ class InteropDashboard extends PolymerElement {
...
-          display: flex;
-          border-radius: 3px;
-          background: white;
         }
 
         .score-table {
-          height: 100%;
           width: 100%;
           border-collapse: collapse;
+          margin-top: 3em;
```
DanielRyanSmith, line 414:
```suggestion
          margin-top: 2.5em;
```

Looking at [the version from this PR](https://table-markup-dot-wptdashboard-staging.uk.r.appspot.com/interop-2023) compared to [the staging environment](https://staging.wpt.fyi/interop-2023), it seems like there are some small visual changes, but this would get partially back to the spacing we were at previously.

DanielRyanSmith (approved):
Thanks for the change!

Issue #3137, opened 2023-02-01
Interop dashboard's interop-data.json appears to be missing 2023 data, resulting in empty page


patrickangle:
As of this morning and having cleared my browser caches, the interop dashboard doesn't display any data for me. Looking in dev tools, it appears that the 2023 data is missing from `interop-data.json`, and thus can't be found to populate the page.

DanielRyanSmith:
Thank you for reporting this @patrickangle. A fix has been deployed now. Can you check if https://wpt.fyi/interop-2023 for us to make sure this problem is no longer an issue?

patrickangle:
Yep, looks good now! Thank you 😄

PR #3138, opened 2023-02-01
Bypass any caching of interop-data.json

foolip:
Could fix https://github.com/web-platform-tests/wpt.fyi/issues/3137, but not the root cause.

### DIFF
--- a/webapp/components/interop.js
+++ b/webapp/components/interop.js
@@ -30,7 +30,7 @@ class InteropDataManager {
 
   async fetchYearData() {
     // prepare all year-specific info for reference.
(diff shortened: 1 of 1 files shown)


DanielRyanSmith (approved):

foolip:
https://wpt.fyi/static/interop-data.json?cachebust has the 2023 data for me, so that's good. (So does https://wpt.fyi/static/interop-data.json)

foolip:
Let's do https://github.com/web-platform-tests/wpt.fyi/pull/3139 instead.

PR #3139, opened 2023-02-01
rename interop-data file

DanielRyanSmith:
Renaming the `interop-data.json` file to fix caching problems.


foolip (approved):
We should keep the old file to not break things transiently, but approving so you can get this landed if needed.

foolip (approved):

DanielRyanSmith:
> We should keep the old file to not break things transiently, but approving so you can get this landed if needed.
Got it, added back the file with the original name.

PR #3154, opened 2023-02-10
Update interop-2023-contain tests url

### LINKED ISSUE #3153: Fix CSS containment tests link on Interop dashboard
The current link only includes 2023 tests, we should include the 2022 ones too: https://wpt.fyi/results/css?label=master&label=experimental&product=chrome&product=firefox&product=safari&aligned&view=interop&q=label%3Ainterop-2022-contain%20or%20label%3Ainterop-2023-contain

@DanielRyanSmith Can you please take a look? I think changing the json files is problematic because we might hit caching issues again, so I'll leave it to you.

DanielRyanSmith:
Fixes #3153

If any caching problems occur, the should rectify themselves over the coming days. We can investigate further if this link does direct properly by Monday.

### DIFF
--- a/webapp/static/interop-data.json
+++ b/webapp/static/interop-data.json
@@ -378,7 +378,7 @@
                 "description": "Containment",
                 "mdn": "https://developer.mozilla.org/docs/Web/CSS/contain",
                 "spec": "https://drafts.csswg.org/css-contain/#contain-property",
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
> Isn't there interop-data_v2.json as well?

Yes 😅 thank you!
Definitely need to make final changes to avoid the confusion of this quick fix soon.

PR #3170, opened 2023-02-20
README: Correct syntax for and-conjuction

zcorpan:
Explicit `and` doesn't appear to work.

## Description
Fix docs to not suggest using `and` keyword in the search query, since it doesn't work. Omitting it works.

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -14,21 +14,17 @@ By default, a search query will be implicitly treated as an `exists` query (a di
 across each of the runs separately). However, there are several other root query types that
 can be invoked by wrapping the query(s), including explicitly wrapping with `exists`.
 
-If multiple root queries are used, they are implicitly combined with `AND`, i.e. each query
-must separately be true. e.g.
+If multiple root queries are used, then each query must separately be true. e.g.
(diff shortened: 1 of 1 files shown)


foolip:
@KyleJu can you check if there isn't some support for this in the parser? If not I'd expect "and" to be treated as a path filter, and that doesn't seem to happen.

In other words, I think the feature exists, but it doesn't work as documented.

KyleJu:
> @KyleJu can you check if there isn't some support for this in the parser? If not I'd expect "and" to be treated as a path filter, and that doesn't seem to happen.
> 
> In other words, I think the feature exists, but it doesn't work as documented.

We definitely have [`and` support](https://github.com/web-platform-tests/wpt.fyi/blob/7142bfde7f958b0add956753d33d249622e37acb/webapp/components/test-search.js#L88). I suspect that it is a parser issue.

KyleJu:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/3171.  I will merge this PR until `and` issue is fixed

KyleJu (approved):

PR #3176, opened 2023-02-27
Update app-engine.md deployment instructions

jcscottiii:
The deployment steps currently assume you are either:
1.  logged in, or
  - (works if deploying from local but not docker since the docker instance has a separate GCP profile),
2.  have a client-secret.json from the IAM GCP panel
  - (works if for local or docker given the user has downloaded a client-secret beforehand)

This update ensures case number 1 works when running the docker instance by making users to log in using gcloud auth login


past (approved):
Looks good, thanks!

Issue #3033, opened 2022-11-21
Dockerfile Maintenance Improvements


jcscottiii:
In #2945, security updates are provided by dependabot. Dependabot also can provide updates for Docker images. Prior to doing that, we need to perform some maintenance items on the Dockerfiles:

- [x] Have the root [Dockerfile](https://github.com/web-platform-tests/wpt.fyi/blob/main/Dockerfile) and search cache [Dockerfile](https://github.com/web-platform-tests/wpt.fyi/blob/main/api/query/cache/service/Dockerfile) pin the patch version of the go version. (e.g.  `golang:1.16.15-buster`) This is necessary because only tracking major & minor hides the fact that the docker tag can be updated
- [x] Switch the results-processor [Dockerfile](https://github.com/web-platform-tests/wpt.fyi/blob/main/results-processor/Dockerfile) to use the official python [docker](https://hub.docker.com/_/python/tags?page=1&name=3.7) image. Ensure it is tagged with major.minor.patch.
  - [x]  Currently, we are using python 3.7. Optionally, consider tackling #3032 beforehand
- [x] Uncomment the docker section in the dependabot YAML.

PR #3180, opened 2023-03-03
Make some Dockerfile maintenance improvements

past:
- Upgrade to Go 1.20.1 and pin the patch version
- Upgrade to Python 3.9 and use the official Python image
- Get rid of the obsolete dependency on Python 2
- Upgrade to Debian 11 for local development

Further upgrades to Python 3.10+ would need an updated official Go image that doesn't exist yet. This should resolve most of issue #3033, modulo the dependabot change that will come in a future PR.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # vim: set expandtab sw=4
-FROM golang:1.19-buster
+FROM golang:1.20.1-bullseye
 
 # Create a non-priviledged user to run browsers as (Firefox and Chrome do not
 # like to run as root).
@@ -14,8 +14,7 @@ RUN apt-get update -qqy && apt-get install -qqy --no-install-suggests \
         firefox-esr \
         lsb-release \
         openjdk-11-jdk \
-        python-crcmod \
-        python3.7 \
+        python3.9 \
         sudo \
         tox \
         wget \
--- a/Makefile
+++ b/Makefile
@@ -205,8 +205,7 @@ curl: apt-get-curl
 gcc: apt-get-gcc
 git: apt-get-git
 psmisc: apt-get-psmisc
-python3: apt-get-python3.7
-python: apt-get-python
+python3: apt-get-python3.9
 tox: apt-get-tox
 unzip: apt-get-unzip
 wget: apt-get-wget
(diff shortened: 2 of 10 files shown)


past:
The errors stem from the old images in CI. 'make test' succeeds when run locally in a rebuilt docker image.

jcscottiii:
Verified what @past said. The tests did run successfully locally.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # vim: set expandtab sw=4
-FROM golang:1.19-buster
+FROM golang:1.20.1-bullseye
```
jcscottiii, line 2: For the Go upgrade: Could you complete these [steps](https://github.com/web-platform-tests/wpt.fyi/blob/main/docs/upgrading-go.md)? This will allow the main tooling Dockerfile here to match the Go version deployed for 1) the webapp [[1](https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/web/app.yaml#L5)][[2](https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/web/app.staging.yaml#L5)] and 2) the [searchcache](https://github.com/web-platform-tests/wpt.fyi/blob/main/api/query/cache/service/Dockerfile).

Also, could you change this [line](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L18) to mention that we are using bullseye instead of buster.

Also, I see that you pinned the patch version 🎉 .
Could you update the wording [here](https://github.com/web-platform-tests/wpt.fyi/blob/main/webapp/web/app.staging.yaml#L5) and [here](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L22).

jcscottiii (changes requested):
Two small requests (commented on file):
- making the Go version consistent across all services
- Doc updates

Dockerfile
past: Good catch! One question: in your last sentence you asked to updated the wording in 2 places, but the first link points to app.staging.yaml. Perhaps you meant [here](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L3)?

jcscottiii: Whoops! Thanks! That's the right link. Should read as: 

Could you update the wording [here](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L3) and [here](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L22).

past: I made all of the proposed changes with the exception of [this one](https://github.com/web-platform-tests/wpt.fyi/blame/main/docs/upgrading-go.md#L22). It turns out that patch versions are not supported in the [go directive](https://go.dev/ref/mod#go-mod-file-go).

jcscottiii (approved):

PR #3188, opened 2023-03-07
Use python3-crcmod apt-get package for results-processor and developer Dockerfiles

jcscottiii:
Use python3-crcmod apt-get package for results-processor and developer Dockerfiles

In the results-processor, `virtualenv` is not there by default. added `python3-virtualenv` to install it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN apt-get update -qqy && apt-get install -qqy --no-install-suggests \
         lsb-release \
         openjdk-11-jdk \
         python3.9 \
+        python3-crcmod \
         sudo \
         tox \
         wget \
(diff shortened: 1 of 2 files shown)


jcscottiii:
Results-processor dockerfile
```sh
$ root@396ed8121333:/app# gsutil version -l | grep crcmod
compiled crcmod: True
```

Developer dockerfile

```sh
$ user@dd594171ba73:~/wpt.fyi$ gsutil version -l | grep crcmod
compiled crcmod: True
```

KyleJu:
Failed due to a disk size issue. I delete a version on staging and it should be fine now

jcscottiii:
@KyleJu Thanks!

KyleJu (approved):

KyleJu:
> @KyleJu Thanks!

Anytime!

PR #3207, opened 2023-03-15
Update score for Accessibility Testing (60%)

zcorpan:
See https://github.com/web-platform-tests/interop-2023-accessibility-testing/issues/3


DanielRyanSmith:
This file is being deleted in an open PR and this data will be maintained in `webapp/components/interop-data.js`. Sorry for the confusion! We intend this to be the final change to where we store this information.

zcorpan:
@DanielRyanSmith ok thanks, rebased.

PR #3205, opened 2023-03-14
move interop data outside of static (fixes #3132)

racheljay:
## Description
This PR moves the data that was previously in `components/static/interop-data.json` to a `.js` file outside of the static directory as outlined in the discussion for #3142.

## Changes
`components/static/interop-data.json` was previously removed and this PR adds that data to `components/interop-data.js`. It is now imported directly into `interop.js` and accessed by `fetchYearData()`.

PR #3208, opened 2023-03-15
remove interop-data_v2.json

racheljay:
## Description
This PR removes the JSON file that was made redundant by #3205 to avoid the confusion of having it in two places.


DanielRyanSmith (approved):

DanielRyanSmith:
Thanks for taking care of that!

PR #3161, opened 2023-02-14
Add support for Node.js (fixes #3144)

past:
Adds support for Node.js as a browser/product, which was requested in #3144.

PR #3213, opened 2023-03-18
Add "WebKit" as a product

gsnedders:
This largely mimics what was done in #3161; let me know if I've missed anything.

### DIFF
--- a/webapp/components/product-info.js
+++ b/webapp/components/product-info.js
@@ -46,7 +46,7 @@ const versionPatterns = Object.freeze({
 
 // The set of all browsers known to the wpt.fyi UI.
 const AllBrowserNames = Object.freeze(['android_webview', 'chrome_android', 'chrome_ios', 'chrome',
-  'deno', 'edge', 'firefox', 'flow', 'node.js', 'safari', 'servo', 'webkit', 'webkitgtk']);
+  'deno', 'edge', 'firefox', 'flow', 'node.js', 'safari', 'servo', 'webkitgtk', 'webkit']);
 
 // The list of default browsers used in cases where the user has not otherwise
 // chosen a set of browsers (e.g. which browsers to show runs for). Stored as
--- a/webapp/components/product-info.js
+++ b/webapp/components/product-info.js
@@ -17,6 +17,7 @@ const DisplayNames = (() => {
   m.set('node.js', 'Node.js');
   m.set('servo', 'Servo');
   m.set('uc', 'UC Browser');
+  m.set('webkit', 'WebKit');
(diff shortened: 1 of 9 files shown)


gsnedders:
Also: does it still require any work on the admin side to add things to the backend?

gsnedders:
That said, we might want to call this "macOS WebKit" given how naming is going for Chrome (see also #3214).

past:
The changes look good to me and I agree that "macOS WebKit" would be better. If you have sorted out the authentication stuff for uploading results, I don't think you need anything else on the backend side.

KyleJu (approved):
LGTM after addressing the naming!

DanielRyanSmith:
It looks like there is a test that needs to be updated in `components/test/test-search.html`

gsnedders:
Seemingly:

```js
    test('browser test status eq', () => {
      // All known browsers.
      for (const browser of AllBrowserNames) {
        assertQueryParse(browser + ':ok', {exists: [{product: browser, status: 'OK'}]});
      }
```

My guess is that `webapp/components/test-search.js` isn't coping with having one browser name be a prefix of another.

Does anyone still around here understand Ohm well enough to quickly know the solution here?

webapp/components/product-info.js
```diff
@@ -45,7 +46,7 @@ const versionPatterns = Object.freeze({
 
 // The set of all browsers known to the wpt.fyi UI.
 const AllBrowserNames = Object.freeze(['android_webview', 'chrome_android', 'chrome_ios', 'chrome',
-  'deno', 'edge', 'firefox', 'flow', 'node.js', 'safari', 'servo', 'webkitgtk']);
+  'deno', 'edge', 'firefox', 'flow', 'node.js', 'safari', 'servo', 'webkit', 'webkitgtk']);
```
DanielRyanSmith, line 49:
```suggestion
  'deno', 'edge', 'firefox', 'flow', 'node.js', 'safari', 'servo', 'webkitgtk', 'webkit']);
```

I don't know why, but rearranging these made the test pass for me. Maybe someone smarter than me can figure out why. 😅

KyleJu:
CI passed. Merging!

gsnedders:
> LGTM after addressing the naming!

We never did address this! But it sounded like you were hoping to make progress on #3214 soonish?

KyleJu:
> macOS WebKit


> We never did address this! But it sounded like you were hoping to make progress on #3214 soonish?

That is my bad! We do intend to address it in Q2, but the renaming should be done in this PR. Would you mind sending a follow-up PR?

Issue #3117, opened 2023-01-27
Sort functionality for Interop dashboard


jensimmons:
@foolip had the idea a while ago to add little triangles (or other UI) to allow people to sort by each column of the data tables. So — sort alphabetically by area name (the default, and as it is now). Sort by % number order for each of the three browsers. Or sort by % number order for Interop score.

It'd be especially helpful to be able to quickly sort and see what's in most need of work for a particular browser. Anyone could sort and see the order of most-interoperable to least-interoperable.

I'd love it if this is possible. I do not believe it's a blocker to launching. We could always add the functionality later.

foolip:
Oh right, I forgot about this! I think this would be very useful, whether looking for how best to improve an individual browser's score, or the overall interop score. cc @DanielRyanSmith

nt1m:
With PR #3125 merged, pasting this snippet in the console:

```
[...document.querySelector("interop-dashboard").shadowRoot.querySelectorAll("th")].forEach(n => n.addEventListener("click", (e) => {
  let index = [...e.target.closest("thead > tr").querySelectorAll("th")].findIndex(node => n == node);
  let tbody = e.target.closest("table").querySelector("tbody");
  let bodyItems = [...tbody.querySelectorAll("tr")];
  bodyItems.sort((a, b) => {
    let aContent = a.children[index].textContent;
    let bContent = b.children[index].textContent;
    if (isNaN(parseFloat(aContent)))
      return aContent.localeCompare(bContent);
    return parseFloat(aContent) > parseFloat(bContent);
  });

  tbody.textContent = "";
  tbody.append(...bodyItems);
}));
```

works. 

Perhaps it would be possible to integrate such code into a component.

Worth noting that `:has()` makes it awkward with`aContent.localeCompare(bContent);`, since it will always appear first heh.

DanielRyanSmith:
I currently don't have time to implement this, but if anyone is passionate about adding the functionality and wants to go for it, I am available to review any PRs as needed. 🙂

jensimmons:
I could really use this functionality this week! Still wishing for it.

foolip:
Integrating the sort JS would be easy, but I'm not sure how to trigger it. There's no obvious place to place little arrows.

nt1m:
<img width="695" alt="image" src="https://user-images.githubusercontent.com/1795135/224392059-19a5a063-a5c6-41e0-846b-b4dc528a2ed1.png">

[...]

PR #3227, opened 2023-03-23
Interop Dashboard Sort Functionality 

racheljay:
## Description
This PR adds sort functionality to the Interop 2023 dashboard in the Active Focus Areas table, a feature requested in #3117. By default the data will sort in ascending order by feature name. When you click the toggle above any data column, it will sort the features in ascending order. If you click a toggle on any column that is sorted in ascending order, it will re-sort into descending order.

## Changes
This PR adds a row of buttons in the Active Focus Areas table header, as well as a click handler (`handleSortClick`) and a sorting method (`sortRows`)

interopsort

### DIFF
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -622,6 +622,13 @@ class InteropDashboard extends PolymerElement {
     return `${(score / 10).toFixed(1)}%`;
   }
 
+  getNumericalBrowserScoreByFeature(browserIndex, feature) {
+    const scores = this.stable ? this.scores.stable : this.scores.experimental;
+    const score = scores[browserIndex][feature];
+    const roundedScore = Math.round(score * 100) / 100;
+    return roundedScore / 10;
+  }
+
   getBrowserScoreTotal(browserIndex) {
     return this.totals[browserIndex];
   }
@@ -696,7 +703,7 @@ class InteropDashboard extends PolymerElement {
   }
 
   getSortIcon(index) {
-    index = index - 1
+    index = index - 1;
     if (this.sortColumn !== index && this.isSortedAsc === true) {
       return '/static/expand_less.svg';
     } else if (this.sortColumn === index && this.isSortedAsc === true) {
@@ -704,7 +711,7 @@ class InteropDashboard extends PolymerElement {
     } else if (this.sortColumn === index && this.isSortedAsc === false) {
       return '/static/expand_less.svg';
     }
-      return '/static/expand_less.svg';
+    return '/static/expand_less.svg';
 
   }
 
@@ -724,7 +731,7 @@ class InteropDashboard extends PolymerElement {
     const individualScores = [];
     for (let i = 0; i < rows.length; i++) {
       const feature = rows[i];
-      individualScores[i] = [feature, parseFloat(this.getBrowserScoreForFeature(sortColumn, feature))];
+      individualScores[i] = [feature, this.getNumericalBrowserScoreByFeature(sortColumn, feature)];
     }
     individualScores.sort((a, b) => a[1] - b[1]);
     for (let i = 0; i < individualScores.length; i++) {
@@ -733,20 +740,22 @@ class InteropDashboard extends PolymerElement {
   }
 
   sortRows(rows, index, sortColumn, isSortedAsc) {
-    if(index !== 0) return rows;
-      const sortedFeatureOrder = [];
-      // For the first column, sort alphabetically by name
-      if(sortColumn === -1) {
-        this.alphabeticalSort(rows, sortedFeatureOrder)
+    if(index !== 0) {
+      return rows;
+    }
+    const sortedFeatureOrder = [];
+    // For the first column, sort alphabetically by name
+    if(sortColumn === -1) {
+      this.alphabeticalSort(rows, sortedFeatureOrder);
       // For the other columns, sort numerically by score
-      } else if (sortColumn >= 0) {
-       this.numericalSort(rows, sortedFeatureOrder, sortColumn)
-      }
-      // Reverse current sort order
-      if (isSortedAsc === false) {
-        sortedFeatureOrder.reverse();
-      }
-      return sortedFeatureOrder;
+    } else if (sortColumn >= 0) {
+      this.numericalSort(rows, sortedFeatureOrder, sortColumn);
+    }
+    // Reverse current sort order
+    if (isSortedAsc === false) {
+      sortedFeatureOrder.reverse();
+    }
+    return sortedFeatureOrder;
   }
 
   handleSortClick(e) {
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -622,6 +622,8 @@ class InteropDashboard extends PolymerElement {
     return `${(score / 10).toFixed(1)}%`;
   }
 
+  // getNumericalBrowserScoreByFeature returns the same score as
+  // getBrowserScoreForFeature but as a number instead of a string
   getNumericalBrowserScoreByFeature(browserIndex, feature) {
     const scores = this.stable ? this.scores.stable : this.scores.experimental;
     const score = scores[browserIndex][feature];
@@ -708,7 +710,7 @@ class InteropDashboard extends PolymerElement {
       return '/static/expand_less.svg';
     } else if (this.sortColumn === index && this.isSortedAsc === true) {
       return '/static/expand_more.svg';
-    } else if (this.sortColumn === index && this.isSortedAsc === false) {
+    } else if (this.sortColumn === index && !this.isSortedAsc) {
       return '/static/expand_less.svg';
     }
     return '/static/expand_less.svg';
@@ -752,7 +754,7 @@ class InteropDashboard extends PolymerElement {
       this.numericalSort(rows, sortedFeatureOrder, sortColumn);
     }
     // Reverse current sort order
-    if (isSortedAsc === false) {
+    if (!isSortedAsc) {
       sortedFeatureOrder.reverse();
     }
     return sortedFeatureOrder;
@@ -766,7 +768,7 @@ class InteropDashboard extends PolymerElement {
       this.isSortedAsc = true;
     } else if (this.sortColumn === i && this.isSortedAsc === true) {
       this.isSortedAsc = false;
-    } else if (this.sortColumn === i && this.isSortedAsc === false) {
+    } else if (this.sortColumn === i && !this.isSortedAsc) {
       this.isSortedAsc = true;
     }
     this.sortColumn = i;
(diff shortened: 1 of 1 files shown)


webapp/components/interop-dashboard.js
```diff
@@ -344,10 +345,19 @@ class InteropDashboard extends PolymerElement {
...
                   </tr>
+                  <tr>
+                  <template is="dom-if" if="[[showSortToggle(itemsIndex)]]">
+                    <td><paper-icon-button class="sort-button" id="col-0" on-click="handleSortClick" src="[[getSortIcon(0, sortToggled)]]"></paper-icon-button></td>
+                    <td><paper-icon-button class="sort-button" id="col-1" on-click="handleSortClick" src="[[getSortIcon(1, sortToggled)]]"></paper-icon-button></td>
+                    <td><paper-icon-button class="sort-button" id="col-2" on-click="handleSortClick" src="[[getSortIcon(2, sortToggled)]]"></paper-icon-button></td>
+                    <td><paper-icon-button class="sort-button" id="col-3" on-click="handleSortClick" src="[[getSortIcon(3, sortToggled)]]"></paper-icon-button></td>
+                    <td><paper-icon-button class="sort-button" id="col-4" on-click="handleSortClick" src="[[getSortIcon(4, sortToggled)]]"></paper-icon-button></td>
+                  </template>
+                </tr>
```
DanielRyanSmith, line 356: You can move the `<tr>` tag into the template so that an empty row will not be created if it's not needed.

webapp/components/interop-dashboard.js
```diff
@@ -456,6 +466,28 @@ class InteropDashboard extends PolymerElement {
...
       dataManager: Object,
       scores: Object,
+      tableState: {
+        type: Object,
+        value: {
+          sortColumn: -1,
+          sortOrder: 'asc',
+          sorted: true
+        }
+      },
```
DanielRyanSmith, line 476: `tableState` can be broken into separate variables here, and the `sortOrder` property can be changed to a boolean if renamed to something like `isSortedAscending`, which should make for easier comparisons. I'm also not seeing any uses of the `sorted` property in the code (but I may have missed something).

webapp/components/interop-dashboard.js
```diff
@@ -456,6 +466,28 @@ class InteropDashboard extends PolymerElement {
...
+      toggleStates: {
+        type: Array,
+        value: [
+          true,
+          false,
+          false,
+          false,
+          false
+        ]
+      },
```
DanielRyanSmith, line 486: After testing this, it looks like this variable to control the arrow direction always has 1 or less value equal to true. I think you can do away with this array altogether if you change the logic of your `getSortIcon()` method to only display a downward arrow if the index is equal to the sort column and the sort order is ascending.

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
     this.$.toggleStable.setAttribute('aria-pressed', true);
     this.$.toggleExperimental.setAttribute('aria-pressed', false);
   }
+
+  getSortIcon(index) {
+    if (this.toggleStates[index]) {
+      return '/static/expand_more.svg';
+    }
+    return '/static/expand_less.svg';
+  }
```
DanielRyanSmith, line 717: This method technically takes in `sortToggled` but it is not listed explicitly as an argument. I think you should get the same functionality by passing the new `sortColumn` variable to this method.

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
...
+
+  sortRows(rows, index, tableState) {
+    if(index === 0) {
```
DanielRyanSmith, line 720: You can flip this logic to:
```js
if (index !== 0) {
  return rows;
}
```
This is fundamentally the same but removes the need to indent most of the method in an `if` block.

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
...
   }
+
+  getSortIcon(index) {
+    if (this.toggleStates[index]) {
+      return '/static/expand_more.svg';
+    }
+    return '/static/expand_less.svg';
+  }
+
+  sortRows(rows, index, tableState) {
```
DanielRyanSmith, line 719: This method is pretty complex and is sorting two different ways based on whether you're sorting alphabetically or by score. Maybe each of the sorts can be broken into separate functions?

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
...
+      // Alphabetize by focus area column
+      if(tableState.sortColumn === -1) {
+        const rowNames = [];
+        for(let i = 0; i < rows.length; i++) {
+          const feature = rows[i];
+          rowNames[i] = [feature, this.getRowInfo(feature, 'description')];
+        }
+        rowNames.sort((a, b) => {
+          a[1] - b[1];
+        });
```
DanielRyanSmith, line 732: Something is wrong here, because technically this sort function is not returning any value (no return statement). Oddly enough, it seems like the table is being sorted correctly, but I think that is because the values for the interop tables are already sorted alphabetically in `interop-data.js`, and this sort call has no effect.

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
...
+        for (let i = 0; i < rowNames.length; i++) {
+          sortedFeatureOrder[i] = rowNames[i][0];
+        }
+        // List in ascending order by score
+      } else if (tableState.sortColumn >= 0) {
+        const individualScores = [];
+
+        for (let i = 0; i < rows.length; i++) {
+          const feature = rows[i];
+          individualScores[i] = [feature, parseFloat(this.getBrowserScoreForFeature(tableState.sortColumn, feature))];
```
DanielRyanSmith, line 743: This is _technically_ doing what you want, but `getBrowserScoreForFeature()` is returning a string-formatted version of the browser score and you are having to parse that formatted string back into a number. Ideally, you should take the way that `getBrowserScoreForFeature()` is obtaining the number and implement it separately without needing to handle all of this typecasting.
Another possibility is creating a `formatBrowserScore()` method that takes the browser number score and formats it like a string. Then, you could refactor `getBrowserScoreForFeature()` to return the raw browser score number. (Hopefully I explained that well)

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +708,74 @@ class InteropDashboard extends PolymerElement {
...
+      this.tableState.sortOrder = 'dec';
+      this.toggleStates[i + 1] = false;
+    } else if (this.tableState.sortColumn === i && this.tableState.sortOrder === 'dec') {
+      this.tableState.sortOrder = 'asc';
+      this.toggleStates.fill(false);
+      this.toggleStates[i + 1] = true;
+    }
+    this.tableState.sortColumn = i;
+
+    this.sortToggled = !this.sortToggled;
```
DanielRyanSmith, line 778: I think the reason it felt like the `sortToggled` variable is needed is because the `sortColumn` variable lives as a property in the `tableState` object. Polymer doesn't explicitly know when you've changed values on object properties, so it will not trigger a re-render unless the instance variables explicitly change. If/when `sortColumn` is moved to an actual instance variable rather than inside the `tableState` object, Polymer should know that it needs to re-render anything that uses `sortColumn`.

DanielRyanSmith (changes requested):
I tried this out locally and the functionality is exactly as I would have expected. 🙂
However, there are some aspects of the code that need to be addressed to get this merge-ready.
There are also some UX changes that likely need to be handled. Specifically that the arrows should not displace the table and the column headers can be clicked to sort the rows. However, that can optionally be addressed in a later PR.

racheljay:
@DanielRyanSmith Thanks for the feedback. I'm working on making these changes now. I'll let you know if I have any questions.

webapp/components/interop-dashboard.js
racheljay: I reworked this one a bit but it might be a little convoluted still. Let me know if you think it can be further simplified. I wasn't able to get the correct behavior for double clicking a toggle without passing in `sortColumn` as well as `isSortedAsc`

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +701,75 @@ class InteropDashboard extends PolymerElement {
...
+    const sortedFeatureOrder = [];
+    // For the first column, sort alphabetically by name
+    if(sortColumn === -1) {
+      this.alphabeticalSort(rows, sortedFeatureOrder);
+      // For the other columns, sort numerically by score
+    } else if (sortColumn >= 0) {
+      this.numericalSort(rows, sortedFeatureOrder, sortColumn);
+    }
+    // Reverse current sort order
+    if (isSortedAsc === false) {
```
DanielRyanSmith, line 755:
```suggestion
    if (!isSortedAsc) {
```

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +701,75 @@ class InteropDashboard extends PolymerElement {
...
+
+  handleSortClick(e) {
+    const i = parseInt(e.target.id.split('-')[1]) - 1;
+
+    if (this.sortColumn !== i) {
+      this.sortColumn = i;
+      this.isSortedAsc = true;
+    } else if (this.sortColumn === i && this.isSortedAsc === true) {
+      this.isSortedAsc = false;
+    } else if (this.sortColumn === i && this.isSortedAsc === false) {
```
DanielRyanSmith, line 769:
```suggestion
    } else if (this.sortColumn === i && this.isSortedAsc) {
      this.isSortedAsc = false;
    } else if (this.sortColumn === i && !this.isSortedAsc) {
```

webapp/components/interop-dashboard.js
```diff
@@ -672,5 +701,75 @@ class InteropDashboard extends PolymerElement {
...
   }
+
+  getSortIcon(index) {
+    index = index - 1;
+    if (this.sortColumn !== index && this.isSortedAsc === true) {
+      return '/static/expand_less.svg';
+    } else if (this.sortColumn === index && this.isSortedAsc === true) {
+      return '/static/expand_more.svg';
+    } else if (this.sortColumn === index && this.isSortedAsc === false) {
+      return '/static/expand_less.svg';
```
DanielRyanSmith, line 712:
```suggestion
    if (this.sortColumn !== index && this.isSortedAsc) {
      return '/static/expand_less.svg';
    } else if (this.sortColumn === index && this.isSortedAsc) {
      return '/static/expand_more.svg';
    } else if (this.sortColumn === index && !this.isSortedAsc) {
      return '/static/expand_less.svg';
```

webapp/components/interop-dashboard.js
DanielRyanSmith: Yes, that sounds about right - you'll need them passed as arguments so that Polymer knows they have changed, and that it needs to re-evaluate what the value is.

racheljay:
@DanielRyanSmith I think I've implemented all the changes requested. Please let me know if there's anything I've left out.

DanielRyanSmith (approved):
This looks ready to me besides a few nits. Optionally, it would also be nice if you threw in some comments explaining the new methods you added.

Thanks for the update! I might follow up with some small UX changes, but this is a highly requested feature you've worked on. Much appreciated. 🙂

PR #3235, opened 2023-03-28
Do not zero out tests with harness errors in Interop view

DanielRyanSmith:
In the results analysis script, we currently do not zero out test scores that contained harness errors. This change updates wpt.fyi's aggregations to match the analysis script. This should remove anamolies that exist today on the "Test Totals" section of the page.

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -726,13 +726,14 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
           return [0, 0];
         }
         // Take the passes / total subtests to get a percentage passing.
-        const percentPassed = rs[i].passes / rs[i].total;
-        passes += percentPassed;
-      // If this is a new summary, aggregate using the new process.
-      } else if (passingStatus || (rs[i].total > 0 && rs[i].total === rs[i].passes)) {
-        // If we have a total of 0 subtests but the status is passing,
-        // mark as 100% passing.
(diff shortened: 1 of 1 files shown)


DanielRyanSmith:
For reference of the problem, the original plan for interop scores was to mark the scores of all tests that have a testharness.js error with a result of 0. Interop 2022 did not contain tests with testharness.js errors unless they were explicitly investigated and determined to still be relevant despite these testharness.js errors. Interop 2023 has some small number of tests that display harness errors. However, the logic for marking tests with harness errors as a 0 overall score still existed in the wpt.fyi test results display logic. Since we are not marking them as 0 in the results-analysis script, there is a small % difference in the scores displayed on the dashboard (where we pull the from the results-analysis data) and on the wpt.fyi test results page (where we aggregate from  test results summaries).

@foolip, This should remove anomalies we are seeing in the Interop scores vs. the test results page. For example, [Pointer & Mouse Events test results scores](https://wpt.fyi/results/?label=experimental&label=master&product=chrome&product=firefox&product=safari&aligned&view=interop&q=label%3Ainterop-2023-events) have a a score discrepancy between the dashboard and test results page, but this change will will cause the results page to match the Interop Dashboard. I'd like to land this change to mirror the numbers, but If we want to discuss how we are handling this aggregation overall, I'm free to talk as needed :^)

CC @nairnandu

KyleJu (approved):

foolip:
Post-merge LGTM, I took a look yesterday and agree this makes sense. But then I closed the tab I guess :)

PR #3236, opened 2023-03-28
Adjust table sort UX

DanielRyanSmith:
This change takes the newly-added sorting functionality and makes some UX changes to make the table headers clickable rather than adding a new row of clickable arrows.

This PR:
- Moves the new click handler logic to the table headers rather than a new row of arrows.
- Displays an arrow at the top right of the column sorted by, indicating the direction of the sort.

---
The table is sorted alphabetically ascending by default.
Screenshot 2023-03-28 at 1 04 29 PM

Clicking a new column header will sort it in descending order.
Screenshot 2023-03-28 at 1 04 40 PM

Clicking a column header twice will sort the table in ascending order.
Screenshot 2023-03-28 at 1 04 49 PM

### DIFF
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -107,11 +107,16 @@ class InteropDashboard extends PolymerElement {
           cursor: pointer;
         }
 
+        .sort-icon-focus-areas {
+          position: absolute;
+          top: 4px;
+          width: 20px;
+        }
+
         .sort-icon {
           position: absolute;
           top: 4px;
           right: -4px;
-          float: right;
           width: 20px;
         }
 
@@ -316,9 +321,7 @@ class InteropDashboard extends PolymerElement {
                     <tr class="section-header">
                       <th class="sortable-header" on-click="sortByName">
                         {{section.name}}
-                        <template is="dom-if" if="[[shouldShowSortIcon(0, sortColumn)]]">
-                          <img class="sort-icon" src="[[getSortIcon(isSortedAsc)]]" />
-                        </template>
+                        <img class="sort-icon-focus-areas" src="[[getSortIcon(0, sortColumn, isSortedAsc)]]" />
                       </th>
                       <th class="sortable-header" on-click="sortByChrome">
                         <template is="dom-if" if="[[stable]]">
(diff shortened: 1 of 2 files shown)


webapp/components/interop-dashboard.js
```diff
@@ -727,51 +805,77 @@ class InteropDashboard extends PolymerElement {
...
-    const i = parseInt(e.target.id.split('-')[1]) - 1;
-
-    if (this.sortColumn !== i) {
-      this.sortColumn = i;
-      this.isSortedAsc = true;
-    } else if (this.sortColumn === i && this.isSortedAsc) {
+  };
+
+  /**
+   * Column sort handlers.
```
KyleJu, line 844: Optional: You can rewrite the HTML with dom-repeat and pass the index as i

KyleJu (approved):

webapp/components/interop-dashboard.js
DanielRyanSmith: I thought this as well, and it was my first approach to the implementation, but could not figure out how to do this in the `on-click` while passing the index, and I think we haven't passed arguments into an `on-click` function anywhere else in the app. But I do assume that there is a better way of doing this.

DanielRyanSmith:
> It is a bit unclear to user that we can click on section headers to sort. Should we add arrows next to all browsers, rather than only one? It would be consistent with the /results page

My hope was that it would be good enough to add the `cursor: pointer` CSS attribute when hovering, but perhaps this is not obvious enough. I can give this a shot

> The alignment is far on the test name. What if we put the arrow on the left of section headers? Or on the top right corner of the "Active Focus Areas" so its placement is consistent?

I think this is a good idea. My implementation was to just simply display the arrow "at the top right of the header cell". Placing it to the right of the content in the cell could be more intuitive.

DanielRyanSmith:
This is the change I came up with:

The arrows now persist on non-selected columns, but they are a light-grey rather than their original black color.
Screenshot 2023-03-29 at 1 28 35 PM

Selecting a new column will make that column's arrow black, and the previously selected column's arrow will turn to grey.
Screenshot 2023-03-29 at 1 28 45 PM

PR #3246, opened 2023-04-04
add back interop-data.json

DanielRyanSmith:
Adds back the `interop-data.json` file by request, as it is being referenced by outside tooling. Also, adds a test to ensure that the data in `webapp/components/interop-data.js` matches the data in `webapp/static/interop-data.json`, so that any changes are made to both files.

### DIFF
--- a/webapp/components/test/interop-data.html
+++ b/webapp/components/test/interop-data.html
@@ -1,0 +1,23 @@
+<!doctype html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+  <script src="../../node_modules/wct-browser-legacy/browser.js"></script>
+</head>
+
+<body>
+<script type="module">
+import {interopData} from '../interop-data.js';
+// Check that the data in the JavaScript file and in the JSON file match.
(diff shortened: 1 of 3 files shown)


webapp/components/test/interop-data.html
```diff
@@ -0,0 +1,23 @@
...
+<head>
+  <meta charset="utf-8">
+  <script src="../../node_modules/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+  <script src="../../node_modules/wct-browser-legacy/browser.js"></script>
+</head>
+
+<body>
+<script type="module">
+import {interopData} from '../interop-data.js';
+// Check that the data in the JavaScript file and in the JSON file match.
```
past, line 12: Maybe add some words to explain that interop-data.json is used by Mozilla infra and cannot be deleted? Unless you can think of a better place to ensure someone a few months from now will not attempt to remove it again.

past (approved):
Looks good to me, thanks.

jcscottiii (approved):
LGTM with @past's suggestion

webapp/components/test/interop-data.html
DanielRyanSmith: Agreed. Added a comment to the files.

PR #3248, opened 2023-04-05
Fix sorting on main results page

DanielRyanSmith:
This change fixes the sorting functionality on the main results page, which was not working properly due to a bad reference to the `this` keyword.


jcscottiii (approved):
Verified that sorting works on the deployed version https://mainpage-sort-fix-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=experimental&label=master&aligned

LGTM

Issue #3009, opened 2022-10-21
Pasting paths into search box lowercases them, breaking them


gsnedders:
I was trying to search for `shadow-dom/DocumentOrShadowRoot-prototype-elementFromPoint.html or
shadow-dom/MouseEvent-prototype-offsetX-offsetY.html`, however in both STP 156 and Chrome 109 this gets lowercased and finds nothing.

Oddly, copying the URL gives me the search query before it gets lowercased… and that does manage to find the tests!

KyleJu:
> Hi @KyleJu, I would like to work on this issue. Just to clarify, we should only need to modify `webapp/components/test-search.js` right?

Yup It is likely a frontend issue and webapp/components/test-search.js should be the place

LeonardYam:
Hi, I want to check my understanding of the search box.

Right now the search box supports two functions:
1.  Parse a search into the required structured query object using `Ohm` to pass into the `api/search` endpoint.
2. Navigate to a given file-path. E.g. Entering `/encoding/encodeInto.any.html` should navigate you to the results for that test. (this process is actually automatic and does not require a query commit)

Side-notes:
- In `webapp/components/test-search.js`, how does `value="{{ queryInput::input }}"` work? From my understanding, this is equivalent to having an oninput listener that does `this.queryInput = e.target.value` and that this is a [virtual method call](https://github.com/tc39/proposal-bind-operator), but I can't find where the input method is binded.
- For the navigation/auto-complete function, it involves a lot of querySelector calls and code-coupling with `webapp/components/wpt-results.js` and `webapp/components/wpt-app.js`. Is there a better way to perform this operation?

foolip:
I ran into this now when trying to search for "enumerateDevices" tests for https://github.com/web-platform-tests/interop/issues/532. The UI keeps changing it to "enumeratedevices" which doesn't find anything. Editing the URL does work though:
https://wpt.fyi/results/?label=master&label=experimental&aligned&q=%28path%3A%2Faudio-output%20or%20path%3A%2Fmediacapture-streams%29%20enumerateDevices

@KyleJu can you take a look at https://github.com/web-platform-tests/wpt.fyi/pull/3577?

Issue #2980, opened 2022-09-27
Unable to combine path substring matches


foolip:
A search like [`path:/dom/ or path:/fetch/`](https://wpt.fyi/results/?label=master&label=experimental&aligned&view=subtest&q=path%3A%2Fdom%2F%20or%20path%3A%2Ffetch%2F) works to find all all of the tests in dom/ or fetch/.

However, a search like [`at-supports or css-supports`](https://wpt.fyi/results/?label=master&label=experimental&aligned&view=subtest&q=at-supports%20or%20css-supports) _doesn't_ list all of the tests that have "at-supports" or "css-supports" in the path, instead it seems to just list the tests with "at-supports".

Since path substring matching is the default there's no atom that can be used to work around this, it seems.

Reported by @drott.

PR #3249, opened 2023-04-06
Fix case-sensitive path matching for the search box

LeonardYam:
## Description

Currently, navigating to a file-path with uppercase characters like `/encoding/encodeInto.any.html` does not work.  
Somewhat related issue: #3009 

## Review Information

Enter any file-path which contains upper-case characters into the search-box.

Local build vs production:

https://user-images.githubusercontent.com/11293086/230372141-ed486db7-25b1-483f-adec-f58e5a7f8a77.mp4

## Changes
```javascript
matches = matches
          .filter(p => p.toLowerCase())
          .filter(p => p.includes(query))
``` 
Auto-complete options are generated from `matches`. I believe the intention here was to map all paths to a lower-case string and perform query matching. However, converting all paths to lowercase would cause further issues later (since the file-path is case-sensitive). Thus, the change was to perform case-insensitive query matching while keeping case-sensitive paths.

```javascript
if (autocompleteSelection.value === path) {
  this.dispatchEvent(new CustomEvent('autocomplete', {
    detail: {path: path},
  }));
```
`path` refers to `queryInput` which is the text within the search-box. Currently, `queryInput` is being converted to lowercase and hence the equality-check fails for any path which contains uppercase characters. The path passed into detail should also be changed to use the case-sensitive value in autocompleteSelection.

## Notes
I am unable to test whether these changes affects complex queries due to #2941. This would also mean that fixes on #2980 and other bugs would be unable to proceed.

An alternative solution to this would be to ensure `queryInput` does not get lowercased, however due to the above issue, keeping it lowercase for now would be the safer option.

### DIFF
--- a/webapp/components/test-search.js
+++ b/webapp/components/test-search.js
@@ -525,7 +525,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
   }
 
   handleKeyDown(e) {
-    // Prevent tab key from working
+    // Prevent tab key navigation on search bar.
     if (e.keyCode === 9) {
       e.preventDefault();
       return false;
--- a/webapp/components/test-search.js
+++ b/webapp/components/test-search.js
@@ -514,7 +514,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
   }
 
   commitQuery() {
-    this.query = this.queryInput;    
+    this.query = this.queryInput;
     this.dispatchEvent(new CustomEvent('commit', {
       detail: {
         query: this.query,
--- a/webapp/components/test-search.js
+++ b/webapp/components/test-search.js
@@ -487,8 +487,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
       let matches = Array.from(paths);
       if (query) {
         matches = matches
-          .filter(p => p.toLowerCase())
-          .filter(p => p.includes(query))
+          .filter(p => p.toLowerCase().includes(query))
           .sort((p1, p2) => p1.indexOf(query) - p2.indexOf(query));
       }
       for (const match of matches.slice(0, 10 - datalist.children.length)) {
@@ -515,7 +514,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
   }
 
   commitQuery() {
-    this.query = this.queryInput;
+    this.query = this.queryInput;    
     this.dispatchEvent(new CustomEvent('commit', {
       detail: {
         query: this.query,
@@ -526,6 +525,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
   }
 
   handleKeyDown(e) {
+    // Prevent tab key from working
     if (e.keyCode === 9) {
       e.preventDefault();
       return false;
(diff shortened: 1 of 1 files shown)


webapp/components/test-search.js
```diff
@@ -487,8 +487,7 @@ class TestSearch extends WPTFlags(PolymerElement) {
       let matches = Array.from(paths);
       if (query) {
         matches = matches
-          .filter(p => p.toLowerCase())
-          .filter(p => p.includes(query))
+          .filter(p => p.toLowerCase().includes(query))
```
DanielRyanSmith, line 490: Good catch! The original version definitely couldn't be the intended behavior, as it arbitrarily converted the strings to lowercase and only filtered empty strings.

webapp/components/test-search.js
```diff
@@ -526,13 +525,15 @@ class TestSearch extends WPTFlags(PolymerElement) {
   }
 
   handleKeyDown(e) {
+    // Prevent tab key from working
```
DanielRyanSmith, line 528:
```suggestion
    // Prevent tab key navigation on search bar.
```

webapp/components/test-search.js
```diff
@@ -551,9 +552,9 @@ class TestSearch extends WPTFlags(PolymerElement) {
       if (autocompleteSelection.getAttribute('atom')) {
         return;
       }
-      if (autocompleteSelection.value === path) {
+      if (autocompleteSelection.value.toLowerCase() === path) {
```
DanielRyanSmith, line 555:
```suggestion
      if (autocompleteSelection.value.toLowerCase() === path.toLowerCase()) {
```

DanielRyanSmith (changes requested):
I was having trouble replicating this issue, but then I noticed you're using Firefox as your browser to test the site. This seems to be a Firefox-specific problem, as it doesn't trigger the `handleChange()` method after selecting from the search box options (which I assume you trigger instead in the video by clicking out of the search box). This problem is caused because the query is changed to lowercase in the search box before `handleChange()` is called.

This PR in its current form breaks the implementation on Chrome and Safari, but after playing around, the suggested changes below should work on all browsers.

I've submitted some changes that break other browsers' view on this site before, so I have some experience. This is a good reminder to test site changes on each of the major browsers, and a reason why the Interop effort is so important. 🙂

Thanks for identifying a problem and bringing forward a solution @LeonardYam!

LeonardYam:
Ah, thanks for pointing it out, I'll remember to test on Chrome and Safari in the future!

I have committed the relevant changes and this should be ready for merging @DanielRyanSmith.

Note:
Will #2941 be resolved anytime soon? I would be willing to pick it up since working on bugs related to search queries requires this to be fixed first.

DanielRyanSmith (approved):
Thanks again!

DanielRyanSmith:
About #2941, I do not think that this will be resolved any time soon, but there are ways to run the search cache locally, although it is intensive in terms of processing power. The instructions for that are in a readme file in a searchcache subfolder

PR #3250, opened 2023-04-06
Display a banner on dashboards for previous Interop years

DanielRyanSmith:
This PR adds a banner that displays on Interop dashboards for previous years noting that the user is viewing interop scores from a previous year, and contains a link to view the current interop year's dashboard.

The banner is displayed for Interop 2021 and 2022.
Screenshot 2023-04-06 at 3 53 46 PM

The banner does not display for Interop 2023.
Screenshot 2023-04-06 at 3 53 54 PM

### DIFF
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -35,6 +35,17 @@ class InteropDashboard extends PolymerElement {
           text-align: center;
         }
 
+        .previous-year-banner {
+          height: 40px;
+          background-color: #DEF;
+          text-align: center;
+          padding-top: 16px;
+        }
+
+        .previous-year-banner p {
+          margin: 0;
+        }
+
         .grid-container {
           margin: 0 2em;
           display: grid;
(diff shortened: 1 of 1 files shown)


webapp/components/interop-dashboard.js
```diff
@@ -583,6 +602,14 @@ class InteropDashboard extends PolymerElement {
     this.features = Object.entries(this.getYearProp('focusAreas'))
       .map(([id, info]) => Object.assign({ id }, info));
 
+    // Determine the current Interop year. It is assumed that
+    // the current year is the latest year defined in interop-data.
+    const currentInteropYear = Math.max(
```
KyleJu, line 607: We can use `getAllYears()` method here

webapp/components/interop-dashboard.js
```diff
@@ -267,6 +278,12 @@ class InteropDashboard extends PolymerElement {
         }
 
       </style>
+      <div class="previous-year-banner" hidden$=[[isCurrentYear]]>
```
KyleJu, line 281: Minor nit: isCurrentYear -> this default value for this should be true. I suspect that there might be a minor time gap when isCurrentYear is initialized and rendered here vs when ready() is called

DanielRyanSmith: Good idea - I've set it to have a default value of `true`.

webapp/components/interop-dashboard.js
DanielRyanSmith: Agreed, and changed!

KyleJu (approved):

jcscottiii (approved):

PR #3275, opened 2023-04-26
Detect interop-* triage metadata

KyleJu:
A part of https://github.com/web-platform-tests/wpt-metadata/issues/3954.  Add an "interop" PR label to metadata that is labelled "interop-*".

### Staging Test
https://github.com/web-platform-tests/wpt-metadata/pull/4104

### DIFF
--- a/shared/triage_metadata.go
+++ b/shared/triage_metadata.go
@@ -11,6 +11,7 @@ import (
 	"errors"
 	"math/rand"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/google/go-github/v47/github"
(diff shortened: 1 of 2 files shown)


shared/triage_metadata.go
```diff
@@ -246,6 +264,17 @@ func appendTestName(test string, metadata MetadataResults) {
 	}
 }
 
+func containsInterop(metadata MetadataResults) bool {
+	for _, links := range metadata {
+		for _, link := range links {
+			if strings.Contains(link.Label, "interop") {
```
foolip, line 270: Just to call out the behavior, this will catch labels like `interop-2023-has` but also `lets-go-interoperability` that contain "interop" anywhere. That's OK I think.

foolip (approved):

DanielRyanSmith (approved):

shared/triage_metadata.go
KyleJu: I am happy to tighten the logic in a follow-up PR!

PR #3287, opened 2023-04-27
Dynamically add stable or experimental labels to Interop Dashboard URLs

### LINKED ISSUE #3286: Link to stable results from stable Interop 2023 results table
The feature table in https://wpt.fyi/interop-2023?stable links to https://wpt.fyi/results/css/css-backgrounds?label=experimental&label=master&product=chrome&product=firefox&product=safari&aligned&view=interop&q=label%3Ainterop-2023-cssborderimage and similar. Note the `label=experimental` in here.

These links should use `label=stable` instead to show the results used to compute the results.

@meyerweb raised this in a meeting today

DanielRyanSmith:
Fixes #3286

This change adds the stable or experimental labels to test results links on the Interop dashboard based on the view that the user is viewing. Previously, the "experimental" label was used for all links, even when viewing stable data.

### DIFF
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -639,12 +639,26 @@ class InteropDashboard extends PolymerElement {
       return '';
     }
 
-    if (stable) {
-      testsURL += '&label=stable';
-    } else {
-      testsURL += '&label=experimental';
+    // Test results are defined as absolute paths from this origin.
+    let url; 
+    try {
+      url = new URL(testsURL, window.location.origin);
+    } catch (e) {
+      console.error(e, e.stack);
+      return '';
     }
-    return testsURL;
+    // Test results URLs can have multiple 'label' params. Grab them all.
+    const existingLabels = url.searchParams.getAll('label');
+    // Make sure there isn't an existing stable or experimental label param.
+    const newLabels = existingLabels.filter(val => val !== 'stable' && val !== 'experimental');
+    // Add the stable/experimental label depending on the dashboard view.
+    newLabels.push(stable ? 'stable' : 'experimental');
+    // Delete the existing label params and re-add them.
+    url.searchParams.delete('label');
+    for (const labelValue of newLabels)
+      url.searchParams.append('label', labelValue);
+    
+    return url.toString();
   }
 
   // Get the tests URL for a row and add the stable/experimental label.
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -649,7 +649,7 @@ class InteropDashboard extends PolymerElement {
     }
     // Test results URLs can have multiple 'label' params. Grab them all.
     const existingLabels = url.searchParams.getAll('label');
-    // Make sure there isn't an existing stable or experimental label param.
+    // Remove any existing stable or experimental label param.
     const newLabels = existingLabels.filter(val => val !== 'stable' && val !== 'experimental');
     // Add the stable/experimental label depending on the dashboard view.
     newLabels.push(stable ? 'stable' : 'experimental');
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -615,10 +615,8 @@ class InteropDashboard extends PolymerElement {
 
   featureLinks(feature, stable) {
     const data = this.getYearProp('focusAreas')[feature];
-    let testsURL = data?.tests;
-    if (testsURL) {
-      testsURL = this.formatTestsURL(testsURL, stable);
-    }
+    const testsURL = this.formatTestsURL(data?.tests, stable);
+
     return [
       { text: 'Spec', href: data?.spec },
       { text: 'MDN', href: data?.mdn },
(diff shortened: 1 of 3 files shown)


webapp/components/interop-dashboard.js
```diff
@@ -630,6 +634,26 @@ class InteropDashboard extends PolymerElement {
     return this.getYearProp('focusAreas')[name][prop];
   }
 
+  // Add the stable or experimental label to a tests URL depending on the view.
+  formatTestsURL(testsURL, stable) {
+    // Don't try to add a label if the URL is undefined or empty.
+    if (!testsURL) {
+      return testsURL;
```
KyleJu, line 641:
```suggestion
      return '';
```
Let's return ''  so it does not return weird values that might  cause issues

webapp/components/interop-dashboard.js
```diff
@@ -630,6 +634,26 @@ class InteropDashboard extends PolymerElement {
...
+    }
+
+    if (stable) {
+      testsURL += '&label=stable';
```
KyleJu, line 645: I would assume that "?" delimiter is always present in the testURL, e.g.   /results/ vs /results/? /results/&label=stable would fail here

webapp/components/interop-dashboard.js
```diff
@@ -630,6 +634,26 @@ class InteropDashboard extends PolymerElement {
...
+    } else {
+      testsURL += '&label=experimental';
+    }
+    return testsURL;
+  }
+
+  // Get the tests URL for a row and add the stable/experimental label.
+  getTestsURL(name, stable) {
+    return this.formatTestsURL(this.getRowInfo(name, 'tests'), stable);
```
KyleJu, line 654: Same comment about the **?** delimiter in the test URL here

webapp/components/interop-dashboard.js
DanielRyanSmith: Yes, these are always wpt.fyi paths to test results pages with an interop label, so the query string should always already be present

webapp/components/interop-dashboard.js
```diff
@@ -613,12 +613,16 @@ class InteropDashboard extends PolymerElement {
     return feature === this.feature;
   }
 
-  featureLinks(feature) {
+  featureLinks(feature, stable) {
     const data = this.getYearProp('focusAreas')[feature];
+    let testsURL = data?.tests;
```
jcscottiii, line 618: This section will set testsURL to undefined if `data?` is undefined. However down below in `formatTestsURL`, if data exists but `data.tests` does not exist, you return an empty string [here](https://github.com/web-platform-tests/wpt.fyi/pull/3287/files#diff-770c7c8879ccc187effa2395cb35a77882a85b9eb65fc536ae0545552e74a1f1R641). Maybe it would be nice to have a constant at the top like `EMPTY_URL = ''`;
```suggestion
    const testsURL = data?.tests : this.formatTestsURL(data?.tests, stable) : EMPTY_URL;
```

and 
```js
  formatTestsURL(testsURL, stable) {
    // Don't try to add a label if the URL is undefined or empty.
    if (!testsURL) {
      return EMPTY_URL;
    }
```

This way you're returning the same data type in the event there's missing data and the dom-if [here](https://github.com/web-platform-tests/wpt.fyi/pull/3287/files#diff-770c7c8879ccc187effa2395cb35a77882a85b9eb65fc536ae0545552e74a1f1R496) will work as expected.

webapp/components/interop-dashboard.js
DanielRyanSmith: After trying this, I think using the URL object might not work in our specific case because we allow for multiples of the same query string parameter (e.g. multiple "label" params). Using the `set` method might overwrite a `&label=master` search param.

webapp/components/interop-dashboard.js
DanielRyanSmith: I agree that there's data type ambiguity here with my implementation like you mentioned. I removed the logic check and always invoke `formatTestsURL()` instead of checking if `testsURL` is valid beforehand, which will now always return an empty string if it isn't valid from `formatTestsURL()`. This should fix the problem. Let me know if you think this is a good approach!

jcscottiii: That works!

webapp/components/interop-dashboard.js
jcscottiii: I looked at the URLSearchParams [API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams).
Yeah, you're right, I wish there was a setAll method.

Here's a [fiddle](https://jsfiddle.net/m8zx7c1r/4/) with it working but it adds a lot of lines of code. (Probably can be optimized)

jcscottiii: I would update the implementation to match. I like @foolip's suggestion because search parameter modification shouldn't be something we deal with manually if we can avoid it. 
I did some searching to make sure we weren't the only ones looking for this same solution.


You should put a TODO note that it could be simplified and add these two links:
- https://github.com/whatwg/url/issues/762
- https://github.com/whatwg/url/issues/461

jcscottiii (approved):

KyleJu (approved):

KyleJu:
Thanks for addressing the comments!

PR #3288, opened 2023-04-28
Tighten the interop label criteria

KyleJu:
Follow-up for #3275. Only label it when it prefixes with **interop-**


foolip (approved):

foolip:
Thanks for fixing @KyleJu!

PR #3290, opened 2023-05-01
Update summary file format in documentation

### LINKED ISSUE #2940: api/results documentation does not reflect current behaviour
In https://github.com/web-platform-tests/wpt.fyi/tree/main/api#apiresults, it says

```
The summary JSON is in the format
{
  "/path/to/test.html": [1, 1],
}
Where the array contains [number of passes, total tests].
```

while it is actually something like that `"/path/to/test.html": {"s": "O", "c": [1, 1]},`

It would be good to update documentation

DanielRyanSmith:
Fixes #2940

This change updates the documentation that describes wpt.fyi summary files and their new v2 format.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -178,13 +178,54 @@ Performs an HTTP redirect for the results summary JSON blob of the given TestRun
 
 __Response format__
 
-The summary JSON is in the format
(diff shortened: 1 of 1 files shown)


KyleJu (approved):
Thanks for updating it Daniel!

PR #3293, opened 2023-05-02
Read Interop scores of previous years from static files

DanielRyanSmith:
This change stops the Interop dashboards from previous years from updating after that interop year has ended. Instead, the 2021 and 2022 dashboards will reference scores from static csv files within the repository that only contain score data for their respective years.

In order for the Interop 2022 score to match the final scores for that year, logic has been added to factor in 10% of the investigation score to be weighted as part of each overall browser score.

Screenshot 2023-05-02 at 2 15 12 PM

Screenshot 2023-05-02 at 2 15 31 PM


DanielRyanSmith:
As a sanity check, I compared the frozen score numbers with the [2021 EOY blog post](https://web.dev/compat2021-holiday-update/) and the [2022 EOY blog post](https://web.dev/interop-2022-wrapup/). The scores are mostly the same with small increases, most likely from the blog posts being written in mid December and the browser scores being frozen on January 1st.

webapp/components/interop-data.js
```diff
@@ -16,7 +16,7 @@ export const interopData = {
         'score_as_group': false
       }
     ],
-    'csv_url': 'https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/interop-2021/interop-2021-{stable|experimental}-v2.csv',
+    'csv_url': '/static/interop-2021-{stable|experimental}-v2.csv',
```
jcscottiii, line 19: I am thinking we should have a comment or note in the docs on where we get this CSV from in the future.
Previously, we had the direct link, so a developer would know exactly where to get next year's CSV.
Now, that we are shipping it with our code, a question a future developer would have is: Do I need to generate this CSV? If so, how? (But the answer would be to refer to the results-analysis repo.)

webapp/components/interop-summary.js
```diff
@@ -221,7 +221,13 @@ class InteropSummary extends PolymerElement {
     this.updateSummaryScore(scoreNumbers[0], scores[scores.length - 1][summaryFeatureName]);
     // Update the rest of the browser scores.
     for (let i = 1; i < scoreNumbers.length; i++) {
-      this.updateSummaryScore(scoreNumbers[i], scores[i - 1][summaryFeatureName]);
+      let browserScore = scores[i - 1][summaryFeatureName];
+      // For Interop 2022, 10% of the browser score was based on the investigation scores.
+      if (this.year === '2022') {
+        const investigationScore = this.dataManager.getYearProp('investigationTotalScore');
```
jcscottiii, line 227: Just curious: How hard would it be to modify the CSV to hard code the new browser score so we don't need to calculate it?

While we are committing this, we are still technically generating the browser score on every load.

jcscottiii: If it would be too much effort, I could be convinced to keep this. I just wanted to weigh the options.
On one hand, I would rather not get in the habit of one-off calculations per year for these frozen scores.
On the other hand, we could always revisit this when we freeze next year's score. And see if this will also apply then.

DanielRyanSmith: > Just curious: How hard would it be to modify the CSV to hard code the new browser score so we don't need to calculate it?

It's possible, but would require writing new logic to read final numbers from new locations. I'm not opposed to it, but  I feel like there is not a lot of gain for us to do so, since the user experience would be mostly the same, and the calculations are on the client side and inexpensive.

> we could always revisit this when we freeze next year's score.

I think this might be the best path forward. There is talk of spinning off previous years into separate files, which might make this approach of having final numbers already calculated a bit easier. Perhaps we can create an issue to address this when the preparations for next year roll around?

jcscottiii: Ahh. Thanks for answering that.

And yeah create the issue / add this to an existing issue.

jcscottiii (approved):

webapp/components/interop-data.js
```diff
@@ -16,7 +16,15 @@ export const interopData = {
...
-    'csv_url': 'https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/interop-2021/interop-2021-{stable|experimental}-v2.csv',
+    /**
+     * Interop scores are "frozen" after the end of the year.
+     * Once an interop year is completed, results are generated one more time
+     * from the results-analysis script for the full year, and those scores
+     * are placed for reference in the webapp/static directory. The score
+     * is no longer updated and is referenced from this location.
+     * More information at https://github.com/web-platform-tests/results-analysis
+    **/
+    'csv_url': '/static/interop-2021-{stable|experimental}-v2.csv',
```
foolip, line 27: Can we drop the v2 suffix now that the URLs are changing anyway? (I think we added them to make the switch between v1 and v2 smooth, and possibly as cache busting.)

webapp/components/interop-summary.js
foolip, line 227: It looks like this mixes in the investigation score equally throughout the year and not based on when the investigation score changes as the dashboard originally did. Is that right?

foolip:
A few comments on the details, but I think the most important thing here is how the static files were generated. Can you document where the files were copied from, or what script (at which commit of which repo?) was used to generate them?

I'm thinking especially about the fact that tests may have been renamed after 2022, so if we use the current wpt-metadata repo to get the test list, we may not have exactly the same test list as was used at the end of 2022. Ideally, we should check that all of the tests we use as input to the scoring script exist in the last set of runs for 2022. If not, then most likely a rename has caused some test to be ignored.

webapp/components/interop-data.js
DanielRyanSmith: Yes, that's a good idea. It might also be worthwhile making this change eventually for the `results-analysis` repo.

webapp/components/interop-summary.js
DanielRyanSmith: Good catch! That is correct that this change only adjusts the scores at the end and will not reflect incremental changes in the graph. I will revert to the old implementation for 2022 in order to keep the graph consistent.

DanielRyanSmith:
> A few comments on the details, but I think the most important thing here is how the static files were generated. Can you document where the files were copied from, or what script (at which commit of which repo?) was used to generate them?

I was considering adding this here, but maybe this is something we should document in the [results-analysis repository](https://github.com/web-platform-tests/results-analysis) - maybe a section on generating EOY interop numbers. Do you think that is better documented in that repo?

> I'm thinking especially about the fact that tests may have been renamed after 2022, so if we use the current wpt-metadata repo to get the test list, we may not have exactly the same test list as was used at the end of 2022. Ideally, we should check that all of the tests we use as input to the scoring script exist in the last set of runs for 2022. If not, then most likely a rename has caused some test to be ignored.

The static files have been generated using the existing `results-analysis` script, which is the same script today that populates the values on the dashboard for previous years. There's a possibility that we could try to replicate a more historically accurate version of the metadata and script, but it would take some additional work to try and find historical metadata and use that in the score generation script, which we don't have functionality for at the moment.

The scores seemed quite close to their original values, although it's possible (or even likely) there are some small differences. I'm receptive to trying to pinpoint these historical numbers and try to find discrepancies, but it likely wouldn't have a large effect on overall scores visible in the old dashboards. It depends on if we decide it's important to take a deep dive into this to maintain historical accuracy.

foolip:
Thanks @DanielRyanSmith, I've reviewed https://github.com/web-platform-tests/results-analysis/pull/172. When it comes to correctness, I don't think we should worry about changes to the script, only tests labeled in the wpt-metadata repo.

For Interop 2021, we already have the original test list in files in https://github.com/web-platform-tests/results-analysis/tree/f805a0eaeab7568ffc53d2df9b363ebc5b76bf68/compat-2021. Can you hack the script to use those tests and see if the scores change at all? If they do, that means there's some difference. But if the difference is a rounding error, we can decide not to care. In https://github.com/web-platform-tests/results-analysis/pull/172 I've asked to dump the tests used to score the static tests into a file, and with that we can understand the difference in the future if needed.

For Interop 2022, we never had the test list in files. I hacked up https://github.com/web-platform-tests/wpt-metadata/blob/foolip/interop-2022-list/index.js on top of the last commit of 2022 and it produced (after sorting) the CSV file in https://gist.github.com/foolip/cedfdade30b86f934a673856a6d4b5cd with 1030 tests.

Running the script again with the current wpt-metadata repo does show some changes to the labeled tests:

```diff
@@ -1,4 +1,11 @@
-interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=0
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=1
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=2
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=3
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=4
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=5
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=6
+interop-2022-cascade,css/css-cascade/all-prop-revert-layer-noop.html?include=7
 interop-2022-cascade,css/css-cascade/all-prop-revert-layer.html
 interop-2022-cascade,css/css-cascade/layer-basic.html
 interop-2022-cascade,css/css-cascade/layer-cssom-order-reverse.html
@@ -86,6 +93,7 @@
 interop-2022-color,css/css-color/lch-006.html
 interop-2022-color,css/css-color/lch-007.html
 interop-2022-color,css/css-color/lch-008.html
+interop-2022-color,css/css-color/nested-color-mix-with-currentcolor.html
 interop-2022-color,css/css-color/oklab-001.html
 interop-2022-color,css/css-color/oklab-002.html
 interop-2022-color,css/css-color/oklab-003.html
@@ -553,7 +561,7 @@
 interop-2022-forms,css/css-ui/appearance-auto-non-html-namespace-001.html
 interop-2022-forms,css/css-ui/appearance-button-001.html
 interop-2022-forms,css/css-ui/appearance-checkbox-001.html
-interop-2022-forms,css/css-ui/appearance-cssom-001.html
+interop-2022-forms,css/css-ui/appearance-cssom-001.html?exclude=Invalid
 interop-2022-forms,css/css-ui/appearance-initial-value-001.html
 interop-2022-forms,css/css-ui/appearance-listbox-001.html
 interop-2022-forms,css/css-ui/appearance-menulist-001.html
@@ -687,7 +695,6 @@
 interop-2022-forms,html/semantics/forms/the-input-element/datetime-local.html
 interop-2022-forms,html/semantics/forms/the-input-element/datetime.html
 interop-2022-forms,html/semantics/forms/the-input-element/defaultValue-clobbering.html
-interop-2022-forms,html/semantics/forms/the-input-element/email-set-value.html
 interop-2022-forms,html/semantics/forms/the-input-element/email.html
 interop-2022-forms,html/semantics/forms/the-input-element/files.html
 interop-2022-forms,html/semantics/forms/the-input-element/hidden-charset-case-sensitive.html
```

I've checked and most of these changes were made after 2022 ended, which means the new test names won't be found when scoring 2022.

@DanielRyanSmith can you look into scoring 2022 with the test lists taken from the linked CSV file? If it can also dump the files used into a format similar to 2021 and 2023 that would be great, so we end up having committed tests lists per year all in one place.

Issue #2617, opened 2021-08-12
Search cache error "Failed to fetch test runs" on staging


foolip:
Steps to reproduce:
- Start at https://staging.wpt.fyi/results/?label=master&label=experimental&aligned
- Search for "(safari:!pass&safari:!ok)", which changes the URL to https://staging.wpt.fyi/results/?label=master&label=experimental&aligned&q=%28safari%3A%21pass%26safari%3A%21ok%29

You should see this:
![image](https://user-images.githubusercontent.com/498917/129237481-e9eeedd1-b45e-4e34-851b-014179f28654.png)

The same steps starting from https://wpt.fyi/results/?label=master&label=experimental&aligned doesn't have the same problem, so this might be a regression in staging.

@KyleJu can you look into this? When was the latest prod push and have any interesting changes happened since then?

This was reported by @gsnedders originally with this example:
https://staging.wpt.fyi/results/?q=%28safari%3A%21pass%26safari%3A%21ok%29%20%28firefox%3Apass%7Cfirefox%3Aok%29%20%28chrome%3Apass%7Cchrome%3Aok%29&run_id=5755468253233152&run_id=6204903965851648&run_id=5894387494027264

KyleJu:
I was able to search both https://staging.wpt.fyi/results/?q=%28safari%3A%21pass%26safari%3A%21ok%29%20%28firefox%3Apass%7Cfirefox%3Aok%29%20%28chrome%3Apass%7Cchrome%3Aok%29&run_id=5755468253233152&run_id=6204903965851648&run_id=5894387494027264 and `(safari:!pass&safari:!ok)` without any issues on staging. 

My last prod push is last night 11 PM PST time so they should be in sync

gsnedders:
There's definitely something up with the cache; https://staging.wpt.fyi/results/?q=%28safari%3A%21pass%26safari%3A%21ok%29%20%28firefox%3Apass%7Cfirefox%3Aok%29%20%28chrome%3Apass%7Cchrome%3Aok%29%20%21sharedworker&run_id=5755468253233152&run_id=6204903965851648&run_id=5894387494027264 now fails for example

KyleJu:
The staging deployment using 9c46dad0d0a839f8aa059349fd0f2f8a2228704a also failed at any search queries. searchcache on wpt.fyi is working correctly. Searchcache returns `Status Code: 502`.  I suspected this line from logs

[...]

PR #3313, opened 2023-05-16
Fix AND conjunction in query

### LINKED ISSUE #3171: and conjunction is not working properly
wpt.fyi search synx support [`and` key word](https://github.com/web-platform-tests/wpt.fyi/blob/7142bfde7f958b0add956753d33d249622e37acb/webapp/components/test-search.js#L88). It appears to be broken and I suspect that it is a parser issue

KyleJu:
Fix #3171. The and-conjunction likely stopped working since https://github.com/web-platform-tests/wpt.fyi/pull/1057. The issue never surfaced because `AND` is implied between query terms, and the top-level queries work properly without it.

### Issues
and-conjunction does not work when it queries two or more distinct products. For example,  a simple query like `edge:pass and triaged:edge` works, but `edge:pass and triaged:chrome` does not work.

### Cause
https://github.com/web-platform-tests/wpt.fyi/pull/1057 introduces the notion of `exists` root node, and `exists` is implied [unless another root node is specified](https://github.com/web-platform-tests/wpt.fyi/tree/main/api/query#root-queries), e.g. all, none. A query is parsed differently when `and` is explicit vs implicit:

```
      assertQueryParse('chrome:pass firefox:timeout', {
        exists: [
          {product: 'chrome', status: 'PASS'},
          {product: 'firefox', status: 'TIMEOUT'},
        ],
      });
      assertQueryParse('chrome:pass AND firefox:timeout', {
        exists: [{
          and: [
            {product: 'chrome', status: 'PASS'},
            {product: 'firefox', status: 'TIMEOUT'},
          ]
        }],
      });
```

By definition & implementation of `exists`, each of query objects of `exists` must be true for a single run. Without the explicit `and`, the query objects of `exists` root are `{product: 'chrome', status: 'PASS'}`, `{product: 'firefox', status: 'TIMEOUT'}`.  With `and`,  it is a single query object  `and: [{product: 'chrome', status: 'PASS'},{product: 'firefox', status: 'TIMEOUT'},]`. In `BindToRuns` method, there does not exists `ONE` run where it can be both chrome and firefox. Thus it fails to bind to any runs.

###  Solutions
For top-level queries (means without nested queries) with a single query object, e.g. Or/And, bind them directly without the `exists` logic.

### Future Issues
When there are nested `AND` with two or more distinct products, the query still will not work as expected, because they are bound to one run at `BindToRuns` , e.g. `edge:pass or (chrome:pass firefox:fail)`. However, they never worked to begin with. I believe `exists` root node is still correct by its definiton, but should not be attached to all queries by default. 

Alternatively,  either user somehow rewrites this query with a different root node, e.g. all, none, etc. or we should remove  `exists` node by default.

### Follow-up
I will file a bug against the future issues mentioned above

### DIFF
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -118,6 +118,11 @@ type AbstractExists struct {
 // each specific/individual run.
 func (e AbstractExists) BindToRuns(runs ...shared.TestRun) ConcreteQuery {
 	queries := make([]ConcreteQuery, len(e.Args))
+	// When the nested query is a single query, e.g. And/Or, bind that query directly.
+	if len(e.Args) == 1 {
+		return e.Args[0].BindToRuns(runs...)
+	}
+
 	for i, arg := range e.Args {
 		var query ConcreteQuery
 		// Exists queries are split; one run must satisfy the whole tree.
--- a/api/query/atoms_test.go
+++ b/api/query/atoms_test.go
@@ -624,6 +624,38 @@ func TestStructuredQuery_combinednotlink(t *testing.T) {
 		AbstractQuery: AbstractExists{Args: []AbstractQuery{AbstractAnd{Args: []AbstractQuery{TestNamePattern{Pattern: "cssom"}, AbstractNot{Arg: AbstractLink{Pattern: "chromium.bug"}}}}}}}, rq)
 }
 
+func TestStructuredQuery_existsSimple(t *testing.T) {
+	var rq RunQuery
+	err := json.Unmarshal([]byte(`{
+		"run_ids": [0, 1, 2],
+		"query": {
+			"exists": [
+				{"and": [
+					{"pattern": "cssom"}
+				  ]
+				}
+			]
+		}
+	}`), &rq)
+	assert.Nil(t, err)
+	assert.Equal(t, RunQuery{RunIDs: []int64{0, 1, 2},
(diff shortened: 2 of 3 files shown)


DanielRyanSmith (approved):
Ran some of these queries in the staging environment:
`chrome:timeout and background-fetch/abort.https.window.html`
`chrome:timeout and firefox:fail and background-fetch/abort.https.window.html`
`chrome:timeout and safari:timeout and firefox:pass`
`firefox:timeout and safari:notrun`
`edge:missing and chrome:pass`
`chrome:pass and edge:missing and firefox:fail and safari:fail`

All give appropriate results, but do not work in production. Looks great! Thanks for the fix Kyle. 🙂

jcscottiii (approved):

KyleJu:
> Ran some of these queries in the staging environment: `chrome:timeout and background-fetch/abort.https.window.html` `chrome:timeout and firefox:fail and background-fetch/abort.https.window.html` `chrome:timeout and safari:timeout and firefox:pass` `firefox:timeout and safari:notrun` `edge:missing and chrome:pass` `chrome:pass and edge:missing and firefox:fail and safari:fail`
> 
> All give appropriate results, but do not work in production. Looks great! Thanks for the fix Kyle. 🙂

Interesting. That means it also fixed #2617. These queries did not work at all on staging

KyleJu:
Filed https://github.com/web-platform-tests/wpt.fyi/issues/3323 as a follow-up

PR #3332, opened 2023-05-23
Update score for Accessibility Testing (70%)

zcorpan:
See https://github.com/web-platform-tests/interop-2023-accessibility-testing/issues/3#issuecomment-1546395755

### DIFF
--- a/webapp/components/interop-data.js
+++ b/webapp/components/interop-data.js
@@ -300,7 +300,8 @@ export const interopData = {
         'name': 'Accessibility Testing',
         'url': 'https://github.com/web-platform-tests/interop-2023-accessibility-testing',
         'scores_over_time': [
-          { 'date': '2023-03-08', 'score': 600 }
+          { 'date': '2023-03-08', 'score': 600 },
+          { 'date': '2023-05-13', 'score': 700 }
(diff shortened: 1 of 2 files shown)


zcorpan:
https://github.com/web-platform-tests/wpt.fyi/actions/runs/5057204533/jobs/9075667648?pr=3332 failed, but not clear to me why

KyleJu:
Looks like one of the interop tests failed:
```
chrome 113               ✖ components/test/interop-data.html » InteropData » should match interop-data.json exactly

  expected { Object (2021, 2022, ...) } to deeply equal { Object (2021, 2022, ...) }
    <unknown> at      assert.deepEqual at /components/chai/chai.js:2054:0
    <unknown> at   Context.<anonymous> at interop-data.html:16:0
```

DanielRyanSmith:
We keep an identical copy of the data at `webapp/static/interop-data.json` per request, as this file is used by some of Mozilla's tools. If you make this change in that file as well, it should fix the test failure. I'll update the test message to be more verbose about the fix.

KyleJu (approved):

zcorpan:
Fixed

PR #3341, opened 2023-06-01
Remove two-way binnding for triage mode

KyleJu:
Replace the two-way binding annotation with a one-way binding and an event handler for `isTriageMode`. 

`isTriageMode` is used to enable the triage metadata mode on `wpt-app`. Rather than an automatic upward data flow, `github-login` will explicitly fire a `triagemode` event when the toggle is clicked.

Tested on [staging](https://remove-github-binding-dot-wptdashboard-staging.uk.r.appspot.com/results/background-fetch?label=master&label=experimental&aligned)

### DIFF
--- a/webapp/components/wpt-header.js
+++ b/webapp/components/wpt-header.js
@@ -55,7 +55,7 @@ class WPTHeader extends WPTFlags(PolymerElement) {
           <a href="/">web-platform-tests dashboard</a>
         </h1>
         <template is="dom-if" if="[[githubLogin]]">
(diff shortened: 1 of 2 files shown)


KyleJu:
Fix: need to change the two-way data binding symbols {{}}  to  [[]] in the parent dom

jcscottiii (approved):
Thanks for showing https://github.com/web-platform-tests/wpt.fyi/pull/3341/commits/81dffa47dbf12662efbb76ddc7180c41a1fea0bc

That makes sense. I have updated my PR to do that same.

PR #3342, opened 2023-06-01
Use one-way binding for the path property

past:
As part of the migration from the unsupported Polymer to Lit, remove the two-way binding for the path property.


DanielRyanSmith (approved):
This looks correct to me. After viewing the files, I can't identify any additional use the two-way data binding offered here (from what I know 🙂).

past:
@KyleJu I manually tested the changes on [staging](https://past-path-binding-conv-dot-wptdashboard-staging.uk.r.appspot.com/results/?label=experimental&label=master&aligned) and I don't see any issues. Can you take an extra look before we merge this?

KyleJu (approved):
LGTM with a comment to address:

The biggest impact of this change is supposed to be navigation, but there is a redundant use of path across the whole app, which makes the two-way binding of path obsolete. The navigation works correctly after some testing, but could you include [this bit](https://github.com/web-platform-tests/wpt.fyi/blob/695181d110a9f1d6b2b6e9360ac31a43beb1f45e/webapp/views/wpt-app.js#L156)? It should be `path="[[subroute.path]]"`

PR #3350, opened 2023-06-05
Remove notify for testruns.

jcscottiii:
TestRunsQueryLoader is actually extended and used everywhere. So no parent objects actually need to be notified with an event listener.

---

Removed the notify from TestRunsQueryLoader
TestRunsQueryLoader is a base class that is extended by:
- TestRunsBase ('wpt-results-base') which is the parent class for
  - TestFileResultsTable ('test-file-results-table'). Not wrapped or extended by anything else.. CONCLUSION SAFE.
- TestRunsUIBase, which is the parent class for:
  - WPTRuns ('wpt-runs') - Has no parents. Not extended anymore. Only used by webapp/template/test-runs.html. notify does not go up.
  - WPTApp ('wpt-app'). Conclusion: SAFE
  - WPTResults ('wpt-results'). Used in wpt-app. Conclusion: NEED TO INVESTIGATE. Turns out that wpt-app has the testRuns already.

### DIFF
--- a/webapp/components/test-runs.js
+++ b/webapp/components/test-runs.js
@@ -19,7 +19,6 @@ const TestRunsQueryLoader = (superClass) =>
         // Fetched + parsed JSON blobs for the runs
         testRuns: {
           type: Array,
-          notify: true,
         },
         nextPageToken: String,
         displayedProducts: Array,
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -154,7 +154,7 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
                      is-loading="{{resultsLoading}}"
                      structured-search="[[structuredSearch]]"
                      path="{{subroute.path}}"
-                     test-runs="{{testRuns}}"
+                     test-runs="[[testRuns]]"
                      test-paths="{{testPaths}}"
                      search-results="{{searchResults}}"
                      subtest-row-count={{subtestRowCount}}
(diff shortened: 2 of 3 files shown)


jcscottiii:
placed a breakpoint on the following places:
- https://github.com/web-platform-tests/wpt.fyi/blob/dfd9d1f4ac2fc1b5b71b1f69c87883cfd4b1da59/webapp/components/test-file-results.js#L107
- https://github.com/web-platform-tests/wpt.fyi/blob/dfd9d1f4ac2fc1b5b71b1f69c87883cfd4b1da59/webapp/components/test-file-results.js#L118
- https://github.com/web-platform-tests/wpt.fyi/blob/dfd9d1f4ac2fc1b5b71b1f69c87883cfd4b1da59/webapp/components/test-file-results.js#L160
- https://github.com/web-platform-tests/wpt.fyi/blob/e55b24cd2e71b59c3594b3ee53220cf7d0c61b7e/webapp/components/test-file-results-table.js#L301
- https://github.com/web-platform-tests/wpt.fyi/blob/695181d110a9f1d6b2b6e9360ac31a43beb1f45e/webapp/views/wpt-results.js#L528

Loaded this page  https://testruns-remove-notify-dot-wptdashboard-staging.uk.r.appspot.com/results/accelerometer/Accelerometer_insecure_context.html?label=experimental&label=master&aligned

Saw that testRuns were populated.

KyleJu (approved):
Thank you for the detailed writeup!

PR #3411, opened 2023-07-13
Update history endpoint

racheljay:
This PR alters the existing `/api/history` endpoint to accept POST requests rather than GET requests. This is due to GET requests supporting a finite URL length. The POST request body will give us more flexibility with what information can be sent along with the request.

### DIFF
--- a/api/test_history.go
+++ b/api/test_history.go
@@ -1,16 +1,45 @@
 package api
 
 import (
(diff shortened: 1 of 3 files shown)


DanielRyanSmith (approved):
Looks great!

KyleJu (approved):

PR #3433, opened 2023-08-02
Use Chrome For Testing for downloading ChromeDriver

jcscottiii:
Currently, all the go_chrome_test github actions are failing since this PR https://github.com/web-platform-tests/wpt.fyi/pull/3419. Turns out that we cannot use the old way for versions >114. https://stackoverflow.com/a/76729048

The likely culprit is that the distro where we download Chrome updated to 115. And then https://chromedriver.storage.googleapis.com stopped after 114. (You can visit that url to see the keys available) 

In the future, we can download the chrome version too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -148,13 +148,18 @@ chrome: wget
 		sudo dpkg --install $${ARCHIVE} 2>/dev/null; \
 	fi
 
-# https://sites.google.com/a/chromium.org/chromedriver/downloads/version-selection
-chromedriver: wget unzip chrome
+# Pull ChromeDriver from Chrome For Testing
+# Need to create the CHROMEDRIVER_PATH and then move the files in because the
+# directory structure in chromedriver_linux64.zip has changed.
+# https://stackoverflow.com/a/76729048
+chromedriver: wget unzip chrome jq
 	if [[ ! -f "$(CHROMEDRIVER_PATH)" ]]; then \
-		CHROME_VERSION=$$(google-chrome --version | grep -ioE "[0-9]+\.[0-9]+\.[0-9]+"); \
-		CHROMEDRIVER_VERSION=$$(curl https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$${CHROME_VERSION}); \
-		wget -q https://chromedriver.storage.googleapis.com/$${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip; \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -148,13 +148,18 @@ chrome: wget
...
+# directory structure in chromedriver_linux64.zip has changed.
+# https://stackoverflow.com/a/76729048
+chromedriver: wget unzip chrome jq
 	if [[ ! -f "$(CHROMEDRIVER_PATH)" ]]; then \
-		CHROME_VERSION=$$(google-chrome --version | grep -ioE "[0-9]+\.[0-9]+\.[0-9]+"); \
-		CHROMEDRIVER_VERSION=$$(curl https://chromedriver.storage.googleapis.com/LATEST_RELEASE_$${CHROME_VERSION}); \
-		wget -q https://chromedriver.storage.googleapis.com/$${CHROMEDRIVER_VERSION}/chromedriver_linux64.zip; \
-		sudo unzip chromedriver_linux64.zip -d $$(dirname $(CHROMEDRIVER_PATH)); \
+		CHROME_VERSION=$$(google-chrome --version | grep -ioE "[0-9]+\.[0-9]+\.[0-9]+\.[0-9]+"); \
+		CHROMEDRIVER_URL=$$(curl -s https://googlechromelabs.github.io/chrome-for-testing/known-good-versions-with-downloads.json | jq -r ".versions[] | select(.version == \"$${CHROME_VERSION}\") | .downloads.chromedriver[] | select(.platform == \"linux64\") | .url"); \
```
mathiasbynens, line 158: This code assumes that for any non-CfT Google Chrome binary, there is a corresponding ChromeDriver binary release. Unfortunately this is not _yet_ true in a few edge cases — e.g. A/B testing releases, where different Chrome versions are released per platform, could currently result in the latest CfT having a lower patch version than the non-CfT version.

To make this code 100% robust, you’d still apply [Version Selection](https://chromedriver.chromium.org/downloads/version-selection) like before, but now using the new CfT JSON endpoints. Concretely, revert back to `grep -ioE "[0-9]+\.[0-9]+\.[0-9]+")` to get just the `MAJOR.MINOR.BUILD` parts of the version number, and then run it through https://googlechromelabs.github.io/chrome-for-testing/latest-patch-versions-per-build-with-downloads.json to find the corresponding ChromeDriver version there.

PR #3434, opened 2023-08-03
Get data from local datastore instead of from mock JSON

racheljay:
This PR adds logic to populate temporary fake data into the local environment datastore. Previously it has been pulling mock data from a static JSON file. It is now pulling from and processing data from a mock data set in the local version of the datastore.


util/populate_dev_data.go
```diff
@@ -318,3 +321,152 @@ func FetchPendingRuns(wptdHost string) ([]shared.PendingTestRun, error) {
 	err := shared.FetchJSON(url, &pendingRuns)
 	return pendingRuns, err
 }
+
+func addFakeHistoryData(store shared.Datastore) {
+	// browser_name,browser_version,date,test_name,subtest_name,status
```
DanielRyanSmith, line 326: This looks fine as a temporary change. Can we add a TODO comment here to import real data here when staging is populated with real data?

DanielRyanSmith (commented):
NIce work with taking the raw db format and converting it to be ingested client side 👍
It looks like this change breaks the timeline charts using fake data on staging/prod. Can we add a change to submit the old mocked JSON if no history is pulled from the database?

BertsLifee (approved):

BertsLifee (approved):

BertsLifee (approved):

DanielRyanSmith (approved):
Looks great! This should work just as intended once the raw data exists in staging 🙂

KyleJu:
WOOHOO Great work!  Don't forget to add tests for the `testHistoryHandler` function!

PR #3440, opened 2023-08-08
change testHistoryEntry format

racheljay:
This PR removes the `BrowserVersion` item in the `TestHistoryEntry` struct because it is not being used. It also renames the item `Browser` to `BrowserName` to better match naming conventions elsewhere in the project.

### DIFF
--- a/api/test_history.go
+++ b/api/test_history.go
@@ -10,6 +10,12 @@ import (
 	"github.com/web-platform-tests/wpt.fyi/shared"
 )
 
+// Subtest represents the final format for subtest data.
+type Subtest map[string]string
+
+// Browser represents the final format for browser data.
+type Browser map[string][]Subtest
+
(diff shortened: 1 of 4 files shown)


api/test_history.go
```diff
@@ -83,10 +83,10 @@ func testHistoryHandler(w http.ResponseWriter, r *http.Request) {
 
 	for _, run := range runs {
 
-		_, ok := testsByBrowser[run.Browser]
+		_, ok := testsByBrowser[run.BrowserName]
```
KyleJu, line 86: nit: move the type declaration to the top of the file, `type Subtest map[string]string` and `type Browser map[string][]Subtest`

KyleJu (approved):

DanielRyanSmith (approved):

api/test_history.go
racheljay: done 👍

Issue #3351, opened 2023-06-07
Labeling directories


foolip:
For https://github.com/web-platform-tests/interop-2023-accessibility-testing/issues/44, it would be useful to be able to label directories, to be able to replace this query of multiple directories and one test: [`path:/accessibility or path:/accname or path:/html-aam or path:/wai-aria or path:/svg-aam or path:/graphics-aria or path:/graphics-aam or path:/dpub-aam or path:/html/dom/aria-element-reflection.html`](https://wpt.fyi/results/?label=experimental&label=master&aligned&q=path%3A%2Faccessibility%20or%20path%3A%2Faccname%20or%20path%3A%2Fhtml-aam%20or%20path%3A%2Fwai-aria%20or%20path%3A%2Fsvg-aam%20or%20path%3A%2Fgraphics-aria%20or%20path%3A%2Fgraphics-aam%20or%20path%3A%2Fdpub-aam%20or%20path%3A%2Fhtml%2Fdom%2Faria-element-reflection.html)

Labeling is currently done in [wpt-metadata](https://github.com/web-platform-tests/wpt-metadata) with support in the wpt.fyi UI to add labels to tests, but not directories.

Option 1: We already support linking a bug to a whole directory in wpt-metadata ([example](https://github.com/web-platform-tests/wpt-metadata/blob/3794a05315b45116293aae9f30800a793ee54ce4/css/css-grid/masonry/META.yml)) and it we could do the same for labels. https://github.com/web-platform-tests/wpt-metadata/pull/4303 is an example of what this would look like in wpt-metadata.

Option 2: We could support labels in WPT itself, via its META.yml files. That would make this similar to https://github.com/web-platform-tests/wpt.fyi/issues/372, which is about spec links extracted from WPT. @KyleJu is working on this now.

In either case, the idea is that `label:accessibility` would work as a search atom, and it would not be necessary to individually label all of the tests in each directory.

foolip:
@cookiecrook did https://github.com/web-platform-tests/wpt-metadata/pull/4297, so no need for my draft https://github.com/web-platform-tests/wpt-metadata/pull/4303.

It looks from https://wpt.fyi/results/?label=master&label=experimental&aligned&q=label%3Aa11y like this doesn't work in the search cache. @KyleJu can you help check how labels and bugs are different? Given that we have `triaged:chrome` which is based on bug links I guess the search cache does understand '*' as a wildcard for tests in some contexts, but not for labels?

[...]

PR #3447, opened 2023-08-10
Support wildcards in label search

KyleJu:
Fix https://github.com/web-platform-tests/wpt.fyi/issues/3351. Support searching wildcards in label.  

This is intended for use case https://github.com/web-platform-tests/wpt-metadata/pull/4297, and https://github.com/web-platform-tests/wpt-metadata/pull/4303

### DIFF
--- a/api/query/cache/index/filter.go
+++ b/api/query/cache/index/filter.go
@@ -324,6 +324,18 @@ func (tl TestLabel) Filter(t TestID) bool {
 	}
 
 	labels, ok := tl.metadata[name]
+	dir := filepath.Dir(name)
+	// Dir terminates with either '.' (when the top-level is a file) or '/'
+	// (when the top-level is a directory).
+	for !ok && len(dir) > 1 {
+		labels, ok = tl.metadata[dir+"/*"]
+		if ok {
+			break
+		}
+
+		dir = filepath.Dir(dir)
+	}
+
 	if !ok {
 		return false
 	}
--- a/api/query/cache/index/index_filter_test.go
+++ b/api/query/cache/index/index_filter_test.go
@@ -833,6 +833,65 @@ func TestBindExecute_TestLabel(t *testing.T) {
 	assert.Equal(t, expectedResult, srs[0])
 }
 
+func TestBindExecute_LabelWithWildcards(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	loader := NewMockReportLoader(ctrl)
+	idx, err := NewShardedWPTIndex(loader, testNumShards)
+	assert.Nil(t, err)
+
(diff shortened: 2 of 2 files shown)


KyleJu:
A screenshot for a wildcard search:
Screenshot 2023-08-11 at 9 27 35 AM (2)

api/query/cache/index/index_filter_test.go
```diff
@@ -833,6 +833,65 @@ func TestBindExecute_TestLabel(t *testing.T) {
...
+						Test:   "/d/e/f",
+						Status: "FAIL",
+					},
+				},
+			},
+		},
+	})
+	metadata := map[string][]string{
+		"/foo/bar/b.html": {"random"},
+		"/a/*":  {"interop1", "INTEROP2"},
```
jcscottiii, line 863: There's a possible edge case. Here's the scenario: If you modify `matchingTestName` to equal `/a` and this metadata line to `"/*":  {"interop1", "INTEROP2"},`, you will get zero results as-is.

The question I have: Is it even possible for the metadata to produce this scenario (if the appropriate meta.yml was modified)? If not, you can ignore this question.

*I don't think this would be a useful scenario to even have in the metadata (since it's just a root level wildcard) but I didn't want to assume.*

jcscottiii (approved):
LGTM. Just one question

api/query/cache/index/index_filter_test.go
KyleJu: Great question! It is a possible edge case, but it wouldn't make sense to have a root level wildcard, since it would label all WPT tests.  However, we should safeguard this case in the wpt-metadata repository.

api/query/cache/index/filter.go
```diff
@@ -324,6 +324,18 @@ func (tl TestLabel) Filter(t TestID) bool {
 	}
 
 	labels, ok := tl.metadata[name]
+	dir := filepath.Dir(name)
+	// Dir terminates with either '.' (when the top-level is a file) or '/'
+	// (when the top-level is a directory).
+	for !ok && len(dir) > 1 {
```
KyleJu, line 330: On a second look, I want to point out a drawback of this for loop implementation. Since a test can have labels on a test level and directory level, this for loop will terminate as soon as `A` set of labels are found.  However, if a test `/a/b/c` is labelled to `foo` and `/a/*` is labelled to `bar`,  `label:bar` will not return the intended result, due to early termination at `for !ok ...`. This drawback will affect all wildcards implementation in this file.

KyleJu: I will file a bug

KyleJu: https://github.com/web-platform-tests/wpt.fyi/issues/3448

Issue #3446, opened 2023-08-10
Make sure history graphs update when the page is exited


racheljay:
We are currently getting some weird behavior where, once the "show new history" button is pressed, the generated history data will persist to other tests. We need to ensure that the history generated is reset after the page is exited.

PR #3459, opened 2023-08-16
Refresh history data when page is changed

racheljay:
This PR solves issue #3446 where the same data was persisting for different tests. I solved this by checking for and clearing any existing data before running the query for new data. I also added logic to hide the history timelines when a new test page is visited after viewing a previous one.

### DIFF
--- a/webapp/components/new-test-results-history-grid.js
+++ b/webapp/components/new-test-results-history-grid.js
@@ -97,7 +97,7 @@ class TestResultsGrid extends PolymerElement {
   }
 
   displayCharts(showTestHistory, path) {
-    if (!path || !showTestHistory) {
+    if (!path || !showTestHistory || path.split('?')[0].slice(-5) !== ".html") {
       return;
     }
 
@@ -234,7 +234,7 @@ class TestResultsGrid extends PolymerElement {
   // get test history and aligned run data
   async getTestHistory(path) {
     // If there is existing data, clear it to make sure nothing is cached
-    if(this.historicalData && path.split('?')[0].slice(-5) === ".html") {
+    if(this.historicalData) {
       this.historicalData = {};
     }
 
--- a/webapp/components/new-test-results-history-grid.js
+++ b/webapp/components/new-test-results-history-grid.js
@@ -89,7 +89,7 @@ class TestResultsGrid extends PolymerElement {
   static get observers() {
     return [
       'displayCharts(showTestHistory, path)',
-      'updateForPath(showTestHistory, path)'
+      'clearCharts(path, this.historicalData)',
     ];
   }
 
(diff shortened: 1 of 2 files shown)


webapp/components/new-test-results-history-grid.js
```diff
@@ -88,21 +88,24 @@ class TestResultsGrid extends PolymerElement {
...
 
-  displayCharts(showTestHistory) {
-    if (!showTestHistory || this.historicalData !== undefined) {
+  displayCharts(showTestHistory, path) {
+    if (!path) {
+      return;
+    }
+    if (!showTestHistory) {
       return;
     }
```
DanielRyanSmith, line 105:
```suggestion
    if (!path || !showTestHistory) {
      return;
    }
```

webapp/components/new-test-results-history-grid.js
```diff
@@ -231,17 +234,22 @@ class TestResultsGrid extends PolymerElement {
   }
 
   // get test history and aligned run data
-  async getTestHistory() {
-    if (!this.path) {
+  async getTestHistory(path) {
+    if (!path) {
       throw new Error('Test path is undefined');
     }
```
DanielRyanSmith, line 240: This check is redundant now that the value is checked for in `displayCharts()`, so this block can probably just be removed

DanielRyanSmith (approved):
LGTM with some nits

webapp/components/new-test-results-history-grid.js
racheljay: done 👍

webapp/components/new-test-results-history-grid.js
racheljay: done 👍

webapp/components/new-test-results-history-grid.js
```diff
@@ -88,21 +88,22 @@ class TestResultsGrid extends PolymerElement {
...
   }
 
   static get is() {
     return 'new-test-results-history-grid';
   }
 
-  displayCharts(showTestHistory) {
-    if (!showTestHistory || this.historicalData !== undefined) {
+  displayCharts(showTestHistory, path) {
+    if (!path || !showTestHistory) {
```
KyleJu, line 100: We should only reload data when it is a subtest path

webapp/components/new-test-results-history-grid.js
```diff
@@ -88,21 +88,22 @@ class TestResultsGrid extends PolymerElement {
...
   }
 
   static get is() {
     return 'new-test-results-history-grid';
   }
 
-  displayCharts(showTestHistory) {
-    if (!showTestHistory || this.historicalData !== undefined) {
+  displayCharts(showTestHistory, path) {
+    if (!path || !showTestHistory || path.split('?')[0].slice(-5) !== ".html") {
```
racheljay, line 100: @KyleJu I believe this is now accurately checking for subtests urls. Can you let me know if there are any edge cases you can think of that I'm not covering?

KyleJu: Could you use this method for checking instead? https://github.com/web-platform-tests/wpt.fyi/blob/7d1370ea05466fc84533ebed782d348756e102d5/webapp/components/path.js#L70

racheljay: Done 👍

KyleJu (approved):

PR #3460, opened 2023-08-16
Add Chromium to wpt.fyi browsers

DanielRyanSmith:
This change adds support for Chromium runs, which have not been ingested by wpt.fyi since the renaming of "chrome nightly" to "chromium nightly" in the WPT repo.

### DIFF
--- a/shared/browsers.go
+++ b/shared/browsers.go
@@ -14,7 +14,7 @@ var defaultBrowsers = []string{
 
 // An extra list of known browsers.
 var extraBrowsers = []string{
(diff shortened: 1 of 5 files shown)


DanielRyanSmith:
The good news is that the Chromium runs have been ingested, but are just not able to display due to "chromium" not being listed as a valid product. The staging environment displays the Chromium runs that currently exist 🙂 

https://2023-08-15-add-chromiu-dot-wptdashboard-staging.uk.r.appspot.com/runs?max-count=100&product=chromium

jcscottiii (approved):

Issue #3450, opened 2023-08-11
Adjust Test History Timeline Chart Height


racheljay:
Currently the new history timeline charts are styled to display in a fixed window. We need to style them to adjust in size so that they grow to fit however many rows are generated. Possibly additional logic could be added for extra long test suites to display the charts individually rather than all at once.

PR #3471, opened 2023-08-21
height based on subtest name

racheljay:
This PR fixes issue #3450. Previous iterations of the timeline had a fixed height. The height of the timeline chart is now calculated by the number of subtests, unless there are more than 20, in which case it will add a scroll bar. This PR also fixes a bug where some charts were breaking due to missing data.


KyleJu (approved):

DanielRyanSmith (approved):

DanielRyanSmith:
Unrelated to this PR, but I'm seeing that the sections of the timeline graph are no longer navigating properly to the anticipated run ID link. That is probably worth creating a separate issue and PR for. 👍

Issue #3448, opened 2023-08-11
Missing search results due to wildcard implementation


KyleJu:
See [this comment for more details(https://github.com/web-platform-tests/wpt.fyi/pull/3447#discussion_r1291688758). On a high level, the for loop `for !ok && len(dir) > 1` should not terminate as soon as a result is found, because a test can be triaged to multiple links and labels due to wildcards. It should find all labels and test links, as we it looks up all parent directories.

KyleJu:
A comment for potential [performance improvements](https://github.com/web-platform-tests/wpt.fyi/pull/3472#issuecomment-1690312389)

PR #3472, opened 2023-08-23
Probe directory tree for wildcard matching

KyleJu:
Fix https://github.com/web-platform-tests/wpt-metadata/issues/4665 and a part of https://github.com/web-platform-tests/wpt.fyi/issues/3448.  Probe directory tree for wildcard matching, regardless if a match has found. 

It should not have much performance impact before and after the change.


jcscottiii (approved):
The code LGTM.  Could you monitor the performance of this? Now we are always walking up to the root (which is needed for this implementation). If it is bad, we may need a way to improve this look up

Issue #3490, opened 2023-09-02
Deploy 667f4ce2 to production


past:
Previous deployment was #3474 (9f00b3c2)

Changelist 9f00b3c2...667f4ce2

Changes:
667f4ce2 Bump github.com/google/uuid from 1.3.0 to 1.3.1 (#3478)
e7610aa6 Bump puppeteer from 21.1.0 to 21.1.1 in /webdriver (#3477)
5a3bfc2c change testName to test_name (#3489)
70bc6771 Add history handler tests (#3485)
9fa7a53f Bump cloud.google.com/go/datastore from 1.13.0 to 1.14.0 (#3479)
0eec8c03 Bump chai from 4.3.7 to 4.3.8 in /webdriver (#3476)
0420f1a6 Bump python in /results-processor (#3480)
cf152a9c Bump eslint from 8.47.0 to 8.48.0 in /webapp (#3482)
d1a625c2 Bump @babel/eslint-parser from 7.22.10 to 7.22.11 in /webapp (#3483)
edea0584 height based on subtest name (#3471)

This push is happening as part of the regular weekly push.

Pushing all three services - webapp, processor, and searchcache.

past:
Some checks failed for 667f4ce2, but one was an internal gcloud error while deploying to staging and the other is the intermittent #3090.

KyleJu:
> Some checks failed for [667f4ce](https://github.com/web-platform-tests/wpt.fyi/commit/667f4ce23203236033edc70eb5b9a00c9d0daf75), but one was an internal gcloud error while deploying to staging and the other is the intermittent #3090.

Let me rerun the staging deployment on main

past:
Deployment is now complete.

PR #3491, opened 2023-09-02
Add a new script to automate prod deployment

past:
This shell script automates steps 1-3 of the production deployment instructions. I used it successfully for #3490 and I hope to add the remaining steps in the future.

### DIFF
--- a/util/deploy-production.sh
+++ b/util/deploy-production.sh
@@ -1,0 +1,77 @@
+#!/bin/bash
+
+# Helper script for deploying to production.
+# Needs the following packages to be installed: google-cloud-cli, gh, jq
+
+set -e
+
+usage() {
+  USAGE="Usage: deploy-production.sh [-f]
+    -f : Always deploy (even if checks have failed)"
+  echo "${USAGE}"
+}
+
+while getopts ':fh' flag; do
+  case "${flag}" in
+    f) FORCE_DEPLOY='true' ;;
+    h|*) usage && exit 0;;
+  esac
+done
+
+GH_OWNER="web-platform-tests"
+GH_REPO="wpt.fyi"
+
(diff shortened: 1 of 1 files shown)


util/deploy-production.sh
```diff
@@ -0,0 +1,77 @@
...
+    echo "No new changes to deploy."
+    exit 0
+fi
+CHANGE_COUNT=$(echo "$CHANGELIST"|wc -l)
+echo -e "There are $CHANGE_COUNT changes to deploy:\n$CHANGELIST"
+
+# Verfiy that all commit checks passed.
+FAILED_CHECKS=$(gh api /repos/"$GH_OWNER"/"$GH_REPO"/commits/HEAD/check-runs | jq -r '.check_runs | map(select(.conclusion == "failure"))')
+FAILURES=$(echo "$FAILED_CHECKS" | jq -r 'length')
+if [[ "${FAILURES}" != ""  ]];
```
jcscottiii, line 38: Given FAILURES is an int, shouldn't this be the following:

```
if [[ "${FAILURES}" != "0"  ]];
```

jcscottiii:
```sh
$ FAILURES=0
$ if [[ "${FAILURES}" != ""  ]]; then  echo "Found $FAILURES failures. EXITING!"; else echo "no failures found"; fi
Found 0 failures. EXITING!


# Modify the conditional
$ if [[ "${FAILURES}" != "0"  ]]; then  echo "Found $FAILURES failures"; else echo "no failures found"; fi
no failures found

```

past: Indeed, good catch. Fixed.

jcscottiii (approved):

PR #3505, opened 2023-09-11
Process history script

DanielRyanSmith:
This change adds the script that generates test history to the repository. The script is hosted on a GCP cloud function, and invoked daily on a scheduled task. The script is stored here for version control and will need to be redeployed to the cloud function with subsequent changes.


scripts/README.md
```diff
@@ -0,0 +1,55 @@
+
```
KyleJu, line 1:
```suggestion
```

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    """
+    test_history_entry: TestHistoryEntry = TestHistoryEntry.query().get()
+    if test_history_entry is not None:
+        raise JSONGenerationError(
+            'TestHistoryEntry entities exist in Datastore. '
+            'New JSON files should not be generated if data already exists.')
+
+
+# default parameters used for cloud functions.
+def main(args=None, topic=None) -> str:
```
KyleJu, line 413: We also need a function that deletes old data, e.g. maintaining a one-year window, cleaning all the data before current date - one year.

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    client = ndb.Client(project=PROJECT_NAME)
+    with client.context():
```
KyleJu, line 415: Can we add a verbose flag `-v` and add more debug logging? It will be easier to figure out problems through logging on cloud function, when something goes wrong

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    but the database has not been cleared of existing entries.
+    """
+    pass
+
+
+def get_processing_start_date() -> MostRecentHistoryProcessed:
+    most_recent_processed: MostRecentHistoryProcessed = (
+        MostRecentHistoryProcessed.query().get())
+
+    if most_recent_processed is None:
```
KyleJu, line 396: Does it mean that it will raise an error when user runs it the first time?

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    runs_list: list[MetadataDict] = resp.json()
+
+    # If we have no runs to process in this date interval,
+    # we can skip this interval for processing from now on.
+    if len(runs_list) == 0:
+        print('No runs found for this interval.')
+        update_recent_processed_date(date_entity, end_interval_string)
+        # If we've processed up to (now - 24 hours), then return null,
+        # which signals we're done processing.
+        if end_interval == yesterday:
```
KyleJu, line 351: Shouldn't we return here regardless of `end_interval == yesterday condition`? runs_list has nothing to sort in line 360

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+            return None
+
+    # Sort by revision -> then time start, so that the aligned runs are
+    # processed in groups with each other.
+    # Note that this technically doesn't have an impact if only 1 set of
+    # aligned runs are processed, but this sort will allow this script to
+    # function properly if multiple aligned run sets were to be processed
+    # together.
+    runs_list.sort(key=lambda run: run['revision'])
```
KyleJu, line 360: IIUC, this line does not do anything. What you need is a first/secondary sort, e.g.  [attrgetter](https://docs.python.org/3/howto/sorting.html#operator-module-functions). run_list is being resorted next line, without specifying a tie-breaker rule

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    runs_list.sort(key=lambda run: run['time_start'])
+
+    # Print the dates just to get info on the list of runs we're working with.
+    print('Runs to process:')
+    for run in runs_list:
```
KyleJu, line 365: Should we check if it is a set of 4? or somewhat mark missing results if it is not 4?

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+
+        if should_process_run(run_metadata):
+            process_single_run(run_metadata)
+        else:
+            print('Run has already been processed')
+
+        revisions_processed[revision][browser_name] = True
+        # If all runs for this revision have been processed, we can update
+        # the most recently processed date to the run's start time.
+        if (revisions_processed[revision]['chrome'] and
```
KyleJu, line 304:
```suggestion
      if (revisions_processed[revision]['chrome'] and
```
This part should be outside of for loop

KyleJu: There needs a else statement if a browser is missing, and we should call `update_recent_processed_date` regardless because it will always be missing

scripts/build_test_history.py
```diff
@@ -0,0 +1,447 @@
...
+    entities_to_write: list[TestHistoryEntry] = []
+    unique_entities_to_write: set[tuple[str, str]] = set()
+    # Iterate through each test.
+    for test_data in run_data['results']:
+        # Format the test name.
+        test_name = re.sub(r'\s', ' ', test_data['test'])
+
+        create_entity_if_needed(
+            test_name,
+            subtest_name='',
```
KyleJu, line 154:
```suggestion
            subtest_name='',
```
why do we record a (test_name,  '') pair?

scripts/build_test_history.py
KyleJu: Looks like it will be marked as missing next time it runs

scripts/build_test_history.py
DanielRyanSmith: This will raise an error if there's no date entity in datastore that tells us what date to start the processing on. We created the entity manually in the datastore UI to start, but that doesn't seem intuitive. 😅  I've now I also added a flag that can be set to write this entity or set it to a different date 🙂 

```sh
# Set history processing start date to the beginning of 2023
python scripts/build_test_history.py --set-history-start-date=2023-01-01T00:00:00.000Z
```

scripts/build_test_history.py
DanielRyanSmith: True! I added another statement to make sure we return here regardless

scripts/build_test_history.py
DanielRyanSmith: Python has a stable sorting method, so it maintains the relative order of elements with equal keys - however, it's still definitely better and intuitive to do it as you mentioned 👍 I've added updated this to use attrgetter now

scripts/build_test_history.py
DanielRyanSmith: Hopefully we always get 4 runs, otherwise they're not technically aligned, but I also added a check now that throws an error if we get any number other than 4 👍

scripts/build_test_history.py
DanielRyanSmith: Not that we have the check mentioned above (4 runs required), it shouldn't be possible for us to come across a condition when one of these runs is missing

scripts/build_test_history.py
DanielRyanSmith: An empty string for the subtest name represents the status of the overall test

scripts/build_test_history.py
DanielRyanSmith: I agree this is useful - perhaps we can add this at a later date?

scripts/build_test_history.py
DanielRyanSmith: Added!

KyleJu (approved):

PR #3516, opened 2023-09-25
Add support for Firefox for Android as a product

jgraham:
This is called` firefox_android` in wptrunner. 

We currently use the geckoview logo so that it's distinguishable from desktop Firefox, although that logo is relatively unknown, so in the future we should look for a way to get something based on the actual Firefox logo.


KyleJu:
chrome 117               ✖ components/test/test-search.html » <test-search> » Parser/interpreter » browser test status eq

Looks like the [test_search test](https://github.com/web-platform-tests/wpt.fyi/blob/31c3916b7402a280c69fcf17e752ced883899bfe/webapp/components/test/test-search.html#L27) failed. Let me take a look shortly

KyleJu:
> chrome 117 ✖ components/test/test-search.html » » Parser/interpreter » browser test status eq
> 
> Looks like the [test_search test](https://github.com/web-platform-tests/wpt.fyi/blob/31c3916b7402a280c69fcf17e752ced883899bfe/webapp/components/test/test-search.html#L27) failed. Let me take a look shortly

there were issues with [parsing expression grammars](http://en.wikipedia.org/wiki/Parsing_expression_grammar) in [omh](https://github.com/ohmjs/ohm/tree/main) not being able to parse firefox_android, while it works for chrome_android, etc.  I messed around with the matching precedence and got it to work by switching the order of `firefox` and `firefox_android`. Not an ideal fix but I have to understand [the grammar precedence](https://github.com/ohmjs/ohm/blob/39ccead882ad35dad73d99dea854ff60e7146291/doc/patterns-and-pitfalls.md?plain=1#L79) better to find out the root cause

KyleJu (approved):

KyleJu:
@jgraham ready to be submitted

Issue #3090, opened 2022-12-27
Flaky web_component_test action


jcscottiii:
The web_component_test CI action is a little flaky.
An example [run](https://github.com/web-platform-tests/wpt.fyi/actions/runs/3789867524/jobs/6444124312):

```
chrome 108               ✓ components/test/wpt-app.html » <wpt-app> » WPTApp.prototype.* » computePathIsRootDir  » root dir
chrome 108               Tests failed: Error thrown outside of test function: uncaptured fetch: http://localhost:8081/api/versions?product=chrome
                             window.fetch at wpt-app.html:21
        ProductBuilder.versionsURLUpdated at /components/wptdashboard/components/product-builder.js:235
         Object.runObserverEffect [as fn] at /components/@polymer/polymer/lib/mixins/property-effects.js:245
                               runEffects at /components/@polymer/polymer/lib/mixins/property-effects.js:146
        ProductBuilder._propertiesChanged at /components/@polymer/polymer/lib/mixins/property-effects.js:2160
          ProductBuilder._flushProperties at /components/@polymer/polymer/lib/mixins/properties-changed.js:424
                     ProductBuilder.ready at /components/@polymer/polymer/lib/mixins/property-effects.js:2106
                     ProductBuilder.ready at /components/@polymer/polymer/lib/mixins/element-mixin.js:759
         ProductBuilder._enableProperties at /components/@polymer/polymer/lib/mixins/properties-changed.js:399
  TemplateInstance.__enableOrFlushClients at /components/@polymer/polymer/lib/mixins/property-effects.js:2038
chrome 108               CALL quit()
chrome 108               DELETE /session/:sessionID
chrome 108               RESPONSE quit()
chrome 108               BrowserRunner complete
Test run ended in failure: Error thrown outside of test function: uncaptured fetch: http://localhost:8081/api/versions?product=chrome
                             window.fetch at wpt-app.html:21
        ProductBuilder.versionsURLUpdated at /components/wptdashboard/components/product-builder.js:235
         Object.runObserverEffect [as fn] at /components/@polymer/polymer/lib/mixins/property-effects.js:245
                               runEffects at /components/@polymer/polymer/lib/mixins/property-effects.js:146
        ProductBuilder._propertiesChanged at /components/@polymer/polymer/lib/mixins/property-effects.js:2160
          ProductBuilder._flushProperties at /components/@polymer/polymer/lib/mixins/properties-changed.js:424
[...]

PR #3569, opened 2023-10-28
Remove the unfinished chrome_ios product

past:
Remove the Chrome iOS product since the work to upload test runs will not be done any time soon.

Also remove an unused variable that was causing errors in hterm.


past:
The failed check is the intermittent #3090.

WeizhongX (approved):
Thank you, Panos!

Issue #3584, opened 2023-11-02
Deploy 388a0071 to production


past:
Previous deployment was #3559 (96ac7347)

Changelist 96ac7347...388a0071

Changes:
388a0071 Bump eslint from 8.51.0 to 8.52.0 in /webapp (#3565)
24931121 Bump cloud.google.com/go/cloudtasks from 1.12.1 to 1.12.3 (#3571)
ea161d0c Bump google.golang.org/grpc from 1.58.3 to 1.59.0 (#3563)
431598d1 Bump mypy from 1.6.0 to 1.6.1 in /results-processor (#3560)
51ed2809 Update the mobile testing score (#3566)
8a13f910 Remove the unfinished chrome_ios product (#3569)
3cdefdde Add missing functionality to deploy-production.sh (#3544)
cfa10755 Bump urllib3 from 1.26.17 to 1.26.18 in /results-processor (#3554)

This push is happening as part of the regular weekly push.

Pushing all three services - webapp, processor, and searchcache.

past:
Deployment is now complete.

PR #3585, opened 2023-11-02
Add a flag to skip bug creation during deployment

past:
This is useful for restarting a production deployment that was interrupted due to a transient or intermittent error. A common case is an internal error during 'make deploy_production' that disappears when retrying.

This was used successfully for #3584.


jcscottiii (approved):
LGTM!

PR #3607, opened 2023-11-20
update interop 2023 accessibility investigation to 100%

cookiecrook:
Follow on from the 98% PR, which was not yet in main.
https://github.com/web-platform-tests/wpt.fyi/pull/3605

so I cherry-picked those diffs... let me know if that causes a problem. hopefully not. thanks.


cookiecrook:
@KyleJu

jcscottiii (approved):

PR #3636, opened 2023-12-19
Ignore version and os_build conflicts

past:
Since we don't control the OS versions we get from Azure, we can't guarantee there will be no conflicting versions.

### DIFF
--- a/results-processor/wptreport.py
+++ b/results-processor/wptreport.py
@@ -37,7 +37,8 @@ DEFAULT_PROJECT = 'wptdashboard'
 RELEASE_CHANNEL_LABELS = frozenset({'stable', 'beta', 'experimental'})
 # Ignore inconsistent browser minor versions for now.
 # TODO(Hexcles): Remove this when the TC decision task is implemented.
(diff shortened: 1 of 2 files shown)


jgraham:
The GH actions failure here looks unrelated. AFAICT this change is good, but I don't really know this code well.

jcscottiii:
I have a question:

This looks like it will cause "version" and "os_build" conflicts to be ignored for all browsers. Is it safe to ignore those conflicts for all browsers? (I'm not sure of the answer to this myself)
- If not, we should add logic to ignore version and os_build only for safari.
- If yes, this PR LGTM

JonWBedard:
This is clearly correct for Safari.

In general, I can't see how respecting conflicts for OS version and OS build can possibly be correct. The situation that motivated this discussion seems possible for any browser. A new release of the underlying OS (especially a minor one, which OS build, in particular, is necessarily sensitive to) does not imply a new browser version.

jcscottiii (approved):

jgraham (approved):

PR #3647, opened 2024-01-07
Add default Interop dashboard URL route

### LINKED ISSUE #3641: Update dashboard URL to wpt.fyi/interop
Currently there isn't a canonical URL for whatever the latest version of the Interop dashboard is. We instead only have per-year URLs e.g. `wpt.fyi/interop-2023`. It would make sense to me to use the year-specific URLs for the archives, but have a single canonical URL for the overall project.

DanielRyanSmith:
Fixes #3641

This change makes the `/interop` route redirect to the interop dashboard of the current active interop effort.

### DIFF
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -37,8 +37,11 @@ func RegisterRoutes() {
 	shared.AddRoute("/runs", "test-runs", testRunsHandler)
 	shared.AddRoute("/test-runs", "test-runs", testRunsHandler) // Legacy name
 
-	// Dashboard for the interop effort, by year. Defaults to current year.
-	shared.AddRoute("/{name:(?:compat|interop-?)}{year:[0-9]*}", "interop-dashboard", interopHandler)
+	// Dashboard for the interop effort, by year.
+	shared.AddRoute("/{name:(?:compat|interop-)}{year:[0-9]+}", "interop-dashboard", interopHandler)
+
+	// Redirect to current year's interop effort.
+	shared.AddRoute("/interop", "interop-dashboard", interopHandler)
 
 	// Admin-only manual results upload.
 	shared.AddRoute("/admin/results/upload", "admin-results-upload", adminUploadHandler)
--- a/webapp/routes.go
+++ b/webapp/routes.go
@@ -37,11 +37,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/runs", "test-runs", testRunsHandler)
 	shared.AddRoute("/test-runs", "test-runs", testRunsHandler) // Legacy name
 
-	// Dashboard for the interop effort, by year.
-	shared.AddRoute("/{name:(?:compat|interop-)}{year:[0-9]+}", "interop-dashboard", interopHandler)
-
-	// Default interop dashboard route that defaults to current year.
-	shared.AddRoute("/interop", "interop-dashboard", interopHandler)
+	// Dashboard for the interop effort, by year. Defaults to current year.
+	shared.AddRoute("/{name:(?:compat|interop-?)}{year:[0-9]*}", "interop-dashboard", interopHandler)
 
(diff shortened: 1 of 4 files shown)


webapp/routes.go
```diff
@@ -40,6 +40,9 @@ func RegisterRoutes() {
 	// Dashboard for the interop effort, by year.
 	shared.AddRoute("/{name:(?:compat|interop-)}{year:[0-9]+}", "interop-dashboard", interopHandler)
 
+	// Default interop dashboard route that defaults to current year.
+	shared.AddRoute("/interop", "interop-dashboard", interopHandler)
```
KyleJu, line 44:
```suggestion
	shared.AddRoute("/{name:(?:compat|interop-)}{year:[0-9]+}", "interop-dashboard", interopHandler)
	shared.AddRoute("/interop", "interop-dashboard", interopHandler)
```
I think putting the two together is more clear, similar to testRunsHandler. Also merge the comment

KyleJu: Actually a step further, given that /interop and /interop-202x are a part of each other, I think it is better to merge /interop to the regex `{name:(?:compat|interop-)}{year:[0-9]+}`. wdyt?

KyleJu:
To clarify here: this change is a redirect from /interop to /interop-2023. Is this the single canonical URL behaviour that jgraham is looking for? Or they want /interop without the redirect?

webapp/routes.go
DanielRyanSmith: I think that sounds like a good idea 🙂
I'll add it to the existing regex

DanielRyanSmith:
> To clarify here: this change is a redirect from /interop to /interop-2023. Is this the single canonical URL behaviour that jgraham is looking for? Or they want /interop without the redirect?

This is a good question. I'll defer to @jgraham, but my first thought is that it might be useful to specifically redirect to `/interop-{year}`, so that it is fairly obvious that there is a URL that links to the specific year. If we do not make it obvious that `/interop` will refer to the active year only, there might be more instances of users' old links directing to a version of the interop dashboard they did not intend.

jgraham:
Well I think ideally I'd like `/interop` to be a real page, but maybe the content should be slightly different to the current dashboard, so for now a redirect seems OK.

webapp/routes.go
```diff
@@ -37,8 +37,8 @@ func RegisterRoutes() {
 	shared.AddRoute("/runs", "test-runs", testRunsHandler)
 	shared.AddRoute("/test-runs", "test-runs", testRunsHandler) // Legacy name
 
-	// Dashboard for the interop effort, by year.
-	shared.AddRoute("/{name:(?:compat|interop-)}{year:[0-9]+}", "interop-dashboard", interopHandler)
+	// Dashboard for the interop effort, by year. Defaults to current year.
+	shared.AddRoute("/{name:(?:compat|interop-?)}{year:[0-9]*}", "interop-dashboard", interopHandler)
```
KyleJu, line 41: /interop2023 will be a legit URL now. Would that be an issue for the endpoint? Do we parse the "-"?

DanielRyanSmith: I was thinking this was fine, as the site will still function and redirect as intended in this case, but there are also instances of odd behavior (e.g. `/interop-`, `/compat`, and `/compat-` all redirect to the current dashboard now).

Based on jgraham's comment, I think it might be best to separate the regex out, as it's becoming a somewhat complex statement, and it's possible that `/interop` will one day be more of a landing page that explains the Interop effort as a whole, rather than just navigating to the dashboard.

KyleJu: Yea I agree. Let's separate it out. Thanks for going back and forth!

KyleJu (approved):
LGTM after separating the /interop endpoint

PR #3663, opened 2024-01-17
Add WebFeaturesData container [Part 1/4]

jcscottiii:
This commit introduces WebFeaturesData which is a struct that consumers of the web features data can use to check if a feature matches with a web feature.

This commit supports the initial v1 of the manifest file which will be generated during each WPT release

This commit excludes the GitHubDownloader to actually download from GitHub in order to make the PRs easier to review

### DIFF
--- a/shared/web_features.go
+++ b/shared/web_features.go
@@ -1,0 +1,109 @@
+// Copyright 2024 The WPT Dashboard Project. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+package shared
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+)
+
+// WebFeaturesData is the public data type that represents the data parsed from
+// a given manifest file
+type WebFeaturesData map[string]map[string]interface{}
+
+// ErrUnknownWebFeaturesManifestVersion indicates that the parser does not know how to parse
+// this version of the web features file.
+var ErrUnknownWebFeaturesManifestVersion = errors.New("unknown web features manifest version")
+
+// ErrBadWebFeaturesManifestJSON indicates that there was an error parsing the given
+// v1 manifest file.
+var ErrBadWebFeaturesManifestJSON = errors.New("invalid json when reading web features manifest")
+
+// ErrUnexpectedWebFeaturesManifestV1Format indicates that there was an error parsing the given
+// v1 manifest file.
+var ErrUnexpectedWebFeaturesManifestV1Format = errors.New("unexpected web features manifest v1 format")
+
+// TestMatchesWithWebFeature performs two checks.
+// If the given test path is present in the data. If not, return false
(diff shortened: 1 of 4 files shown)


DanielRyanSmith (approved):
Kyle might have some more opinions on this, but it otherwise LGTM

shared/web_features.go
```diff
@@ -0,0 +1,109 @@
...
+	"strings"
+)
+
+// WebFeaturesData is the public data type that represents the data parsed from
+// a given manifest file
+type WebFeaturesData map[string]map[string]interface{}
+
+// ErrUnknownWebFeaturesManifestVersion indicates that the parser does not know how to parse
+// this version of the web features file.
+var ErrUnknownWebFeaturesManifestVersion = errors.New("unknown web features manifest version")
```
KyleJu, line 22: These Err* variables can be private, as it won't be used outside of the packages?

shared/web_features.go
```diff
@@ -0,0 +1,109 @@
...
+// TestMatchesWithWebFeature performs two checks.
+// If the given test path is present in the data. If not, return false
+// If it is present, check if the given web feature applies to that test.
+func (d WebFeaturesData) TestMatchesWithWebFeature(test, webFeature string) bool {
+	if len(d) == 0 {
+		return false
+	}
+	if webFeatures, ok := d[test]; ok {
+		_, found := webFeatures[strings.ToLower(webFeature)]
+
```
KyleJu, line 41:
```suggestion
```

shared/web_features_manifest_util.go
```diff
@@ -0,0 +1,43 @@
...
+	"context"
+	"io"
+)
+
+// GetWPTWebFeaturesManifest is the entrypoint for handling web features manifest downloads.
+// It includes two generic steps: Downloading and Parsing.
+func GetWPTWebFeaturesManifest(ctx context.Context, downloader WebFeaturesManifestDownloader, parser WebFeatureManifestParser) (WebFeaturesData, error) {
+	manifest, err := downloader.Download(ctx)
+	if err != nil {
+		return nil, nil
```
KyleJu, line 17:
```suggestion
		return nil, err
```

KyleJu (approved):
LGTM with minor comments

shared/web_features.go
jcscottiii: This is true. We can make them private.

shared/web_features.go
jcscottiii: I used the golangci-lint tool (currently it only applies to the api folder now, eventually it should apply to other folders. I decided to use it on these files to help with the migration to it whenever it happens). It comes up with a [nlreturn](https://github.com/ssgreg/nlreturn) lint error. When the code block is above a certain size, it requires a new line before the return.

shared/web_features_manifest_util.go
jcscottiii: Good catch! I added a test to catch this too.

PR #3664, opened 2024-01-17
Add the GitHub Web Features Manifest Downloader [Part 2/4]

jcscottiii:
This contains the actual implementation to download the manifest from the latest GitHub release of WPT

Other changes:
- Also adds a new dependency to help with the mocking of the GitHub library


shared/web_features_manifest_github_download.go
```diff
@@ -0,0 +1,134 @@
...
+
+// ErrNoWebFeaturesManifestFileFound when a given GitHub release does not contain a manifest file to download.
+var ErrNoWebFeaturesManifestFileFound = errors.New("web features manifest not found in release")
+
+// ErrMissingBodyDuringWebFeaturesManifestDownload when a http call to download the file contains no body.
+// This should not happen often.
+var ErrMissingBodyDuringWebFeaturesManifestDownload = errors.New("empty body when downloading web features manifest")
+
+// ErrUnableToRetrieveGitHubRelease indicates the request to retrieve the latest release failed
+var ErrUnableToRetrieveGitHubRelease = errors.New("failed to retrieve latest github release")
```
KyleJu, line 32:
```suggestion
var ErrUnableToRetrieveGitHubRelease = errors.New("failed to retrieve latest GitHub release")
```

shared/web_features_manifest_github_download.go
```diff
@@ -0,0 +1,134 @@
...
+
+// ErrGitHubAssetDownloadFailedToComplete indicates the download request failed for a given release asset
+var ErrGitHubAssetDownloadFailedToComplete = errors.New("request to download github asset failed")
```
KyleJu, line 35:
```suggestion
var ErrGitHubAssetDownloadFailedToComplete = errors.New("request to download GitHub asset failed")
```

shared/web_features_manifest_github_download.go
```diff
@@ -0,0 +1,134 @@
...
+func (d GitHubWebFeaturesManifestDownloader) Download(ctx context.Context) (io.ReadCloser, error) {
+	release, _, err := d.gitHubClient.Repositories.GetLatestRelease(
+		ctx,
+		"jcscottiii", // REPLACE WITH SourceOwner BEFORE MERGING,
+		WebFeatureManifestRepo)
+	if err != nil {
+		return nil, errors.Join(ErrUnableToRetrieveGitHubRelease, err)
+	}
+
+	// Find the asset URL for the manifest file
```
KyleJu, line 79:
```suggestion
	// Find the asset URL for the manifest file.
```

shared/web_features_manifest_github_download.go
```diff
@@ -0,0 +1,134 @@
...
+			assetURL = *asset.BrowserDownloadURL
+
+			break
+		}
+	}
+	if assetURL == "" {
+		return nil, ErrNoWebFeaturesManifestFileFound
+	}
+
+	// Download the file
```
KyleJu, line 92: Missing . at the end of the comment. Ditto below

KyleJu (approved):
LGMT with nits

shared/web_features_manifest_github_download_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"encoding/json"
+	"errors"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-github/v47/github"
+	"github.com/migueleliasweb/go-github-mock/src/mock"
```
KyleJu, line 22: Is this a widely used library that has ongoing contributions in the future? My concern is that e.g. if GitHub API upgrades to v48 with breaking changes, the mock might break

KyleJu (changes requested):
Remove LGTM before a comment about the mock library

shared/web_features_manifest_github_download_test.go
KyleJu: Alternatively, we can generate our [own mock(https://github.com/web-platform-tests/wpt.fyi/blob/main/shared/sharedtest/metadata_util_mock.go) through declaring an interface in MockGen

jcscottiii: That is always a valid concern. I picked this library because it was recommended by the GitHub API client we use.

https://github.com/google/go-github?tab=readme-ov-file#testing-code-that-uses-go-github

It allows you see the URL path to check exactly what the github client sends off

But we could create a small interface for the GetLatestRelease method. I can do that.

KyleJu (approved):
Thanks for addressing the mock comment!

PR #3665, opened 2024-01-17
Add support to search for web features in searchcache [Part 3/4]

jcscottiii:
The service will now interpret a query that takes the form of feature:<web-feature-key>

Similarly to the searchcacheMetadataFetcher, there is now a searchcacheWebFeaturesManifestFetcher

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -372,7 +372,7 @@ feature label, regardless of browsers.
 
 Where web-feature-name is a string, case-insensitive and matches the filename base
 for any of the .yml files in the
-[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory..
+[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory.
 
  E.g.
 
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -362,3 +362,20 @@ Search triaged tests with a label interop-2022:
 
 See [Meta qualities](#meta-qualities) above for more information on other
 meta qualities than `"different"`.
+
+#### feature
+
+`feature` query atoms perform a search for tests that have a matching
+feature label, regardless of browsers.
+
+    {"feature": [web-feature-name]}
+
+Where web-feature-name is a string, case-insensitive and matches the filename base
+for any of the .yml files in the
+[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory..
+
+ E.g.
+
+Search the [nesting](https://github.com/web-platform-dx/web-features/blob/main/feature-group-definitions/nesting.yml) feature:
+
+    feature:nesting
--- a/api/query/atoms.go
+++ b/api/query/atoms.go
@@ -365,6 +365,26 @@ func (t AbstractTestLabel) BindToRuns(_ ...shared.TestRun) ConcreteQuery {
 	}
 }
 
+// AbstractTestWebFeature represents the root of a web_feature query, which matches test-level
+// metadata to a searched web feature.
+type AbstractTestWebFeature struct {
+	WebFeature      string
+	manifestFetcher searchcacheWebFeaturesManifestFetcher
+}
+
+// BindToRuns for AbstractTestWebFeature fetches test-level metadata; it is independent of test runs.
+// nolint:ireturn // TODO: Fix ireturn lint error
+func (t AbstractTestWebFeature) BindToRuns(_ ...shared.TestRun) ConcreteQuery {
+	t.manifestFetcher = searchcacheWebFeaturesManifestFetcher{}
+
+	data, _ := t.manifestFetcher.Fetch()
+
+	return TestWebFeature{
+		WebFeature:      t.WebFeature,
+		WebFeaturesData: data,
+	}
+}
+
 // MetadataQuality represents the root of an "is" query, which asserts known
 // metadata qualities to the results.
 type MetadataQuality int
@@ -1003,6 +1023,27 @@ func (t *AbstractTestLabel) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// UnmarshalJSON for AbstractTestWebFeature attempts to interpret a query atom as
+// {"feature":<web_feature_string>}.
+func (t *AbstractTestWebFeature) UnmarshalJSON(b []byte) error {
+	var data map[string]*json.RawMessage
+	if err := json.Unmarshal(b, &data); err != nil {
+		return err
+	}
+	webFeatureMsg, ok := data["feature"]
+	if !ok {
+		return errors.New(`Missing web feature pattern property: "feature"`)
+	}
+	var webFeature string
+	if err := json.Unmarshal(*webFeatureMsg, &webFeature); err != nil {
+		return errors.New(`Property "feature" is not a string`)
+	}
+
+	t.WebFeature = webFeature
+
+	return nil
+}
+
 // UnmarshalJSON for AbstractTriaged attempts to interpret a query atom as
 // {"triaged":<browser name>}.
 func (t *AbstractTriaged) UnmarshalJSON(b []byte) error {
@@ -1182,6 +1223,12 @@ func unmarshalQ(b []byte) (AbstractQuery, error) {
 	}
 	{
 		var t AbstractTestLabel
+		if err := json.Unmarshal(b, &t); err == nil {
+			return t, nil
+		}
+	}
+	{
+		var t AbstractTestWebFeature
 		if err := json.Unmarshal(b, &t); err == nil {
 			return t, nil
 		}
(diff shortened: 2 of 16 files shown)


api/query/README.md
```diff
@@ -362,3 +362,20 @@ Search triaged tests with a label interop-2022:
...
+#### feature
+
+`feature` query atoms perform a search for tests that have a matching
+feature label, regardless of browsers.
+
+    {"feature": [web-feature-name]}
+
+Where web-feature-name is a string, case-insensitive and matches the filename base
+for any of the .yml files in the
+[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory..
```
KyleJu, line 375:
```suggestion
[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory.
```

api/query/atoms.go
```diff
@@ -365,6 +365,26 @@ func (t AbstractTestLabel) BindToRuns(_ ...shared.TestRun) ConcreteQuery {
...
+// AbstractTestWebFeature represents the root of a web_feature query, which matches test-level
+// metadata to a searched web feature.
+type AbstractTestWebFeature struct {
+	WebFeature      string
+	manifestFetcher searchcacheWebFeaturesManifestFetcher
+}
+
+// BindToRuns for AbstractTestWebFeature fetches test-level metadata; it is independent of test runs.
+// nolint:ireturn // TODO: Fix ireturn lint error
+func (t AbstractTestWebFeature) BindToRuns(_ ...shared.TestRun) ConcreteQuery {
```
KyleJu, line 377: This function needs a test in api/query/atoms_test.go, a similar setup to [TestStructuredQuery_bindTestLabel](https://github.com/web-platform-tests/wpt.fyi/blob/701c625dc22f6dedea3964de971c4c9d64191504/api/query/atoms_test.go#L1135).

KyleJu: In the test setup, searchcacheWebFeaturesManifestFetcher{} will be passed in from `t.manifestFetcher`, instead of being created directly here. To make the test work, a mock file need to be generated for searchcacheWebFeaturesManifestFetcher through MockGen. 
```
if t.metadataFetcher == nil {
		t.metadataFetcher = searchcacheMetadataFetcher{}
	}
```

api/query/atoms.go
```diff
@@ -1003,6 +1023,27 @@ func (t *AbstractTestLabel) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// UnmarshalJSON for AbstractTestWebFeature attempts to interpret a query atom as
+// {"feature":<web_feature_string>}.
+func (t *AbstractTestWebFeature) UnmarshalJSON(b []byte) error {
```
KyleJu, line 1028: Missing tests for this function

api/query/cache/index/filter.go
```diff
@@ -342,6 +349,20 @@ func (tl TestLabel) Filter(t TestID) bool {
...
+func (twf TestWebFeature) Filter(t TestID) bool {
+	name, _, err := twf.tests.GetName(t)
+	if err != nil {
+		return false
+	}
+	// Check if there's any data
+	if twf.webFeaturesData == nil {
+		return false
+	}
+	// Get the Web Features for that exact test path
```
KyleJu, line 362:
```suggestion
	// Get the Web Features for that exact test path.
```

api/query/cache/index/filter.go
```diff
@@ -342,6 +349,20 @@ func (tl TestLabel) Filter(t TestID) bool {
 	return false
 }
 
+// Filter interprets a TestWebFeature as a filter function over TestIDs.
+func (twf TestWebFeature) Filter(t TestID) bool {
+	name, _, err := twf.tests.GetName(t)
+	if err != nil {
+		return false
+	}
+	// Check if there's any data
```
KyleJu, line 358:
```suggestion
	// Check if there's any data.
```

KyleJu (commented):
Note to myself: Need to finish reviews for polling and cache logic

api/query/atoms.go
jcscottiii: I missed the atoms_test.go file completely. Will do this.

api/query/atoms_test.go
```diff
@@ -919,11 +974,11 @@ func TestStructuredQuery_bindExistsWithTwoProducts(t *testing.T) {
 			})
 		if i%2 == 0 { // Evens are edge
 			or1.Args = append(or1.Args,
-						RunTestStatusEq{
-							Run:    int64(i),
-							Status: 1,
-						},
-				)
+				RunTestStatusEq{
```
jcscottiii, line 977: Note for reviewers: this was done automatically as part of go fmt. Nothing was changed here

api/query/atoms.go
jcscottiii: In order to prevent the `if variable == nil{ then initialize}` logic (which would lead to an non-testable branch), we can instead embed AbstractTestWebFeature with an intermediate struct during the initialization of AbstractTestWebFeature. And during the initialization, we can also have a default implementation of the interface set.


You can see that happening[ here](https://github.com/web-platform-tests/wpt.fyi/pull/3665/files#diff-6b9fa200c6337e946daaabc6bc1eaba8123f839e7606201c94f28c5b61d903a8R1239-R1245). The intermediate type is called TestWebFeatureAtom. So now the UnmarshalJSON is handled by that. And if successful, it gets embedded into the AbstractTestWebFeature which is used like the other abstract atoms.

You can see in the atoms test file that now, we can override the interface with our mock still without `if variable == nil{ then initialize}`

Also, I made AbstractTestWebFeature and TestWebFeature (and as a result also TestWebFeatureAtom) all exported to be similar to what exists. Although, these and the existing ones don't seem to be needed outside of their current package.

I did not make my new interface `webFeaturesManifestFetcher` exported nor in the shared package since right now, it is only planned for the searchcache.

jcscottiii:
This is ready for another review @KyleJu

KyleJu:
Chatted offline. The downloading Manifest bit can be moved to the shared/ directory

jcscottiii:
Note to reviewers:

Moving all the Github logic to shared allowed me to:
- unexport a lot of the things
- add a new test file for the existing poll.go file.

jcscottiii:
@KyleJu This is ready for another round

shared/web_features_manifest_github_download.go
```diff
@@ -141,3 +142,60 @@ func (s *gitHubDownloadStream) Close() error {
...
 }
+
+// GitHubWebFeaturesClient is the entrypoint to retrieving web features from GitHub.
+type GitHubWebFeaturesClient struct {
+	downloader webFeaturesManifestDownloader
+	parser     webFeatureManifestParser
+}
+
+// gitHubWebFeaturesClientOptions contains all the non-required options that
+// can be used to configure an instance of GitHubWebFeaturesClient
```
KyleJu, line 153:
```suggestion
// can be used to configure an instance of GitHubWebFeaturesClient.
```

api/query/web_features_manifest_cache.go
```diff
@@ -0,0 +1,59 @@
...
+// Currently, the separate goroutine in the poll folder will use this.
+func SetWebFeaturesDataCache(newData shared.WebFeaturesData) {
+	webFeaturesDataCacheLock.Lock()
+	defer webFeaturesDataCacheLock.Unlock()
+	webFeaturesDataCache = newData
+}
+
+// GetWebFeaturesDataCache safely retrieves the data cache.
+func GetWebFeaturesDataCache() shared.WebFeaturesData {
+	webFeaturesDataCacheLock.RLock()
```
KyleJu, line 24: I don't see a reason using read lock here, but I guess it is a standard implementation from what I read.

KyleJu (approved):
LGTM. Thanks for addressing all the comments!

api/query/web_features_manifest_cache.go
jcscottiii: The polling services (`StartMetadataPollingService` and `StartWebFeaturesManifestPollingService`) both write to their respective in-memory cache (`query.MetadataMapCached` and `query.webFeaturesDataCache`). And requests to searchcache potentially read from that. There is a potential data race that can occur with the existing metadata cache. By using the lock around the web features cache, it prevents that.

We need to add a similar lock to the MetadataMapCached.

I added some race tests to help exercise the data race to this example PR https://github.com/web-platform-tests/wpt.fyi/pull/3674

jcscottiii:
@KyleJu I added one more [commit]([94a8b46](https://github.com/web-platform-tests/wpt.fyi/pull/3665/commits/94a8b46a22e8c383409b4b841b6626674354bb26)). I think it should be fine but don't want to assume.

api/query/web_features_manifest_cache.go
KyleJu: Aha thanks for the example. I think I didn't exactly distinguish the difference between data race and race condition, where data race pertains to the memory access conflict. Appreciate it!

KyleJu (approved):
LGTM again

PR #3667, opened 2024-01-19
Display Chrome and Edge as separate entities in Interop 2024

DanielRyanSmith:
This change adds the functionality to have a variable number of browsers display on the interop dashboard, and separates Edge from Chrome into its own entity and score (for Interop 2024 only).

This change also displays Chrome Canary for the Interop 2024 dashboard.

Screenshot 2024-01-22 at 10 31 09 AM
Screenshot 2024-01-22 at 10 31 18 AM

Screenshot 2024-01-22 at 10 33 57 AM


foolip:
@DanielRyanSmith can you make all circles the same size, so that the top two circles are aligned about the middle two below?

foolip:
Thank you @DanielRyanSmith! This looks good to me, and I'll run it past the interop team.

webapp/components/interop-dashboard.js
```diff
@@ -336,54 +338,40 @@ class InteropDashboard extends PolymerElement {
                   <!-- First score table header with sort functionality -->
                   <template is="dom-if" if="[[isFirstTable(itemsIndex)]]">
                     <tr class="section-header">
-                      <th class="sortable-header" on-click="sortByName">
+                      <th class="sortable-header">
                         {{section.name}}
-                        <img class="sort-icon-focus-areas" src="[[getSortIcon(0, sortColumn, isSortedAsc)]]" />
+                        <img class="sort-icon-focus-areas" src="[[getSortIcon(-2, sortColumn, isSortedAsc)]]" />
```
KyleJu, line 343: It is fine to use a negative index for Focus area column, until we have more types of columns. What if we create a separate function that calls getSortIcon(0,  sortColumn, isSortedAsc) without the negative number hacks?

webapp/components/interop-dashboard.js
```diff
@@ -864,7 +855,28 @@ class InteropDashboard extends PolymerElement {
     return columnNumber === sortColumn;
   }
 
+  adjustSortIconIndex(index) {
```
KyleJu, line 858: Per comment above, instead of using a negative number, create a separate function for e.g. getFocusAreaColumnIcon().

webapp/components/interop-summary.js
```diff
@@ -235,6 +233,42 @@ class InteropSummary extends PolymerElement {
...
+  }
+
+  getBrowserIconName(browserInfo, isStable) {
+    if (isStable) {
+      return browserInfo.tableName;
+    }
+    return `${browserInfo.tableName} ${browserInfo.experimentalName}`;
+  }
+
+  // Returns the browser full name as a list of strings so we can
```
KyleJu, line 266:
```suggestion
  // Returns the browser full names as a list of strings so we can
```

KyleJu (approved):
LGTM with minor comments

webapp/components/interop-dashboard.js
DanielRyanSmith: Good idea - I've implemented this!

PR #3668, opened 2024-01-20
Update Interop color logic for summary scores

DanielRyanSmith:
Updates the Interop dashboard to use a color based on a gradient when displaying the summary scores.

---

Example with 2024 Stable dashboard
Screenshot 2024-02-05 at 1 53 12 PM
Screenshot 2024-02-05 at 1 53 18 PM

---
Interop 2023 dashboard
Screenshot 2024-02-05 at 1 54 50 PM
Screenshot 2024-02-05 at 1 54 55 PM

### DIFF
--- a/webapp/components/interop-summary.js
+++ b/webapp/components/interop-summary.js
@@ -287,29 +287,13 @@ class InteropSummary extends PolymerElement {
   calculateColor(score) {
     const gradient = [
       // Red.
-      {
-        scale: 0,
-        backgroundColor: [238, 43, 43],
-        textColor: [238, 43, 43],
-      },
+      { scale: 0, color: [250, 0, 0] },
       // Orange.
-      {
-        scale: 33.33,
-        backgroundColor: [250, 125, 0],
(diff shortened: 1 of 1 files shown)


foolip (approved):
Looks like this would work.

If you want to have some fun, since `color-mix()` was part of Interop 2023 and now supported everywhere, can you see what this would be like if we generate something like `color-mix(in srgb, #568f24 50%, #388E3C)` when the score is between 75 and 95? Totally optional :)

foolip:
I was too quick to approve. There was also an issue of contrast between text and background. Perhaps `color-mix()` could also be used to darken the colors instead of using alpha to achieve the effect?

zcorpan:
Mixing with 30% black seems to do the trick.

```
color-mix(in lch, ${c[0]} 70%, black);
```

https://software.hixie.ch/utilities/js/live-dom-viewer/saved/12338

DanielRyanSmith:
Thanks for the help everyone! I updated this PR with the suggested changes. 🙂

PR #3670, opened 2024-01-22
[Interop 2024] Change score rounding

KyleJu:
Fix 3640. Define rounding rules for the top-level scores.

- Round down when a score is above the 97.5% breakpoint for all years
- Before 2024, use existing `Math.floor`
- Starting 2024, use `Math.round`.


DanielRyanSmith (approved):

webapp/components/interop-summary.js
```diff
@@ -197,9 +197,24 @@ class InteropSummary extends PolymerElement {
     return scores !== null && scores !== undefined;
   }
 
+  // roundScore defines the rounding rules for the top-level scores.
+  roundScore(score) {
+    // Round down when a score is above the 975 breakpoint.
```
foolip, line 202: The option we settled on in the interop team is to special-case 100 only. So we would round the score, and if it's 100 but `score` is not 1000, we change it to 99.

KyleJu: Resolved!

foolip: Thanks Kyle!

PR #3710, opened 2024-02-09
Update the links for 2024 investigation areas

stof:
Closes https://github.com/web-platform-tests/interop/issues/630

This adds the missing URL for the accessibility area as the repo has now been created.
I also updated links for the other 2 investigation areas as the repositories have been renamed (the old name redirects to the new one, so the links were not broken)

### DIFF
--- a/webapp/components/interop-data.js
+++ b/webapp/components/interop-data.js
@@ -651,17 +651,17 @@ export const interopData = {
     'investigation_scores': [
       {
         'name': 'Accessibility Testing',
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (commented):
If you make the same changes to `webapp/components/interop-data.js`, That should fix the test issues. 🙂

DanielRyanSmith (approved):

stof:
I updated the other file. Sorry for forgetting the need to update 2 files instead of one (I knew about that, I even already commented on some PRs in the past to explain how to solve this issue by updating both files).

PR #3726, opened 2024-02-22
Use one-way binding for wpt-bsf

past:
This is another step in the migration from Polymer to Lit.

### DIFF
--- a/webapp/components/wpt-bsf.js
+++ b/webapp/components/wpt-bsf.js
@@ -151,14 +151,14 @@ class WPTBSF extends LoadingState(PolymerElement) {
     };
     this.enterChart = () => {
       this.isInteracting = true;
-      const event = new CustomEvent('is-interacting-changed', {
+      const event = new CustomEvent('interactingchanged', {
         detail: { value: this.isInteracting }
       });
       this.dispatchEvent(event);
     };
     this.exitChart = () => {
       this.isInteracting = false;
-      const event = new CustomEvent('is-interacting-changed', {
+      const event = new CustomEvent('interactingchanged', {
         detail: { value: this.isInteracting }
       });
       this.dispatchEvent(event);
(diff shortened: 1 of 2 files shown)


past:
Verified through devtools debugging that `WPTApp.prototype.{enterBSF,exitBSF}` and `WPTBSF.prototype.{enterChart,exitChart}` are called in the same order and with the same arguments, and that `WPTApp.bsfIsInteractingChanged` is called with the appropriate value.

webapp/components/wpt-bsf.js
```diff
@@ -152,9 +151,17 @@ class WPTBSF extends LoadingState(PolymerElement) {
     };
     this.enterChart = () => {
       this.isInteracting = true;
+      const event = new CustomEvent('is-interacting-changed', {
```
KyleJu, line 154: we don't use "-" for the event name.  How about showbsf?
```suggestion
      const event = new CustomEvent('showbsf, {
```

past: I used the same event name that Polymer fires when the isInteractive property changes, but I can use one without dashes (interactingchanged) if that is our convention. Note that this event has nothing to do with the `showBSF` parameter or whether the BSF chart is shown more broadly. As the relevant code in `wpt-app.js` indicates, it is guarding analytics event submission when the user is still interacting with the chart in order to avoid a known issue.

webapp/components/wpt-bsf.js
```diff
@@ -152,9 +151,17 @@ class WPTBSF extends LoadingState(PolymerElement) {
...
+        detail: { value: this.isInteracting }
+      });
+      this.dispatchEvent(event);
     };
     this.exitChart = () => {
       this.isInteracting = false;
+      const event = new CustomEvent('is-interacting-changed', {
```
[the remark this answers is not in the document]
past: Used `interactingchanged` as above.

webapp/components/wpt-bsf.js
KyleJu: Aha. Thanks for the explanation.

KyleJu (approved):

PR #3736, opened 2024-02-29
Upgrade node.js to version 18

past:
Node.js 16 [reached the end of life](https://github.com/nodejs/Release) last November.


jcscottiii (approved):
Checked:
- URL for script is valid
- Script does download
- And matches the format found on https://github.com/nodesource/distributions


LGTM! Thanks Panos!

PR #3753, opened 2024-03-19
Add support for view=test query parameter

jcscottiii:
Backend changes:
- Add "test" to the allowlist values for the view url parameter
- Add TestParseViewParam to assert behavior for existing view params and the new "test" parameter

Frontend changes:
- Add new Test View button
- Add test_view_ prefix. This prefix is used to keep track of the new view=test mode.
- Add new helper method isTestView() to support this new view
- Add helper method isInteropView to retroactively support the already supported interop view.
- Create getNodeTotalProp and getNodePassProp to encapsulate the logic of which property from the aggregation node to use.

About view=test mode
- Test pass is only counted if the number of subtest passes == the total subtests
- This should **not** display percentages. Only the counts from the first bullet point when at  directory level.
- When on the individual test level, it should only show `PASS` when subtest passes == the total subtests
  - Other cases:
  - When the status == OKAY and subtest passes != the total subtests, still show FAIL
  - Show the status for other cases that come from the existing STATUS_ABBREVIATIONS mapping
  - Fallback to showing FAIL for anything else

For reviewers:
- Double check to make sure I accounted for the legacy v1 summary case? Or am I overthinking and the v1 case does not matter?

High level view:
image

Test View:
image

Mixed View:
image

Feature Flag (default is off and the feature does not work then):
image

### DIFF
--- a/webapp/components/wpt-flags.js
+++ b/webapp/components/wpt-flags.js
@@ -47,6 +47,7 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'queryBuilder',
       'queryBuilderSHA',
       'showBSF',
+      'showViewEqTest',
       'structuredQueries',
       'triageMetadataUI',
       'webPlatformTestsLive',
@@ -256,6 +257,11 @@ class WPTFlagsEditor extends FlagsEditorClass(/*environmentFlags*/ false) {
     <paper-item>
       <paper-checkbox checked="{{showBSF}}">
         Enable Browser Specific Failures graph
+      </paper-checkbox>
+    </paper-item>
+    <paper-item>
+      <paper-checkbox checked="{{showViewEqTest}}">
+        Enable view=test query parameter
       </paper-checkbox>
     </paper-item>
 `;
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -1079,10 +1079,7 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
 
   isTestView() {
     // If the `showViewEqTest` feature flag is not active, return false immediately.
-    if (!this.showViewEqTest) {
-      return false;
-    }
-    return this.view === 'test';
+    return this.showViewEqTest && this.view === 'test';
   }
 
   getTotalsClass(totalInfo) {
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -55,6 +55,13 @@ const STATUS_ABBREVIATIONS = {
 };
 const PASSING_STATUSES = ['O', 'P'];
 
+// ViewEnum contains the different values for the `view` query parameter.
+const ViewEnum = {
+  Subtest: 'subtest',
+  Interop: 'interop',
+  Test: 'test'
+}
+
 class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathInfo(LoadingState(TestRunsUIBase)))))) {
   static get template() {
     return html`
@@ -1004,22 +1011,22 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
   }
 
   testViewButtonClass(view) {
-    return (view === 'test') ? 'selected' : 'unselected';
+    return (view === ViewEnum.Test) ? 'selected' : 'unselected';
   }
 
   interopButtonClass(view) {
-    return (view === 'interop') ? 'selected' : 'unselected';
+    return (view === ViewEnum.Interop) ? 'selected' : 'unselected';
   }
 
   defaultButtonClass(view) {
-    return (view !== 'interop' && view !== 'test') ? 'selected' : 'unselected';
+    return (view !== ViewEnum.Interop && view !== ViewEnum.Test) ? 'selected' : 'unselected';
   }
 
   clickInterop() {
     if (this.isInteropView()) {
       return;
     }
-    this.view = 'interop';
+    this.view = ViewEnum.Interop;
   }
 
   clickTestView() {
@@ -1031,14 +1038,14 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     if (this.isTestView()) {
       return;
     }
-    this.view = 'test';
+    this.view = ViewEnum.Test;
   }
 
   clickDefault() {
     if (this.isDefaultView()) {
       return;
     }
-    this.view = 'subtest';
+    this.view = ViewEnum.Subtest;
   }
 
   changeView(view) {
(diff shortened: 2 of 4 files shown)


webapp/views/wpt-results.js
```diff
@@ -1039,14 +1070,29 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+  isInteropView() {
+    return this.view === 'interop';
+  }
+
+  isTestView() {
+    // If the `showViewEqTest` feature flag is not active, return false immediately.
+    if (!this.showViewEqTest) {
+      return false;
+    }
+    return this.view === 'test';
```
DanielRyanSmith, line 1085: Optional logic nit
```suggestion
    return this.showViewEqTest && this.view === 'test';
```

webapp/views/wpt-results.js
```diff
@@ -255,6 +255,9 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     <template is="dom-if" if="[[shouldDisplayToggle(canViewInteropScores, pathIsATestFile)]]">
```
DanielRyanSmith, line 255: This will only display the button for test view in places where interop scores can be viewed. Is that intentional?

webapp/views/wpt-results.js
```diff
@@ -1164,18 +1210,66 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     return `${this.getTestNumbersDisplay(passes, total, isDir)} (${cellDisplay}%)`;
   }
 
+  // Formats the numbers shown on the results page for the test view.
+  formatCellDisplayTestView(passes, total, status, isDir) {
+
+    // At the test level:
+    // 1. Show PASS is passes == total for subtests.
+    // 2. Show FAIL if status is undefined (legacy summaries) or 'O' (because showing OKAY would be misleading).
+    // 3. Show FAIL otherwise.
```
DanielRyanSmith, line 1219: This looks like the correct logic here to me. As long as condition #2 will override condition #1 if not met. 🙂

DanielRyanSmith (commented):
Looking pretty good here. 🙂 I have some clarifying questions.

webapp/views/wpt-results.js
DanielRyanSmith: Looking again, This might need to be readjusted so that the status is also taken into consideration at the test level. As in, if passes === total, the test status should also be a passing status also to be considered a true pass.

webapp/views/wpt-results.js
jcscottiii: For now yes. If this is something that is accepted by default, I can add an else template case that only shows the default and test buttons

webapp/views/wpt-results.js
jcscottiii: If the status does not equal Pass, what should happen?

Let's take https://storage.googleapis.com/wptd/820b6449e01db26e2074e318388f480c7d6d957c/chrome-123.0.6312.86-linux-20.04-815b03e53e-summary_v2.json.gz for example

Are you saying that I should make sure `"s"` equals `"O"` or `"P"`? And then for the "else" case, don't increment testPass? Would the "else" case ever happen?

jcscottiii: If we check both, in the event only status is passing or subtests count === subtests passing, what should I display?

webapp/components/wpt-flags.js
```diff
@@ -258,6 +259,11 @@ class WPTFlagsEditor extends FlagsEditorClass(/*environmentFlags*/ false) {
         Enable Browser Specific Failures graph
       </paper-checkbox>
     </paper-item>
+    <paper-item>
```
KyleJu, line 262: Is it necessary to hide it behind the flag? view=test is an optional parameter and doesn't show in the default view

webapp/views/wpt-results.js
DanielRyanSmith: After syncing about this separately, the pass/fail of the status is already calculated into the overall subtest pass/fail, so this logic is handling things as expected. 🙂

webapp/views/wpt-results.js
```diff
@@ -255,6 +255,9 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
     <template is="dom-if" if="[[shouldDisplayToggle(canViewInteropScores, pathIsATestFile)]]">
       <div class="channel-area">
         <paper-button id="toggleInterop" class\$="[[ interopButtonClass(view) ]]" on-click="clickInterop">Interop View</paper-button>
+        <template is="dom-if" if="[[showViewEqTest]]">
```
KyleJu, line 258: If not using a flag, showViewEqTest can instead check the presence of view=test parameter

webapp/views/wpt-results.js
```diff
@@ -1164,18 +1207,66 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
+      } else if ((status === undefined) || (status === 'O')) {
+        return "FAIL";
+      } else if (status in STATUS_ABBREVIATIONS) {
+        return STATUS_ABBREVIATIONS[status];
+      } else {
+        return "FAIL";
+      }
+    }
+
+    const formatPercent = parseFloat((passes / total * 100).toFixed(0));
```
KyleJu, line 1229: lt is hard to compare float without loss of precision, in general you should instead check abs(formatPercent - 100.0) < EPSILON, where EPSILON is the precision loss due to the float representation, or use string equality directly, without `parseFloat` from the result of toFixed(0).

webapp/views/wpt-results.js
```diff
@@ -988,21 +1003,37 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
 
   clickInterop() {
-    if (!this.isDefaultView()) {
+    if (this.isInteropView()) {
       return;
     }
     this.view = 'interop';
   }
 
+  clickTestView() {
```
KyleJu, line 1025: nit: consider declaring a string constant for `interop`, `test` on the top

KyleJu (approved):
LGTM with minor comments

webapp/views/wpt-results.js
jcscottiii: Good catch. This was actually unused. Removed.

webapp/views/wpt-results.js
jcscottiii: Done!

webapp/views/wpt-results.js
```diff
@@ -55,6 +55,13 @@ const STATUS_ABBREVIATIONS = {
 };
 const PASSING_STATUSES = ['O', 'P'];
 
+// ViewEnum contains the different values for the `view` query parameter.
+const ViewEnum = {
```
KyleJu, line 59: nit: VIEW_ENUM

PR #3807, opened 2024-04-29
Disable local debug mode by default

past:
This should fix the related code scanning [alert](https://github.com/web-platform-tests/wpt.fyi/security/code-scanning/3).

I bet nobody is using the debugger at the moment, but at least leaving the parameter in the source will remind anyone who wants to do so in the future.

### DIFF
--- a/results-processor/README.md
+++ b/results-processor/README.md
@@ -41,3 +41,10 @@ update dependencies, run the following commands:
 pip3.9 install --user pip-tools
 python3.9 -m piptools compile requirements.in
 ```
+
+## Local debugging
+
+Debugging is disabled both in production and when running locally by default.
+To enable debugging when running locally pass `debug=True` to the `app.run()`
+call in the last line of
+[`main.py`](https://github.com/web-platform-tests/wpt.fyi/blob/main/results-processor/main.py).
--- a/results-processor/main.py
+++ b/results-processor/main.py
@@ -122,4 +122,4 @@ def task_handler():
 # Run the script directly locally to start Flask dev server.
 if __name__ == '__main__':
     logging.basicConfig(level=logging.DEBUG)
-    app.run(debug=True)
+    app.run(debug=False)


jcscottiii:
The flask [docs](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=run#flask.Flask.run) mention that indeed we should never use app.run() in production.

However, that flow is never executed when running in app engine. The [Dockerfile](https://github.com/web-platform-tests/wpt.fyi/blob/e644825770347a28448dacbaf456583be172ac28/results-processor/Dockerfile#L39) runs explicitly on the [app](https://github.com/web-platform-tests/wpt.fyi/blob/e644825770347a28448dacbaf456583be172ac28/results-processor/main.py#L32) object in the main.py file. As a result, it never reaches the __name__ == __main__ conditional. 

A few options we have:
1. Keep the change you have. But only if no one is really using it. So I would ask in our meeting today if anyone uses debug mode first.
2. Leverage the [FLASK_DEBUG](https://flask.palletsprojects.com/en/1.1.x/api/?highlight=run#flask.Flask.debug) environment variable and do something like this:
   - ```
    logging.basicConfig(level=logging.DEBUG)
    app.run()  # debug should be handled automatically when running "FLASK_DEBUG=true python3 main.py"
     ```

In either option, I would add a section to the [README](https://github.com/web-platform-tests/wpt.fyi/blob/e644825770347a28448dacbaf456583be172ac28/results-processor/README.md) on how to enable debugging mode since we would be turning it off by default.

I would go with option 2. But could be persuaded to option 1 if no one is really using debug mode.

We could also dismiss the alert. But if we ever accidentally change the Dockerfile to run main.py directly like we do locally, we would want that alert then. And I'm not sure if GitHub would resurface that alert.

past:
I'm happy to make the doc changes, but judging by the dearth of code changes to results-processor in the last several years I sincerely doubt that anyone has used the debugger apart from the original authors. Let's discuss at the meeting.

past:
I went with your option 1, since nobody seems to care much about debugging this.

PR #3837, opened 2024-05-21
Temporary workaround for Linter err

KyleJu:
https://github.com/golangci/golangci-lint/issues/4733. Fix linter failures temporarily


KyleJu:
I changed the line from nolint:ireturn to nolint:all. I suspect that the definition of ireturn has changed in [golangci-lint](https://github.com/golangci/golangci-lint) 1.58.2

PR #3850, opened 2024-06-06
Expand linting to all JS files under webapp/components/

past:
There is a discrepancy in the paths `npm run lint` and `npm run lint-fix` use that appears to be accidental. I couldn't find anything in the commit history that indicates otherwise. Regardless, we should lint as much as possible, so this change expands the linting to every JS file under components.


past:
Note that this is based on top of #3848 which contributes the first two commits here (PRs aren't great for reviewing dependent changes).

jcscottiii (approved):

PR #3896, opened 2024-07-09
Upgrade actions/checkout to v4

past:
I missed one more warning in our Actions page. Turns out that checkout v4 has been released.


past:
The web_components_test failure is #3090.

DanielRyanSmith (approved):

PR #3898, opened 2024-07-11
Upgrade to a maintained gomock fork

past:
https://github.com/golang/mock is archived and recommends switching to https://github.com/uber-go/mock which is a maintained fork.

This PR was mainly generated using:
`find . -name "*.go"| xargs sed -i 's/github.com\/golang\/mock/go.uber.org\/mock/'` 
and fixing up the Makefile. In order to properly apply this change to an existing VM or locally you'd need to run these commands beforehand:
```
go clean github.com/golang/mock/mockgen
which mockgen|xargs rm
```
to avoid conflicts with the old tool.


KyleJu (approved):
Tests passed so the upgrade worked as expected

PR #3907, opened 2024-07-16
Address final feedback in RFC 190 for view=test

jcscottiii:
During the review of [RFC 190](https://github.com/web-platform-tests/rfcs/pull/190), we agreed that we should:
- Remove the button when in interop and allow this to be reachable by query parameter only.

This change does that by removing the button added in https://github.com/web-platform-tests/wpt.fyi/pull/3753

The idea is to allow users to accomplish their
workflows with view=test without confusing others. Another RFC would be needed to add a button.

Additionally, since this RFC has been accepted, we can remove the feature flag gating it.

### DIFF
--- a/webapp/components/wpt-flags.js
+++ b/webapp/components/wpt-flags.js
@@ -47,7 +47,6 @@ Object.defineProperty(wpt, 'ClientSideFeatures', {
       'queryBuilder',
       'queryBuilderSHA',
       'showBSF',
-      'showViewEqTest',
       'structuredQueries',
       'triageMetadataUI',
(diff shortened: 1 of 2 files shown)


jcscottiii:
Reviewers can try it out by going to https://jcscottiii-enable-view-dot-wptdashboard-staging.uk.r.appspot.com/ and manually append `view=test` query parameter when examining runs.

https://jcscottiii-enable-view-dot-wptdashboard-staging.uk.r.appspot.com/results/?sha=056fdef895&label=master&max-count=1&view=test

KyleJu (approved):

PR #3908, opened 2024-07-17
Add support for Ladybird

tcl3:
This PR adds support for the [Ladybird web browser](https://ladybird.org).

I've followed what was done in #3213 and #3161. We don't currently have our icon available in a browser-logos npm package, so I've put our icon directly into `webapp/static`.

Please let me know if there's anything else I need to do.


KyleJu (approved):
The change is straightforward. Since there is no easy way to test it from fork, let's test it on staging before the prod release!

PR #3909, opened 2024-07-17
Upgrade Go to version 1.21

past:
One of the last few PRs caused go to demand we run `go mod tidy`, which updated the Go version in go.mod. I went ahead and updated everything else in order to make that happen.

### DIFF
--- a/api/query/cache/service/Dockerfile
+++ b/api/query/cache/service/Dockerfile
@@ -1,6 +1,5 @@
 # Production deployment spec for query cache service.
 
-# Base golang 1.20.1 image.
 FROM golang:1.21.12-bullseye as builder
 
 RUN apt-get update
--- a/api/query/cache/service/Dockerfile
+++ b/api/query/cache/service/Dockerfile
@@ -1,7 +1,7 @@
 # Production deployment spec for query cache service.
 
 # Base golang 1.20.1 image.
-FROM golang:1.20.14-bullseye as builder
+FROM golang:1.21.12-bullseye as builder
 
 RUN apt-get update
 RUN apt-get install -qy --no-install-suggests git
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # vim: set expandtab sw=4
-FROM golang:1.20.14-bullseye
+FROM golang:1.21.12-bullseye
 
 # Create a non-priviledged user to run browsers as (Firefox and Chrome do not
 # like to run as root).
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/web-platform-tests/wpt.fyi
 
-go 1.20
+go 1.21
 
 require (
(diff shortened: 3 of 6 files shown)


api/query/cache/service/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Production deployment spec for query cache service.
 
 # Base golang 1.20.1 image.
```
jcscottiii, line 3: We can remove this outdated comment.
```suggestion
```

jcscottiii (approved):
LGTM w/ one suggestion.


For the future:
For the docker images, we may want to migrate from [bullseye](https://hub.docker.com/layers/library/golang/1.21.12-bullseye/images/sha256-56af38adb9c99f95a02879c2d7b145932cb25884ee0c2d9deed932a83d169ca5?context=explore) to [bookworm](https://hub.docker.com/layers/library/golang/1.21.12-bookworm/images/sha256-a3769bd3ea33489e49767136d224f3496d96a2bdfbcac5d640a3f1c68c78b8ff?context=explore) as it is now marked as `oldstable` according to [debian](https://www.debian.org/releases/). But it is in LTS, so we have time.

I think you may have mentioned this, but were you planning on migrating to go1.22 soon after?
I see that app engine [supports](https://cloud.google.com/appengine/docs/standard/lifecycle/support-schedule#go) it.

Go 1.23 is expected to [release](https://tip.golang.org/doc/go1.23) in August. As a result, Go 1.21 will fall out of support [then](https://go.dev/doc/devel/release#:~:text=two%20newer%20major%20releases).

past:
> LGTM w/ one suggestion.
> 
> For the future: For the docker images, we may want to migrate from [bullseye](https://hub.docker.com/layers/library/golang/1.21.12-bullseye/images/sha256-56af38adb9c99f95a02879c2d7b145932cb25884ee0c2d9deed932a83d169ca5?context=explore) to [bookworm](https://hub.docker.com/layers/library/golang/1.21.12-bookworm/images/sha256-a3769bd3ea33489e49767136d224f3496d96a2bdfbcac5d640a3f1c68c78b8ff?context=explore) as it is now marked as `oldstable` according to [debian](https://www.debian.org/releases/). But it is in LTS, so we have time.
> 
> I think you may have mentioned this, but were you planning on migrating to go1.22 soon after? I see that app engine [supports](https://cloud.google.com/appengine/docs/standard/lifecycle/support-schedule#go) it.
> 
> Go 1.23 is expected to [release](https://tip.golang.org/doc/go1.23) in August. As a result, Go 1.21 will fall out of support [then](https://go.dev/doc/devel/release#:~:text=two%20newer%20major%20releases).

That's right, I plan to update to 1.22 shortly and maybe 1.23 soon thereafter. My branch with the full dependency updates migrates to 1.22, but I'm taking it one step at a time in order to debug some breakage with deprecated packages.

PR #3910, opened 2024-07-19
Replace obsolete packr with native embed package

past:
[Packr](https://github.com/gobuffalo/packr/) is no longer maintained and they recommend switching to [embed](https://pkg.go.dev/embed) which is a native file embedding feature of Go. This set of changes transitions the code to embed.

Misc changes:
- `golangci-lint` updated to the latest version due to spurious errors in third-party libraries after the `embed` change
- One whitespace change from running `gofmt -s -w api/checks/summaries/compile.go` as demanded by the latest `golangci-lint`
- Addition of some `-build_flags=--mod=mod` `mockgen` parameters that were introduced in #3898 

In order to properly apply this change to an existing VM or locally you'd need to run these commands beforehand:
```
go clean [github.com/gobuffalo/packr/v2](http://github.com/gobuffalo/packr/v2)
which packr2|xargs rm
find . -name '*-packr.go'|xargs rm
find . -name 'packrd'|xargs rmdir
```
to clean up after the old tool.


jcscottiii (commented):
Thanks for doing this.

Here are some files that need updating:
- https://github.com/web-platform-tests/wpt.fyi/blob/d12a4236725eb19b7a13870fc9bb043bea221116/util/tools.go#L9
  - remove
- https://github.com/web-platform-tests/wpt.fyi/blob/d12a4236725eb19b7a13870fc9bb043bea221116/webapp/web/.gcloudignore#L41
  - rename to embed
- https://github.com/web-platform-tests/wpt.fyi/blob/d12a4236725eb19b7a13870fc9bb043bea221116/util/server-watch.sh#L7
  - rename to embed


Question:
- In the `go_build_dev` Makefile [target](https://github.com/web-platform-tests/wpt.fyi/blob/d12a4236725eb19b7a13870fc9bb043bea221116/Makefile#L66), we skip packr. At first, glance, I don't see an equivalent for embed. We may need to write a bug to look into this functionality later.

Also, I see the deployment is failing with this error:
```
api/checks/summaries/compile.go:24:12: pattern templates/*.md: no matching files found
webapp/dynamic_components_handler.go:16:12: pattern dynamic-components/templates/*: no matching files found
```

We may need to get rid of these lines: https://github.com/web-platform-tests/wpt.fyi/blob/d12a4236725eb19b7a13870fc9bb043bea221116/webapp/web/.gcloudignore#L41-L45

I'm guessing because packr happened locally before we uploaded the bits via `go generate`. whereas go embed does not run with go generate. Instead, it happens during build time. And build time happens in the cloud on app engine. So we need to upload those files now.

past:
All good points, thanks for pointing me to .gcloudignore!

past:
@jcscottiii do you want to take a look at the last 5 commits that you haven't reviewed yet? Note that the fourth commit is just the result of rebasing this PR on top of #3991.

As far as I can tell every problem is now resolved. All tests pass locally and in CI, testing the website works both locally and in the [staged URL](https://past-packr-expand-dot-wptdashboard-staging.uk.r.appspot.com/), and the check failure seems transient: all [recent builds](https://pantheon.corp.google.com/cloud-build/builds;region=us-east4?inv=1&invt=Abbs4A&project=wptdashboard-staging) completed successfully.

PR #3953, opened 2024-08-16
Remove a reduntant nil check

past:
Fixes a gosimple [lint error](https://github.com/web-platform-tests/wpt.fyi/actions/runs/10410088455/job/28831116906?pr=3948): "api/query/query.go:102:5: S1009: should omit nil check; len() for []int64 is defined as zero (gosimple)"

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/web-platform-tests/wpt.fyi
 
-go 1.21
+go 1.21.13
 
 require (
 	cloud.google.com/go/cloudtasks v1.12.13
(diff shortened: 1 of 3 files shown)


jcscottiii (approved):

past:
go_cloud_test started before I updated the key, but staging deployment started afterwards and failed with an internal error as it frequently does.

go.mod
jcscottiii, line 1: This makes sense. I did not know that they required the patch version as of 1.21 until looking into this.

image

Thanks for this.

PR #3968, opened 2024-08-26
Fix the int conversion warning in linter

KyleJu:
Fix the int conversion warning in Linter, suggested in https://github.com/securego/gosec/issues/1185#issuecomment-2304307264. Nolint the intended use of int64 to int conversion

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,7 @@ GO_FILES := $(shell find $(WPTD_PATH) -type f -name '*.go')
 GO_TEST_FILES := $(shell find $(WPTD_PATH) -type f -name '*_test.go')
 # Golangci version should be updated periodically.
 # See: https://golangci-lint.run/usage/install/#other-ci
-GOLANGCI_LINT_VERSION := v1.52.2
+GOLANGCI_LINT_VERSION := v1.60.3
 
 build: go_build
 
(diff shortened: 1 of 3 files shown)


past:
According to [this comment](https://github.com/golangci/golangci-lint/pull/4941#issuecomment-2308537266) the latest version of golangci-lint should not have many false positives. Can you try first to update [ours](https://github.com/web-platform-tests/wpt.fyi/blob/main/Makefile#L30) and see if the errors are gone?

KyleJu:
> According to [this comment](https://github.com/golangci/golangci-lint/pull/4941#issuecomment-2308537266) the latest version of golangci-lint should not have many false positives. Can you try first to update [ours](https://github.com/web-platform-tests/wpt.fyi/blob/main/Makefile#L30) and see if the errors are gone?

The false positive case mentioned is an ["uint8 -> int" overflow false positive](https://github.com/securego/gosec/issues/1185#issuecomment-2302209267), or "byte to int". In our cases, we are casting int 64 to int 32 intentionally, which doesn't fall into the false positive.

Regardless I will update the latest golangci-lint version. Thanks for the comment!

past (approved):
Ideally we would stop doing the error-prone type casting and use int (or uint64) consistently, but we don't need to block the linter fix on this change.

PR #4005, opened 2024-09-17
Make GitHub wpt org admins have admin access to wpt.fyi

gsnedders:
This should make it less nebulous as to who is and isn't an admin.

### DIFF
--- a/shared/github_oauth.go
+++ b/shared/github_oauth.go
@@ -163,6 +163,10 @@ func (gaci githubAccessControlImpl) IsValidAdmin() (bool, error) {
 	if !valid {
 		return false, errors.New("Invalid access token")
 	}
+	membership, _, err := gaci.oauthGHClient.Organizations.GetOrgMembership(gaci.ctx, "", "web-platform-tests")
+	if err == nil && *membership.Role == "admin" {
+		return true, nil
+	}
 	key := gaci.ds.NewNameKey("Admin", gaci.user.GitHubHandle)
 	var dst struct{}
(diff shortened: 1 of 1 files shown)


KyleJu:
I am unsure about this change - WPT.fyi has [a few places](https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt.fyi%20IsValidAdmin&type=code) where only designated admins can use. E.g. admin handlers where we set the UI and server side flags. Also I noticed that wpt-pr-bot is an admin in [the wpt org](https://github.com/web-platform-tests/wpt/settings/access). I am unsure about the bot account having the admin access.

It is not difficult to add an admin to the project, when requested. What do you think @past?

KyleJu:
> I am unsure about this change - WPT.fyi has [a few places](https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt.fyi%20IsValidAdmin&type=code) where only designated admins can use. E.g. admin handlers where we set the UI and server side flags. Also I noticed that wpt-pr-bot is an admin in [the wpt org](https://github.com/web-platform-tests/wpt/settings/access). I am unsure about the bot account having the admin access.
> 
> It is not difficult to manually add an admin to the project, when requested. What do you think @past?

Had an offline discussion with the team. We all agreed on having a single source of truth for the admin list - either maintaining a list in wpt.fyi Datastore or fetching from the WPT org, not both. If adding the wpt org admins is necessary, what if we add all wpt org admins to the wpt.fyi admin list, but with limited privileges? e.g. access to the admin handler should be limited.

PR #4008, opened 2024-09-20
Upgrade Go to version 1.22

past:
The next step in the Go module updates saga. This includes a lint warning fix and running `go get -u ./... ;
go mod tidy`.


past:
The checks fail because they run on webplatformtests/wpt.fyi:latest instead of webplatformtests/wpt.fyi:past-upgrade-modules. Every test passes locally.

KyleJu:
> The checks fail because they run on webplatformtests/wpt.fyi:latest instead of webplatformtests/wpt.fyi:past-upgrade-modules. Every test passes locally.

That is correct. However, in [the Go 1.21 update PR](https://github.com/web-platform-tests/wpt.fyi/pull/3909),  the check also uses webplatformtests/wpt.fyi:latest but the CI passes. How did that pass?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/web-platform-tests/wpt.fyi
 
-go 1.21.13
```
jcscottiii, line 3: I am curious if go 1.22 will allow us to keep go1.12.13 here for now (but keep the rest of the changes) Then in the next PR, we could bump this. It might be something to try.

jcscottiii (approved):

past:
> That is correct. However, in [the Go 1.21 update PR](https://github.com/web-platform-tests/wpt.fyi/pull/3909), the check also uses [webplatformtests/wpt.fyi:latest](https://github.com/web-platform-tests/wpt.fyi/actions/runs/9989846596/job/27609309899) but the CI passes. How did that pass?

Because that PR didn't change the go.mod version format and Go 1.20 was still able to parse it. Essentially PRs that touch Docker file may or may not pass, depending on how backwards-compatible these changes are.

go.mod
past: I tried it, but the module updates in this PR require 1.22, so checks fail demanding updates to go.mod. 

Maybe it could work if we split the runtime and module updates in two separate PRs. I can try that in 1.23, once it is available in AppEngine.

Issue #3981, opened 2024-09-06
Add more tests to api/ghactions


gsnedders:
See https://github.com/web-platform-tests/wpt.fyi/pull/3970#issuecomment-2332769826

PR #4010, opened 2024-09-20
Use non-conflicting route for GitHub actions

### LINKED ISSUE #4009: ghactions.notifyHandler doesn't get called for requests to /api/checks/github-actions
See the discussion at https://github.com/web-platform-tests/wpt/issues/48223. [This run](https://github.com/web-platform-tests/wpt/actions/runs/10925326720/job/30347359745) should have ended up at ghactions.notifyHandler, but instead was handled by checks.updateCheckHandler.

past:
This should fix #4009. According to [mux docs](https://pkg.go.dev/github.com/gorilla/mux#readme-matching-routes) "Routes are tested in the order they were added to the router. If two routes match, the first one wins." The route `/api/checks/github-actions` is picked up by the handler for route `/api/checks/{commit}`. Presumably the route addition happens as packages are initialized in lexicographic order and `checks` comes before `ghactions`.

Tested locally with cURL, but #3981 would have helped here.

### DIFF
--- a/api/ghactions/routes.go
+++ b/api/ghactions/routes.go
@@ -12,5 +12,5 @@ func RegisterRoutes() {
 	// the results of a GitHub Actions workflow run.
 	// The endpoint is insecure, because we'll only try to fetch (specifically) a
 	// web-platform-tests/wpt build with the given ID.
-	shared.AddRoute("/api/checks/github-actions/", "github-actions-notify", notifyHandler).Methods("POST")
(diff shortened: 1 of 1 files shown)


gsnedders:
This makes this differ to the Azure Pipelines endpoint, which feels surprising. Maybe we should just reorder them in `webapp/web/main.go`?

past:
Oh, that would probably work, too.

past:
> Oh, that would probably work, too.

Turns out it doesn't and I don't know why.

past:
Closed in favor of #4017.

PR #4028, opened 2024-10-05
Upgrade to Debian Bookworm and Python 3.11

past:
This PR upgrades the Debian distro to Bookworm, which requires an update in the Python (3.11) and Java (17) versions. 

It also comes with an update to the Google Cloud CLI package names and repository secret installation (the recommended approach in the deprecation section of `man apt-key`). This is a prerequisite for upgrading to Go 1.23.

Had to run `pip-compile requirements.in` as part of the process.


past:
The check failures are due to the jobs running with the image `webplatformtests/wpt.fyi:latest` instead of `webplatformtests/wpt.fyi:past-debian-bookworm`.

jcscottiii:
I wrote up a suggestion to help with those times when the deploy fails in these cases: https://github.com/web-platform-tests/wpt.fyi/issues/4029

jcscottiii (approved):
LGTM

PR #4040, opened 2024-10-07
Add labels to interop-data.{js,json}

gsnedders:
Without the labels present here, consumers of this JS/JSON data had two choices: one was to rely on the fact this data is implicitly in this file (within the `tests` URL), and the other was to pull down the category-data.json file from the results-analysis repo which explicitly contains the focus area to label mapping (but does not contain data like "what is an active focus area").


DanielRyanSmith (approved):
Yeah, it's as you said that this label mapping was previously only available in the results-analysis repo.

This will clear things up a bit for anyone is looking here on what this data means. 🙂

Issue #4020, opened 2024-09-25
GitHub Actions results not appearing on the dashboard


gsnedders:
From https://github.com/web-platform-tests/wpt/issues/48223#issuecomment-2375384661:

> So:
> * https://github.com/web-platform-tests/wpt/actions/runs/11039052681/job/30670828741 succeeded (which implies that the results receiver got through the uploader and all without giving any error), got 200 back
> * https://wpt.fyi/status doesn't list any pending or invalid github-actions submissions
> * https://wpt.fyi/runs?max-count=100&from=2024-09-18T00%3A00&product=safari%5Bgithub-actions%5D doesn't have any runs

This seems surprising that it succeeded to upload, and yet the upload doesn't appear as pending, invalid, or show up as complete.

We seem to have logging going into Google Cloud Logging, but I don't have any access to that.

cc @past @DanielRyanSmith @jcscottiii @KyleJu

KyleJu:
Checking the log, I can confirm that `/api/checks/github-actions/` was called with `run_id=11039052681` and returned 200. Also I see the logging 
`Adding %!s(*string=0xc0002990a0) for web-platform-tests/wpt run 11039052681 to upload...`  from [this line](https://github.com/web-platform-tests/wpt.fyi/blob/a24fc662deecac9bab58d5e6b0a240bff3a3cc17/api/ghactions/notify.go#L132), but it didn't log properly:

`log.Infof("Adding %s for %s/%s run %v to upload...", artifact.Name, owner, repo, runID)`

It should be `*artifact.Name` instead.  It could be an issue with the artifacts. Is there a link that I can inspect the wpt results from the artifacts?

What if we

gsnedders:
@KyleJu https://github.com/web-platform-tests/wpt/actions/runs/11039052681 has downloads links for all the artifacts.

gsnedders:
I guess I'm more interested in any logging from the request we issue for https://wpt.fyi/api/results/create? Because that's presumably working too and giving 200 back.

https://github.com/web-platform-tests/wpt.fyi/blob/a24fc662deecac9bab58d5e6b0a240bff3a3cc17/api/receiver/create_run.go#L115-L117 is one case where we could silently fail?

But it seems like it should presumably be managing to schedule the task, it's just for whatever reason it never completes.

KyleJu:
Confirmed that `/api/results/create` has no error log around the time this run was created. 

[...]

PR #4057, opened 2024-10-15
[Processor] Correct cap for secret access

KyleJu:
Fix an issue in https://github.com/web-platform-tests/wpt.fyi/issues/4020. The self.datastore.get(key)['Secret'] should be capitalized. See [here](https://pantheon.corp.google.com/datastore/databases/-default-/entities;kind=Token;ns=__$DEFAULT$__/query/kind?inv=1&invt=Abe2ew&project=wptdashboard)

### DIFF
--- a/results-processor/processor.py
+++ b/results-processor/processor.py
@@ -118,7 +118,7 @@ class Processor(object):
     def _secret(self, token_name):
         _log.info('Reading secret: %s', token_name)
         key = self.datastore.key('Token', token_name)
-        return self.datastore.get(key)['secret']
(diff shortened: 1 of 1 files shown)


jcscottiii (approved):
LGTM! Thanks for catching this.

I verified that it is indeed capitalized. 


image

gsnedders:
Ah, this made me believe it's lowercase:

https://github.com/web-platform-tests/wpt.fyi/blob/76e0bfe1a31065621c7c055c9e5e7ac18e0d7998/shared/models.go#L466-L469

PR #4060, opened 2024-10-15
Get almost the entire results-processor typed

gsnedders:
This should be relatively self-evident to review, and CI should verify whether or not this works.


KyleJu (approved):
I asked @jrobbins for help since he has more experience with mypy, but I also looked more into the use of cast() in this PR. LGTM

PR #4061, opened 2024-10-15
Fix processing of archives with no parent directory

gsnedders:
The ZIP files created by GitHub Actions don't have any parent directory, and just directly contain wpt_report.json, etc.

Previously, we were searching for filenames starting with "/", which obviously doesn't match if there is no parent directory.

I believe this will fix the issue described in https://github.com/web-platform-tests/wpt.fyi/issues/4020#issuecomment-2415214323, which will hopefully allow us to close that issue.


results-processor/processor.py
```diff
@@ -187,13 +188,14 @@ def _download_archive(self, archive_url):
...
                     continue
-                path = z.extract(f, path=self._temp_dir)
-                if re.match(r'^.*/wpt_report.*\.json$', f.filename):
+                basename = posixpath.basename(f.filename)
+                if fnmatch.fnmatchcase(basename, 'wpt_report*.json'):
+                    path = z.extract(f, path=self._temp_dir)
                     self.results.append(path)
-                if re.match(r'^.*/wpt_screenshot.*\.txt$', f.filename):
+                elif fnmatch.fnmatchcase(basename, 'wpt_screenshot*.txt'):
+                    path = z.extract(f, path=self._temp_dir)
```
past, line 198: Why did you move (and duplicate) the path assignment into the conditional? Do the if/elif conditions have side effects?

past (approved):

results-processor/processor.py
gsnedders: Mostly just avoiding decompressing things unnecessarily, try to mitigate some of the risk of any sort of decompression bomb by only decompressing the files we actually care about.

PR #4080, opened 2024-10-21
Add interop-2023-webcomponents as a previous focus area in 2024

gsnedders:
This appears to have been omitted, while I believe everything else is present.


DanielRyanSmith (approved):
Good catch!

PR #4085, opened 2024-10-22
Add interop and investigation scores to mobile results

DanielRyanSmith:
Small change to add the "Interop" and "Investigations" scores to the mobile results view, as well as the Investigations scores table.

### DIFF
--- a/webapp/components/interop-dashboard.js
+++ b/webapp/components/interop-dashboard.js
@@ -238,13 +238,10 @@ class InteropDashboard extends WPTFlags(PolymerElement) {
           padding: 0 5px;
         }
 
-        .warning {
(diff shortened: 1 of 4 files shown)


webapp/components/interop-dashboard.js
```diff
@@ -244,6 +244,10 @@ class InteropDashboard extends WPTFlags(PolymerElement) {
           border-radius: 5px;
         }
 
+        #mobileWarning {
+          margin-bottom: 32px;
+        }
```
past, line 249: Perf/simplicity nit: since you are adding this style, better add the .warning bits here as well, since the latter is now redundant.

past (approved):
LGTM

Issue #3840, opened 2024-05-25
Local build is broken


atjn:
Hi :) I am trying to follow the guide on setting up local development of the wpt.fyi dashboard. I am running into a lot of problems and I have no idea how to proceed.

I have executed the exact commands detailed in: https://github.com/web-platform-tests/wpt.fyi/tree/f319b11328efa5312fb7f472ca9222df58e2fb9f?tab=readme-ov-file#development
The only exception is that I need to run `sudo` in front of the commands, otherwise Docker won't run. I cannot see how that would create the following errors, I run many other Docker containers this way with no problems.

## Error 1
Git decides to crash because the directory `/home/user/wpt.fyi`  does not have the correct owner. I fixed this temporarily by adding the following to `/Makefile`:
```diff
@@ -56,6 +56,7 @@ go_build: git mockgen packr2 github_action_go_setup
        make webapp_node_modules_prod
        go generate ./...
        # Check all packages without producing any output.
+       git config --global --add safe.directory /home/user/wpt.fyi
        go build -v ./...
        # Build the webapp.
        go build -v ./webapp/web
```

## Error 2
The Google cloud emulator crashes because it cannot find `env.yaml` in the `data_dir`. I fixed this by adding the following to `util/docker-dev/web_server.sh`:
```diff
@@ -21,4 +21,4 @@ if [ "${DOCKER_STATUS}" != "0" ]; then
 fi
 
 info "Starting web server. Port forwarded to host: ${WPTD_HOST_WEB_PORT}"
-wptd_exec_it "\$(gcloud beta emulators datastore env-init) && util/server-watch.sh"
+wptd_exec_it "\$(gcloud config set project wptdashboard) && \$(gcloud beta emulators datastore start) && \$(gcloud beta emulators datastore env-init) && util/server-watch.sh"
```

## Error 3
At this point it seems like the setup script runs correctly, but trying to access `loclhost:8080` or `localhost:8081` results in a `connection reset` error. I do not know how to fix that.

My operating system: Linux, Ubuntu 22.04 LTS
CPU: Modern Intel AMD64

I hope you can help me figure out what is wrong. I am happy to share any other details.

Here is the full log of what I did to see error 3 after I had applied the fixes for error 1 and 2.
<details>
<summary>Full log</summary>

```
atjn:~/Documents/GitHub/wpt.fyi$ sudo ./util/docker-dev/run.sh
[sudo] password for atjn: 
[...]

PR #4089, opened 2024-10-27
Add subtle striping for table in `<wpt-results/>`

### LINKED ISSUE #1783: Make it easier to distinguish rows on the result page
Two ideas:
* Highlight the current (hovered/focused) row
* Use alternating background colours in the table

(Suggested by @george-steel )

YoshiRulz:
## Description
Resolves #1783/#2217 by adding a grey background to every other row in the table.
Before/after:
comparison
Untested because I'm on NixOS and Docker isn't trivial, though based on #3840 I'm guessing it wouldn't have worked anyway. The above screenshot was taken by adding the rule via the dev tools.

## Review Information
idk man

## Changes
Added a single CSS rule. edit: two now

## Requirements
N/A

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -93,6 +93,9 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
         top: 0;
         z-index: 1;
       }
+      path-part {
+        vertical-align: bottom;
+      }
       .path {
         margin-bottom: 16px;
       }
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -74,6 +74,9 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
         width: 100%;
         border-collapse: collapse;
       }
+      tr:nth-child(2n) {
+        background-color: #7F7F7F1F;
+      }
       tr.spec {
         background-color: var(--paper-grey-200);
       }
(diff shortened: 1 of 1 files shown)


KyleJu:
@YoshiRulz Hi James! First of all, thank you for the fix! One minor follow-up: the test names are not vertically centered around the cells. Could you center them?

image

YoshiRulz:
I had a look and it seems non-trivial (`path-part { margin: auto 0; }` and similar don't work), but in any case that problem isn't related to this changeset and only happens when the viewport is narrow. It is possible to align them to the bottom of the cell with `vertical-align` though, which IMO reads better, shall I do that?

KyleJu:
> I had a look and it seems non-trivial (`path-part { margin: auto 0; }` and similar don't work), but in any case that problem isn't related to this changeset and only happens when the viewport is narrow. It is possible to align them to the bottom of the cell with `vertical-align` though, which IMO reads better, shall I do that?

You are right! I couldn't get it to move after fiddling with the margin and padding in `path-part`, so feel free to align them to the bottom of the cell with vertical-align. Let me spend more time on it before the end of the week!

Another thing is that on the homepage, it uses a different shade of grey, #EEEEEE

Screenshot 2024-10-30 11 09 43 AM

I would like to make it consistent! Thanks again!

KyleJu:
Staging builds:
Screenshot 2024-11-19 11 39 47 AM

Screenshot 2024-11-19 11 39 47 AM

KyleJu (approved):

PR #4055, opened 2024-10-14
Reapply "Upgrade to Debian Bookworm and Python 3.11 (#4028)"

past:
This is reapplying the changes from #4028 while fixing the selenium issues.

PR #4098, opened 2024-11-01
Use the new docker image on PRs that change it

### LINKED ISSUE #3187: PRs that update the Docker image should run tests on that image
Kyle tells me that we used to have this in Travis CI, but we might have left it behind after the migration to GitHub actions.

Right now PRs that modify any of the docker configuration files (e.g. #3180) fail in CI because the GitHub actions use the latest docker image from merges to the main branch. A sketch of the desired solution here would be to change [`.github/actions/make-in-docker/action.yml`](https://github.com/web-platform-tests/wpt.fyi/blob/main/.github/actions/make-in-docker/action.yml#L9) to:

image: docker://webplatformtests/wpt.fyi:${{ vars.GITHUB_REF_NAME }}

And perhaps also [set `tag_names: true`](https://github.com/elgohr/Publish-Docker-Github-Action#tag_names) in [docker-update.yml](https://github.com/web-platform-tests/wpt.fyi/blob/main/.github/workflows/docker-update.yml#L22)

past:
Fixes #3187.

The idea that I had in that issue doesn't work for a number of reasons. I will note that docker-update.yml runs not only for `Dockerfile` changes, but also for `.github/workflows/docker-update.yml` changes, which doesn't apply in this case.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - name: Docker config check
+        uses: dorny/paths-filter@v3
+        id: dockerchanges
+        with:
+          filters: |
+            dockerconf:
+              - 'Dockerfile'
+        # run only if Docker configuration was changed
+      - if: steps.dockerchanges.outputs.dockerconf == 'true'
+        name: Rebuild Docker image
+        run: docker build -t webplatformtests/wpt.fyi:latest .
       - uses: ./.github/actions/make-in-docker
         with:
           target: lint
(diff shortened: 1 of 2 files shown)


past:
Tested in #4055.

jcscottiii (approved):
LGTM. I wonder if we could use [composite actions](https://docs.github.com/en/actions/sharing-automations/creating-actions/creating-a-composite-action#:~:text=Composite%20actions%20allow%20you%20to%20collect%20a%20series%20of%20workflow%20job%20steps%20into%20a%20single%20action%20which%20you%20can%20then%20run%20as%20a%20single%20job%20step%20in%20multiple%20workflows) to help DRY this up. But I'm not too familiar with it. Maybe something to look into in the future.

past:
It's an interesting idea, thanks. I don't have cycles for this right now, but I'll keep it in mind for the future.

Issue #4107, opened 2024-11-06
Some Focus Area percentages on dashboard don't match the results percentage


cookiecrook:
Some Focus Area percentages on dashboard don't match the results percentage

For example, the [dashboard](https://wpt.fyi/interop-2024) lists the Accessibility Focus Area at 97.8% and 99.7% respectively for Firefox and Safari…

<img width="693" alt="screen shot of dashboard listing scores described in the issue prose: 97.8% and 99.7%" src="https://github.com/user-attachments/assets/17a0903d-5169-4f73-a87c-e4d6839dc8da">

But if you click through to the [Focus Area results](https://wpt.fyi/results/?label=experimental&label=master&aligned&view=interop&q=label%3Ainterop-2024-accessibility), the scores are 98% and 99.9%.  (Bottom row right.)

<img width="589" alt="screen shot of results listing scores described in the issue prose: 98 and 99.9%." src="https://github.com/user-attachments/assets/3e42953e-0772-4059-b17b-e6ec6ce04846">

I haven't dug into which is correct, but it seems clear one of them is wrong, and the dashboard percentages should be consistent throughout.

KyleJu:
Assign @DanielRyanSmith just so this issue has an owner. Feel free to discuss this in our meeting!

cookiecrook:
Any movement on this @DanielRyanSmith? Thanks.

DanielRyanSmith:
Sorry, I missed this until now - I'm not quite sure immediately what is happening here yet, since safari does not have any flaky tests that might cause this (that was my initial guess).

I see Safari with a 99.7% on the dashboard and a 99.9% on the results page.

![Screenshot 2024-12-02 7 18 37 PM](https://github.com/user-attachments/assets/14700a5a-7940-4d0e-b378-b5deacea2231)
---
![Screenshot 2024-12-02 7 18 53 PM](https://github.com/user-attachments/assets/5a4e1748-8a56-405b-9b0d-76eef820f519)

If I had to guess again, it might be the fact that some tests/subtests under the label "interop-2024-accessibility" were added/removed at some point through the year, and the interop score aggregation script is using an incorrect number of tests to calculate the passing percentage.

cookiecrook:
Thanks for digging in... The total number has fluctuated a bit as invalid or disputed tests were removed or modified.

DanielRyanSmith:
[...]

PR #4142, opened 2024-12-04
Update rounding logic for results page

DanielRyanSmith:
Fixes https://github.com/web-platform-tests/wpt.fyi/issues/4107

The test results page was handling score rounding slightly differently than the interop dashboard (rounding to the nearest whole percent rather than nearest tenth of a percent). In order to avoid rounding up to 100% in cases where there were still some tests/subtests failing, [some logic was added to instead display the score as 99.9% in those cases](https://github.com/web-platform-tests/wpt.fyi/blob/b2c8fe3da11a55d5ad22f8d2d89b2f45c1fad94f/webapp/views/wpt-results.js#L1191-L1193). This created a situation in which the interop dashboard and the results page looked close enough to each other (99.7% vs. 99.9%) that it seemed as if there was a calculation error on one of the pages. I hadn't looked at this logic in some time and also assumed this was some calculation error. :sweat_smile: 

This very simple change makes it so the results page rounds the score in the same way as the interop dashboard, so this confusion will not come up again.

### Before
C66AS3UueX5yAnL

### After
7hqcLeTCmKGhfNA

### DIFF
--- a/webapp/views/wpt-results.js
+++ b/webapp/views/wpt-results.js
@@ -1176,7 +1176,7 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
       return `${this.getTestNumbersDisplay(passes, total, isDir)} subtests`;
     }
 
-    const formatPercent = parseFloat((passes / total * 100).toFixed(0));
+    const formatPercent = parseFloat((passes / total * 100).toFixed(1));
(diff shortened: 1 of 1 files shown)


jcscottiii (approved):
LGTM. A possible improvement would be to have a tooltip or alternative text that is initially hidden. When triggered (on hover or something else), it shows the precise number.

PR #4184, opened 2024-12-28
Use one-way bindings for the flags component

past:
Added a test for the persistence of flags. This is another step in the migration from Polymer to Lit.


KyleJu (approved):
Verified on the staging instance. Thanks!

PR #4186, opened 2025-01-03
Use one-way binding for loading-state

past:
Also, make some code health improvements while I'm here:
- Mark private methods with an underscore
- Use console.error for error messages
- Use numerice comparison for numeric variables
- Make isLoading read-only
- Add comments

### DIFF
--- a/webapp/components/loading-state.js
+++ b/webapp/components/loading-state.js
@@ -4,46 +4,74 @@
  * found in the LICENSE file.
  */
 
-/*
-LoadingState is a behaviour component for indicating when information is
-still being loaded (generally, fetched).
-*/
-
-
+/**
+ * LoadingState is a behaviour component for indicating when information is
+ * still being loaded (generally, fetched).
+ */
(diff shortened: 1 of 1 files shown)


webapp/components/loading-state.js
```diff
@@ -4,46 +4,74 @@
...
+        observer: '_loadingCountChanged',
       },
+      /**
+       * Whether the component is currently loading data.
+       * Computed based on `loadingCount`.
+       */
       isLoading: {
         type: Boolean,
-        computed: 'computeIsLoading(loadingCount)',
-        notify: true,
```
KyleJu, line 24: Note: check if removing notify: true will break anything

KyleJu: isLoading isn't used as two-way data bindings in the codebase anywhere[1]. It is safe

[1] https://github.com/search?q=repo%3Aweb-platform-tests%2Fwpt.fyi%20isLoading&type=code

KyleJu (approved):

PR #4187, opened 2025-01-04
Use one-way bindings for wpt-metadata and wpt-amend-metadata

past:
This migration was a bit larger than usual due to the fact that the metadata component and the triage component interact so closely. The main changes are:

- Removal of `notify: true` and change double angle brackets to double brackets.
- Use get/set to let Polymer properly detect array property updates.
- DOM manipulation logic is now encapsulated within methods to better adhere to Polymer's data flow principles.

### DIFF
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -195,7 +195,7 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     </template>
   </tbody>
 </table>
-<wpt-amend-metadata id="amend" selected-metadata="{{selectedMetadata}}" path="[[path]]"></wpt-amend-metadata>
+<wpt-amend-metadata id="amend" selected-metadata="[[selectedMetadata]]" path="[[path]]"></wpt-amend-metadata>
 `;
   }
 
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -295,6 +295,7 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     this.toggleDiffFilter = () => {
       this.onlyShowDifferences = !this.onlyShowDifferences;
     };
+    this.addEventListener('selected-metadata-changed', this.selectedMetadataChanged);
   }
 
   computeDisplayedProducts(testRuns) {
@@ -457,7 +458,13 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
   }
 
   clearSelectedCells(selectedMetadata) {
-    this.handleClear(selectedMetadata);
+    if (this.selectedMetadata.length === 0) {
+      this.handleClear(this.selectedMetadata);
+    }
+  }
+
+  selectedMetadataChanged(e) {
+    this.selectedMetadata = e.detail.value;
   }
 
   handleTriageMode(isTriageMode) {
--- a/webapp/components/test-file-results-table.js
+++ b/webapp/components/test-file-results-table.js
@@ -457,7 +457,7 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     return ['FAIL', 'ERROR', 'TIMEOUT'].includes(status);
   }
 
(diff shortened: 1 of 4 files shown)


webapp/components/wpt-metadata.js
```diff
@@ -100,7 +100,7 @@ class WPTMetadata extends PathInfo(LoadingState(PolymerElement)) {
               ></wpt-metadata-node>
             </template>
           </iron-collapse>
-          <paper-button id="metadata-toggle" onclick="[[openCollapsible]]">
```
KyleJu, line 103: nit: this variable should be removed in the file.

webapp/components/test-file-results-table.js
```diff
@@ -195,7 +195,7 @@ class TestFileResultsTable extends WPTFlags(Pluralizer(AmendMetadataMixin(WPTCol
     </template>
   </tbody>
 </table>
-<wpt-amend-metadata id="amend" selected-metadata="{{selectedMetadata}}" path="[[path]]"></wpt-amend-metadata>
+<wpt-amend-metadata id="amend" selected-metadata="[[selectedMetadata]]" path="[[path]]"></wpt-amend-metadata>
```
KyleJu, line 198: we still need to listen to the change in `selectedMetadata` in test-file-results-table.js, because it is used to clear the selected cells UI when the triage UI is dismissed through `clearSelectedCells`. I will post a recorded video clips to demonstrate that

webapp/views/wpt-results.js
```diff
@@ -401,11 +401,11 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
...
                     path="[[path]]"
                     search-results="[[searchResults]]"
-                    metadata-map="{{metadataMap}}"
-                    label-map="{{labelMap}}"
+                    metadata-map="[[metadataMap]]"
+                    label-map="[[labelMap}]]"
                     triage-notifier="[[triageNotifier]]"></wpt-metadata>
     </template>
-    <wpt-amend-metadata id="amend" selected-metadata="{{selectedMetadata}}" path="[[path]]"></wpt-amend-metadata>
+    <wpt-amend-metadata id="amend" selected-metadata="[[selectedMetadata]]" path="[[path]]"></wpt-amend-metadata>
```
KyleJu, line 408: ditto. Still need a EventListener for selectedMetadata

webapp/views/wpt-results.js
```diff
@@ -401,11 +401,11 @@ class WPTResults extends AmendMetadataMixin(Pluralizer(WPTColors(WPTFlags(PathIn
       <wpt-metadata products="[[displayedProducts]]"
                     path="[[path]]"
                     search-results="[[searchResults]]"
-                    metadata-map="{{metadataMap}}"
-                    label-map="{{labelMap}}"
+                    metadata-map="[[metadataMap]]"
+                    label-map="[[labelMap}]]"
```
KyleJu, line 405: wpt-results still need to add EventListener for `metadataMap` and `labelMap` because the data is being fetched inside of <wpt-metadata>. Once it is fetched, <wpt-metadata> notifies wpt-results that data has been fetched. Then wpt-results updates the inline labels UI in the test results.

KyleJu:
[Screen recording 2025-01-17 4.15.36 PM.webm](https://github.com/user-attachments/assets/6f059b58-026e-49bd-8871-8a8d580d46bd)

How the upward data binding for `selectedMetadata` is being used

webapp/components/test-file-results-table.js
past: Good catch! It's actually the method in `wpt-results.js` that clears the cells, but I changed it here for consistency.

webapp/views/wpt-results.js
past: Excellent point, fixed.

webapp/components/wpt-metadata.js
past: Done.

webapp/views/wpt-results.js
past: OK, done.

past:
This is ready now for another review pass.

KyleJu (approved):
Everything looks good. Thanks for the changes!

PR #4204, opened 2025-01-17
Chromium bug searches should use the new URL

past:
I noticed that the triage UI search functionality is currently broken for Chromium, in that the query to search for bugs mentioning the selected file name results in a search for all open bugs. This is caused by the migration from Monorail to Buganizer, which uses slightly different paths.

I took the opportunity to update all Chromium bug tracker URLs to the new format, with the exception of those in `metadata_util_test.go` that are trying to match the data included in an old archive from wpt-metadata. Arguably the test should be rewritten to not rely on explicit string values in those fields, but it doesn't seem very useful to update the actual archive as well to really eliminate every old URL instance.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -576,7 +576,7 @@ __`product`__ : browser[version[os[version]]]. e.g. `chrome-63.0-linux`
 
 #### Post Examples
 - POST /api/metadata?product=chrome\&product=firefox \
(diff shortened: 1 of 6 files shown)


past:
This PR is based on top of #4187 to avoid merge conflicts, and I intend to rebase it once that one lands.

KyleJu (approved):

PR #4220, opened 2025-01-29
Check Chromium revision pinning scripts into repo

DanielRyanSmith:
This adds the source code for the GCP cloud functions that are run to automatically update the pinned Chromium revision in the WPT repository.

Subsequent changes will come with deployment and update instructions.


scripts/check_chromium_revision.py
```diff
@@ -0,0 +1,133 @@
...
+PR_NUMBER = "50375"
+REPO_OWNER = "web-platform-tests"
+REPO_NAME = "wpt"
+
+
+def trigger_ci_tests() -> str | None:
+    # Reopen the PR to run the CI tests.
+    s = requests.Session()
+    s.headers.update({"Authorization": f"token {get_token()}"})
+    url = f"https://api.github.com/repos/{REPO_OWNER}/{REPO_NAME}/pulls/{PR_NUMBER}"
```
jcscottiii, line 30: We should either:
1. Use the GitHub client python library
2. Manually [pin](https://docs.github.com/en/rest/about-the-rest-api/api-versions?apiVersion=2022-11-28) the `X-GitHub-Api-Version` header in each API call.

scripts/update_chromium_revision.py
```diff
@@ -0,0 +1,97 @@
...
+PR_NUMBER = "50375"
+
+
+def all_passing_checks() -> bool:
+    """Check if all CI tests passed."""
+    s = requests.Session()
+    s.headers.update({
+        "Authorization": f"token {get_token()}"
+    })
+    url = f"https://api.github.com/repos/{REPO_OWNER}/{REPO_NAME}/commits/{get_sha()}/check-suites"
```
jcscottiii, line 27: We should either:
1. Use the GitHub client python library
2. Manually [pin](https://docs.github.com/en/rest/about-the-rest-api/api-versions?apiVersion=2022-11-28) the `X-GitHub-Api-Version` header in each API call.

Right now, we are expecting the response to be a certain format. If the API version changes and the shape of the response changes, we will start to get KeyError. (for example when we go to check the conclusion.

scripts/README.md
```diff
@@ -1,4 +1,24 @@
...
+for all major platforms (Win/Mac/Linux) and trigger the WPT CI check suite to
+run against this new revision.
+
+
+_update_chromium_revision.py_
+
+The purpose of this script is to check the WPT CI check suite to see if all
+tests passed for the new revision, and to update the pinned revision if so.
+
+The current PR used for running the check suites is at https://github.com/web-platform-tests/wpt/pull/50375
```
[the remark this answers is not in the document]
jcscottiii: Any instructions on how to create these PRs in the future?

scripts/check_chromium_revision.py
```diff
@@ -0,0 +1,133 @@
...
+
+DEFAULT_TIMEOUT = 600.0
+BUCKET_NAME = 'wpt-versions'
+NEW_REVISION_FILE = 'pinned_chromium_revision_NEW'
+OLD_REVISION_FILE = 'pinned_chromium_revision'
+PLATFORM_INFO = [
+    ("Win_x64", "chrome-win.zip"),
+    ("Win", "chrome-win.zip"),
+    ("Linux_x64", "chrome-linux.zip"),
+    ("Mac", "chrome-mac.zip")
```
jcscottiii, line 18: Outside the scope of this: But I wonder how this will work when we want Android given Chrome on Android runs on Chromium CI. We probably still want some coverage. Might be worth bringing this up to Weizhong.

scripts/check_chromium_revision.py
```diff
@@ -0,0 +1,133 @@
...
+]
+SNAPSHOTS_PATH = "https://storage.googleapis.com/chromium-browser-snapshots/"
+PR_NUMBER = "50375"
+REPO_OWNER = "web-platform-tests"
+REPO_NAME = "wpt"
```
jcscottiii, line 23: Thoughts on making these environment variables? A user shouldn't really need to change the code just to update these variables. Instead, they could quickly deploy the app with updated variables. You could do more of these variables but this set would be a good start. 

[[1](https://12factor.net/config#:~:text=The%20twelve%2Dfactor%20app%20stores%20config%20in%20environment%20variables%20(often%20shortened%20to%20env%20vars%20or%20env))]

jcscottiii (commented):
Thanks for adding these scripts. I added some comments to expand on how to deploy.

Also, you will find a comment on pinning our GitHub API version.

Other comments are nits or questions.

Let me know if you have any questions!

scripts/check_chromium_revision.py
DanielRyanSmith: I can definitely make these environment variable instead 🙂

scripts/check_chromium_revision.py
DanielRyanSmith: I'm not very familiar with the Chrome Android work, but is this shouldn't affect run for Chrome Android because it's considered a different product and does not use the pinned revision here (I might also be misunderstanding your comment 😅)

scripts/update_chromium_revision.py
DanielRyanSmith: Good idea. I think I'll approach using #2 and make changes once I confirm everything functions as expected.

scripts/README.md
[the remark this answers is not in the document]
DanielRyanSmith: > Could you add the command(s) to deploy these scripts in each section?

I've been doing it through the GCP UI, but I'll look up and add instructions for this.

> Also, do you know how to generate the GIT_CHECK_PR_STATUS_TOKEN stored in secret manager that is used in both of these scripts?

These are instructions I need to add as well, but I was thinking this might be better suited to live in our rotation docs.

> Also, do you know about the instructions for [process_test_history](https://pantheon.corp.google.com/functions/details/us-central1/process_test_history?env=gen2&project=wptdashboard)?

The instructions file `build_test_history.py` is what describes this, but I now realize I've not coordinated the names correctly here. It should probably be that "build_test_history" should be renamed to "process_test_history" to match the Cloud Function name.

DanielRyanSmith:
Thanks for taking such a good look at this James! 🙂

scripts/README.md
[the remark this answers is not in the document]
DanielRyanSmith: > Any instructions on how to create these PRs in the future?

Do you mean creating new cloud functions in the future? Or making changes to these files?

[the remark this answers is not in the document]
jcscottiii: > I've been doing it through the GCP UI, but I'll look up and add instructions for this.

That's good to know. I thought you were deploying it with the CLI. So I was curious what arguments you passed in. You could just reference this [document](https://cloud.google.com/functions/docs/deploy#console) and call out any gotchas that need to be filled in.

> These are instructions I need to add as well, but I was thinking this might be better suited to live in our rotation docs.

That works!

> The instructions file build_test_history.py is what describes this, but I now realize I've not coordinated the names correctly here. It should probably be that "build_test_history" should be renamed to "process_test_history" to match the Cloud Function name.

I saw that name and I was wondering if they were the same. Yeah could you either rename the file here or change the name in GCP?

> Do you mean creating new cloud functions in the future? Or making changes to these files?
I meant this PR https://github.com/web-platform-tests/wpt/pull/50375. But I guess let me ask these questions first:
1. How often do you change this PR?
2. Have we had other PRs like this in the past?
3. How did you know what changes to make in the first place for this PR?

scripts/update_chromium_revision.py
```diff
@@ -0,0 +1,107 @@
...
+        'against a new available Chromium revision. If all tests pass, the new '
+        'revision will be pinned for use.\\n\\nLast revision checked: '
+        f'{new_revision.decode('utf-8')}\\nCheck run date: {date.today()}'
+        f'\\nOutcome: **{outcome}**'
+    )
+
+    body = '{"body":"' + body + '"}'
+    s = requests.Session()
+    s.headers.update({'Authorization': f'token {get_token()}'})
+    url = f'https://api.github.com/repos/{repo_owner}/{repo_name}/pulls/{pr_number}'
```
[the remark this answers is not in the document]
DanielRyanSmith: Ah, I missed pinning these as well - and that's a great idea, I'll do that 🙂

scripts/check_chromium_revision.py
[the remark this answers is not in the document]
DanielRyanSmith: This is worded perfectly, so I added it directly 😎

jcscottiii (approved):

Issue #4219, opened 2025-01-29
Move the webapp to use App Engine's flexible runtime


jcscottiii:
# Problem & Background
During the weekly deployment to production, the webapp can repeatably timeout. This can lead to lengthy deployments for the engineer on call. However, the search cache and results processor seem to have higher success rates. The leading theory behind this is because they both use App Engine's [flexible runtime instead of the standard runtime](https://cloud.google.com/appengine/docs/the-appengine-environments).
- Processor ([prod](https://github.com/web-platform-tests/wpt.fyi/blob/81b71a1840ce16d2d5cc23cef48b53edb9d53a6c/results-processor/app.yaml#L2-L3))([staging](https://github.com/web-platform-tests/wpt.fyi/blob/81b71a1840ce16d2d5cc23cef48b53edb9d53a6c/results-processor/app.staging.yaml#L6-L7))
- Search cache ([prod](https://github.com/web-platform-tests/wpt.fyi/blob/81b71a1840ce16d2d5cc23cef48b53edb9d53a6c/api/query/cache/service/app.yaml#L2-L3))([staging](https://github.com/web-platform-tests/wpt.fyi/blob/main/api/query/cache/service/app.staging.yaml))

# Proposal
**Move the webapp service to use App Engine's flexible runtime**

Steps:
- Make a new Dockerfile. Can be similar to Search Cache's [Dockerfile](https://github.com/web-platform-tests/wpt.fyi/blob/main/api/query/cache/service/Dockerfile).
  - We cannot reuse the same dockerfile because of this issue: https://issuetracker.google.com/issues/286782972
  -  Can most likely just call [make go_build](https://github.com/web-platform-tests/wpt.fyi/blob/81b71a1840ce16d2d5cc23cef48b53edb9d53a6c/Makefile#L55C1-L61) 
- Configure the staging and prod yamls to use the flexible runtime
- Take a look at this [document](https://cloud.google.com/appengine/docs/flexible/migrating) for any gotchas
- Take a look at this [document](https://cloud.google.com/appengine/docs/flexible/custom-runtimes/build)

# Other added benefits
- We can debug the Dockerfile being built in [GCP Cloud Build](https://pantheon.corp.google.com/cloud-build/builds?referrer=search&project=wptdashboard)
  - Example Search Cache [build](https://pantheon.corp.google.com/cloud-build/builds;region=global/ddb0e71a-24b4-4026-a2c2-8ef665b908cc?project=wptdashboard)
[...]

PR #4226, opened 2025-02-03
Use the flexible runtime for the webapp

past:
Implements the proposal in #4219. I tested the new Dockerfile locally running the following from the top-level directory:
```
docker build -f webapp/web/Dockerfile --build-arg=LOCAL_SRC=. -t webapp .
```


jcscottiii:
I see that searchcache does this interesting thing where it creates a temporary wpt.fyi directory beforehand

https://github.com/web-platform-tests/wpt.fyi/blob/81b71a1840ce16d2d5cc23cef48b53edb9d53a6c/Makefile#L210-L217

We should modify the conditional to include webapp too (since we are copying how the searchcache image is built). That should fix the error of where cloud builder can't find wpt.fyi

Long term we can investigate if we even need that temporary directory.

Also, something to consider for a future PR. Change the final image type from gcr.io/distroless/static-debian11 to gcr.io/distroless/static-debian12 since debian11 is [superseded](https://www.debian.org/releases/bullseye/#:~:text=Debian%2011%20has%20been%20superseded%20by%20Debian%2012%20(bookworm).) by debian12. This change would go in both the searchcache and webapp images. But to reduce the amount of variables in this PR, feel free to do it later.

jcscottiii:
I took a look. The package.json is missing because of the .gcloudignore. (it wasn't found in the [manifest.json](https://pantheon.corp.google.com/storage/browser/_details/staging.wptdashboard-staging.appspot.com/ae/5b564127-80aa-4db9-9498-9401e35faf0b/manifest.json;tab=live_object?project=wptdashboard-staging))

https://github.com/web-platform-tests/wpt.fyi/blob/922416d0b85d4861ef5cb6f5e143efea2589442a/webapp/web/.gcloudignore#L12-L17

We can probably use the simpler one used  for searchcache since the docker file does the multi-stage compilation into a single binary

https://github.com/web-platform-tests/wpt.fyi/blob/922416d0b85d4861ef5cb6f5e143efea2589442a/api/query/cache/service/.gcloudignore#L1-L6

webapp/web/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+FROM golang:1.22.11-bookworm as builder
+
+RUN apt-get update
+RUN apt-get install -qy --no-install-suggests git sudo
+WORKDIR /go/src/wpt.fyi
+
+# Default to the existing path when running inside the development docker container.
+ARG LOCAL_SRC=wpt.fyi
+COPY $LOCAL_SRC .
+RUN make go_build
```
jcscottiii, line 12:
```suggestion
RUN CGO_ENABLED=0  make go_build
```

This will create a statically linked binary.

past: It's weird that this causes a "file not found" error and not a runtime linker error, but hey, it works.

past:
Deployed successfully, but the deployed version doesn't appear to be working yet. More investigation tomorrow.

past:
I think I need to copy some more files, like node modules, to the final container.

jcscottiii:
One time instructions:

```sh
docker network create wptfyi-network
docker pull google/cloud-sdk:latest
docker run -d \
  --name datastore-emulator \
  --network wptfyi-network \
  -p 8081:8081 \
  google/cloud-sdk:latest \
  gcloud beta emulators datastore start --project=local --host-port 0.0.0.0:8081 --no-store-on-disk
```

Iterative instructions:

```sh
make package_service PROJECT=wptdashboard-staging APP_PATH=webapp/web/app.staging.yaml
pushd webapp/web
docker build -t wpt-web-app-img .
docker rm -f  wpt-web-app; docker run \
  --name wpt-web-app \
  --network wptfyi-network \
  -p 8082:8080 \
  -e DATASTORE_EMULATOR_HOST=datastore-emulator:8081 \
  -e DATASTORE_PROJECT_ID=local \
  -e GAE_DEPLOYMENT_ID=local \
  -i wpt-web-app-img
# ctrl-c to stop then type popd to go back and start the build again.
popd
```
Open the browser to http://localhost:8082. Even though it says 8080.

If you want to inspect the files, it becomes harder since we are using a distro less image. You could temporarily swap to a debug image.  More about it [here](https://iximiuz.com/en/posts/docker-debug-slim-containers/) in the `Temporary` section.

```diff
-FROM gcr.io/distroless/static-debian12
+FROM gcr.io/distroless/static-debian12:debug
```


Clean up instructions (when finally done):
```sh
docker rm -f  wpt-web-app datastore-emulator
docker network  rm  wptfyi-network
```

jcscottiii:
It is up now! But I would click around to make sure nothing else is broken.

past:
I think this is working as intended now both locally and in the staging deployment. Besides the TODO for nicehttp, I have another TODO for figuring out what to do for the VPC connector. It doesn't seem to hurt keeping it in for now, any thoughts on that?

jcscottiii:
@past I think it is fine to keep those VPC connections for now. We can always clean up in the future.

PR #4229, opened 2025-02-06
Turn on navigation to Interop 2025

DanielRyanSmith:
Sets Interop 2025 as the default Interop year, and allows navigation to the 2025 dashboard.


webapp/components/interop-data.js
```diff
@@ -1537,6 +1537,7 @@ export const interopData = {
      * https://github.com/web-platform-tests/results-analysis
     **/
     'csv_url': 'https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/interop-2025/interop-2025-{stable|experimental}-v2.csv',
+    'mobile_csv_url': 'https://api.github.com/repos/jgraham/interop-results/contents/2025/latest/aligned/mobile-{stable|experimental}-current.csv?ref=main',
```
jcscottiii, line 1540: I think you want the raw.githubusercontent.com URL for the mobile url. Also, there are no stable results for now. I only see experimental results for stable.

https://github.com/jgraham/interop-results/tree/4cb9d304ea32acd6c2d35a86ed486b8b0f0a02e7/2025/latest/aligned

webapp/static/interop-data.json
```diff
@@ -1506,6 +1506,7 @@
       "interop-2025-webrtc"
     ],
     "csv_url": "https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/interop-2025/interop-2025-{stable|experimental}-v2.csv",
+    "mobile_csv_url": "https://api.github.com/repos/jgraham/interop-results/contents/2025/latest/aligned/mobile-{stable|experimental}-current.csv?ref=main",
```
jcscottiii, line 1509: Same for this.

jcscottiii (commented):
One question about the mobile_csv_url in the two files.

webapp/components/interop-data.js
DanielRyanSmith: This is the URL used for the mobile results for 2024 as well, but we handle the mobile url differently to obtain the raw results view:

https://github.com/web-platform-tests/wpt.fyi/blob/17237071421b7951932448f1e2f2a261233ffc0f/webapp/components/interop-data-manager.js#L389-L396

There may be a better way of doing this, but this is what I found that worked 😅 

Also, it is true that there are not stable results, but the link configuration is set up so that we can obtain stable results if they become available as scores.

jcscottiii: Actually given the fact we will get stable results sometime this year, ignore my initial concern. My follow up question: If someone selects stable, what will happen given that no stable results exist yet?


Also, thanks for the code. That helped. I ran it through a [REPL](https://jsfiddle.net/#&togetherjs=59EQGRfdUO) with the current JSON payload and it makes sense.

jcscottiii (approved):

PR #4234, opened 2025-02-06
Add specifications for Interop 2025

DanielRyanSmith:
This PR adds all required changes for Interop 2025, but does not make the 2025 dashboard visible client-side.

### DIFF
--- a/webapp/components/interop-summary.js
+++ b/webapp/components/interop-summary.js
@@ -201,6 +201,11 @@ class InteropSummary extends PolymerElement {
       const interopDiv = this.shadowRoot.querySelector('#interopSummary');
       interopDiv.style.display = 'none';
       summaryDiv.style.minHeight = '275px';
+    // This min-height is added to this section to ensure that the section below
+    // is not moved after the user selects between STABLE and EXPERIMENTAL
+    // (experimental browser names are longer and add additional lines).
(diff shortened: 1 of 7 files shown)


webapp/components/interop-summary.js
```diff
@@ -201,8 +201,8 @@ class InteropSummary extends PolymerElement {
       const interopDiv = this.shadowRoot.querySelector('#interopSummary');
       interopDiv.style.display = 'none';
       summaryDiv.style.minHeight = '275px';
-    } else if (this.year === '2024') {
-      summaryDiv.style.minHeight = '350px';
+    } else if (parseInt(this.year) >= 2024) {
```
jcscottiii, line 204: I wonder if we should add a new field to the data called `current_years` or `live_years` Because I imagine eventually we will want to hide 2024 but keep 2025. Instead of changing 1) the logic here and  2) the data (which we were already going to change), we could just change the data.

jcscottiii (approved):
LGTM with one consideration

webapp/components/interop-summary.js
DanielRyanSmith: Actually, I removed the TODO and added a proposed fix - let me know how you feel about it. My initial thought was to add these constants to interop-data, but I think it's better to save that file for data actually related to Interop focus areas.

jcscottiii (approved):

PR #4256, opened 2025-02-12
Increase the min number of instances and the cool down period

past:
The current webapp VM configuration is:
```
automatic_scaling:
  cool_down_period: 120s
  min_num_instances: 2
  max_num_instances: 20
  cpu_utilization:
    target_utilization: 0.5
```

This change removes the `instance_class` that is only applicable to the standard environment (c.f. [flex docs](https://cloud.google.com/appengine/docs/flexible/reference/app-yaml?tab=go)), doubles the minimum number of instances and increases the cool down period to avoid health check errors during instance initialization.

### DIFF
--- a/webapp/web/app.staging.yaml
+++ b/webapp/web/app.staging.yaml
@@ -4,7 +4,11 @@
 
 runtime: custom
 env: flex
-instance_class: F4_1G
+
+automatic_scaling:
+  min_num_instances: 4
+  max_num_instances: 20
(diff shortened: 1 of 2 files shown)


past:
The GCP [recommendation](https://cloud.google.com/appengine/docs/flexible/known-issues#overprovisioning) is to overprovision by 50%, but I doubled it just to be safe.

DanielRyanSmith (approved):

PR #4264, opened 2025-02-13
Adjust more timeouts

jcscottiii:
This change disables most NGINX timeout checks (e.g., client body, header, read timeouts) and sets the keepalive timeout to a high value. The goal is to let the Go application handle timeout configurations and prevent resource exhaustion on the NGINX side. This approach is temporary until the Go application can be migrated away from `nicehttp`.

Why this change:
- Previously, some requests were failing at 60 seconds, likely due to the NGINX timeout settings and the limited capacity of the smaller instances. This change disables or increases various NGINX timeouts (e.g., client body, header, read timeouts) and sets a high keepalive timeout to allow the Go application to handle timeout logic. I have not seen any requests taking longer than 60 seconds today after yesterday's changes. So this is a proactive change.
- This should prevent requests from failing with the larger instances, even under increased demand. It might also allow us to reduce the instance size slightly in the future.

### DIFF
--- a/webapp/web/nginx.conf
+++ b/webapp/web/nginx.conf
@@ -73,6 +73,23 @@ http {
             proxy_set_header X-Forwarded-Proto $scheme;
             # Disable checking of the body size
             client_max_body_size 0;
+            # Disable timeout checks. Allow the application to determine the appropriate timeouts
+            client_body_timeout 0;
+            client_header_timeout 0;
+            send_timeout 0;
+            proxy_read_timeout 0;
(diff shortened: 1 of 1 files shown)


jcscottiii:
Turns out these aren't working as expected.

jcscottiii:
@past @DanielRyanSmith sorry for all the notifications. Now this is ready again.

DanielRyanSmith (commented):
Thanks for assisting with this. 🙂

DanielRyanSmith (approved):

PR #4266, opened 2025-02-14
Fix the info about investigation areas

stof:
## Description
For the dashboard to work, names need to be the same between the table sections and the investigation scores.
For now, all investigation areas have `#` as link on the dashboard because the corresponding info is not found (they would also fail to display a score, even though this is not yet different as no investigation area has a score registered yet)

## Changes
I updated the `investigation_scores` section to use the same names that table sections.

## Requirements
n/a

### DIFF
--- a/webapp/components/interop-data.js
+++ b/webapp/components/interop-data.js
@@ -1447,22 +1447,22 @@ export const interopData = {
     ],
     'investigation_scores': [
       {
-        'name': 'Accessibility Testing',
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Thanks for catching this! I'm handling a new deployment now so I'll get this change in as well. 🙂

stof:
sorry for missing the mismatch between `Gaming` and `Gamepad`. Good catch.

PR #4289, opened 2025-03-03
Bump flags 0.2.2 to flg 3.0.0

past:
A month ago flags [migrated](https://github.com/dpup/node-flags?tab=readme-ov-file#package-name-update) to flg, and a couple weeks ago @vercel/flags [became](https://vercel.com/changelog/npm-i-flags) flags. This was enough to confuse dependabot into generating #4287, but this PR implements the proper migration.

### DIFF
--- a/webdriver/package-lock.json
+++ b/webdriver/package-lock.json
@@ -4,15 +4,15 @@
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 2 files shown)


KyleJu (approved):
Great catch!!

PR #4294, opened 2025-03-04
Add & use isTestMissing helper when calculating passes in view=test

### LINKED ISSUE #4293: [bug] test=view should ignore missing tests
There is an edge case where we need to consider missing tests.

After #4290 , I saw that if you go to  this [page](https://wpt.fyi/results/css/css-anchor-position?label=master&aligned&view=test&q=feature%3Aanchor-positioning%20%21is%3Atentative)

You'll see this:

Image

But if you go to the [test](https://wpt.fyi/results/css/css-anchor-position/anchor-position-005-print.html?label=master&aligned&view=test&q=feature%3Aanchor-positioning%20%21is%3Atentative) drill down, the test is missing on Safari.

Image

From the endpoint, you'll see that Safari is indeed missing among the four browsers in the payload:

Image

jcscottiii:
Moves isMissing logic to a common method. It will continue to be used for calculating whether we have a single subtest. Additionally, it will be used with the new logic for view=test. Without this check, we will add missing tests as passing tests.

More details in #4293 

Fixes #4293

### DIFF
--- a/webapp/components/test/wpt-results.html
+++ b/webapp/components/test/wpt-results.html
@@ -383,6 +383,17 @@ suite('<wpt-results>', () => {
       const formatted = trf.formatCellDisplay(passes, total, status, isDir);
       expect(formatted).to.equal('PASS');
     });
+
+    test('should display Missing for missing test (test view, file)', () => {
+      const passes = 0;
+      const total = 0;
+      const status = '';
+      const isDir = false;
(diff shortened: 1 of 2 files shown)


jcscottiii:
Before:
image


After:
image

You can see that Safari now reflects the correct count.

KyleJu (approved):

PR #4295, opened 2025-03-04
Use the same number of processor instances in staging as in prod

past:
We often get alerts due to the results-arrival queue depth exceeding the alert threshold for a long time. Given that task queue configuration is the same across staging and production, let's try to use more CPU for staging to avoid that problem.

### DIFF
--- a/results-processor/app.staging.yaml
+++ b/results-processor/app.staging.yaml
@@ -7,7 +7,7 @@ runtime: custom
 env: flex
 
 manual_scaling:
-  instances: 3
+  instances: 5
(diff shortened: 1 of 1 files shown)


past:
@KyleJu was there any reason apart from cost to have a stricter limit for the staging environment?

KyleJu:
> @KyleJu was there any reason apart from cost to have a stricter limit for the staging environment?

Nope, it was due to the cost of running wpt.fyi at the time

KyleJu (approved):

PR #4321, opened 2025-03-11
Actually fetch favicon

jonathan-j-lee:
Follow-up of #416, which doesn't actually add the favicon to the page source. This still appears to be the case on the live site.


KyleJu (approved):

KyleJu:
Thank you @jonathan-j-lee!

PR #4340, opened 2025-03-26
golangci-lint migration

### LINKED ISSUE #4339: golangci-lint migration
golangci-lint is failing with the following errors:
```
Error: can't load config: unsupported version of the configuration: "" See https://golangci-lint.run/product/migration-guide for migration instructions
```

KyleJu:
Fix #4339. Migrate [go-lint](https://golangci-lint.run/product/migration-guide/) and fix new lint errors:
- Changes in .golangci.yaml are made through `golangci-lint migrate`

For new lint errors:
-  Remove embedded field 
- Lower cap the first word of error messages
- Use switch statement in webhook.go and string replacement suggestions in compile.go
- For api/query/cache/lru/lru.go, add the correct parenthesis to compare time

### DIFF
--- a/api/checks/webhook.go
+++ b/api/checks/webhook.go
@@ -18,13 +18,13 @@ import (
 const requestedAction = "requested"
 const rerequestedAction = "rerequested"
 
-// WebhookGithubEvent represents the allowed GitHub webhook event types.
-type WebhookGithubEvent string
+// webhookGithubEvent represents the allowed GitHub webhook event types.
+type webhookGithubEvent string
 
 const (
-	EventCheckSuite  WebhookGithubEvent = "check_suite"
-	EventCheckRun    WebhookGithubEvent = "check_run"
-	EventPullRequest WebhookGithubEvent = "pull_request"
+	eventCheckSuite  webhookGithubEvent = "check_suite"
+	eventCheckRun    webhookGithubEvent = "check_run"
+	eventPullRequest webhookGithubEvent = "pull_request"
 )
 
 var runNameRegex = regexp.MustCompile(`^(?:(?:staging\.)?wpt\.fyi - )(.*)$`)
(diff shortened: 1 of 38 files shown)


KyleJu:
@jcscottiii a similar change to https://github.com/GoogleChrome/webstatus.dev/pull/1317 yesterday

api/checks/webhook.go
```diff
@@ -71,6 +71,7 @@ func checkWebhookHandler(w http.ResponseWriter, r *http.Request) {
 
 	var processed bool
 	api := NewAPI(ctx)
+	// nolint:staticcheck
```
jcscottiii, line 74: We should avoid using this nolint. If I remember correctly, this check will force us to use switch-case which then allows us to use the exhaustive linter and have exhaustive switch-case statements.

Could have something like this:
```diff
+// WebhookGithubEvent represents the allowed GitHub webhook event types.
+type WebhookGithubEvent string
+
+const (
+        EventCheckSuite  WebhookGithubEvent = "check_suite"
+        EventCheckRun    WebhookGithubEvent = "check_run"
+        EventPullRequest WebhookGithubEvent = "pull_request"
+)
+
func checkWebhookHandler(w http.ResponseWriter, r *http.Request) {
....
...
...
...
-        event := r.Header.Get("X-GitHub-Event")
+       inputEvent := WebhookGithubEvent(event)
-        switch event {
-        case "check_suite", "check_run", "pull_request":
+        switch inputEvent {
+        case EventCheckSuite, EventCheckRun, EventPullRequest:
            break
         default:
....
....
...
...
..
-        // nolint:staticcheck
-        if inputEvent == EventCheckSuite {
+
+        switch inputEvent {
+        case EventCheckSuite:
                 processed, err = handleCheckSuiteEvent(api, payload)
-        } else if inputEvent == EventCheckRun {
+        case EventCheckRun:
                 processed, err = handleCheckRunEvent(api, payload)
-        } else if inputEvent == EventPullRequest {
+        case EventPullRequest:
                 processed, err = handlePullRequestEvent(api, payload)
         }
```

And we don't need a default case for the second switch-case since you already filtered them out

jcscottiii (commented):
Two comments

api/checks/webhook.go
KyleJu: Thanks for the suggestions! I wanted to address all the no-lint in a separate PR and keep this PR a low-risk change. But it is better not to introduce no-lint at all. I lower-cased all the constants to keep them private instead.

jcscottiii (approved):

KyleJu:
since [9ee8e3b](https://github.com/web-platform-tests/wpt.fyi/pull/4340/commits/9ee8e3b986d78c2775e732552263bc6d91fead35)'s staging is deployed successfully, I will merge the PR

PR #4345, opened 2025-04-01
Bump taskcluster to v83

jrobbins:
## Description
Updating a golang dependency as part of team rotation duties.

This one seems pretty simple to me.  Hopefully, I am not missing something.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/web-platform-tests/wpt.fyi
 
-go 1.23.6
+go 1.23.7
 
 require (
 	cloud.google.com/go/cloudtasks v1.13.5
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/web-platform-tests/wpt.fyi
 
-go 1.23.6
+go 1.23.7
```
KyleJu, line 3: All the go version in Dockerfile need to be bumped as well, e.g. webapp/web/Dockerfile

KyleJu: For future reference, there is a Golang upgrade guide here, https://github.com/web-platform-tests/wpt.fyi/blob/main/docs/upgrading-go.md

jrobbins: > All the go version in Dockerfile need to be bumped as well, e.g. webapp/web/Dockerfile

OK, I think I got the top-level Dockerfile.  And, the search one had already been updated.

KyleJu (approved):

PR #4360, opened 2025-04-15
Update README for feature search with new links

DanielRyanSmith:
This wording might need changed. :sweat_smile:

### DIFF
--- a/api/query/README.md
+++ b/api/query/README.md
@@ -362,12 +362,13 @@ feature label, regardless of browsers.
 
     {"feature": [web-feature-name]}
 
-Where web-feature-name is a string, case-insensitive and matches the filename base
-for any of the .yml files in the
-[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory.
+Where web-feature-name is a string, case-insensitive and matches the feature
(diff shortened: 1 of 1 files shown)


api/query/README.md
```diff
@@ -362,12 +362,13 @@ feature label, regardless of browsers.
 
     {"feature": [web-feature-name]}
 
-Where web-feature-name is a string, case-insensitive and matches the filename base
-for any of the .yml files in the
-[feature-group-definitions](https://github.com/web-platform-dx/web-features/tree/main/feature-group-definitions) directory.
```
jcscottiii, line 367: Thoughts on changing the old link to https://github.com/web-platform-dx/web-features/tree/main/features [1] ?

I wonder if we should point there instead because the WEB_FEATURES.yml files are scattered throughout the repository. Whereas [1] shows all the names in a list. 

But can be convinced to keep what you have.

DanielRyanSmith: Yes, I think that is fine. I was operating under the assumption that it might be worthwhile to point users where this mapping is coming from, but it's easier to see what is searchable by looking at the features directory. 🙂

DanielRyanSmith:
It looks like my branch was not up-to-date anyway and the URL has already been changed. this can simply be closed!

PR #4398, opened 2025-05-15
Improve support for parsing WebKit Nightly versions and SHAs

clopez:
webkitgtk and wpewebkit nightly versions have the following format: `wpewebkit-2.49.1 (294826@main)` where `2.49.1 (294826@main)` is the version.
This is the standard naming on the WebKit repository where each commit on the main WebKit branch has a unique identifier of `number@main`.

However, the current version/sha parser expects the WPT Git SHA identifier to be passed after an `@`.

This patch improves the parser to ignore the first `@main` in the identifier string, allowing it to get the right version and also the optional SHA.

### DIFF
--- a/shared/params.go
+++ b/shared/params.go
@@ -77,13 +77,43 @@ func ParseProductSpecs(specs ...string) (products ProductSpecs, err error) {
 	return products, nil
 }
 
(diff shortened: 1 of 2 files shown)


DanielRyanSmith (approved):
Thanks for taking care of this!

PR #4523, opened 2025-08-26
Update the score for the mobile testing investigation area

stof:
## Description
Update the score for the mobile testing investigation area to 12%

## Review Information
See https://github.com/web-platform-tests/interop-mobile-testing/issues/36 for the meeting notes where the score was computed to 12% in June.


jcscottiii (approved):
Thanks for this!

PR #4527, opened 2025-08-29
Delete old versions after deployment

jstenback:
This changes the order in which deletion of old versions of wpt.fyi is deleted from the infrastructure. Prior to this change the deletion happened at the beginning of the deployment. The deletion itself takes many minutes per service, with prompts inbetween, so the time from start of deplyment to the new image actually being available in the infrastructure was longer than it needed to be. With this change, the deployment starts right away, and once the deployment is done, we ask whether old versions should be deleted and do those deletions w/o prompting per service, meaning the person doing the deployment doesn't need to sit and wait etc.

### DIFF
--- a/util/deploy-production.sh
+++ b/util/deploy-production.sh
@@ -17,13 +17,6 @@ usage() {
   echo "${USAGE}"
 }
 
-# Deletes the service passed as a parameter.
-delete_oldest_version() {
-  OLDEST_REV=$(gcloud app --project=wptdashboard versions list --sort-by=last_deployed_time --filter="service=$1" --limit=1 --format=json | jq -r '.[] | .id')
-  echo "Deleting $1 service version $OLDEST_REV"
-  gcloud app --project=wptdashboard versions delete --service=$SERVICE ${QUIET:+--quiet} $OLDEST_REV
-}
-
 while getopts ':bfqh' flag; do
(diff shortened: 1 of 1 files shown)


util/deploy-production.sh
```diff
@@ -143,3 +136,32 @@ fi
...
+
+  if [[ $DELETE == "y" ]]; then
+    echo "Found $VERSIONS for the default service, deleting the oldest version of all services."
+
+    OLDEST_REV=$(gcloud app --project=wptdashboard versions list --sort-by=last_deployed_time --filter="service=$SERVICE" --limit=1 --format=json | jq -r '.[] | .id')
+    for SERVICE in $SERVICES; do
+      echo "Deleting $SERVICE service version $OLDEST_REV"
+      gcloud app --project=wptdashboard versions delete --service=$SERVICE --quiet $OLDEST_REV
+    done
+  elif (($VERSIONS > 3)); then
```
jcscottiii, line 163: Currently, this branch is nested inside the `VERSIONS == 3` branch. This should be unnested to be at the same level:

```
if (($VERSIONS == 3)); then
  ..
elif (($VERSIONS > 3)); then
..
fi
```

jcscottiii (approved):
LGTM with one consideration.

PR #4528, opened 2025-08-29
Adjust log level for failing to cache empty search results

jstenback:
Changes the log severity for empty search results from ERROR to INFO.

Previously, a POST request to /api/search that yielded no results would log an error. This caused the Google App Engine request log to be marked with ERROR severity, even for successful HTTP 200 responses.

This change lowers the log level to INFO to avoid treating empty search results as an error condition, ensuring that request logs accurately reflect the status of the operation.


marco0204-create (commented):
APPROVE

PR #4728, opened 2026-01-31
Allow root-level parenthetical expressions in query grammar

gsnedders:
Adds support for wrapping root-level queries in parentheses (e.g., `(all(...))`  or `(exists(...) and path:/foo)`), making the grammar more flexible and allowing better grouping for complex queries.

This makes complex and/or combinations possible that weren't before, closing a gap in what the query language can express compared with structured queries.

This builds on top of #4727.


gsnedders:
It may well make the most sense to just land both of these in #4730 at this point?

PR #4740, opened 2026-02-06
Use 2025 label for 2026 anchor positioning

DanielRyanSmith:
Per discussion, no new tests are being added to the anchor positioning focus area, and the 2025 label will simply be reused. This means that the 2026 label is being removed from all tests, and we should just use the 2025 label in links to tests to represent the focus area.


DanielRyanSmith:
CC @nairnandu

jcscottiii (approved):

PR #4763, opened 2026-02-18
Correcting the hyperlink for view transitions to point at the root instead of the CSS-view-transitions folder

nairnandu:
Correcting the hyperlink for view transitions to point at the root instead of the CSS-view-transitions folder


DanielRyanSmith:
There are some unrelated errors in the CI that are affecting this PR - I'm working on the fix now. 🙂

DanielRyanSmith (approved):

PR #4764, opened 2026-02-18
Address golangci-lint issues

DanielRyanSmith:
The most recent update to the linter surfaced a bunch of new issues and is blocking the CI. Gemini seems to think that the errors are not a risk and should be ignored by the linter.

This change contains two possible fixes for the linting errors. We either ignore the `gosec` rules entirely in `.golangci.yaml`, or we add an explanation for every error and add it to the lint ignore.

Gemini also suggests a method to check if a URL is safe (see below). I am indifferent to the solution.

---

#### Can we "Fix" the Root Cause?
- For SSRF (G704):
  The "root cause" fix is to validate the URL before using it. We should ensure the URL:
   1. Is a valid URL.
   2. Points to a trusted domain (e.g., *.storage.googleapis.com).
   3. Does not point to internal IP ranges (like 127.0.0.1 or the metadata server 169.254.169.254).

We could implement a helper function like shared.SafeDo(client, req) or shared.ValidateURL(url) and use it everywhere. This would be a "true" fix. If we add a validation check, `gosec` might still complain (it's not that smart), but the code will be objectively safer.

- For XSS (G705):
  This one is trickier because it's a false positive inherent to how gosec works. gosec doesn't understand that you've set an application/json header. It only sees bytes being written to
  a stream.
   * The "fix" here is exactly what I did: set the header and use json.Marshal.
   * Since the linter still complains, the "cleanest" engineering practice is actually to use targeted inline ignores (the // nolint:gosec comments) rather than a global one, because it forces you to acknowledge that this specific line is safe.

### DIFF
--- a/api/checks/jwt.go
+++ b/api/checks/jwt.go
@@ -66,7 +66,7 @@ func getJWTClient(ctx context.Context, appID, installation int64) (*http.Client,
 	}
 	// tokenResponse is the JSON response body.
 	var tokenResponse struct {
-		AccessToken string    `json:"token"`
+		AccessToken string    `json:"token"` // nolint:gosec
 		ExpiresAt   time.Time `json:"expires_at"`
 	}
 	if err := json.Unmarshal(body, &tokenResponse); err != nil {


DanielRyanSmith:
> Will this fix solve some of these problems? https://github.com/securego/gosec/pull/1506

I believe it will. Perhaps we should ignore these rules in the linter until a new update lands?

PR #4831, opened 2026-04-01
Remove old feature flags

gsnedders:
This removes a lot of old feature flags — all of these have been enabled for many years and the code hasn't been changed in years.

My underlying motivation for this was https://github.com/web-platform-tests/wpt.fyi/pull/3577#issuecomment-3801912731 where I couldn't figure out why the tests were failing for years, because we were testing in a different configuration that there's no reason for it to exist.


DanielRyanSmith (approved):

DanielRyanSmith:
Thanks for the cleanup!

PR #4906, opened 2026-05-13
Move BSF graph to Insights page

yulunwu2:
Relocate the Browser-Specific Failures graph from the homepage to the Insights page alongside anomalies and flakes.

### DIFF
--- a/webapp/views/wpt-app.js
+++ b/webapp/views/wpt-app.js
@@ -6,7 +6,6 @@ import '../components/wpt-flags.js';
 import { WPTFlags } from '../components/wpt-flags.js';
 import '../components/wpt-header.js';
 import '../components/wpt-permalinks.js';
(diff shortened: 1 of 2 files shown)


webapp/views/wpt-app.js
```diff
@@ -110,17 +109,9 @@ class WPTApp extends PathInfo(WPTFlags(TestRunsUIBase)) {
       <div class="separator"></div>
 
       <template is="dom-if" if="[[showBSFGraph]]">
```
jgraham: Can't we just remove this whole section for now?

jgraham (approved):
This seems like an improvement to me, although I don't think we a placeholder element on the main page.

webapp/views/wpt-app.js
yulunwu2: Done, thanks for the review!

PR #4921, opened 2026-05-21
Add UI toggle for /third_party tests in BSF chart

yulunwu2:
Adds an `includeThirdParty` query parameter to `/api/bsf` (default false)
that selects between the existing CSV (which now excludes `/third_party`
tests) and the new `-with-third-party.csv` variant. The wpt-bsf component
exposes this as a "Test scope" toggle next to the existing channel toggle,
defaulting to exclude third-party tests by default to prioritize tests
written for WPT and following the WPT subtest convention.

Also hardens fetchBSF.Fetch against malformed-but-successful responses
(empty CSV, missing "date" column), which previously slipped past the
status-code check and surfaced on the frontend as a silent blank chart.

The companion change in results-analysis must land and publish both CSVs
to gh-pages before this is deployed.

### DIFF
--- a/api/bsf_handler.go
+++ b/api/bsf_handler.go
@@ -54,7 +54,13 @@ func (b BSFHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		isExperimental = *val
 	}
 
-	lines, err := b.fetcher.Fetch(isExperimental)
+	includeThirdParty := false
+	val, _ = shared.ParseBooleanParam(q, "includeThirdParty")
+	if val != nil {
+		includeThirdParty = *val
+	}
+
+	lines, err := b.fetcher.Fetch(isExperimental, includeThirdParty)
 	if err != nil {
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 
--- a/shared/fetch_bsf.go
+++ b/shared/fetch_bsf.go
@@ -15,11 +15,17 @@ import (
 
 const (
 	// experimentalBSFURL is the GitHub URL for fetching the experimental BSF data
-	// for Chrome, Firefox and Safari.
+	// for Chrome, Firefox and Safari. First party tests only.
 	experimentalBSFURL = "https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/experimental-browser-specific-failures.csv"
 	// stableBSFURL is the GitHub URL for fetching the stable BSF data
-	// for Chrome, Firefox and Safari.
+	// for Chrome, Firefox and Safari. First party tests only.
 	stableBSFURL = "https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/stable-browser-specific-failures.csv"
+	// experimentalBSFWithThirdPartyURL is the GitHub URL for fetching the experimental BSF data
+	// for Chrome, Firefox and Safari. All tests (first and third party) included.
+	experimentalBSFWithThirdPartyURL = "https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/experimental-browser-specific-failures-with-third-party.csv"
+	// stableBSFWithThirdPartyURL is the GitHub URL for fetching the stable BSF data
+	// for Chrome, Firefox and Safari. All tests (first and third party) included.
+	stableBSFWithThirdPartyURL = "https://raw.githubusercontent.com/web-platform-tests/results-analysis/gh-pages/data/stable-browser-specific-failures-with-third-party.csv"
 )
 
 // BSFData stores BSF data of the latest WPT revision.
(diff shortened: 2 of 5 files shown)


yulunwu2:
Screenshot 2026-05-21 at 2 53 32 PM

shared/fetch_bsf.go
```diff
@@ -90,20 +96,28 @@ func FilterandExtractBSFData(rawBSFdata [][]string, from *time.Time, to *time.Ti
...
 	} else {
-		url = stableBSFURL
+		if excludeThirdParty {
+			url = stableBSFURL
+		} else {
+			url = stableBSFWithThirdPartyURL
+		}
 	}
 
 	resp, err := http.Get(url)
```
DanielRyanSmith, line 123: We should add some status code checks for the response now that we're making changes here (this should have already been in place 😅, but since we're updating the CSV generation, it's more important). If we fail to fetch the CSV files with the new changes, right now we'll return a malformed response. We should handle that more gracefully so it doesn't cause frontend problems.

api/bsf_handler.go
```diff
@@ -54,7 +54,13 @@ func (b BSFHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		isExperimental = *val
 	}
 
-	lines, err := b.fetcher.Fetch(isExperimental)
+	excludeThirdParty := true
+	val, _ = shared.ParseBooleanParam(q, "excludeThirdParty")
+	if val != nil {
+		excludeThirdParty = *val
+	}
```
DanielRyanSmith:
```suggestion
	excludeThirdParty := true
    if val, err := shared.ParseBooleanParam(q, "excludeThirdParty"); err == nil && val != nil {
        excludeThirdParty = *val
    }
```

This would make it so the flag defaults to `true` when parsing fails, rather than `false`.

captainbrosset (approved):

foolip:
@yulunwu2 it looks like the text is still just a little bit taller than the graph, could you try combining the updated line into just "Last update: 012345678abcdef"? Or if it's already available in the script, the date of that commit, and show the date but link to the commit. But not growing the height of the component is the most important part.

yulunwu2:
Screenshot 2026-05-26 at 9 28 27 AM

yulunwu2:
Screenshot 2026-05-26 at 9 28 50 AM

yulunwu2:
Thanks for the review @foolip! Could I get another review for this PR? @jgraham as well for approval.

foolip (approved):
Looks like this will work, but best if @DanielRyanSmith reviews too.

DanielRyanSmith (commented):
Some small changes that might be useful.

DanielRyanSmith (approved):

util/commands.sh
```diff
@@ -7,7 +7,7 @@ WPTD_HOST_GCD_PORT=${WPTD_HOST_GCD_PORT:-"8001"}
 WPTD_PATH="$(git rev-parse --show-toplevel)"
 
 function wptd_chown() {
-  docker exec -u 0:0 "${DOCKER_INSTANCE}" chown -R $(id -u $USER):$(id -g $USER) $1
+  docker exec -u 0:0 "${DOCKER_INSTANCE}" chown -R $(id -u $USER):$(id -g $USER) $1 || true
```
DanielRyanSmith: This seems unrelated and should be removed.

webapp/package-lock.json
```diff
@@ -9499,6 +9499,16 @@
...
+    "node_modules/mocha/node_modules/diff": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/diff/-/diff-3.2.0.tgz",
+      "integrity": "sha512-597ykPFhtJYaXqPq6fF7Vl1fXTKgPdLOntyxpmdzUOKiYGqK7zcnbplj5088+8qJnWdzXhyeau5iVr8HVo9dgg==",
+      "dev": true,
+      "license": "BSD-3-Clause",
+      "engines": {
+        "node": ">=0.3.1"
+      }
+    },
```
DanielRyanSmith: The package-lock updates should be reverted, as they're not related to this change.

api/bsf_handler.go
```diff
@@ -54,7 +54,13 @@ func (b BSFHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		isExperimental = *val
 	}
 
-	lines, err := b.fetcher.Fetch(isExperimental)
+	includeThirdParty := false
+	val, _ = shared.ParseBooleanParam(q, "includeThirdParty")
+	if val != nil {
+		includeThirdParty = *val
+	}
```
DanielRyanSmith, line 61: This is now defaulting to false instead, which is fine if it's intended, but the PR description states otherwise.

DanielRyanSmith (changes requested):
This may be WIP, but I'll add some comment otherwise 🙂

jgraham (approved):
No objections in principle to this change from me (but again, with the proviso that I think BSF failed as a metric and the better solution is to remove it entirely).

captainbrosset:
@yulunwu2 I see the data change took effect at wpt.fyi already, but the UI toggle is not there yet.
Can we make sure this gets merged shortly to avoid displaying a partial graph without the corresponding UI filter?

I understand there are discussions about the BSF metric, and we should keep on working together on this, but as of now, the graph filters out an important part of the platform which many web developers need, without saying why.

api/bsf_handler.go
yulunwu2: updated commit message to "exclude third-party tests"

webapp/package-lock.json
yulunwu2: Fixed, sorry about that

util/commands.sh
yulunwu2: Fixed, sorry about that

yulunwu2:
@captainbrosset I don't have write permissions to the repo, I think someone else needs to land this.

captainbrosset:
@foolip  can you help ?

DanielRyanSmith (approved):

DanielRyanSmith:
This change will be deployed with this week's weekly deployment 🙂

PR #4948, opened 2026-06-15
Add `--force` and `--prod` flags to process_test_history.py

DanielRyanSmith:
## Overview
This PR adds `--force` and `--prod` CLI flags to `process_test_history.py` to support manual runs and production overrides.

## Root Cause / Motivation
- **Usability**: We need a way to force operations (like setting the start date) even if the Datastore is not empty, especially when recovering from partial runs.
- **Environments**: The script had hardcoded staging values. We need to be able to run it against production to catch up production history as well.

## Detailed Changelog
    *   Added `--force` CLI flag to bypass empty Datastore checks.
    *   Added `--prod` CLI flag to switch configuration to production (overriding GCS bucket, Google Cloud Project, and Runs API URL).
    *   Updated `set_history_start_date` and `main` checks to respect `--force`.


jcscottiii (approved):
LGTM. But I think we should start to minimize the number of per environment configs in code in future. And instead, provide them via environment variables. Then you would be able to get rid of the unit test just to check config variation.

- [1] https://www.12factor.net/config
- [2] https://docs.cloud.google.com/appengine/docs/standard/reference/app-yaml?tab=python#:~:text=Optional.%20Overrides%20the,alphanumeric%20or%20%22_%22).
- [3] https://github.com/search?q=repo%3AGoogleChrome%2Fwebstatus.dev+os.getenv&type=code

PR #4949, opened 2026-06-15
Implement deterministic keys for TestHistoryEntry

DanielRyanSmith:
## Overview
This PR implements deterministic key names for `TestHistoryEntry` to ensure database writes are idempotent.

## Root Cause / Motivation
Using random IDs for `TestHistoryEntry` means that re-running the script over already processed runs results in duplicate entries. Making the writes idempotent is critical for safe retries and is a prerequisite for checkpointing.

## Detailed Changelog
    *   Implemented `_get_entry_key_name` helper using SHA-256 hash of TestName and SubtestName prefixed with RunID.
    *   Updated `_build_new_test_history_entry` to use this key name.


scripts/process_test_history.py
```diff
@@ -113,13 +114,23 @@ class MetadataDict(TypedDict):
     labels: list[str]
 
 
+def _get_entry_key_name(run_id: str, test_name: str, subtest_name: str) -> str:
```
jcscottiii, line 118: **Key Length Guard**: While the SHA-256 hash is always 64 characters, the key name also includes the `run_id` (`{run_id}_{hash}`). If `run_id` is exceptionally long (e.g., > 435 characters), the key name could still exceed 500 characters. While `run_id` (typically an `int64` as string) is expected to be short, it is good practice to guard against this.
    * **Recommendation**: Define a constant for the limit at the top of the file, and check it in `_get_entry_key_name` to fail early:
      ```python
      # At the top of the file:
      MAX_KEY_NAME_LENGTH = 500

      # In _get_entry_key_name:
      key_name = f"{run_id}_{name_hash}"
      if len(key_name) > MAX_KEY_NAME_LENGTH:
          raise ValueError(
              f"Key name length {len(key_name)} exceeds Datastore limit of "
              f"{MAX_KEY_NAME_LENGTH} characters: {key_name}"
          )
      return key_name
      ```

jcscottiii (approved):
LGTM with one nit

scripts/process_test_history.py
DanielRyanSmith: Added!

PR #4950, opened 2026-06-15
Parallelize browser processing in process_test_history.py

DanielRyanSmith:
## Overview
This PR parallelizes browser processing and implements batching for Datastore writes.

## Root Cause / Motivation
Processing Chrome, Edge, Firefox, and Safari runs sequentially is slow. Parallelizing them improves throughput. Also, writing entries to Datastore in one giant batch at the end of a run can fail/timeout; writing in smaller batches (500) is more robust.

## Detailed Changelog
    *   Parallelized browser processing using `ThreadPoolExecutor` (4 workers) in `process_runs`.
    *   Updated `process_single_run` to establish thread-local NDB context and check if run was already processed.
    *   Implemented batching (500 entities) for Datastore writes inside `process_single_run`.
    *   Updated `create_entity_if_needed` to return the created entity to support batching.


jcscottiii (approved):
LGTM but Gemini had this note:

  - **Dependency Note**: This PR in isolation introduces a recovery bug: if a run fails midway, some batches are written. On retry, `should_process_run` (which checks if *any* entity exists for the run) will see the partial writes and skip the run, leaving it incomplete.
  - **Mitigation**: This bug is resolved in the next PR (4951) by removing `should_process_run` and relying on deterministic keys for overwriting. If merging, **PR 4950 and PR 4951 must be merged together** (or 4951 immediately after 4950) to avoid this issue in production.

PR #4957, opened 2026-06-17
Fix test history timeline end date using last processed date

### LINKED ISSUE #4953: WPT test history timelines are not updating
The [test history processing Cloud Run Function](https://pantheon.corp.google.com/run/detail/us-central1/process-test-history/observability/metrics?project=wptdashboard) has been broken in GCP for quite some time (since the beginning of the year). The cron job has been repeatedly failing each run. This means that the WPT test history has not been updating since that time. I had thought we had error monitoring set up for this scheduled task, but it seems this is not the case.

Additionally, the UI  shows the [timeline as current up to the present date](https://github.com/web-platform-tests/wpt.fyi/blob/e58d6b30a08f195206c602299270f3bd1039e33a/webapp/components/test-results-history-timeline.js#L181-L183), which is not correct. We track the date of the most recently processed aligned runs, and we should stop the timeline from displaying any date past that. This is likely another reason why this problem has been masked for so long.

We should take the time to future-proof this problem while fixing it.
- [x] Fix the underlying cause of the failures.
- [x] Carefully look at the script for ways to optimize the run time, and implement them.
- [ ] Add error monitoring for the GCP cloud run function.
- [x] Add tests for this script, since none currently exist.
- [x] Update the UI so the timeline ends at the date of the most recently processed aligned run.

DanielRyanSmith:
Resolves #4953

## Overview
This PR fixes the end date of the test results history timeline to use the last processed date instead of `Date.now()`. It also adds a 1-hour padding to the end date on the frontend to ensure visibility of the last run's status changes.

## Root Cause / Motivation
Previously, the timeline used `Date.now()` as the end date. If the test history processing was delayed or stopped, the timeline would still extend to the current time, hiding the lack of recent updates. By using the actual last processed date, we correctly reflect data freshness. Additionally, if a status change occurred in the very last processed run, the last segment would have 0 duration and be invisible; adding a 1-hour padding solves this.

## Detailed Changelog
* **`shared/models.go`**:
    * Added `MostRecentHistoryProcessed` struct to map the Datastore entity that stores the last processed date.
* **`api/test_history.go`**:
    * Query `MostRecentHistoryProcessed` in `testHistoryHandler`.
    * Defined `TestHistoryResponse` struct to include `last_processed` date.
    * Return the `last_processed` date in the API response.
* **`api/test_history_test.go`**:
    * Updated `TestHistoryHandler` test to include `MostRecentHistoryProcessed` in the mock datastore and assert it is returned in the response.
* **`webapp/components/test-results-history-timeline.js`**:
    * Store `last_processed` from API response.
    * Use `last_processed` with a 1-hour padding as the default end date for the timeline.

### DIFF
--- a/webapp/components/test-results-history-timeline.js
+++ b/webapp/components/test-results-history-timeline.js
@@ -166,7 +166,12 @@ class TestResultsTimeline extends PathInfo(PolymerElement) {
     this.dataTable.addColumn({ type: 'date', id: 'End' });
 
     const dataTableRows = [];
-    const now = new Date();
+    // Use the last processed date as the end date of the timeline.
+    // We add 1 hour padding to ensure that if a status change occurred in the
+    // last processed run, it will still be visible as a segment with non-zero duration.
+    const defaultEndDate = this.lastProcessed
+      ? new Date(new Date(this.lastProcessed).getTime() + 60 * 60 * 1000)
+      : new Date();
     this.chartRunIDs[chartIndex] = [];
 
     // Create a row for each subtest
@@ -180,7 +185,7 @@ class TestResultsTimeline extends PathInfo(PolymerElement) {
 
         // Use the next entry as the end date, or use present time if this
         // is the last entry
-        let endDate = now;
+        let endDate = defaultEndDate;
         if (i + 1 !== browserTestData[subtestName].length) {
           const nextDataPoint = browserTestData[subtestName][i + 1];
           endDate = new Date(nextDataPoint.date);
@@ -257,6 +262,7 @@ class TestResultsTimeline extends PathInfo(PolymerElement) {
     if(this.historicalData) {
       this.historicalData = {};
     }
+    this.lastProcessed = null;
 
     const options = {
       method: 'POST',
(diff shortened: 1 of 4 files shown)


szy196 (approved):

webapp/components/test-results-history-timeline.js
```diff
@@ -166,7 +166,12 @@ class TestResultsTimeline extends PathInfo(PolymerElement) {
     this.dataTable.addColumn({ type: 'date', id: 'End' });
 
     const dataTableRows = [];
-    const now = new Date();
+    // Use the last processed date as the end date of the timeline.
+    // We add 1 hour padding to ensure that if a status change occurred in the
+    // last processed run, it will still be visible as a segment with non-zero duration.
+    const defaultEndDate = this.lastProcessed
```
szy196: Gemini gave a suggestion to have a defensive check here:
 ```   
let defaultEndDate = new Date();
    if (this.lastProcessed) {
      const lastProcessedTime = new Date(this.lastProcessed).getTime();
      if (!isNaN(lastProcessedTime)) {
        defaultEndDate = new Date(lastProcessedTime + 60 * 60 * 1000);
      }
    }
```

szy196 (approved):

jcscottiii (approved):

